PR #2, opened 2024-06-15
fix and improve GitHub cache action

bcressey:
**Issue number:**
Failure observed after merging the first PR.

**Description of changes:**
The previous cache action failed because `Makefile.toml` was removed from the kit. Switch it to use the Makefile targets instead.

While we're at it, modify the cache action to cache Rust and Go deps, to see if this helps to speed up builds.

**Testing done:**
Ran through the `make` steps locally.

**Terms of contribution:**


larvacea:
All checks have failed: one of those cache actions does not have the required `key` parameter.

larvacea (approved):
It works. I am a fan. I've been running multiple cache actions for years on my own tiny app.

yeazelm (approved):

PR #7, opened 2024-06-19
Use bottlerocket-settings-sdk for settings models needed in core-kit

cbgbt:
**Description of changes:**
Some software in Bottlerocket uses `modeled-types` shared by Bottlerocket's settings models. These types now have a home in https://github.com/bottlerocket-os/bottlerocket-settings-sdk. This change moves core-kit to refer to the settings SDK for shared types.

**Testing done:**
* Core kit builds

**Terms of contribution:**


bcressey (approved):

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1571,7 +1573,7 @@ impl Runtime {
     fn get_resource_limits(
         &self,
         rlimit_type: &OciDefaultsResourceLimitType,
-        values: &OciDefaultsResourceLimit,
+        values: &OciDefaultsResourceLimitV1,
```
sam-berning, line 1576: I think it's ok to use the setting extension's model for now, but long term these helpers should live with the setting extension right?

sam-berning (approved):

sources/api/schnauzer/src/helpers.rs
cbgbt: Yes. The settings extension API provides a mechanism for exposing these helpers, so the helper code would move to the owning extension.

PR #11, opened 2024-06-20
don't cache debug and nvidia packages

bcressey:
**Issue number:**
N/A

**Description of changes:**
These packages significantly increase the size of the cache, and aren't needed when rebuilding packages.

Also, try to fix another bug that prevents the cache from working, by ignoring errors in that step:
```
Run gh extension install actions/gh-actions-cache
...
Error: Cache with input key 'build-cache' does not exist
Error: Process completed with exit code 1.
```

**Testing done:**
😬 

**Terms of contribution:**

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -34,12 +34,18 @@ jobs:
       # This builds the current packages and kits.
       - run: make ARCH=x86_64
       - run: make ARCH=aarch64
-      # Remove the previous cache.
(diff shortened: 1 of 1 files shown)


larvacea (approved):
Both pieces look reasonable. In particular "(if it exists)" certainly looks as if it should solve the "nope, no such cache key, FAIL" issue.

cbgbt (approved):
😬

PR #15, opened 2024-06-25
apiclient/apiserver: add ephemeral-storage commands

tzneal:
**Issue number:**

Closes #

**Description of changes:**

Adds apiclient/apiserver commands for initializing ephemeral storage and binding directories to that storage

**Testing done:**

Updated containerd package to run the new apiclient commands:

```patch
diff --git a/packages/containerd/containerd.service b/packages/containerd/containerd.service
index fc18c2ba..e483c3c1 100644
--- a/packages/containerd/containerd.service
+++ b/packages/containerd/containerd.service
@@ -9,6 +9,8 @@ Slice=runtime.slice
 EnvironmentFile=/etc/network/proxy.env
 EnvironmentFile=-/etc/containerd/nvidia.env
 ExecStart=/usr/bin/containerd
+ExecStartPre=apiclient ephemeral-storage init
+ExecStartPre=apiclient ephemeral-storage bind --dirs /var/lib/containerd /var/lib/kubelet /var/log/pods
 Type=notify
 Delegate=yes
 KillMode=process
```

I then launched an instance with 0 ephemeral disks, 1 disk, and multiple disks and ensured they all started up correctly. For the instance with > 0 disks, I ensured that our disk was faster:

```sh
bash-5.1# cd /var/log/pods
bash-5.1# dd if=/dev/zero of=./test bs=512k count=4096 oflag=direct
cd4096+0 records in
4096+0 records out
2147483648 bytes (2.1 GB, 2.0 GiB) copied, 1.50426 s, 1.4 GB/s

bash-5.1# cd /var/log
bash-5.1# dd if=/dev/zero of=./test bs=512k count=4096 oflag=direct
4096+0 records in
4096+0 records out
2147483648 bytes (2.1 GB, 2.0 GiB) copied, 18.6687 s, 115 MB/s
```

**Terms of contribution:**

### DIFF
--- a/sources/api/apiclient/src/ephemeral_storage.rs
+++ b/sources/api/apiclient/src/ephemeral_storage.rs
@@ -1,0 +1,87 @@
+use model::ephemeral_storage::{Bind, Filesystem, Init};
+use snafu::ResultExt;
+use std::path::Path;
+
+/// Requests ephemeral storage initialization through the API
+pub async fn initialize<P>(
+    socket_path: P,
+    filesystem: Option<Filesystem>,
+    disks: Option<Vec<String>>,
+) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    let uri = "/actions/ephemeral-storage/init";
+    let opts =
+        serde_json::to_string(&Init { filesystem, disks }).context(error::JsonSerializeSnafu {})?;
+    let method = "POST";
+    let (_status, _body) = crate::raw_request(&socket_path, &uri, method, Some(opts))
+        .await
+        .context(error::RequestSnafu { uri, method })?;
+    Ok(())
+}
+
+/// Requests binding of directories to configured ephemeral storage through the API
+pub async fn bind<P>(socket_path: P, targets: Vec<String>) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    let uri = "/actions/ephemeral-storage/bind";
+    let opts = serde_json::to_string(&Bind { targets }).context(error::JsonSerializeSnafu {})?;
+    let method = "POST";
+    let (_status, _body) = crate::raw_request(&socket_path, &uri, method, Some(opts))
+        .await
+        .context(error::RequestSnafu { uri, method })?;
+    Ok(())
+}
+
+/// Lists the ephemeral disks available for configuration
+pub async fn list_disks<P>(socket_path: P, format: Option<String>) -> Result<String>
+where
+    P: AsRef<Path>,
+{
+    list(socket_path, "list-disks", format).await
+}
+
+/// Lists the ephemeral disks available for configuration
+pub async fn list_dirs<P>(socket_path: P, format: Option<String>) -> Result<String>
+where
+    P: AsRef<Path>,
+{
+    list(socket_path, "list-dirs", format).await
+}
+async fn list<P>(socket_path: P, item: &str, format: Option<String>) -> Result<String>
+where
+    P: AsRef<Path>,
+{
+    let mut query = Vec::new();
+    if let Some(query_format) = format {
+        query.push(format!("format={}", query_format));
+    }
+
+    let uri = format!("/actions/ephemeral-storage/{}?{}", item, query.join("&"));
+    let method = "GET";
+    let (_status, body) = crate::raw_request(&socket_path, &uri, method, None)
+        .await
+        .context(error::RequestSnafu { uri, method })?;
+    Ok(body)
+}
+mod error {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub enum Error {
+        #[snafu(display("Failed {} request to '{}': {}", method, uri, source))]
+        Request {
+            method: String,
+            uri: String,
+            #[snafu(source(from(crate::Error, Box::new)))]
+            source: Box<crate::Error>,
+        },
+        #[snafu(display("Failed to serialize parameters"))]
+        JsonSerialize { source: serde_json::Error },
+    }
+}
+pub use error::Error;
+pub type Result<T> = std::result::Result<T, error::Error>;
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -6,8 +6,9 @@
 // library calls based on the given flags, etc.)  The library modules contain the code for talking
 // to the API, which is intended to be reusable by other crates.
 
-use apiclient::{apply, exec, get, reboot, report, set, update, SettingsInput};
+use apiclient::{apply, ephemeral_storage, exec, get, reboot, report, set, update, SettingsInput};
 use log::{info, log_enabled, trace, warn};
+use model::ephemeral_storage::Filesystem;
 use serde::{Deserialize, Serialize};
 use simplelog::{
     ColorChoice, ConfigBuilder as LogConfigBuilder, LevelFilter, TermLogger, TerminalMode,
@@ -15,8 +16,10 @@ use simplelog::{
 use snafu::ResultExt;
 use std::env;
 use std::ffi::OsString;
+use std::iter::Peekable;
 use std::process;
 use std::str::FromStr;
+use std::vec::IntoIter;
 use unindent::unindent;
 
 const DEFAULT_METHOD: &str = "GET";
@@ -48,6 +51,7 @@ enum Subcommand {
     Set(SetArgs),
     Update(UpdateSubcommand),
     Report(ReportSubcommand),
+    EphemeralStorage(EphemeralStorageSubcommand),
 }
 
 /// Stores user-supplied arguments for the 'apply' subcommand.
@@ -140,6 +144,33 @@ struct UpdateApplyArgs {
 #[derive(Debug)]
 struct UpdateCancelArgs {}
 
+/// Stores the 'ephemeral-storage' subcommand specified by the user.
+#[derive(Debug)]
+enum EphemeralStorageSubcommand {
+    Init(EphemeralStorageInitArgs),
+    Bind(EphemeralStorageBindArgs),
+    ListDisks(EphemeralStorageFormatArgs),
+    ListDirs(EphemeralStorageFormatArgs),
+}
+
+/// Stores user-supplied arguments for the 'ephemeral-storage init' subcommand.
+#[derive(Debug)]
+struct EphemeralStorageInitArgs {
+    disks: Option<Vec<String>>,
+    filesystem: Option<Filesystem>,
+}
+
+/// Stores user-supplied arguments for the 'ephemeral-storage bind' subcommand.
+#[derive(Debug)]
+struct EphemeralStorageBindArgs {
+    targets: Vec<String>,
+}
+/// Stores user-supplied arguments for the 'ephemeral-storage list-disks/list-dirs' subcommand.
+#[derive(Debug)]
+struct EphemeralStorageFormatArgs {
+    format: Option<String>,
+}
+
 /// Informs the user about proper usage of the program and exits.
 fn usage() -> ! {
     let msg = &format!(
@@ -166,6 +197,12 @@ fn usage() -> ! {
             report cis                 Retrieve a Bottlerocket CIS benchmark compliance report.
             report cis-k8s             Retrieve a Kubernetes CIS benchmark compliance report.
             report fips                Retrieve a FIPS Security Policy compliance report.
+            ephemeral-storage init     Initialize ephemeral storage
+            ephemeral-storage bind     Bind directories to previously initialized ephemeral storage.
+            ephemeral-storage list-disks
+                                       List the discovered ephemeral disks that can be initialized.
+            ephemeral-storage list-dirs
+                                       List the directories that can be bound to ephemeral storage.
 
         raw options:
             -u, --uri URI              Required; URI to request from the server, e.g. /tx
@@ -223,7 +260,28 @@ fn usage() -> ! {
 
         report cis-k8s options:
             -f, --format               Format of the CIS report (text or json). Default format is text.
-            -l, --level                CIS compliance level to report on (1 or 2). Default is 1."#,
+            -l, --level                CIS compliance level to report on (1 or 2). Default is 1.
+
+        ephemeral-storage init options:
+            -t, --filesystem           Filesystem to initialize the array as (ext4 or xfs). Default is
+                                       xfs. If a single disk is provided, it is mounted directly without
+                                       constructing an array. If no ephemeral disks are found, this
+                                       operation does nothing.
+            --disks DISK [DISK ...]    Local disks to configure for storage. Default is all ephemeral
+                                       disks.
+
+        ephemeral-storage bind options:
+            --dirs DIR [DIR ...]       Directories to bind to configured ephemeral storage
+                                       (e.g. /var/lib/containerd). If no ephemeral disks are found
+                                       this operation does nothing.
+
+        ephemeral-storage list-disks options:
+            -f, --format               Format of the disk listing (text or json). Default format is text.
+
+        ephemeral-storage list-dirs options:
+            -f, --format               Format of the directory listing (text or json). Default format is text.
+
+            "#,
         socket = constants::API_SOCKET,
         method = DEFAULT_METHOD,
     );
@@ -272,6 +330,7 @@ fn parse_args(args: env::Args) -> (Args, Subcommand) {
 
             // Subcommands
             "raw" | "apply" | "exec" | "get" | "reboot" | "report" | "set" | "update"
+            | "ephemeral-storage"
                 if subcommand.is_none() && !arg.starts_with('-') =>
             {
                 subcommand = Some(arg)
@@ -292,6 +351,7 @@ fn parse_args(args: env::Args) -> (Args, Subcommand) {
         Some("report") => (global_args, parse_report_args(subcommand_args)),
         Some("set") => (global_args, parse_set_args(subcommand_args)),
         Some("update") => (global_args, parse_update_args(subcommand_args)),
+        Some("ephemeral-storage") => (global_args, parse_ephemeral_storage_args(subcommand_args)),
         _ => usage_msg("Missing or unknown subcommand"),
     }
 }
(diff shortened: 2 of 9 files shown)


sources/models/src/ephemeral_storage.rs
```diff
@@ -0,0 +1,32 @@
+//! The 'ephemeral_storage' module holds types used to communicate between client and server for
+//! 'apiclient ephemeral-storage'.
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+
+/// Supported filesystems for ephemeral storage
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Filesystem {
+    XFS,
+    EXT4,
```
bcressey:
```suggestion
    Xfs,
    Ext4,
```

sources/api/apiclient/src/main.rs
```diff
@@ -152,20 +183,24 @@ fn usage() -> ! {
...
+            update apply                 Applies available updates.
+            update cancel                Deactivates an applied update.
+            reboot                       Reboots the host.
+            exec                         Execute a command in a host container.
+            report cis                   Retrieve a Bottlerocket CIS benchmark compliance report.
+            report cis-k8s               Retrieve a Kubernetes CIS benchmark compliance report.
+            report fips                  Retrieve a FIPS Security Policy compliance report.
+            ephemeral-storage init       Initialize ephemeral storage
+            ephemeral-storage bind       Bind directories to previously initialized ephemeral storage.
+            ephemeral-storage list-disks List the discovered ephemeral disks that can be initialized.
```
bcressey: nit: I might instead move the "List the discovered ..." text to the next line, at the original offset. That way all the other help text doesn't need to be realigned.

sources/api/apiclient/src/main.rs
```diff
@@ -223,7 +258,28 @@ fn usage() -> ! {
 
         report cis-k8s options:
             -f, --format               Format of the CIS report (text or json). Default format is text.
-            -l, --level                CIS compliance level to report on (1 or 2). Default is 1."#,
+            -l, --level                CIS compliance level to report on (1 or 2). Default is 1.
+
+        ephemeral-storage init options:
+            -fs, --filesystem          Filesystem to initialize the array as (ext4 or xfs). Default is
```
bcressey: nit: short option should be one character. Maybe also use `-t` in case we ever want `-f` to mean `--force`.
```suggestion
            -t, --filesystem-type      Filesystem to initialize the array as (ext4 or xfs). Default is
```

sources/api/apiclient/src/main.rs
```diff
@@ -223,7 +258,28 @@ fn usage() -> ! {
...
+                                       xfs. If a single disk is provided, its mounted directly without
```
bcressey:
```suggestion
                                       xfs. If a single disk is provided, it is mounted directly without
```

sources/api/apiclient/src/ephemeral_storage.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Requests ephemeral storage initialization through the API
+pub async fn initialize<P>(
+    socket_path: P,
+    filesystem: Option<Filesystem>,
+    disks: Option<Vec<String>>,
+) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    let uri = "/actions/ephemeral-storage/init";
```
bcressey, line 14: We try to keep `sources/api/openapi.yaml` up to date, though it's not used for codegen purposes.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -648,6 +666,81 @@ async fn get_fips_report(query: web::Query<HashMap<String, String>>) -> Result<H
...
+        "json" => Ok(
+            EphemeralListResponse(allowed.iter().map(|x| String::from(*x)).collect())
+                .respond_to(&req),
+        ),
+        "text" => Ok(HttpResponse::Ok()
+            .content_type("application/text")
+            .body(text_response)),
+        _ => Ok(HttpResponse::BadRequest().body("unsupported format")),
+    }
+}
```
bcressey: It's tempting to factor this into a helper function for preparing the response, since it's mostly shared between the two. Why is the "json" arm different for `list_ephemeral_storage_dirs`?

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+static MDADM: &str = "/usr/sbin/mdadm";
+static BLKID: &str = "/usr/sbin/blkid";
+static MKFSXFS: &str = "/usr/sbin/mkfs.xfs";
+static MKFSEXT4: &str = "/usr/sbin/mkfs.ext4";
+
+/// Name of the array (if created) and filesystem label. Selected to be 12 characters so it
+/// fits within both the xfs and ext4 volume label limit.
+static NAME: &str = "br-ephemeral";
+
+/// initialize prepares the ephemeral storage for formatting and formats its.  For multiple disks
```
bcressey:
```suggestion
/// initialize prepares the ephemeral storage for formatting and formats it.  For multiple disks
```

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+/// preparation is the creation of a RAID0 array, for a single disk this is a no-op. The array or disk
+/// is then formatted with the specified filesystem (default=xfs) if not formatted already.
+pub fn initialize(fs: Option<Filesystem>, disks: Option<Vec<String>>) -> Result<()> {
+    let known_disks = ephemeral_devices()?;
+    let known_disks_hash = HashSet::<_>::from_iter(known_disks.iter());
```
bcressey, line 28: I find `OsString` fairly painful to work with and would usually convert them immediately into `String` by way of `.to_string_lossy()` or similar methods. That might cut down on some of the other transforms you have to do.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+    );
+
+    info!("initializing ephemeral storage disks={:?}", disks);
+    // with a single disk, there is no need to create the array
+    let device_name = match disks.len() {
+        1 => disks.first().expect("non-empty").clone(),
+        _ => {
+            let scan_output = mdadm_scan()?;
+            // no previously configured array found, so construct a new one
```
bcressey, line 71: Do we not need to run `mdadm --assemble`? Or that happens automatically during boot?

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+    for dir in &dirs {
+        ensure!(
+            allowed_dirs.contains(dir.as_str()),
+            error::InvalidParameterSnafu {
+                parameter: dir,
+                reason: "specified bind directory not in allow list",
+            }
+        )
+    }
+    std::fs::create_dir_all(mount_point).context(error::MkdirSnafu {})?;
```
bcressey, line 117: If we create a `/mnt/ephemeral` (or `/mnt/.ephemeral`) in `filesystem.spec`, then we would be able to mount local storage without having a writable `/mnt` yet (which is something I'd eventually like to support).

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+
+static MOUNT: &str = "/usr/bin/mount";
+static MDADM: &str = "/usr/sbin/mdadm";
+static BLKID: &str = "/usr/sbin/blkid";
+static MKFSXFS: &str = "/usr/sbin/mkfs.xfs";
+static MKFSEXT4: &str = "/usr/sbin/mkfs.ext4";
+
+/// Name of the array (if created) and filesystem label. Selected to be 12 characters so it
+/// fits within both the xfs and ext4 volume label limit.
+static NAME: &str = "br-ephemeral";
```
bcressey: nit: convention would be to uppercase this, and maybe it can just be "EPHEMERAL"

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+/// they do not exist.
+pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
+    // handle the no local instance storage case
+    if ephemeral_devices()?.is_empty() {
+        info!("no ephemeral disks found, skipping ephemeral storage binding");
+        return Ok(());
+    }
+
+    let device_name = resolve_device_by_label()?;
+    let mount_point = format!("/mnt/{}", NAME);
```
bcressey: We might want to use `.ephemeral` as the name, to indicate that it's not really something we expect anyone to mess with, while trying to avoid collisions with existing use of `/mnt`.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+            .status()
+            .context(error::MountArrayFailureSnafu {})?
+            .success(),
+        error::MountArrayFailureStatusSnafu {}
+    );
+
+    for dir in dirs {
+        // construct a directory name (E.g. /var/lib/kubelet => _var_lib_kubelet) that will be
+        // unique between the binding targets
+        let directory_name = OsString::from(dir.replace('/', "_"));
```
bcressey: maybe prefix this with a `.` also

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+            Command::new(MOUNT)
+                .args([
+                    OsStr::new("--rbind"),
+                    mount_destination.as_ref(),
+                    &source_dir
+                ])
+                .status()
+                .context(error::BindDirectoryFailureSnafu {
+                    dir: String::from_utf8_lossy(source_dir.as_encoded_bytes())
+                })?
```
bcressey: This particular error should only happen if the `mount` command is not found, or permission is denied, so the error should show what command was attempted rather than the directory.

See the various `CommandFailure` definitions we have in error enums across the project for a fairly nice way to do the reporting for these errors.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        info!("sharing mounts for {:?}", source_dir);
+        // mount --make-rshared /var/lib/kubelet
+        ensure!(
+            Command::new(MOUNT)
+                .args([OsStr::new("--make-rshared"), &source_dir])
+                .status()
+                .context(error::ShareMountsFailureSnafu {
+                    dir: String::from_utf8_lossy(source_dir.as_encoded_bytes())
+                })?
```
bcressey: This can reuse the "command not found" error.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+            error::BindDirectoryFailureStatusSnafu {
+                dir: String::from_utf8_lossy(source_dir.as_encoded_bytes())
+            }
+        );
+
+        info!("sharing mounts for {:?}", source_dir);
+        // mount --make-rshared /var/lib/kubelet
+        ensure!(
+            Command::new(MOUNT)
+                .args([OsStr::new("--make-rshared"), &source_dir])
```
bcressey: I'd recommend doing this in a separate loop once all the mounts have happened, to simplify reasoning about what state the system might be in.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+        // mount --rbind /mnt/ephemeral/_var_lib_kubelet /var/lib/kubelet
+        info!("binding {:?} to {:?}", source_dir, mount_destination);
+        ensure!(
+            Command::new(MOUNT)
+                .args([
+                    OsStr::new("--rbind"),
+                    mount_destination.as_ref(),
+                    &source_dir
+                ])
+                .status()
```
bcressey: If we're binding to multiple targets, we may error out in the second or Nth mount, and if the command were attempted again, we'd end up repeatedly binding the first mount. It might be good to check whether this directory is already a bind mount from the ephemeral storage, via parsing `findmnt -n -o TARGET,FSROOT <dir>` output. And skip it if so.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+        let entry = entry.context(error::DiscoverEphemeralSnafu {})?;
+        filenames.push(entry.path().into_os_string());
+    }
+    Ok(filenames)
+}
+
+/// allowed_bind_dirs returns a set of the directories that can be bound to ephemeral storage, which
+/// varies based on the variant
+pub fn allowed_bind_dirs(variant: &str) -> HashSet<&'static str> {
+    let mut allowed = HashSet::from(["/var/lib/containerd"]);
```
bcressey: It would be nice to allow `/var/lib/host-containerd` here also.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+    if variant.contains("k8s") {
+        allowed.insert("/var/lib/kubelet");
+        allowed.insert("/var/log/pods");
+    }
+    if variant.contains("ecs") {
+        allowed.insert("/var/log/ecs");
+    }
```
bcressey:
```suggestion
    if variant.contains("ecs") {
        allowed.insert("/var/lib/docker");
        allowed.insert("/var/log/ecs");
    }
```

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+            Command::new(MOUNT)
+                .args([OsStr::new("--make-rshared"), &source_dir])
+                .status()
+                .context(error::ShareMountsFailureSnafu {
+                    dir: String::from_utf8_lossy(source_dir.as_encoded_bytes())
+                })?
+                .success(),
+            error::ShareMountsFailureStatusSnafu {
+                dir: String::from_utf8_lossy(source_dir.as_encoded_bytes())
+            }
```
bcressey: For the error case where the command ran but failed, it'd be good to collect the process output and print that as additional context.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,379 @@
...
+
+/// resolve_array_by_name resolves the by-id link for the raid array
+fn resolve_array_by_id() -> Result<OsString> {
+    let device_name = OsString::from(format!("/dev/disk/by-id/md-name-{}", NAME));
+
+    // read the symlink from known named location to get the current device that we need
+    // to use
+    let device_name = std::fs::canonicalize(device_name)
+        .context(error::CanonicalizeFailureSnafu {})?
+        .into_os_string();
```
bcressey: nit: this could be a helper shared by the `resolve_*` functions

bcressey (commented):
Nice, thanks for contributing this!

sources/api/apiserver/src/server/ephemeral_storage.rs
tzneal: Since I'm not collecting it, it goes to stdout/stderr and shows up in `journalctl -u apiserver`.

sources/api/apiserver/src/server/ephemeral_storage.rs
tzneal: Yeah, there's 64-md-raid-assembly.rules which handles that.

sources/api/apiserver/src/server/ephemeral_storage.rs
tzneal: Switched to /mnt/.ephemeral.  I tried creating it in filesystem.spec, but it didn't work, there's some stuff going on with /mnt getting remounted that I didn't track down.

sources/api/apiserver/src/server/ephemeral_storage.rs
tzneal: Didn't follow this part, lets discuss when you're back.

sources/api/apiserver/src/server/ephemeral_storage.rs
bcressey: > I tried creating it in filesystem.spec, but it didn't work, there's some stuff going on with /mnt getting remounted that I didn't track down.

Oh right - we'd need a `prepare-mnt.service` equivalent to [prepare-var.service](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/release/prepare-var.service) and [prepare-opt.service](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/release/prepare-opt.service).

sources/api/apiserver/src/server/mod.rs
tzneal: This got cleaned up when I switched away from OsString.

sources/api/openapi.yaml
```diff
@@ -735,3 +746,97 @@ paths:
...
+        400:
+          description: "Bad request input"
+        422:
+          description: "Unprocessable request"
+        500:
+          description: "Server error"
+  /ephemeral-storage/bind:
+    post:
+      summary: "Bind directories to previously initialized ephemeral storage"
+      operationId: "init"
```
bcressey:
```suggestion
      operationId: "bind"
```

sources/api/openapi.yaml
```diff
@@ -735,3 +746,97 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: string
+        400:
+          description: "Bad request input"
+        422:
+          description: "Unprocessable request"
+        500:
+          description: "Server error"
```
bcressey: nit: file should end in a newline

sources/api/apiclient/src/main.rs
```diff
@@ -668,6 +728,137 @@ fn parse_fips_arguments(args: Vec<String>) -> FipsReportArgs {
...
+    while let Some(arg) = iter.next() {
+        match arg.as_ref() {
+            "--dirs" => {
+                targets = collect_non_args(&mut iter);
+                if targets.is_empty() {
+                    usage_msg("Did not give argument to --dirs")
+                }
+            }
+            x => usage_msg(format!("Unknown argument '{}'", x)),
+        }
```
bcressey: Should we support multiple `--dirs` arguments? Same would apply to `--disks`.
```
apiclient ephemeral-storage bind \
  --dirs /var/lib/containerd \
  --dirs /var/lib/kubelet \
  --dirs /var/log/pods
```

Right now, any subsequent instances of the argument would clobber the earlier ones.

(We could also accept `--dir` or `-d` as synonyms to make the multi-argument form more natural.)

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,380 @@
...
+                        parameter: "disks",
+                        reason: format!("unknown disk {:?}", disk),
+                    }
+                )
+            }
+            disks
+        }
+        None => {
+            // if there are no disks specified, and none are available we treat the init as a
+            // no-op to allow "ephemeral storage init"/"ephemeral-storage bind" to work on instances
```
bcressey: nit:
```suggestion
            // no-op to allow "ephemeral-storage init"/"ephemeral-storage bind" to work on instances
```

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,380 @@
...
+    #[snafu(visibility(pub(super)))]
+    pub enum Error {
+        #[snafu(display("Failed to execute '{:?}': {}", command, source))]
+        ExecutionFailure {
+            command: &'static str,
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Failed to discover ephemeral disk"))]
+        DiscoverEphemeral { source: std::io::Error },
```
bcressey:
```suggestion
        #[snafu(display("Failed to discover ephemeral disks from {}: {}", path, source))]
        DiscoverEphemeral { source: std::io::Error, path: PathBuf },
```

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -0,0 +1,380 @@
...
+
+        #[snafu(display("Failed to mount array (non-zero exit)"))]
+        MountArrayFailure {},
```
bcressey: These "non-zero exit" errors could use more context, including the error message from the command:
```suggestion
        #[snafu(display("Failed to mount {} to {}: {}", what, where, output))]
        MountArrayFailure { what: PathBuf, where: PathBuf, output: std::process::Output },
```

sources/api/apiclient/src/main.rs
tzneal: I prefer `--dirs /var/lib/containerd /var/lib/kubelet /var/log/pods` since its less repetitive, but its also unique compared to other commands.  I did update it so that you can repeat `--dirs` (e.g. `--dirs /var/lib/containerd /var/lib/kubelet --dirs /var/log/pods` captures all three args).

If I change it to not support multiple arguments per option, it would probably need to be:

`--dir /var/lib/containerd --dir /var/lib/kubelet --dir /var/log/pods`

I'm good either way, your call.

bcressey (approved):

sources/api/apiclient/src/main.rs
bcressey: > I did update it so that you can repeat --dirs

Thanks, this works for me!

arnaldo2792 (approved):

PR #20, opened 2024-06-27
Fix segfault in host-ctr

larvacea:
Bottlerocket invokes the host-ctr executable with an optional "--registry-config" parameter. If this parameter was omitted (for instance, in customer code), host-ctr would segfault. Now it will behave as it would for an empty registry-config file.

**Issue number:**

Closes #4059

**Description of changes:**

Protect two locations where we tried to dereference the registryConfig pointer. The pointer is nil if host-ctr is invoked
without the optional `--registry-config` parameter.

**Testing done:**

Built an image and demonstrated that the command in issue #4059 could load a container from a public ECR repository without segfaulting.

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -1105,6 +1105,8 @@ func pullImage(ctx context.Context, source string, client *containerd.Client, re
 				Error("failed to read registry config")
 			return nil, err
 		}
+	} else {
+		registryConfig = NewBlankRegistryConfig()
 	}
 
 	// Pull the image
@@ -1225,9 +1227,12 @@ func withDynamicResolver(ctx context.Context, ref string, registryConfig *Regist
 		}
 	// For Amazon ECR Public registries, we should try and fetch credentials before resolving the image reference
 	case strings.HasPrefix(ref, "public.ecr.aws/"):
-		// ... not if the user has specified their own registry credentials for 'public.ecr.aws'; In that case we use the default resolver.
-		if _, found := registryConfig.Credentials["public.ecr.aws"]; found {
-			return defaultResolver
+		// if we have registryConfig, and the user specified credentials for
+		// 'public.ecr.aws', then use defaultResolver.
+		if registryConfig != nil {
+			if _, found := registryConfig.Credentials["public.ecr.aws"]; found {
+				return defaultResolver
+			}
 		}
 
 		// Try to get credentials for authenticated pulls from ECR Public
--- a/sources/host-ctr/cmd/host-ctr/registry.go
+++ b/sources/host-ctr/cmd/host-ctr/registry.go
@@ -45,6 +45,13 @@ func NewRegistryConfig(registryConfigFile string) (*RegistryConfig, error) {
 	return &config, toml.Unmarshal(raw, &config)
 }
 
+// NewBlankRegistryConfig sets up a RegistryConfig without unmarshalling any toml
+func NewBlankRegistryConfig() *RegistryConfig {
+	config := RegistryConfig{}
+	// no config to unmarshal
+	return &config
+}
+
 // registryHosts returns the registry hosts to be used by the resolver.
 // Heavily borrowed from containerd CRI plugin's implementation.
 // See https://github.com/containerd/containerd/blob/1407cab509ff0d96baa4f0eb6ff9980270e6e620/pkg/cri/server/image_pull.go#L332-L405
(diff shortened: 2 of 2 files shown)


larvacea:
The linter complains about seven locations with deprecated function calls. Six are calls to log.L or log.G, and one to a deprecated method in containerd.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1152,9 +1152,11 @@ func withDynamicResolver(ctx context.Context, ref string, registryConfig *Regist
 		}
 	// For Amazon ECR Public registries, we should try and fetch credentials before resolving the image reference
 	case strings.HasPrefix(ref, "public.ecr.aws/"):
-		// ... not if the user has specified their own registry credentials for 'public.ecr.aws'; In that case we use the default resolver.
-		if _, found := registryConfig.Credentials["public.ecr.aws"]; found {
-			return defaultResolver
+		if registryConfig != nil {
```
yeazelm, line 1232: Isn't the right behavior here to just return the defaultResolver if `registryConfig` is nil?

sources/host-ctr/cmd/host-ctr/registry.go
```diff
@@ -58,10 +58,12 @@ func registryHosts(registryConfig *RegistryConfig, authorizerOverride *docker.Au
...
-		if _, ok := registryConfig.Mirrors[host]; ok {
-			endpoints = registryConfig.Mirrors[host].Endpoints
-		} else {
-			endpoints = registryConfig.Mirrors["*"].Endpoints
+		if registryConfig != nil {
+			if _, ok := registryConfig.Mirrors[host]; ok {
+				endpoints = registryConfig.Mirrors[host].Endpoints
+			} else {
+				endpoints = registryConfig.Mirrors["*"].Endpoints
+			}
```
yeazelm: Haven't you created a new problem that `endpoints` can now be unset? I think we probably want the else clause to be for the first if statement:

```suggestion
		if registryConfig != nil {
			if _, ok := registryConfig.Mirrors[host]; ok {
				endpoints = registryConfig.Mirrors[host].Endpoints
			} 
		} else {
		     endpoints = registryConfig.Mirrors["*"].Endpoints

```

sources/host-ctr/cmd/host-ctr/main.go
larvacea: This case branch is "this is the public ECR," and (if possible) the code here is trying to get an auth token for the public ECR. So this if statement at the head of the case branch says "but hey, if `registryConfig` has something to say about the public ECR, well, never mind, we will go with what `registryConfig` says." Thus if `registryConfig` is nil, I am treating it as `registryConfig` does not have anything to say, and we should attempt to get the token.

larvacea: Perhaps a comment is in order. Assuming that my version of "what this thing does" is correct.

sources/host-ctr/cmd/host-ctr/registry.go
larvacea: Likewise, another good candidate for a comment. We start with an empty array of endpoints. Then we may replace that empty array with a non-empty one if registryConfig has an override. Then we append `docker.defaultHost(host)` to the array. The endpoints array always has at least that one `docker.defaultHost()` entry.

larvacea:
I have added comments, hoping to make the code a little less opaque.

sources/host-ctr/cmd/host-ctr/main.go
yeazelm: Ah that makes sense. The diff didn't help here because the top of the function is what creates this `defaultResolver`. After reading more of the context I see how this is the right way to solve this.

yeazelm:
I think I got myself confused a bit reviewing this code so I went digging, I wonder if the right fix is to ensure https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/host-ctr/cmd/host-ctr/main.go#L1025 doesn't proceed as an uninitialized variable if `registryConfigPath` is empty. This seems like a validation of arguments issue. I think the defensive `if` statements you have added make sense, but we should also fix this issue of creating a variable and only creating the right struct if a different variable. Can we look at creating an empty `registryConfig` one isn't passed? This would allow all the passing of this variable downstream to be safer.

larvacea:
Yes, I think it makes sense to build an empty `registryConfig` if `--registry-config` is missing. This is the net effect of what we do today (almost always). We supply `--registry-config /etc/host-containers/host-ctr.toml` in both systemd units where we call `host-ctr`. The file ends up being a single blank line (at least on a default aws-k8s instance).

larvacea:
Updated: added a `NewBlankRegistryConfig` method and use it when we have an empty registry config file path. With this change, `registryConfig` will not be nil when the caller omits `--registry-config`. The locations where we were dereferencing nil are still guarded.

Tested manually using the same `host-ctr run` command supplied by our customer in the original GitHub issue.

yeazelm (approved):

larvacea:
Updated (rebased on develop). Tested: bottlerocket can still pull its containers.

vigh-m (approved):

PR #21, opened 2024-06-27
include kmod-6.1-neuron package in core kit

bcressey:
**Issue number:**
N/A

**Description of changes:**
Noticed that `kmod-6.1-neuron` was not included in the core kit yet. Now it will be!

**Testing done:**
Built core kit, added `kmod-6.1-neuron` to the `aws-k8s-1.30` variant, built an AMI and launched an instance.

**Terms of contribution:**


yeazelm (approved):

bcressey:
⬆️ force push is a hack to fix the failure to build for the aarch64 target. Turns out we don't really have a way *not* to build a package for a given architecture. I flipped a coin and it came up "fix the build failure" instead of "make the entire package a no-op on aarch64".

ytsssun (approved):

yeazelm (approved):

PR #22, opened 2024-06-28
amazon-ssm-agent: Add dynamically-linked agent binaries

### LINKED ISSUE #19: Vendor different compiled versions of the amazon-ssm-agent
**What I'd like:**

Currently, Bottlerocket only vends a statically-linked version of the amazon-ssm-agent. 

Ideally, it would have a version of the agent that is statically-linked and another that is dynamically-linked (with a configurable dependency on `bash`) in separate subpackages so that we can choose the version we want to build for the AMI.

N/a

koooosh:
**Issue number:** #19

Closes #19

**Description of changes:**

Making the following modifications to the `amazon-ssm-agent` package:
- `amazon-ssm-agent.spec`
    - Changing the base packages to vendor dynamically-linked agent binaries patched to have a configurable dependency on `bash`
    - Adding `plugin` subpackages which vendor the original statically-linked agent binaries
- `0001-agent-Add-config-to-make-shell-optional.patch`
    - This is a patch for the SSM agent that adds a new config called `UseShell` which is a boolean set to `false` by default. If this config is `false`, then `bash` is not used when running commands -- instead using `exec`. If `true`, it uses `bash` like the default Agent.
- `amazon-ssm-agent.json`
    - This is the custom config for the SSM agent with the new `UseShell` setting along with defaults installed in `%{_cross_factorydir}%{_cross_sysconfdir}`. Depending on whether they want to use `bash`, downstream users can opt to take this config by adding a tmpfiles.d for that target or create and install their own config.
- `amazon-ssm-agent.service`
    - This is the systemd service for the SSM agent. It is similar to the default Agent service besides changes to the binary path and the `RestartSec` lowered from 90 to 5. Journal logs show no unexpected behavior as a result of this reduction in time.

Also updating the `ecs-agent.spec` in the `ecs-agent` package to use the new `ssm-agent-plugin` subpackage.

**Testing done:**

- Launched an `aws-dev` variant instance with the `amazon-ssm-agent` package added and host-containers disabled.
    - When it is built with `UseShell` set to `false`, running a command with a bash script fails:
        ```
        "status": "Failed",
        ...
        "output": "\n----------ERROR-------\nfailed to run commands: fork/exec #!/usr/bin/bash: no such file or directory",
        ```
        
        While other commands still succeed. Example running `apiclient get os`:

        ```
        "status": "Success",
        ...
        "output": "{\n  \"os\": {\n    \"arch\": \"x86_64\",\n    \"build_id\": \"62833938-dirty\",\n    \"pretty_name\": \"Bottlerocket OS 1.21.0 (aws-dev)\",\n    \"variant_id\": \"aws-dev\",\n    \"version_id\": \"1.21.0\"\n  }\n}\n",
        ```
        
    - When it is built with `UseShell` set to `true`, the bash script runs successfully as expected.

- Regarding the subpackaging of the statically-linked binaries, I launched an `aws-ecs-1` variant instance with the `amazon-ssm-agent-plugin` package added and verified it is able to join a cluster + ECS exec also works.

**Terms of contribution:**

### DIFF
--- a/packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
+++ b/packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
@@ -1,0 +1,93 @@
+From c835d2ddc855439173a8a59828c335d169c03d15 Mon Sep 17 00:00:00 2001
+From: Kush Upadhyay <kushupad@amazon.com>
+Date: Tue, 2 Jul 2024 20:54:29 +0000
+Subject: [PATCH] agent: Add config to make shell optional
+
+Signed-off-by: Kush Upadhyay <kushupad@amazon.com>
+---
+ agent/appconfig/appconfig.go         |  1 +
+ agent/appconfig/contracts.go         |  2 ++
+ agent/plugins/runscript/runscript.go | 36 +++++++++++++++++++---------
+ 3 files changed, 28 insertions(+), 11 deletions(-)
+
+diff --git a/agent/appconfig/appconfig.go b/agent/appconfig/appconfig.go
+index b6abcf1..e214cd5 100644
+--- a/agent/appconfig/appconfig.go
++++ b/agent/appconfig/appconfig.go
+@@ -118,6 +118,7 @@ func DefaultConfig() SsmagentConfig {
+ 		SessionLogsRetentionDurationHours:     DefaultSessionLogsRetentionDurationHours,
+ 		PluginLocalOutputCleanup:              DefaultPluginOutputRetention,
+ 		OrchestrationDirectoryCleanup:         DefaultOrchestrationDirCleanup,
++		UseShell:                              false,
+ 	}
+ 	var agent = AgentInfo{
+ 		Name:                                    "amazon-ssm-agent",
+diff --git a/agent/appconfig/contracts.go b/agent/appconfig/contracts.go
+index 1337398..0a66441 100644
+--- a/agent/appconfig/contracts.go
++++ b/agent/appconfig/contracts.go
+@@ -50,6 +50,8 @@ type SsmCfg struct {
+ 	PluginLocalOutputCleanup string
+ 	// Configure only when it is safe to delete orchestration folder after document execution. This config overrides PluginLocalOutputCleanup when set.
+ 	OrchestrationDirectoryCleanup string
++	// Flag for shell dependency
++	UseShell bool
+ }
+ 
+ // AgentInfo represents metadata for amazon-ssm-agent
+diff --git a/agent/plugins/runscript/runscript.go b/agent/plugins/runscript/runscript.go
+index 48be5e7..d8cbcf1 100644
+--- a/agent/plugins/runscript/runscript.go
++++ b/agent/plugins/runscript/runscript.go
+@@ -174,23 +174,37 @@ func (p *Plugin) runCommands(pluginID string, pluginInput RunScriptPluginInput,
+ 		return
+ 	}
+ 
+-	// Create script file path
+-	scriptPath := filepath.Join(orchestrationDir, p.ScriptName)
+-	log.Debugf("Writing commands %v to file %v", pluginInput, scriptPath)
++	appConfig := p.Context.AppConfig()
+ 
+-	// Create script file
+-	if err = pluginutil.CreateScriptFile(log, scriptPath, pluginInput.RunCommand, p.ByteOrderMark); err != nil {
+-		output.MarkAsFailed(fmt.Errorf("failed to create script file. %v", err))
+-		return
++	var commandName string
++	var commandArguments []string
++
++	if appConfig.Ssm.UseShell {
++
++		// Create script file path
++		scriptPath := filepath.Join(orchestrationDir, p.ScriptName)
++		log.Debugf("Writing commands %v to file %v", pluginInput, scriptPath)
++
++		// Create script file
++		if err = pluginutil.CreateScriptFile(log, scriptPath, pluginInput.RunCommand, p.ByteOrderMark); err != nil {
++			output.MarkAsFailed(fmt.Errorf("failed to create script file. %v", err))
++			return
++		}
++
++		// Construct Command Name and Arguments
++		commandName = p.ShellCommand
++		commandArguments = append(p.ShellArguments, scriptPath)
++	} else {
++
++		// Take only the first element of RunCommand since we prefer single-line commands
++		commandInput := strings.Split(pluginInput.RunCommand[0], " ")
++		commandName = commandInput[0]
++		commandArguments = append(commandInput[1:])
+ 	}
+ 
+ 	// Set execution time
+ 	executionTimeout := pluginutil.ValidateExecutionTimeout(log, pluginInput.TimeoutSeconds)
+ 
+-	// Construct Command Name and Arguments
+-	commandName := p.ShellCommand
+-	commandArguments := append(p.ShellArguments, scriptPath)
+-
+ 	// Execute Command
+ 	exitCode, err := p.CommandExecuter.NewExecute(p.Context, workingDir, output.GetStdoutWriter(), output.GetStderrWriter(), cancelFlag, executionTimeout, commandName, commandArguments, pluginInput.Environment)
+ 
+-- 
+2.40.1
+
--- a/packages/amazon-ssm-agent/amazon-ssm-agent.service
+++ b/packages/amazon-ssm-agent/amazon-ssm-agent.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=Amazon SSM agent
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/amazon-ssm-agent
+KillMode=process
+
+# Restart the agent regardless of whether it crashes (and returns a non-zero result code) or if
+# is terminated normally (e.g. via 'kill -HUP').  Delay restart so that the agent is less likely
+# to restart during a reboot initiated by a script. If the agent exits with status 194 (reboot
+# requested), don't restart at all.
+Restart=always
+RestartPreventExitStatus=194
+RestartSec=5
+
+[Install]
+WantedBy=multi-user.target
--- a/packages/amazon-ssm-agent/amazon-ssm-agent.spec
+++ b/packages/amazon-ssm-agent/amazon-ssm-agent.spec
@@ -9,7 +9,11 @@ Summary: An agent to enable remote management of EC2 instances
 License: Apache-2.0
 URL: https://github.com/aws/amazon-ssm-agent
 Source0: %{gorepo}-%{version}.tar.gz
+Source1: amazon-ssm-agent.service
+Source2: amazon-ssm-agent.json
 Source1000: clarify.toml
+
+Patch0001: 0001-agent-Add-config-to-make-shell-optional.patch
 
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{name}(binaries)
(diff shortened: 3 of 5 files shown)


packages/amazon-ssm-agent/amazon-ssm-agent-tmpfiles.conf
```diff
@@ -0,0 +1 @@
+C /etc/amazon/ssm/amazon-ssm-agent.json - - - - -
```
arnaldo2792: You don't need to provide this, let downstreams provide it.

packages/amazon-ssm-agent/amazon-ssm-agent.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=amazon-ssm-agent
```
arnaldo2792:
```suggestion
Description=Amazon SSM agent
```

packages/amazon-ssm-agent/amazon-ssm-agent.service
```diff
@@ -0,0 +1,18 @@
...
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/amazon-ssm-agent/amazon-ssm-agent
```
arnaldo2792: You should install the binary at `/usr/bin` without the `amazon-ssm-agent`

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -35,11 +40,39 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
...
+Provides: %{name}-plugin(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{name}-plugin)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-plugin-bin)
+
+%description plugin-fips-bin
+%{summary}.
+
 %prep
 %setup -n %{gorepo}-%{version}
+%global __scm git
```
arnaldo2792: You should do what you did in socat:

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/8efc145a8cf7f6337ac5219cdb9b3da431be82e5/packages/socat/socat.spec#L16

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -59,12 +92,47 @@ go build -ldflags "${GOLDFLAGS}" -o ssm-session-worker \
...
 %install
-# Install the SSM agent under 'libexecdir', since it is meant to be used by other programs
+install -D -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}/amazon-ssm-agent.service
+
+# Files copied by tmpfiles.d
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -p -m 0644 %{S:3} %{buildroot}%{_cross_tmpfilesdir}/amazon-ssm-agent-tmpfiles.conf
+
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/amazon/ssm
+install -m 0644 %{S:2} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/amazon/ssm/amazon-ssm-agent.json
```
arnaldo2792: As mentioned, no need to include a default config file.

packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
```diff
@@ -0,0 +1,92 @@
...
+-		output.MarkAsFailed(fmt.Errorf("failed to create script file. %v", err))
+-		return
++	var commandName string
++	var commandArguments []string
++
++	if appConfig.Ssm.UseShell {
++
++		// Create script file path
++		scriptPath := filepath.Join(orchestrationDir, p.ScriptName)
++		log.Infof("Writing commands %v to file %v", pluginInput, scriptPath)
```
arnaldo2792: You changed the level of the log.

packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
```diff
@@ -0,0 +1,92 @@
...
++			output.MarkAsFailed(fmt.Errorf("failed to create script file. %v", err))
++			return
++		}
++
++		// Construct Command Name and Arguments
++		commandName = p.ShellCommand
++		commandArguments = append(p.ShellArguments, scriptPath)
++	} else {
++
++		commandInput := strings.Split(pluginInput.RunCommand[0], " ")
```
arnaldo2792, line 76: Can you add a comment here to help other understand that it is OK to just take the first element `RunCommand`?

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -76,15 +144,33 @@ ln -sf %{version} %{buildroot}%{_cross_fips_libexecdir}/amazon-ssm-agent/bin/lat
...
+%{_cross_unitdir}/amazon-ssm-agent.service
+%{_cross_tmpfilesdir}/amazon-ssm-agent-tmpfiles.conf
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/amazon/ssm
+%{_cross_factorydir}%{_cross_sysconfdir}/amazon/ssm/amazon-ssm-agent.json
 
 %files bin
+%dir %{_cross_bindir}/amazon-ssm-agent
+%{_cross_bindir}/amazon-ssm-agent/amazon-ssm-agent
+%{_cross_bindir}/amazon-ssm-agent/ssm-agent-worker
+%{_cross_bindir}/amazon-ssm-agent/ssm-session-worker
```
bcressey: This will break `ecs exec`, because you've moved the plugin binaries it needs to different packages, and haven't updated `ecs-agent.spec` to pull in the plugin.

packages/amazon-ssm-agent/amazon-ssm-agent.spec
bcressey: > As mentioned, no need to include a default config file.

Why not? Does this config file effectively end up just listing what would be the defaults?

packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
koooosh: Great catch! Will revert this back to debug level.

packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
koooosh: Ack

packages/amazon-ssm-agent/amazon-ssm-agent.service
koooosh: Ack

packages/amazon-ssm-agent/amazon-ssm-agent.service
koooosh: Yes, that's what I did originally. But while testing, I found that if I did this, I was not able to SSM into the instance.

However, this was with host-containers enabled, which we determined later is the incorrect setup, so let me try again.

packages/amazon-ssm-agent/amazon-ssm-agent.spec
koooosh: Ack

packages/amazon-ssm-agent/amazon-ssm-agent.spec
koooosh: I see. Originally, I was installing straight to `/usr/bin` instead of its own directory. But while testing, I found that if I did this, I was not able to SSM into the instance.

However, this was with host-containers enabled, which we determined later is the incorrect setup, so let me try again.

packages/amazon-ssm-agent/amazon-ssm-agent.spec
koooosh: My reason for including the custom config `amazon-ssm-agent.json` and the tmpfiles.d to install it was to make it as easy as possible for downstream users to change this config (and our new setting `UseShell`) as necessary.

But I also understand the case that we do not want to include this in the core-kit itself. I'm fine with either approach.

arnaldo2792: > > As mentioned, no need to include a default config file.
> 
> Why not? Does this config file effectively end up just listing what would be the defaults?

My reasoning was that it wouldn't be possible for downstreams to override this configuration if we provide it through the package, right? RPM will complain if the downstreams provide the same file twice (unless you know a magic trick in RPM to prevent that). What I thought was that we could provide the factory file, with reasonable defaults, and let downstreams decide whether they want to use it with a `tmpfiles.d` file. If they want to provide their own, they can provide their own factory and their own `tmpfiles.d` conf file so that they copy the correct file.

packages/amazon-ssm-agent/amazon-ssm-agent.spec
arnaldo2792: What Ben meant is that in the `ecs-agent.spec` file, we currently depend on the default package `amazon-ssm-agent`.

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/e4f9f14c7701338ba4a6a9ee8b1cbf55d57cefca/packages/ecs-agent/ecs-agent.spec#L86

You should update that dependency and use `amazon-ssm-agent-plugin` to pull in the correct dependency.

koooosh: I see, I can make that change in this PR. In that case, would we still need to update the ECS variants in Bob?

packages/amazon-ssm-agent/amazon-ssm-agent.spec
koooosh: >My reasoning was that it wouldn't be possible for downstreams to override this configuration if we provide it through the package, right?

This config file `amazon-ssm-agent.json` is essentially the default configs + our new config for `UseShell`. So if downstreams want override, they can modify this file (or provide their own `amazon-ssm-agent.json`).

>What I thought was that we could provide the factory file

Can you clarify what you mean by "factory file"?

koooosh: Discussed with @arnaldo2792 offline. We are keeping `amazon-ssm-agent.json` in `%{_cross_factorydir}%{_cross_sysconfdir}`. Customers can opt to take this config by adding a tmpfiles.d for that target or create/install their own config.

koooosh:
New commit makes the following major changes:
- Removes `amazon-ssm-agent-tmpfiles.conf`
- Installs agent binaries directly to `/usr/bin/` instead of its own directory
- Updates `ecs-agent-spec` to use agent-plugin package which has the previous static binaries

packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
```diff
@@ -0,0 +1,93 @@
...
+
+diff --git a/agent/appconfig/appconfig.go b/agent/appconfig/appconfig.go
+index b6abcf1..e214cd5 100644
+--- a/agent/appconfig/appconfig.go
++++ b/agent/appconfig/appconfig.go
+@@ -118,6 +118,7 @@ func DefaultConfig() SsmagentConfig {
+ 		SessionLogsRetentionDurationHours:     DefaultSessionLogsRetentionDurationHours,
+ 		PluginLocalOutputCleanup:              DefaultPluginOutputRetention,
+ 		OrchestrationDirectoryCleanup:         DefaultOrchestrationDirCleanup,
++		UseShell:                              false,
```
arnaldo2792, line 21: I worry that we default to `false` but I understand why (no shell by default!).

arnaldo2792 (approved):

packages/amazon-ssm-agent/0001-agent-Add-config-to-make-shell-optional.patch
koooosh: Yeah, I chose to be more conservative in that regard

yeazelm (approved):

PR #23, opened 2024-06-30
enable host containers and in-place updates to be optional

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/twoliter/pull/318

**Description of changes:**
The `signpost` half of this change allows it to work correctly if only one partition bank is available, which will be true for variants that set the `in-place-updates = false` feature flag. `signpost` has limited utility in this mode, but it's still required to mark the boot as successful and to set the "has boot ever succeeded" flag to ensure we only try to assign the `BOTTLEROCKET-DATA` partition label once.

The `os` half of this change is to make it so software related to host containers or in-place updates is only installed if those features are enabled (which they will be by default). This can be done in two ways:
* negative - "Require **this** or **not-that**, and **this** conflicts with **not-that**"
* positive - "Require **this** if **that**, and **this** requires **that**"

These are more or less equivalent, but I went with the "negative" form to avoid requiring a lockstep Twoliter upgrade here and in downstream repos. Otherwise packages built with this change would not install correctly on variants built with an older Twoliter, because the new image feature flags wouldn't be set.

**Testing done:**
Built an `aws-dev` variant without in-place upgrades or host-containers enabled. Built an unmodified `aws-k8s-1.30` variant.

For the `aws-dev` variant, I confirmed that the software wasn't installed and that `signpost` worked as expected.
```
bash-5.2# updog
bash: updog: command not found

bash-5.2# host-ctr
bash: host-ctr: command not found

bash-5.2# systemctl status migrator
Unit migrator.service could not be found.

bash-5.2# signpost status
OS disk: /dev/vda
Set A:   boot=/dev/vda3 root=/dev/vda4 hash=/dev/vda5 priority=1 tries_left=0 successful=true
Active:  Set A
Next:    Set A

bash-5.2# signpost mark-successful-boot
<succeeds>

bash-5.2# signpost clear-inactive
Inactive partition /dev/vda is not available

bash-5.2# signpost upgrade-to-inactive
Inactive partition /dev/vda is not available

bash-5.2# signpost cancel-upgrade
Inactive partition /dev/vda is not available

bash-5.2# signpost rollback-to-inactive
Inactive partition /dev/vda is not available

bash-5.2# signpost has-boot-ever-succeeded
true
```

For the `aws-k8s-1.30` variant, I upgraded and downgraded across versions and confirmed that worked as expected.

```
<before upgrade>
# signpost status
OS disk: /dev/nvme0n1
Set A:   boot=/dev/nvme0n1p8 root=/dev/nvme0n1p9 hash=/dev/nvme0n1p10 priority=2 tries_left=1 successful=false
Set B:   boot=/dev/nvme0n1p3 root=/dev/nvme0n1p4 hash=/dev/nvme0n1p5 priority=1 tries_left=0 successful=true
Active:  Set B
Next:    Set A

<after upgrade>
# signpost status
OS disk: /dev/nvme0n1
Set A:   boot=/dev/nvme0n1p3 root=/dev/nvme0n1p4 hash=/dev/nvme0n1p5 priority=2 tries_left=0 successful=true
Set B:   boot=/dev/nvme0n1p8 root=/dev/nvme0n1p9 hash=/dev/nvme0n1p10 priority=1 tries_left=0 successful=true
Active:  Set A
Next:    Set A
```

**Terms of contribution:**

### DIFF
--- a/packages/os/storewolf.service
+++ b/packages/os/storewolf.service
@@ -1,7 +1,5 @@
 [Unit]
 Description=Datastore creator
-After=migrator.service
-Requires=migrator.service
 # Block manual interactions with this service, since it could leave the system in an
 # unexpected state
 RefuseManualStart=true
--- a/sources/updater/signpost/src/state.rs
+++ b/sources/updater/signpost/src/state.rs
@@ -20,13 +20,29 @@ pub struct State {
     os_disk: PathBuf,
     // BOTTLEROCKET_PRIVATE partition number
     private_partition_num: u32,
-    sets: [PartitionSet; 2],
+    sets: PartitionSets,
     /// The partition numbers that correspond to the boot partitions in each partition set,
     /// respectively.
     ///
     /// This is used to load the correct partition flags from `table`.
-    boot_partition_nums: [u32; 2],
+    boot_partition_nums: Vec<u32>,
     table: GPT,
+}
+
+#[derive(Debug, Clone)]
+enum PartitionSets {
+    Single(SinglePartitionSet),
+    Dual(DualPartitionSet),
+}
+
+#[derive(Debug, Clone)]
+struct SinglePartitionSet {
+    set: PartitionSet,
+}
+
+#[derive(Debug, Clone)]
+struct DualPartitionSet {
+    sets: [PartitionSet; 2],
     active: SetSelect,
 }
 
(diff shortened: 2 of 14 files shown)


bcressey:
@cbgbt reviewed an earlier version of the `signpost` changes and suggested using an enum to represent the different possible partition sets, which I've done.

packages/os/storewolf.service
```diff
@@ -1,7 +1,5 @@
 [Unit]
 Description=Datastore creator
-After=migrator.service
```
yeazelm, line 3: I didn't catch where this dependency is tracked now? Do we no longer need to ensure `migrator` runs before `storewolf`?

arnaldo2792 (approved):

sources/updater/signpost/src/state.rs
```diff
@@ -103,44 +123,56 @@ impl State {
...
+                SetSelect::B
+            } else {
+                return error::ActiveNotInSetSnafu {
+                    active_partition,
+                    sets,
+                }
+                .fail();
+            };
+            PartitionSets::Dual(DualPartitionSet {
+                sets: sets.try_into().unwrap(),
```
yeazelm: Is there a way to avoid a straight `unwrap`?

yeazelm (approved):

packages/os/storewolf.service
bcressey: It moved to `RequiredBy` in [migrator.service](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/23/files#diff-aeaa16170f631bfaab1d68685d57d5bbc041cb32330500091950a24dbfe8fdf8).

The problem with listing it here is that - if `migrator.service` is not installed - then `storewolf.service` will fail to start because a required service is missing. So we need to invert the dependencies to make it optional.

sources/updater/signpost/src/state.rs
bcressey: Here we know that the vec has two elements and can be converted to a two-element array. I can convert it to an error though. I suppose technically nothing stops anyone from adding or removing another element to the vec inside the branch before this call.

packages/os/storewolf.service
yeazelm: Ok, that makes sense!

yeazelm (approved):

PR #25, opened 2024-07-02
build a kube proxy binary package

### LINKED ISSUE #16: Build `kube-proxy` as part of Kubernetes Package Builds
**What I'd like:**
I'd like to be able to use `kube-proxy` built from the same kubernetes sources as `kubelet`.

This should be as simple as issuing an additional build command in the `kubernetes-*` packages [here](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/8efc145a8cf7f6337ac5219cdb9b3da431be82e5/packages/kubernetes-1.26/kubernetes-1.26.spec#L129):

```
make WHAT="cmd/kube-proxy"
```

And then packaging these into their own RPM. I'd taken a stab at this, but struggled to get the FIPS check to pass, so there may be a bit more to it.

tzneal:
**Issue number:**

Closes #16 

**Description of changes:**

Provides a kube-proxy-1.XX package that is built from the same source as kubelet.

**Testing done:**

Built & launched 1.29 and 1.30 AMIs and verified the installation:

```
bash-5.1# /usr/bin/kubelet --version
Kubernetes v1.29.1-eks-61c0bbb

bash-5.1# /usr/bin/kube-proxy --version
Kubernetes v1.29.1-eks-61c0bbb


bash-5.1# /usr/bin/kubelet --version
Kubernetes v1.30.0-eks-fff26e3

bash-5.1# /usr/bin/kube-proxy --version
Kubernetes v1.30.0-eks-fff26e3
```

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.23/kubernetes-1.23.spec
+++ b/packages/kubernetes-1.23/kubernetes-1.23.spec
@@ -103,6 +103,32 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}kubelet-1.23-bin)
 %description -n %{_cross_os}kubelet-1.23-fips-bin
 %{summary}.
 
+%package -n %{_cross_os}kube-proxy-1.23
+Summary: Container cluster node proxy
+Requires: %{_cross_os}kubelet-1.23
+Requires: %{_cross_os}kube-proxy-1.23(binaries)
+
+%description -n %{_cross_os}kube-proxy-1.23
+%{summary}.
+
+%package -n %{_cross_os}kube-proxy-1.23-bin
+Summary: Container cluster node proxy binaries
+Provides: %{_cross_os}kube-proxy-1.23(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}kube-proxy-1.23)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}kube-proxy-1.23-fips-bin)
+
+%description -n %{_cross_os}kube-proxy-1.23-bin
+%{summary}.
+
+%package -n %{_cross_os}kube-proxy-1.23-fips-bin
+Summary: Container cluster node proxy binaries, FIPS edition
+Provides: %{_cross_os}kube-proxy-1.23(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}kube-proxy-1.23)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}kube-proxy-1.23-bin)
+
+%description -n %{_cross_os}kube-proxy-1.23-fips-bin
+%{summary}.
+
 %prep
 %autosetup -Sgit -c -n %{gorepo}-%{gover} -p1
 
@@ -118,18 +144,22 @@ export FORCE_HOST_GO=1
 # Build codegen programs with the host toolchain.
 make generated_files
 
-# Build kubelet with the target toolchain.
+# Build kubelet and kube-proxy with the target toolchain.
 %set_cross_go_flags
 unset CC
 export KUBE_BUILD_PLATFORMS="linux/%{_cross_go_arch}"
 export %{kube_cc}
 export GOFLAGS="${GOFLAGS} -tags=dockerless"
 export GOLDFLAGS="${GOLDFLAGS}"
+# don't build kube-proxy statically as we use linkermode=external which requires CGO
+export KUBE_CGO_OVERRIDES="kube-proxy"
 make WHAT="cmd/kubelet"
+make WHAT="cmd/kube-proxy"
 
 export KUBE_OUTPUT_SUBPATH="_fips_output/local"
 export GOEXPERIMENT="boringcrypto"
 make WHAT="cmd/kubelet"
+make WHAT="cmd/kube-proxy"
 
 # build the pause container
 cd build/pause/linux/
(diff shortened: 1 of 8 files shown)


packages/kubernetes-1.23/kubernetes-1.23.spec
```diff
@@ -103,6 +103,35 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}kubelet-1.23-bin)
 %description -n %{_cross_os}kubelet-1.23-fips-bin
 %{summary}.
 
+%package -n %{_cross_os}kube-proxy-1.23
+Summary: Container cluster node proxy
+Requires: %{_cross_os}conntrack-tools
+Requires: %{_cross_os}containerd
+Requires: %{_cross_os}ecr-credential-provider-1.23
+Requires: %{_cross_os}aws-signing-helper
+Requires: %{_cross_os}kube-proxy-1.23(binaries)
```
bcressey, line 109: Is there a scenario where it makes sense to install `kube-proxy` but not `kubelet`? If not, you could bind them together, and then you would only need one copy of the license files:
```suggestion
Requires: %{_cross_os}kubelet-1.23
Requires: %{_cross_os}kube-proxy-1.23(binaries)
```

You would also not need to list most of the other dependencies, as they'd be pulled in transitively by the `kubelet` dependency.

packages/kubernetes-1.23/kubernetes-1.23.spec
```diff
@@ -103,6 +103,35 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}kubelet-1.23-bin)
 %description -n %{_cross_os}kubelet-1.23-fips-bin
 %{summary}.
 
+%package -n %{_cross_os}kube-proxy-1.23
+Summary: Container cluster node proxy
+Requires: %{_cross_os}conntrack-tools
+Requires: %{_cross_os}containerd
+Requires: %{_cross_os}ecr-credential-provider-1.23
```
bcressey: What provides these versioned `ecr-credential-provider` packages?

tzneal: Wishful thinking :) I only tested 1.29 and 1.30 which had versioned providers.

packages/kubernetes-1.23/kubernetes-1.23.spec
tzneal: Yeah, that makes sense. I don't see it ever being installed without kubelet.

yeazelm (approved):

packages/kubernetes-1.23/kubernetes-1.23.spec
```diff
@@ -229,4 +261,12 @@ install -p -m 0644 %{S:102} %{buildroot}%{_cross_templatedir}/pod-infra-containe
 %files -n %{_cross_os}kubelet-1.23-fips-bin
 %{_cross_fips_bindir}/kubelet
 
+%files -n %{_cross_os}kube-proxy-1.23
```
monirul, line 264: Is this intentional?
If I am not mistaken, when there’s no list of files or directories after the %files directive, it typically means all files and directories that have been installed to the buildroot during the %install phase are automatically included

tzneal: Without the `%files` directive, the package itself wasn't created.  With this, its an empty package with only dependencies:

```
% rpm -qlp build/rpms/kubernetes-1.28/bottlerocket-kube-proxy-1.28-1.28.7-1.1719972209.77427686.br1.x86_64.rpm
(contains no files)

% rpm -qpR build/rpms/kubernetes-1.28/bottlerocket-kube-proxy-1.28-1.28.7-1.1719972209.77427686.br1.x86_64.rpm
bottlerocket-kube-proxy-1.28(binaries)
bottlerocket-kubelet-1.28
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
```

arnaldo2792: @monirul , this is intentional. In the FIPS world, the new way to compile go packages is by having the main package provide either systemd services, configurations or `tmpfiles.d` drop ins, and use the image feature flag `image-feature-fips`/`image-feature-no-fips` to control which binary is actually installed when the image is built.

packages/kubernetes-1.23/kubernetes-1.23.spec
```diff
@@ -103,6 +103,32 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}kubelet-1.23-bin)
...
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}kube-proxy-1.23)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}kube-proxy-1.23-fips-bin)
+
+%description -n %{_cross_os}kube-proxy-1.23-bin
+%{summary}.
+
+%package -n %{_cross_os}kube-proxy-1.23-fips-bin
+Summary: Container cluster node proxy binaries, FIPS edition
+Provides: %{_cross_os}kube-proxy-1.23(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}kube-proxy-1.23)
```
arnaldo2792, line 126: One question, how will downstreams use this? Do they have to start this as a systemd service or will this binary be found by the `kubelet` somehow and it will call it? If the former, will it make sense to provide the systemd service that starts `kube-proxy`?

arnaldo2792 (approved):

PR #29, opened 2024-07-12
Update kernels

arnaldo2792:
**Issue number:**

N / A

**Description of changes:**
Update kernels to the latest upstream versions:

- 5.10.220
- 5.15.161
- 6.1.96

**Patch changes**

- 5.10 changed one patch:
    - 0085-NFS-Clean-up-nfs_readdir_page_filler.patch
- 5.10 added one patch:
    - 0866-nfs-Leave-pages-in-the-pagecache-if-readpage-failed.patch
- 5.15 changed seven patches:
    - 0103-mptcp-add-new-mptcp_fill_diag-helper.patch
    - 0104-mptcp-add-MPTCP_INFO-getsockopt.patch
    - 0111-mptcp-Support-for-IP_TOS-for-MPTCP-setsockopt.patch
    - 0112-mptcp-sockopt-add-SOL_IP-freebind-transparent-option.patch
    - 0114-mptcp-add-TCP_INQ-cmsg-support.patch
    - 0118-mptcp-getsockopt-add-support-for-IP_TOS.patch
    - 0122-mptcp-support-TCP_CORK-and-TCP_NODELAY.patch
- 5.15 added one patch:
    - 0158-nfs-Leave-pages-in-the-pagecache-if-readpage-failed.patch
 - 6.1 added one patch:
     - 0176-cifs-use-origin-fullpath-for-automounts.patch

**Configuration Changes**

Summary:
```
config-aarch64-5.10-diff:	  0 removed,   0 added,   0 changed
config-aarch64-5.15-diff:	  0 removed,   0 added,   0 changed
config-aarch64-6.1-diff:	  0 removed,   0 added,   0 changed
config-x86_64-5.10-diff:	  2 removed,   2 added,   0 changed
config-x86_64-5.15-diff:	  0 removed,   0 added,   0 changed
config-x86_64-6.1-diff:	  1 removed,   5 added,   6 changed
```

Report:

```
==> /tmp/configs//config-aarch64-5.10-diff <==

==> /tmp/configs//config-aarch64-5.15-diff <==

==> /tmp/configs//config-aarch64-6.1-diff <==

==> /tmp/configs//config-x86_64-5.10-diff <==
-NFSD_V2_ACL y
-NFSD_V3 y
+NFSD_V2 n
+NFS_V4_2_SSC_HELPER y

==> /tmp/configs//config-x86_64-5.15-diff <==

==> /tmp/configs//config-x86_64-6.1-diff <==
-TICK_CPU_ACCOUNTING y
[...]

PR #32, opened 2024-07-13
diff-kernel-config: fix for core-kits

arnaldo2792:
**Issue number:**

N / A 

**Description of changes:**

The script referenced variants, which aren't a thing in the core-kit world

**Testing done:**

I ran the script with the changes in: #29 , and the report was generated:

```bash
config-aarch64-5.10-diff:	  0 removed,   0 added,   0 changed
config-aarch64-5.15-diff:	  0 removed,   0 added,   0 changed
config-aarch64-6.1-diff:	  0 removed,   0 added,   0 changed
config-x86_64-5.10-diff:	  2 removed,   2 added,   0 changed
config-x86_64-5.15-diff:	  0 removed,   0 added,   0 changed
config-x86_64-6.1-diff:	  1 removed,   5 added,   6 changed
```

**Terms of contribution:**

### DIFF
--- a/tools/diff-kernel-config
+++ b/tools/diff-kernel-config
@@ -36,9 +36,7 @@ Compare kernel configurations before and after a series of commits.
 
     -a, --after         new Git revision to compare from
     -b, --before        baseline Git revision to compare against
-    -v, --variant       variant to pick kernel from to compare configs for, may
-                        be given multiple times (optional, defaults to this list:
-                        'aws-k8s-1.23', 'metal-k8s-1.23', 'aws-dev', 'metal-dev')
+    -k, --kernel        the kernel versions
     -o, --output-dir    path to the output directory; must not exist yet
     -r, --resume        resume work on a previous invocation; check which parts
                         already exist in OUTPUT_DIR and skip builds accordingly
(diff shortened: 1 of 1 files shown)


larvacea (approved):
Thanks! Looks great, and it sounds like you got the right answer when you tried it.

tools/diff-kernel-config
```diff
@@ -114,6 +101,7 @@ readonly output_dir
 # checkout.
 [[ -n ${gitrev_before_arg} ]] || usage_error 'require -b|--before'
 [[ -n ${gitrev_after_arg} ]] || usage_error 'require -a|--after'
+[[ -n ${kernel_versions[@]} ]] || usage_error 'require -k|--kernel'
```
koooosh: Builds are failing with this error:
```
diff-kernel-config:104:7: error: Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @). [SC2199]
```

I believe this line is the culprit. Ways to fix this: https://www.shellcheck.net/wiki/SC2199

koooosh (changes requested):

arnaldo2792:
Forced push includes fixes for shellchecks

ginglis13 (approved):

yeazelm (approved):

koooosh (approved):

PR #37, opened 2024-07-16
Add advisories for 2.1.0 and 2.2.0 releases

ginglis13:
**Description of changes:**

Add Bottlerocket Security Advisories for the 2.1.0 release of the bottlerocket-core-kit. This release included patches for the kernels. Also, add the initial `staging` directory in the `advisories` dir to be used for staging advisories ahead of an official release of the core kit.

**Testing done:**

n/a

**Terms of contribution:**

### DIFF
--- a/advisories/2.1.0/BRSA-1j0O73qa.toml
+++ b/advisories/2.1.0/BRSA-1j0O73qa.toml
@@ -1,0 +1,16 @@
+[advisory]
+id = "BRSA-1j0O73qa"
+title = "kernel CVE-2024-36933"
+cve = "CVE-2024-36933"
+severity = "moderate"
+description = "In the Linux kernel, the following vulnerability has been resolved: -  - nsh: Restore skb->{protocol,data,mac_header} for outer header in nsh_gso_segment()."
+
+[[advisory.products]]
+package-name = "kernel-5.15"
+patched-version = "kernel-5.15.160-104.158.amzn2"
+
+[updateinfo]
+author = "giinglis"
+issue-date = 2024-07-08T23:14:32Z
+arches = ["aarch64", "x86_64"]
+version = "2.1.0"
--- a/advisories/2.1.0/BRSA-w8ZsgmV2.toml
+++ b/advisories/2.1.0/BRSA-w8ZsgmV2.toml
@@ -1,0 +1,20 @@
+[advisory]
+id = "BRSA-w8ZsgmV2"
+title = "kernel CVE-2024-36971"
+cve = "CVE-2024-36971"
+severity = "moderate"
+description = "In the Linux kernel, the following vulnerability has been resolved: -  - net: fix __dst_negative_advice() race"
+
+[[advisory.products]]
+package-name = "kernel-5.10"
+patched-version = "kernel-5.10.219-208.866.amzn2"
+
+[[advisory.products]]
+package-name = "kernel-6.1"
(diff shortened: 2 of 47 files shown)


advisories/2.1.0/BRSA-1j0O73qa.toml
```diff
@@ -0,0 +1,16 @@
+[advisory]
+id = "BRSA-1j0O73qa"
```
bcressey: One issue that didn't occur to me until seeing it in this context - we might want to restrict ourselves to lowercase letters as otherwise there could be filename collisions or general weirdness on case-insensitive filesystems.

advisories/2.1.0/BRSA-w8ZsgmV2.toml
```diff
@@ -0,0 +1,20 @@
...
+severity = "moderate"
+description = "In the Linux kernel, the following vulnerability has been resolved: -  - net: fix __dst_negative_advice() race"
+
+[[advisory.products]]
+package-name = "kernel-5.10"
+patched-version = "kernel-5.10.219-208.866.amzn2"
+
+[[advisory.products]]
+package-name = "kernel-6.1"
+patched-version = "kernel-6.1.94-99.176.amzn2023"
```
bcressey: I'm surprised there's just one instance of a CVE that applies to multiple kernels. I'd expect that to be more common if not the majority of cases.

advisories/2.1.0/BRSA-1j0O73qa.toml
ginglis13: Fair enough; I can make that adjustment

advisories/2.1.0/BRSA-w8ZsgmV2.toml
ginglis13: Agreed, so I verified myself manually - this is has become more common as the number of kernel CVEs began to increase; also, checking that past GHSAs on bottlerocket itself (specifically, those that came with the 1.20.3 release) also indicate this being the new norm. Only 3 of the 17 kernel advisories affected more than one kernel series: https://github.com/bottlerocket-os/bottlerocket/security/advisories

advisories/2.1.0/BRSA-1j0O73qa.toml
cbgbt: We should probably subsequently increase the length of the advisory IDs to reduce collision risk.

ginglis13: Our chances of collision would increase from ~ 0.000000002% to ~ 0.0000002%. It is 2 orders of magnitude increase, but still extremely low. Even at 10,000 advisories over the lifetime of the project, the odds are ~0.00002% for ID collision. 

I think I'd prefer some sort of CI action on the advisories dir that would detect a collision for us

cbgbt: Ah, I just remembered doing the math previously and was worried about invalidating our assumptions. But when we did the math, we had been talking about 4/5-digit IDs and not 8. I think you're right, that should be fine.

ginglis13:
^ adjust BRSA IDs / the filenames to use lowercase chars

ginglis13:
^ include kernel BRSAs for 2.2.0 release

vigh-m (approved):

bcressey (approved):

PR #41, opened 2024-07-19
pluto: move hostname variant prevention to the correct place

jmt-lab:
**Description of changes:**

* The NO_HOSTNAME_VARIANTS check was supposed to be in the generate_private_dns_name not the provider_id method. This will make it so `hostname-override` is only generated/set 

**Terms of contribution:**


bcressey (approved):
Looks good to me for Bottlerocket's `aws-k8s-*` variants.

yeazelm (approved):

PR #42, opened 2024-07-22
add a kmod-6.1-neuron-devel package with header files

tzneal:
**Issue number:**

Closes #

**Description of changes:**

Adds a neuron-devel package with a few of the headers from the kernel driver source that are used when building with Neuron.

**Testing done:**

Build a dependent package that consumed these headers.

**Terms of contribution:**

### DIFF
--- a/packages/kmod-6.1-neuron/kmod-6.1-neuron.spec
+++ b/packages/kmod-6.1-neuron/kmod-6.1-neuron.spec
@@ -15,6 +15,13 @@ BuildRequires: %{_cross_os}kernel-6.1-archive
 %description
 %{summary}.
 
+%package devel
+Summary: Files for development using the Neuron drivers
+Requires: %{name}
+
+%description devel
+%{summary}.
+
 %prep
 rpm2cpio %{SOURCE0} | cpio -idmv
 tar -xf %{_cross_datadir}/bottlerocket/kernel-devel.tar.xz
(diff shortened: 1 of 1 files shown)


packages/kmod-6.1-neuron/kmod-6.1-neuron.spec
```diff
@@ -51,9 +58,19 @@ install -p -m 0644 %{S:1} %{buildroot}%{_cross_libdir}/modules-load.d/neuron.con
...
 %license %{neuron_sources}/LICENSE
 %{_cross_attribution_file}
 %{_cross_libdir}/modules/*/extra/neuron.ko.gz
 %{_cross_libdir}/modules-load.d/neuron.conf
 %{_cross_unitdir}/systemd-modules-load.service.d/neuron.conf
+
+%files devel
+%dir %{_cross_includedir}/kmod-6.1-neuron
+%{_cross_includedir}/kmod-6.1-neuron/share/neuron_driver_shared.h
+%{_cross_includedir}/kmod-6.1-neuron/neuron_ioctl.h
```
bcressey: Is this where downstream builds expect the headers?

You could put them in the normal include dirs:
```suggestion
%{_cross_includedir}/share/neuron_driver_shared.h
%{_cross_includedir}/neuron_ioctl.h
```

bcressey (approved):

packages/kmod-6.1-neuron/kmod-6.1-neuron.spec
tzneal: I'll move them, I didn't know if BR wants to separate headers by package or not.

yeazelm (approved):

ginglis13 (approved):

PR #48, opened 2024-07-23
Kit changes for Nvidia settings APIs

monirul:
**Issue number:**

Closes #

**Description of changes:**
This PR contains the changes required to support Nvidia settings APIs, `settings.nvidia-container-runtime` and `settings.kubernetes.nvidia`

**Testing done:**
Yes.
```
bash-5.1# apiclient set settings.kubernetes.device-plugins.nvidia.pass-device-specs=true
bash-5.1# cat /etc/nvidia-k8s-device-plugin/settings.yaml
version: v1
flags:
  migStrategy: "none"
  failOnInitError: true
  plugin:
    passDeviceSpecs: true
    deviceListStrategy: volume-mounts
    deviceIDStrategy: index
bash-5.1# apiclient set settings.kubernetes.device-plugins.nvidia.device-list-strategy=envvar
bash-5.1# cat /etc/nvidia-k8s-device-plugin/settings.yaml
version: v1
flags:
  migStrategy: "none"
  failOnInitError: true
  plugin:
    passDeviceSpecs: true
    deviceListStrategy: envvar
    deviceIDStrategy: index
bash-5.1# apiclient set settings.kubernetes.device-plugins.nvidia.device-id-strategy=uuid
bash-5.1# cat /etc/nvidia-k8s-device-plugin/settings.yaml
version: v1
flags:
  migStrategy: "none"
  failOnInitError: true
  plugin:
    passDeviceSpecs: true
    deviceListStrategy: envvar
    deviceIDStrategy: uuid

bash-5.1# cat /etc/nvidia-container-runtime/config.toml
accept-nvidia-visible-devices-as-volume-mounts = true
accept-nvidia-visible-devices-envvar-when-unprivileged = false

[nvidia-container-cli]
root = "/"
environment = []
ldconfig = "@/sbin/ldconfig"
bash-5.1# apiclient set settings.nvidia-container-runtime.visible-devices-as-volume-mounts=false
bash-5.1# apiclient set settings.nvidia-container-runtime.visible-devices-envvar-when-unprivileged=true
bash-5.1# cat /etc/nvidia-container-runtime/config.toml
accept-nvidia-visible-devices-as-volume-mounts = false
accept-nvidia-visible-devices-envvar-when-unprivileged = true

[nvidia-container-cli]
root = "/"
environment = []
ldconfig = "@/sbin/ldconfig"
```

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
@@ -1,0 +1,13 @@
+[required-extensions]
+kubernetes = "v1"
+std = { version = "v1", helpers = ["default"] }
+
++++
+version: v1
+flags:
+  migStrategy: "none"
+  failOnInitError: true
+  plugin:
+    passDeviceSpecs: {{default true settings.kubernetes.device-plugins.nvidia.pass-device-specs}}
+    deviceListStrategy: {{default "volume-mounts" settings.kubernetes.device-plugins.nvidia.device-list-strategy}}
+    deviceIDStrategy: {{default "index" settings.kubernetes.device-plugins.nvidia.device-id-strategy}}
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-exec-start-conf
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-exec-start-conf
@@ -1,0 +1,8 @@
+[required-extensions]
+kubernetes = "v1"
++++
+[Service]
+{{#if settings.kubernetes.device-plugins.nvidia}}
+ExecStart=
+ExecStart=/usr/bin/nvidia-device-plugin --config-file=/etc/nvidia-k8s-device-plugin/settings.yaml
+{{/if}}
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
@@ -1,7 +1,5 @@
 [Unit]
 Description=Start NVIDIA kubernetes device plugin
-RefuseManualStart=true
-RefuseManualStop=true
 After=kubelet.service
 Wants=kubelet.service
 
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
@@ -1,0 +1,14 @@
+[required-extensions]
+nvidia-container-runtime = "v1"
+std = { version = "v1", helpers = ["default"] }
+
++++
+### generated from the template file ###
+accept-nvidia-visible-devices-as-volume-mounts = {{default true settings.nvidia-container-runtime.visible-devices-as-volume-mounts}}
+accept-nvidia-visible-devices-envvar-when-unprivileged = {{default false settings.nvidia-container-runtime.visible-devices-envvar-when-unprivileged}}
+
+[nvidia-container-cli]
+root = "/"
+path = "/usr/bin/nvidia-container-cli"
+environment = []
+ldconfig = "@/sbin/ldconfig"
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
@@ -19,6 +19,7 @@ Source3: nvidia-oci-hooks-json
 Source4: nvidia-gpu-devices.rules
 Source5: nvidia-container-toolkit-tmpfiles-ecs.conf
 Source6: nvidia-container-toolkit-tmpfiles-k8s.conf
+Source7: nvidia-container-toolkit-config-k8s
 
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}libnvidia-container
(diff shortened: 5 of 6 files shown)


packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -0,0 +1,11 @@
+[required-extensions]
+kubernetes = "v1"
++++
+version: v1
+flags:
+  migStrategy: "none"
+  failOnInitError: true
+  plugin:
+    passDeviceSpecs: {{default true settings.kubernetes.nvidia.device-plugin.pass-device-specs}}
```
cbgbt: In order to use `default`, you need to "import" it from `std` by adding it to the `required-extensions`.


```
std = { version = "v1", helpers = ["default"] }
```

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -0,0 +1,11 @@
...
+    deviceListStrategy: "{{default "volume-mounts" settings.kubernetes.nvidia.device-plugin.device-list-strategy}}"
+    deviceIDStrategy: "{{default "index" settings.kubernetes.nvidia.device-plugin.device-id-strategy}}"
```
cbgbt: Will this result in e.g. "volume-mounts" being double-quoted if the setting is unset?

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
```diff
@@ -1,12 +1,10 @@
 [Unit]
 Description=Start NVIDIA kubernetes device plugin
-RefuseManualStart=true
-RefuseManualStop=true
 After=kubelet.service
 Wants=kubelet.service
 
 [Service]
-ExecStart=/usr/bin/nvidia-device-plugin --device-list-strategy volume-mounts --device-id-strategy index --pass-device-specs=true
+ExecStart=/usr/bin/nvidia-device-plugin --config-file=/etc/nvidia-k8s-device-plugin/settings.yaml
```
cbgbt: You mentioned when chatting that we put a default configuration file in place to keep compatibility with downstream variants that won't add the device plugin settings. That makes sense.

An alternative to overwriting a config tempfile could be to place a systemd dropin that overwrites `ExecStart` based on settings changes. So this file would remain the same, but a template could be rendered to `/etc/sysetmd/system/nvidia-k8s-device-plugin.service.d/exec-start` adding the alternative line with a reference to the rendered config file.

I know we do something like this for `kubelet`, though I'm not sure if there are compelling reasons to prefer one to the other. I suppose anything that gets rendered to `/etc` consumes system memory, so perhaps it's slightly nicer to avoid doing that if we can.

packages/nvidia-container-toolkit/nvidia-container-toolkit-tmpfiles-k8s.conf
```diff
@@ -1 +1 @@
-C /etc/nvidia-container-runtime/config.toml - - - - /usr/share/factory/nvidia-container-runtime/nvidia-container-toolkit-config-k8s.toml
+d /etc/nvidia-container-runtime - - - - -
```
cbgbt: This shouldn't need to be created by tmpfiles.d because the template rendering system will make the directory on our behalf.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-exec-start-conf
```diff
@@ -0,0 +1,8 @@
+[required-extensions]
+kubernetes = "v1"
++++
+[Service]
+{{#if settings.kubernetes.device-plugins.nvidia}}
```
cbgbt, line 5: Did you test that this works correctly on instances without an nvidia device plugin set?

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-tmpfiles.conf
```diff
@@ -0,0 +1,2 @@
+d /etc/nvidia-k8s-device-plugin - - - - -
```
cbgbt: I believe you could similarly drop the tmpfiles requirement here since both files are templates.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
monirul: added

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
monirul: yes.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
monirul: changed as suggested

packages/nvidia-container-toolkit/nvidia-container-toolkit-tmpfiles-k8s.conf
monirul: removed the tmpfiles as suggested.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-exec-start-conf
monirul: Yes. I tested it if `settings.kubernetes.device-plugins.nvidia` is not defined then this file is rendered as 
```
[Service]
```
which are kind of ignored as there is no overriding settings. It works as expected.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-tmpfiles.conf
monirul: deleted as suggested.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
```diff
@@ -1,7 +1,5 @@
 [Unit]
 Description=Start NVIDIA kubernetes device plugin
-RefuseManualStart=true
-RefuseManualStop=true
```
yeazelm, line 4: Are we now manually starting/stopping this?

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-exec-start-conf
yeazelm: This feels weird and to someone just seeing this on the system would appear to be a failure? Can we add a comment in an else statement to say deliberately left this way or something? Is it not better to just remove the `[Service]` as well?

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
bcressey: > Are we now manually starting/stopping this?

We need to be able to restart it when settings change.

bcressey (approved):

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
yeazelm: Ok, that makes sense, I just wanted to make sure that was the right conclusion I made when reading this.

yeazelm (approved):

PR #50, opened 2024-07-24
Add aws-otel-collector

### LINKED ISSUE #52: Add aws-otel-collector package to core-kit
**What I'd like:**
For [AWS OTEL collector](https://github.com/aws-observability/aws-otel-collector/tree/main) to be added as a 3rd party package to the Bottlerocket Core-kit for downstream use cases.

N/a

koooosh:
**Issue number:** #52 

Closes #52 

**Description of changes:**

This change adds the `aws-otel-collector` package to the core-kit to be used by our downstream customers. Specifically, it includes the following:

- `aws-otel-collector.spec` which provides non-FIPS and FIPS versions of the collector binary
- `Cargo.toml` for the new package
- `aws-otel-collector.service` which is the systemd service for the collector. This is similar to [the default service](https://github.com/aws-observability/aws-otel-collector/blob/main/tools/packaging/linux/aws-otel-collector.service) with the following changes:
    - Removing the `EnvironmentFile` field and providing the config path directly in `ExecStart`. This way if downstream customers want to provide their own config, they just have to add a drop-in unit updating the `ExecStart` field.
    - Changing `Restart` from `on-failure` to `always` just for safety
    - Reducing the `RestartSec` from 60s to 5s
    - Removing the `User` and `Group` fields since they are unnecessary
    - Changing `network.target` to `network-online.target configured.target` in relevant settings
- `aws-otel-collector.yaml` which is a minimal config needed for the collector to run
- `aws-otel-collector-tmpfiles.conf` to install above config in `/etc` rather than the default `/opt`
- `0001-change-logger-and-extraconfig-file-paths.patch` to change the log and extraconfig file paths from `/opt` to `/var/log` and `/etc`, respectively
-  Updates to the `bottlerocket-core-kit/Cargo.toml` and root `Cargo.toml` and `Cargo.lock` to add the aws-otel-collector to the kit

**Testing done:**

Launched an `aws-dev` variant instance with the `aws-otel-collector` package added and verified the following:

- Collector service is running successfully:
```bash
bash-5.2# systemctl status aws-otel-collector
● aws-otel-collector.service - AWS OTEL collector
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/aws-otel-collector.service; enabled; preset: enabled)
     Active: active (running) since Wed 2024-07-24 07:24:58 UTC; 14h ago
   Main PID: 1346 (aws-otel-collec)
      Tasks: 9 (limit: 18599)
     Memory: 134.7M
        CPU: 13.451s
     CGroup: /system.slice/aws-otel-collector.service
             └─1346 /usr/bin/aws-otel-collector --config /etc/aws-otel-collector-default-config.yaml
```

- Config file is installed correctly in `/etc`:
```
bash-5.2# ls /etc/ | grep otel
aws-otel-collector-default-config.yaml
```

- Log file is installed in correct location in `/var/log`:
```
bash-5.2# ls /var/log/aws/
aws-otel-collector.log
```

**Terms of contribution:**

### DIFF
--- a/packages/aws-otel-collector/0001-change-logger-and-extraconfig-file-paths.patch
+++ b/packages/aws-otel-collector/0001-change-logger-and-extraconfig-file-paths.patch
@@ -1,0 +1,40 @@
+From 1f1fc80e5869e125f5900aca76c3dbd3f1a24d81 Mon Sep 17 00:00:00 2001
+From: Kush Upadhyay <kushupad@amazon.com>
+Date: Thu, 15 Aug 2024 09:12:55 +0000
+Subject: [PATCH] change logger and extraconfig file paths
+
+Signed-off-by: Kush Upadhyay <kushupad@amazon.com>
+---
+ pkg/extraconfig/extraconfig.go | 2 +-
+ pkg/logger/logger.go           | 2 +-
+ 2 files changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/pkg/extraconfig/extraconfig.go b/pkg/extraconfig/extraconfig.go
+index 9c14d5f7..8ee7cbd9 100644
+--- a/pkg/extraconfig/extraconfig.go
++++ b/pkg/extraconfig/extraconfig.go
+@@ -23,7 +23,7 @@ import (
+ )
+ 
+ var (
+-	UnixExtraConfigPath    = "/opt/aws/aws-otel-collector/etc/extracfg.txt"
++	UnixExtraConfigPath    = "/etc/aws-otel-collector/extracfg.txt"
+ 	WindowsExtraConfigPath = "C:\\ProgramData\\Amazon\\AWSOTelCollector\\Configs\\extracfg.txt"
+ )
+ 
+diff --git a/pkg/logger/logger.go b/pkg/logger/logger.go
+index 9805eeab..7fcbab31 100644
+--- a/pkg/logger/logger.go
++++ b/pkg/logger/logger.go
+@@ -31,7 +31,7 @@ import (
+ )
+ 
+ var (
+-	UnixLogPath      = "/opt/aws/aws-otel-collector/logs/aws-otel-collector.log"
++	UnixLogPath      = "/var/log/aws/aws-otel-collector.log"
+ 	WindowsLogPath   = "C:\\ProgramData\\Amazon\\AWSOTelCollector\\Logs\\aws-otel-collector.log"
+ 	logfile          = getLogFilePath()
+ 	lumberjackLogger = tryNewLumberJackLogger()
+-- 
+2.40.1
+
--- a/packages/aws-otel-collector/aws-otel-collector-tmpfiles.conf
+++ b/packages/aws-otel-collector/aws-otel-collector-tmpfiles.conf
@@ -1,0 +1,1 @@
+C /etc/aws-otel-collector.yaml - - - - -
--- a/packages/aws-otel-collector/aws-otel-collector.service
+++ b/packages/aws-otel-collector/aws-otel-collector.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=AWS OTEL collector
+After=network-online.target configured.target
+Wants=network-online.target configured.target
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/aws-otel-collector --config /etc/aws-otel-collector.yaml
+KillMode=process
+Restart=always
+RestartSec=5s
+
+[Install]
+WantedBy=multi-user.target
--- a/packages/aws-otel-collector/aws-otel-collector.spec
+++ b/packages/aws-otel-collector/aws-otel-collector.spec
@@ -1,0 +1,77 @@
+%global goproject github.com/aws-observability
+%global gorepo aws-otel-collector
+%global goimport %{goproject}/%{gorepo}
+
+Name: %{_cross_os}aws-otel-collector
+Version: 0.40.0
+Release: 1%{?dist}
+Summary: AWS Distro for OpenTelemetry Collector
+License: Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND MIT AND MPL-2.0
+URL: https://github.com/aws-observability/aws-otel-collector
+Source0: %{gorepo}-v%{version}.tar.gz
+Source1: aws-otel-collector.service
+Source2: aws-otel-collector-tmpfiles.conf
+Source3: aws-otel-collector.yaml
+
+# Change log and extraconfig file paths from /opt to /var/log and /etc, respectively
+Patch0001: 0001-change-logger-and-extraconfig-file-paths.patch
+
+BuildRequires: %{_cross_os}glibc-devel
+Requires: %{name}(binary)
+
+%description
+%{summary}.
+
+%package bin
+Summary: Telemetry collector binary
+Provides: %{name}(binary)
+Requires: (%{_cross_os}image-feature(no-fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(fips) or %{name}-fips-bin)
+
+%description bin
+%{summary}.
+
+%package fips-bin
+Summary: Telemetry collector binary, FIPS edition
+Provides: %{name}(binary)
+Requires: (%{_cross_os}image-feature(fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
+
+%description fips-bin
+%{summary}.
+
+%prep
+%autosetup -n %{gorepo}-%{version} -p1
+
+%build
+export GO_MAJOR="1.22"
+
+%set_cross_go_flags
+
+go build -ldflags "${GOLDFLAGS}" -o aws-otel-collector ./cmd/awscollector
+gofips build -ldflags "${GOLDFLAGS}" -o fips/aws-otel-collector ./cmd/awscollector
+
+%install
+install -D -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}/aws-otel-collector.service
+
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -p -m 0644 %{S:2} %{buildroot}%{_cross_tmpfilesdir}
+
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
+install -p -m 0644 %{S:3} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
+
+install -d %{buildroot}{%{_cross_bindir},%{_cross_fips_bindir}}
+install -p -m 0755 aws-otel-collector %{buildroot}%{_cross_bindir}
+install -p -m 0755 fips/aws-otel-collector %{buildroot}%{_cross_fips_bindir}
+
+%files
+%{_cross_attribution_file}
+%{_cross_unitdir}/aws-otel-collector.service
+%{_cross_tmpfilesdir}/aws-otel-collector-tmpfiles.conf
+%{_cross_factorydir}%{_cross_sysconfdir}/aws-otel-collector.yaml
+
+%files bin
+%{_cross_bindir}/aws-otel-collector
+
+%files fips-bin
+%{_cross_fips_bindir}/aws-otel-collector
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-otel-collector"
+version = "0.1.0"
+dependencies = [
+ "glibc",
+]
+
+[[package]]
 name = "aws-signing-helper"
 version = "0.1.0"
 dependencies = [
@@ -54,6 +61,7 @@ dependencies = [
  "acpid",
  "amazon-ssm-agent",
  "aws-iam-authenticator",
+ "aws-otel-collector",
  "aws-signing-helper",
  "bash",
  "binutils",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
   "packages/acpid",
   "packages/amazon-ssm-agent",
   "packages/aws-iam-authenticator",
+  "packages/aws-otel-collector",
   "packages/aws-signing-helper",
   "packages/bash",
   "packages/binutils",
--- a/kits/bottlerocket-core-kit/Cargo.toml
+++ b/kits/bottlerocket-core-kit/Cargo.toml
@@ -15,6 +15,7 @@ path = "../kit.rs"
 acpid = { path = "../../packages/acpid" }
 amazon-ssm-agent = { path = "../../packages/amazon-ssm-agent" }
 aws-iam-authenticator = { path = "../../packages/aws-iam-authenticator" }
+aws-otel-collector = { path = "../../packages/aws-otel-collector" }
 aws-signing-helper = { path = "../../packages/aws-signing-helper" }
 bash = { path = "../../packages/bash" }
 binutils = { path = "../../packages/binutils" }
--- a/packages/aws-otel-collector/Cargo.toml
+++ b/packages/aws-otel-collector/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "aws-otel-collector"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/aws-observability/aws-otel-collector/archive/v0.40.0/aws-otel-collector-v0.40.0.tar.gz"
+sha512 = "4dc90d798fc818f357c93424e3111697e6830a2dc9913fb71e754959ec4f6002be8f079220208cad93b3b6850d337174fa0e52f57d517364d004dd08bae97e14"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
--- a/packages/aws-otel-collector/aws-otel-collector.yaml
+++ b/packages/aws-otel-collector/aws-otel-collector.yaml
@@ -1,0 +1,24 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 127.0.0.1:4317
+
+processors:
+  batch/traces:
+  batch/metrics:
+
+exporters:
+  otlp:
+    endpoint: 127.0.0.1:4318
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch/traces]
+      exporters: [otlp]
+    metrics:
+      receivers: [otlp]
+      processors: [batch/metrics]
+      exporters: [otlp]


packages/aws-otel-collector/aws-otel-collector-default-config.yaml
```diff
@@ -0,0 +1,37 @@
+extensions:
+  health_check:
+
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
```
koooosh: In the collector logs, I see the following warning:
```
{"level":"warn","timestamp":"2024-07-24T07:25:10.048Z","caller":"internal/warning.go:42","message":"Using the 0.0.0.0 address exposes this server to every network interface, which may facilitate Denial of Service attacks. Enable the feature gate to change the default and remove this warning.","kind":"extension","name":"health_check","documentation":"https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks","feature gate ID":"component.UseLocalHostAsDefaultHost"}
```

so I'm not sure if we should include this setting.

packages/aws-otel-collector/aws-otel-collector-default-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+      http:
+        endpoint: 0.0.0.0:4318
```
koooosh: Same as above comment: Should we remove this setting?

packages/aws-otel-collector/aws-otel-collector-default-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+  awsxray:
+    endpoint: 0.0.0.0:2000
```
koooosh: Same as above comment: Should we remove this setting?

packages/aws-otel-collector/aws-otel-collector-default-config.yaml
bcressey: Please set these to 127.0.0.1 instead so the services are only accessible to local clients.

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -0,0 +1,78 @@
+%global goproject github.com/aws-observability
+%global gorepo aws-otel-collector
+%global goimport %{goproject}/%{gorepo}
+
+Name: %{_cross_os}aws-otel-collector
+Version: 0.40.0
+Release: 1%{?dist}
+Summary: An AWS supported service to send telemetry data
```
bcressey: This is not a service and the "AWS supported" statement is confusing in this context. I would use the summary from the GitHub project page:
```suggestion
Summary: AWS Distro for OpenTelemetry Collector
```

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -0,0 +1,78 @@
...
+License: Apache-2.0
```
bcressey: This license statement seems incorrect. You can run `askalono crawl .` in a directory with all the unpacked sources to get its verdict ([askalono](https://github.com/jpeddicord/askalono)). But you are probably looking at an aggregation of all the licenses in [THIRD-PARTY-LICENSES.txt](https://github.com/aws-observability/aws-otel-collector/blob/main/THIRD-PARTY-LICENSES.txt).

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -0,0 +1,78 @@
...
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
+
+%description fips-bin
+%{summary}.
+
+%prep
+%autosetup -n %{gorepo}-%{version} -p0001
+
+%build
+export GO_VERSION="1.22.2"
```
bcressey:
```suggestion
export GO_MAJOR="1.22"
```

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -0,0 +1,78 @@
...
+
+%set_cross_go_flags
+
+go build -ldflags "${GOLDFLAGS}" -o aws-otel-collector ./cmd/awscollector
+gofips build -ldflags "${GOLDFLAGS}" -o fips/aws-otel-collector ./cmd/awscollector
+
+%install
+# Files copied by tmpfiles.d
```
bcressey: From the context - populating `%{_cross_tmpfilesdir}` with a drop-in - this is obvious:
```suggestion
```

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -0,0 +1,78 @@
...
+Summary: Telemetry collector binary, FIPS edition
+Provides: %{name}(binary)
+Requires: (%{_cross_os}image-feature(fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
+
+%description fips-bin
+%{summary}.
+
+%prep
+%autosetup -n %{gorepo}-%{version} -p0001
```
bcressey: nit: this `-p0001` usage seems to be new - here, `socat`, `amazon-ssm-agent` - and I don't prefer it.

It invites the reader to think that it's related to the `Patch0001` patch, but it's actually an argument to the `patch` command that controls prefix stripping. It's weird to pad numeric arguments to CLI programs with leading zeroes.
```suggestion
%autosetup -n %{gorepo}-%{version} -p1
```

packages/aws-otel-collector/aws-otel-collector.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=AWS OTEL collector
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/aws-otel-collector --config /etc/aws-otel-collector-default-config.yaml
```
bcressey: I don't think the file should be named "default" (what if the values are not the default?) and the "config" part is redundant for files in `/etc`:
```suggestion
ExecStart=/usr/bin/aws-otel-collector --config /etc/aws-otel-collector.yaml
```

packages/aws-otel-collector/aws-otel-collector-default-config.yaml
```diff
@@ -0,0 +1,37 @@
+extensions:
+  health_check:
```
bcressey: Not entirely sure it makes sense to ship a config at all here, but if we do then it should be a bare minimum - just the most basic one possible with all optional functionality turned off.

The awsxray, awsemf, http stuff I would cut.

Downstream variants can pretty easily specify a different config file via systemd unit drop-in that overrides the ExecStart command to point to their custom config or rendered template.

packages/aws-otel-collector/aws-otel-collector.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=AWS OTEL collector
+After=network.target
```
bcressey:
```suggestion
After=network-online.target configured.target
Wants=network-online.target configured.target
```

packages/aws-otel-collector/aws-otel-collector-default-config.yaml
koooosh: Ack, will change these endpoints to localhost IP

packages/aws-otel-collector/aws-otel-collector.spec
koooosh: I see. I based this off the Linux `build.spec` file offered by the collector repo: https://github.com/aws-observability/aws-otel-collector/blob/main/tools/packaging/linux/build.spec#L5

When I run `askalono crawl . | grep License: | sort | uniq` on the unpacked collector directory, I get the following excluding errors:
```
License: Apache-2.0 (license header)
License: Apache-2.0 (original text)
License: BSD-2-Clause (original text)
License: BSD-2-Clause-Views (original text)
License: BSD-3-Clause (original text)
License: CC-BY-SA-4.0 (original text)
License: ISC (original text)
License: MIT (original text)
License: MPL-2.0 (original text)
```

Given the license header is listed as `Apache-2.0`, is this correct?

packages/aws-otel-collector/aws-otel-collector.spec
koooosh: Got it. I based this off another example in the core-kit: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/kubernetes-1.30/kubernetes-1.30.spec#L146

I assume we prefer `GO_MAJOR ` so will replace this

packages/aws-otel-collector/aws-otel-collector.spec
koooosh: Ack, I'll change this here and also include commits to `socat` and `amazon-ssm-agent` replacing this.

packages/aws-otel-collector/aws-otel-collector.service
koooosh: Ack, makes sense given we are changing the default config as bare as possible

packages/aws-otel-collector/aws-otel-collector.service
koooosh: Ack, from what I've read online, the reason we may want to do this is to ensure the network is up before we start the OTEL collector service. Is this correct, or is there another reason for this suggestion?

packages/aws-otel-collector/aws-otel-collector-default-config.yaml
koooosh: I believe the collector requires some sort of config which specifies receivers, processors, and exporters; otherwise it will fail. I tried creating a patch to make the config optional but it causes the collector to fail with invalid configuration errors with [ResolverSettings](https://github.com/aws-observability/aws-otel-collector/blob/311c612da46c22aa0e32d3787dd422bbaba8e83d/pkg/config/config_factory.go#L57), which is [an import from the general OTEL collector pkg. ](https://github.com/open-telemetry/opentelemetry-collector/blob/confmap/v0.106.1/confmap/resolver.go#L33)

I've created a minimalist config based on possible core (non-aws specific) components which the collector still functions with here: https://gist.github.com/koooosh/8cddd42a91657552af4709b5d1157bef

Let me know if you notice anything off-putting in that.

packages/aws-otel-collector/aws-otel-collector.spec
koooosh: Went through all the licenses in the THIRD-PARTY file and aggregated this:
```
License: Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND MIT AND MPL-2.0
```

koooosh:
Force push addresses @bcressey comments:
- Updates to `aws-otel-collector.spec`, namely correcting License statement
- Changed `aws-otel-collector-default-config.yaml` to just `aws-otel-collector.yaml` and made the config as minimal as possible with localhost endpoints
- Update settings in `aws-otel-collector.service`
- Updating patch flags in `amazon-ssm-agent` and `socat` for consistency

packages/aws-otel-collector/aws-otel-collector-tmpfiles.conf
```diff
@@ -0,0 +1 @@
+C /etc/aws-otel-collector.yaml - - - - -
```
arnaldo2792, line 1: Don't you have to create the `/var/log/aws` directory, or is something else creating it?

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -65,7 +65,7 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-plugin-bin)
 %{summary}.
 
 %prep
-%autosetup -n %{gorepo}-%{version} -p0001
+%autosetup -n %{gorepo}-%{version} -p1
```
arnaldo2792: This change doesn't relate to your PR. You can cherry pick the commit into another PR, along with the `socat` update.

packages/aws-otel-collector/0001-logger-change-collector-log-file-path.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/pkg/logger/logger.go b/pkg/logger/logger.go
+index 9805eeab..7fcbab31 100644
+--- a/pkg/logger/logger.go
++++ b/pkg/logger/logger.go
+@@ -31,7 +31,7 @@ import (
+ )
+ 
+ var (
+-	UnixLogPath      = "/opt/aws/aws-otel-collector/logs/aws-otel-collector.log"
++	UnixLogPath      = "/var/log/aws/aws-otel-collector.log"
```
arnaldo2792: We should be extra safe and patch this file as well

https://github.com/aws-observability/aws-otel-collector/blob/311c612da46c22aa0e32d3787dd422bbaba8e83d/pkg/extraconfig/extraconfig.go#L26

We should encourage dowstreams to not read configuration files from `/opt`, and instead use standard configuration directories, like `/etc`.

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -0,0 +1,77 @@
...
+install -p -m 0755 fips/aws-otel-collector %{buildroot}%{_cross_fips_bindir}
+
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
+install -p -m 0644 %{S:3} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
+
+%files
+%{_cross_attribution_file}
+%{_cross_unitdir}/aws-otel-collector.service
+%{_cross_tmpfilesdir}/aws-otel-collector-tmpfiles.conf
+%{_cross_factorydir}%{_cross_sysconfdir}/aws-otel-collector.yaml
```
arnaldo2792, line 71: FWIW, I've thought about two options for downstreams to override our default configurations files for those service that don't have an API like this or the SSM agent:

- Override `Exec` directive in systemd services to force the service to look at a different location where the actual configuration file provided by the downstream resides
- Use a `tmpfiles.d` configuration file that installs the downstream overrides at the location where the service will look at: if `tmpfiles.d` detects that multiple configuration files provide the same `target` file, if will user lexicographic order to determine which one to keep and ultimately create

I think we could suggest downstreams to follow the latter, unless others have a better proposal to allow dowstreams override the defaults we provide. (cc: @bcressey , what are your thoughts on this?)

packages/aws-otel-collector/aws-otel-collector-tmpfiles.conf
koooosh: The collector creates it [here](https://github.com/aws-observability/aws-otel-collector/blob/main/pkg/logger/logger.go#L82), so we don't have to.

packages/amazon-ssm-agent/amazon-ssm-agent.spec
koooosh: Ack, I included these fix commits in this PR as the issue was brought up here, but I can create a new one for them and remove them from this PR

packages/aws-otel-collector/0001-logger-change-collector-log-file-path.patch
koooosh: Got it, I'll change the current `/opt/aws/aws-otel-collector/etc/extracfg.txt` path to `/etc/extracfg.txt`

koooosh:
Force push addresses @arnaldo2792 comments:
- Removing socat and amazon-ssm-agent fix commits
- Updating patch with extraconfig path change

packages/aws-otel-collector/aws-otel-collector.service
larvacea: I would use network-online.target in any unit file that's going to do network-stuff (including otel, of course). It's the target that says "hooray, we have an interface and it is configured and you can actually do TCP and DNS stuff on it, good luck, have fun." So for instance, this is a prerequisite for IMDS, and makes sense as a prerequisite for otel collector.

packages/aws-otel-collector/aws-otel-collector.spec
larvacea: I am a fan of `tmpfiles.d` as the clean, simple way to override a default here. It's my uneducated opinion, based on working with `tmpfiles.d` on other packages for configuration things that live in `/etc/something.d`.

larvacea (approved):
I looked through this, and especially looked through the remaining conversations on the pull request, and it looks good to me.

packages/aws-otel-collector/0001-change-logger-and-extraconfig-file-paths.patch
```diff
@@ -0,0 +1,40 @@
...
+
+diff --git a/pkg/extraconfig/extraconfig.go b/pkg/extraconfig/extraconfig.go
+index 9c14d5f7..c962b562 100644
+--- a/pkg/extraconfig/extraconfig.go
++++ b/pkg/extraconfig/extraconfig.go
+@@ -23,7 +23,7 @@ import (
+ )
+ 
+ var (
+-	UnixExtraConfigPath    = "/opt/aws/aws-otel-collector/etc/extracfg.txt"
```
arnaldo2792, line 20: This config should still be under `/etc/aws-otel-collector`.

packages/aws-otel-collector/aws-otel-collector.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=AWS OTEL collector
+After=network-online.target configured.target
+Wants=network-online.target configured.target
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/aws-otel-collector --config /etc/aws-otel-collector.yaml
+KillMode=process
+Restart=on-failure
```
bcressey: Why not always restart? If the process decides it should exit (for whatever reason), we typically want the host to self-heal.

```suggestion
Restart=always
```

packages/aws-otel-collector/aws-otel-collector.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=AWS OTEL collector
```
bcressey, line 2: nit: not sure of this, but the project page gives "ADOT Collector" as the short form of the name
```suggestion
Description=ADOT collector
```

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -0,0 +1,77 @@
...
+%set_cross_go_flags
+
+go build -ldflags "${GOLDFLAGS}" -o aws-otel-collector ./cmd/awscollector
+gofips build -ldflags "${GOLDFLAGS}" -o fips/aws-otel-collector ./cmd/awscollector
+
+%install
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -p -m 0644 %{S:2} %{buildroot}%{_cross_tmpfilesdir}
+
+install -D -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}/aws-otel-collector.service
```
bcressey: nit: better to start out with files being installed in source order

packages/aws-otel-collector/aws-otel-collector.spec
bcressey: > I think we could suggest downstreams to follow the latter, unless others have a better proposal to allow dowstreams override the defaults we provide.

@arnaldo2792 the tmpfiles behavior here is documented but surprising to me (emphasis added):
> If multiple files specify the same path, the entry in the file with the lexicographically **earliest** name will be applied

Elsewhere, for drop-ins and such, the latest file in sorted order takes effect. So this is likely something downstream builds will get wrong, or we'll add a config to a package here and break them if our tmpfiles config has an earlier filename.

You could make the config file path an environment variable, and then downstreams could override just that rather than the entire `Exec` directive using a drop-in.

It'd be annoying to implement, but we could prototype making Bottlerocket's `/etc` an overlayfs, consisting of a read-only set of files under the sys-root, and a tmpfs upper directory. Then packages could install files directly to `/etc`, and downstreams could clobber them with a tmpfiles directive:
```
r /etc/aws-otel-collector.yaml
C /etc/aws-otel-collector.yaml - - - - -
```

packages/aws-otel-collector/0001-change-logger-and-extraconfig-file-paths.patch
koooosh: Ack, will put the extraconfig file under the `/etc/aws-otel-collector` directory

packages/aws-otel-collector/aws-otel-collector.service
koooosh: Yeah, so we expect the collector to always be running (aka never exit cleanly) which is why I left it as `on-failure`.

However, I don't see an issue with changing to `always` just to be safe (also I see that's the general Bottlerocket convention besides for the `wickedd` services), so I can update this.

packages/aws-otel-collector/aws-otel-collector.service
koooosh: Yep, that was the name given in the default service, but I changed it to `AWS OTEL collector` to match the package name we are adding to the core-kit. 

I prefer keeping it as `AWS OTEL collector` due to that reason, but let me know if you feel strongly about changing it to `ADOT collector`!

packages/aws-otel-collector/aws-otel-collector.spec
koooosh: Ack, will adjust the ordering by putting Source 1 install prior to Source 2

koooosh:
Force push updates the following:
- Updates extraconfig file path in patch to: `"/etc/aws-otel-collector/extracfg.txt"`
- Changes `Restart` setting for collector service to `always`
- Rearranges source ordering

packages/aws-otel-collector/aws-otel-collector.spec
arnaldo2792: I had an offline conversation with @bcressey , and we agreed that copying the default config provided by us to `/etc` is OK, and that downstreams can override the config if they provide their own API server services that override the configurations if they need to.

arnaldo2792 (approved):

bcressey (approved):

PR #59, opened 2024-07-30
support hostname-override-source

tzneal:
**Issue number:**

Closes #

**Description of changes:**

Supports the new hostname-override-source setting to we can use the instance-id as the node name with K8s.

**Testing done:**

Built & verified that the following AMIs launched and joined the cluster with the correct node name for the following scenarios:
1.23 - No userdata
1.29 - No userdata
1.29 - hostname-override-source = private-dns-name
1.29 - hostname-override-source = instance-id

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/main.rs
+++ b/sources/api/pluto/src/main.rs
@@ -39,7 +39,7 @@ mod hyper_proxy;
 mod proxy;
 
 use api::AwsK8sInfo;
-use bottlerocket_modeled_types::KubernetesClusterDnsIp;
+use bottlerocket_modeled_types::{KubernetesClusterDnsIp, KubernetesHostnameOverrideSource};
 use imdsclient::ImdsClient;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::fs::File;
@@ -55,8 +55,6 @@ const DEFAULT_DNS_CLUSTER_IP: &str = "10.100.0.10";
 const DEFAULT_10_RANGE_DNS_CLUSTER_IP: &str = "172.20.0.10";
 
 const ENI_MAX_PODS_PATH: &str = "/usr/share/eks/eni-max-pods";
-
-const NO_HOSTNAME_VARIANTS: &[&str] = &["aws-k8s-1.23", "aws-k8s-1.24", "aws-k8s-1.25"];
 
 mod error {
     use crate::{api, ec2, eks};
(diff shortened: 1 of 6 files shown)


sources/api/pluto/src/main.rs
```diff
@@ -346,14 +346,26 @@ async fn generate_provider_id(
...
+    if aws_k8s_info.hostname_override.is_some() {
         return Ok(());
     }
 
+    // no override source provided, and this is one of the no hostname variants
+    if aws_k8s_info.hostname_override_source.is_none()
+        && NO_HOSTNAME_VARIANTS.contains(&aws_k8s_info.variant_id.as_str())
+    {
+        return Ok(());
+    }
```
bcressey: We could get rid of `NO_HOSTNAME_VARIANTS` entirely, and just rely on `hostname-override-source` being left unset, if you make a slight tweak to https://github.com/bottlerocket-os/bottlerocket/pull/4118 and also change the [kubernetes-aws-external-cloud-provider.toml](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/shared-defaults/kubernetes-aws-external-cloud-provider.toml) defaults to be:
```
[settings.kubernetes]
cloud-provider = "external"
hostname-override-source = "private-dns-name"
```

You've already added the migration so this would be an easy 1-line change in that PR.

That particular file is the only difference between defaults settings for `aws-k8s-1.25` and `aws-k8s-1.26`, per:
```
git diff --no-index sources/settings-defaults/aws-k8s-1.2{5,6}
```

bcressey (approved):

sources/api/pluto/src/main.rs
```diff
@@ -365,16 +370,25 @@ async fn generate_private_dns_name(
...
+                    aws_k8s_info.https_proxy.clone(),
+                    aws_k8s_info.no_proxy.clone(),
+                )
+                .await
+                .context(error::Ec2Snafu)?,
+            );
+        }
+        KubernetesHostnameOverrideSource::InstanceID => {
+            aws_k8s_info.hostname_override = Some(instance_id);
+        }
```
bcressey, line 389: nit: it'd be slightly better to fetch the instance ID within this block, as otherwise we pay for the cost of an IMDS interaction and end up throwing away the result when using the private DNS source

bcressey (approved):

sources/api/pluto/src/main.rs
tzneal: We use it both places, in the private DNS lookup we filter by the instance ID so we only get the single response.

ginglis13 (approved):

cbgbt (approved):

PR #62, opened 2024-08-01
Kit changes for Bootstrap Commands

piyush-jena:
**Issue number:**

Closes #

**Description of changes:**
This PR contains core-kit changes to add `bootstrap-commands`. The relevant settings-sdk changes are in [PR 46](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/46)

**Testing done:**
Test case 1: 2 bootstrap commands
```
[settings.bootstrap-commands."002-test-bootstrap-commands"]
commands = [["apiclient", "set", "motd=helloworld2"], ["apiclient", "set", "motd=helloworld3"]]
mode = "once"
essential = true

[settings.bootstrap-commands."001-test-bootstrap-commands"]
commands = [["apiclient", "set", "motd=helloworld1"]]
mode = "always"
essential = true
```
first boot sets `motd` to `helloworld3`
second boot sets `motd` to `helloworld1`

Test case 2: No bootstrap commands
No boot issues.

Note:
3 files have been modified to build successfully. "sources/api/Cargo.toml", "sources/deny.toml", "sources/Cargo.lock". These are dependent on the settings-sdk changes in [PR 46](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/46). They will be reverted back at the end of code review.

**Terms of contribution:**

### DIFF
--- a/packages/os/bootstrap-commands-toml
+++ b/packages/os/bootstrap-commands-toml
@@ -1,0 +1,18 @@
+[required-extensions]
+bootstrap-commands= "v1"
+std = { version = "v1", helpers = ["if_not_null", "toml_encode"]}
++++
+{{#if_not_null settings.bootstrap-commands}}
+{{#each settings.bootstrap-commands}}
+[bootstrap-commands."{{@key}}"]
+{{#if_not_null this.commands}}
+commands = {{ toml_encode this.commands }}
+{{/if_not_null}}
+{{#if_not_null this.mode}}
+mode = "{{{this.mode}}}"
+{{/if_not_null}}
+{{#if_not_null this.essential}}
+essential = {{this.essential}}
+{{/if_not_null}}
+{{/each}}
+{{/if_not_null}}
--- a/packages/os/bootstrap-commands.service
+++ b/packages/os/bootstrap-commands.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Bootstrap Commands
+# We depend on systemd-logind.service for running systemd-inhibit.
+After=systemd-logind.service settings-applier.service apiserver.service
+Requires=systemd-logind.service settings-applier.service apiserver.service
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemd-inhibit --what=shutdown --why="Running bootstrap commands" --mode=delay /usr/bin/bootstrap-commands
+RemainAfterExit=true
+StandardError=journal+console
+SyslogIdentifier=bootstrap-commands
+
+[Install]
+RequiredBy=preconfigured.target
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -32,6 +32,7 @@ Source17: corndog-toml
 Source18: bootstrap-containers-toml
 Source19: host-containers-toml
 Source20: bottlerocket-fips-checks-metadata-json
+Source21: bootstrap-commands-toml
 
 # 1xx sources: systemd units
 Source100: apiserver.service
@@ -52,6 +53,7 @@ Source119: reboot-if-required.service
 Source120: warm-pool-wait.service
 Source122: has-boot-ever-succeeded.service
 Source123: pluto.service
+Source124: bootstrap-commands.service
 
 # 2xx sources: tmpfilesd configs
 Source200: migration-tmpfiles.conf
@@ -59,6 +61,7 @@ Source201: host-containers-tmpfiles.conf
 Source202: thar-be-updates-tmpfiles.conf
 Source203: bootstrap-containers-tmpfiles.conf
 Source204: storewolf-tmpfiles.conf
+Source205: bootstrap-commands-tmpfiles.conf
 
 # 3xx sources: udev rules
 Source300: ephemeral-storage.rules
@@ -74,6 +77,7 @@ BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}apiclient
 Requires: %{_cross_os}apiserver
 Requires: %{_cross_os}bloodhound
+Requires: %{_cross_os}bootstrap-commands
 Requires: %{_cross_os}corndog
 Requires: %{_cross_os}certdog
 Requires: %{_cross_os}ghostdog
@@ -246,6 +250,11 @@ Requires: %{_cross_os}binutils
 %description -n %{_cross_os}driverdog
 %{summary}.
 
+%package -n %{_cross_os}bootstrap-commands
+Summary: Manages bootstrap-commands
+%description -n %{_cross_os}bootstrap-commands
+%{summary}.
+
 %package -n %{_cross_os}bootstrap-containers
 Summary: Manages bootstrap-containers
 Requires: %{_cross_os}host-ctr
@@ -350,6 +359,7 @@ echo "** Output from non-static builds:"
     -p metricdog \
     -p ghostdog \
     -p corndog \
+    -p bootstrap-commands \
     -p bootstrap-containers \
     -p prairiedog \
     -p certdog \
@@ -385,7 +395,7 @@ for p in \
   storewolf settings-committer \
   migrator prairiedog certdog \
   signpost updog metricdog logdog \
-  ghostdog bootstrap-containers \
+  ghostdog bootstrap-commands bootstrap-containers \
   shimpei bloodhound \
   bottlerocket-cis-checks \
   bottlerocket-fips-checks \
@@ -473,14 +483,14 @@ if [ -s "%{_cross_repo_root_json}" ] ; then
 fi
 
 install -d %{buildroot}%{_cross_templatedir}
-install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:14} %{S:15} %{S:16} %{S:17} %{S:18} %{S:19} \
+install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:14} %{S:15} %{S:16} %{S:17} %{S:18} %{S:19} %{S:21} \
   %{buildroot}%{_cross_templatedir}
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
   %{S:100} %{S:102} %{S:103} %{S:105} \
   %{S:106} %{S:107} %{S:110} %{S:111} %{S:112} \
-  %{S:113} %{S:114} %{S:119} %{S:122} %{S:123} \
+  %{S:113} %{S:114} %{S:119} %{S:122} %{S:123} %{S:124} \
   %{buildroot}%{_cross_unitdir}
 
 sed -e 's|PREFIX|%{_cross_prefix}|g' %{S:115} > link-kernel-modules.service
@@ -502,6 +512,7 @@ install -p -m 0644 %{S:201} %{buildroot}%{_cross_tmpfilesdir}/host-containers.co
 install -p -m 0644 %{S:202} %{buildroot}%{_cross_tmpfilesdir}/thar-be-updates.conf
 install -p -m 0644 %{S:203} %{buildroot}%{_cross_tmpfilesdir}/bootstrap-containers.conf
 install -p -m 0644 %{S:204} %{buildroot}%{_cross_tmpfilesdir}/storewolf.conf
+install -p -m 0644 %{S:205} %{buildroot}%{_cross_tmpfilesdir}/bootstrap-commands.conf
 
 install -d %{buildroot}%{_cross_udevrulesdir}
 install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-storage.rules
@@ -639,6 +650,12 @@ install -p -m 0644 %{S:400} %{S:401} %{S:402} %{buildroot}%{_cross_licensedir}
 %files -n %{_cross_os}certdog
 %{_cross_bindir}/certdog
 %{_cross_templatedir}/certdog-toml
+
+%files -n %{_cross_os}bootstrap-commands
+%{_cross_bindir}/bootstrap-commands
+%{_cross_unitdir}/bootstrap-commands.service
+%{_cross_tmpfilesdir}/bootstrap-commands.conf
+%{_cross_templatedir}/bootstrap-commands-toml
 
 %files -n %{_cross_os}bootstrap-containers
 %{_cross_bindir}/bootstrap-containers
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -23,6 +23,8 @@ members = [
 
     "bloodhound",
 
+    "bootstrap-commands",
+
     "bottlerocket-release",
 
     "bottlerocket-variant",
@@ -191,13 +193,13 @@ base64 = "0.22"
 
 [workspace.dependencies.bottlerocket-modeled-types]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.3.0"
-version = "0.3.0"
+tag = "bottlerocket-settings-models-v0.4.0"
+version = "0.4.0"
 
 [workspace.dependencies.bottlerocket-settings-models]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.3.0"
-version = "0.3.0"
+tag = "bottlerocket-settings-models-v0.4.0"
+version = "0.4.0"
 
 [workspace.dependencies.bottlerocket-settings-plugin]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
@@ -206,7 +208,7 @@ version = "0.1.0"
 
 [workspace.dependencies.settings-extension-oci-defaults]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.3.0"
+tag = "bottlerocket-settings-models-v0.4.0"
 version = "0.1.0"
 
 [profile.release]
--- a/sources/api/schnauzer/src/helpers/stdlib/mod.rs
+++ b/sources/api/schnauzer/src/helpers/stdlib/mod.rs
@@ -1130,3 +1130,130 @@ mod test_negate_or_else {
         });
     }
 }
+
+/// `toml_encode` accepts arbitrary input and encodes it as a toml string
+///
+/// # Example
+///
+/// Consider an array of values: `[ "a", "b", "c" ]` stored in a setting such as
+/// `settings.somewhere.foo-list`. In our template we can write:
+/// `{{ toml_encode settings.somewhere.foo-list }}`
+///
+/// This will render `["a", "b", "c"]`.
+///
+/// Similarly, for a string: `"string"`, the template {{ toml-encode "string" }}
+/// will render `"string"`.
+pub fn toml_encode(
+    helper: &Helper<'_, '_>,
+    _: &Handlebars,
+    _: &Context,
+    renderctx: &mut RenderContext<'_, '_>,
+    out: &mut dyn Output,
+) -> Result<(), RenderError> {
+    trace!("Starting toml_encode helper");
+    let template_name = template_name(renderctx);
+    check_param_count(helper, template_name, 1)?;
+
+    // get the string
+    let encode_param = get_param(helper, 0)?;
+    let toml_value: toml::Value =
+        serde_json::from_value(encode_param.to_owned()).with_context(|_| {
+            error::TomlEncodeSnafu {
+                value: encode_param.to_owned(),
+                template: template_name,
+            }
+        })?;
+
+    let result = toml_value.to_string();
+
+    // write it to the template
+    out.write(&result)
+        .with_context(|_| error::TemplateWriteSnafu {
+            template: template_name.to_owned(),
+        })?;
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod test_toml_encode {
+    use crate::helpers::toml_encode;
+    use handlebars::{Handlebars, RenderError};
+    use serde::Serialize;
+    use serde_json::json;
+
+    // A thin wrapper around the handlebars render_template method that includes
+    // setup and registration of helpers
+    fn setup_and_render_template<T>(tmpl: &str, data: &T) -> Result<String, RenderError>
+    where
+        T: Serialize,
+    {
+        let mut registry = Handlebars::new();
+        registry.register_helper("toml_encode", Box::new(toml_encode));
+
+        registry.render_template(tmpl, data)
+    }
+
+    const TEMPLATE: &str = r#"{{ toml_encode settings.foo-string }}"#;
+
+    #[test]
+    fn toml_encode_map() {
+        let result = setup_and_render_template(
+            TEMPLATE,
+            &json!({"settings": {"foo-string": {"hello": "world"}}}),
+        )
+        .unwrap();
+        let expected = r#"{ hello = "world" }"#;
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn toml_encode_empty() {
+        let result =
+            setup_and_render_template(TEMPLATE, &json!({"settings": {"foo-string": []}})).unwrap();
+        let expected = r#"[]"#;
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn toml_encode_empty_string() {
+        let result =
+            setup_and_render_template(TEMPLATE, &json!({"settings": {"foo-string": [""]}}))
+                .unwrap();
+        let expected = r#"[""]"#;
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn toml_encode_toml_injection_1() {
+        let result = setup_and_render_template(
+            TEMPLATE,
+            &json!({"settings": {"foo-string": [ "apiclient set motd=hello', 'echo pwned\""]}}),
+        )
+        .unwrap();
+        let expected = "['''apiclient set motd=hello', 'echo pwned\"''']";
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn toml_encode_toml_injection_2() {
+        let result = setup_and_render_template(
+            TEMPLATE,
+            &json!({"settings": {"foo-string": [ "apiclient set motd=hello\", \"echo pwned\""]}}),
+        )
+        .unwrap();
+        let expected = "['apiclient set motd=hello\", \"echo pwned\"']";
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn toml_encode_toml_injection_3() {
+        let result = setup_and_render_template(
+            TEMPLATE,
+            &json!({"settings": {"foo-string": [ "apiclient set motd=hello\", \"echo pwned\", 'echo pwned2'"]}}),
+        )
+        .unwrap();
+        let expected = "[\"apiclient set motd=hello\\\", \\\"echo pwned\\\", 'echo pwned2'\"]";
+        assert_eq!(result, expected);
+    }
+}
--- a/sources/bootstrap-commands/src/main.rs
+++ b/sources/bootstrap-commands/src/main.rs
@@ -1,0 +1,370 @@
+/*!
+# Bootstrap commands
+
+`bootstrap-commands` ensures that bootstrap commands are executed as defined in the system
+settings. It is called by `bootstrap-commands.service` which runs prior to the execution of
+`bootstrap-containers`.
+
+Each bootstrap command is a set of Bottlerocket API commands. The settings are first rendered
+into a config file. Then, the system is configured by going through all the bootstrap commands
+in lexicographical order and running all the commands inside it.
+
+## Example:
+Given a bootstrap command called `001-test-bootstrap-commands` with the following configuration:
+
+```toml
+[settings.bootstrap-commands.001-test-bootstrap-commands]
+commands = [[ "apiclient", "set", "motd=helloworld"]]
+essential = true
+mode = "always"
+```
+This would set `/etc/motd` to "helloworld".
+
+# Additional Information:
+Certain valid `apiclient` commands that work in a session may fail in `bootstrap-commands`
+due to relevant services not running at the time of the launch of the systemd service.
+
+## Example:
+```toml
+[settings.bootstrap-commands.001-test-bootstrap-commands]
+commands = [[ "apiclient", "exec", "admin", "ls"]]
+essential = true
+mode = "always"
+```
+This command fails because `bootstrap-commands.service` which calls this binary is launched
+prior to `preconfigured.target` while `host-containers@.service` which is a requirement for
+running "exec" commands are launched after preconfigured.target.
+*/
+
+use log::info;
+use serde::Deserialize;
+use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
+use snafu::{ensure, OptionExt, ResultExt};
+use std::collections::BTreeMap;
+use std::env;
+use std::ffi::OsStr;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::{self, Command};
+use std::str::FromStr;
+
+use bottlerocket_modeled_types::{ApiclientCommand, BootstrapMode, Identifier};
+
+const DEFAULT_CONFIG_PATH: &str = "/etc/bootstrap-commands/bootstrap-commands.toml";
+
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct BootstrapCommandConfig {
+    #[serde(default)]
+    bootstrap_commands: BTreeMap<Identifier, BootstrapCommand>,
+}
+
+impl BootstrapCommandConfig {
+    // Gets an iterator for bootstrap_commands, sorted in lexicographical order of their names.
+    fn iter(self) -> impl Iterator<Item = (Identifier, BootstrapCommand)> {
+        self.bootstrap_commands.into_iter()
+    }
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct BootstrapCommand {
+    #[serde(default)]
+    commands: Vec<ApiclientCommand>,
+    #[serde(default)]
+    mode: BootstrapMode,
+    #[serde(default)]
+    essential: bool,
+}
+
+/// Stores user-supplied global arguments
+struct Args {
+    log_level: LevelFilter,
+    config_path: PathBuf,
+}
+
+impl Default for Args {
+    fn default() -> Self {
+        Self {
+            log_level: LevelFilter::Info,
+            config_path: PathBuf::from_str(DEFAULT_CONFIG_PATH).unwrap(),
+        }
+    }
+}
+
+/// Wrapper around process::Command that adds error checking.
+fn run_command<I, S>(bin_path: &str, args: I) -> Result<()>
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    let mut command = Command::new(bin_path);
+
+    command
+        .args(args)
+        .status()
+        .context(error::ExecutionFailureSnafu { command })?;
+
+    Ok(())
+}
+
+fn handle_bootstrap_command<S>(name: S, bootstrap_command: BootstrapCommand) -> Result<()>
+where
+    S: AsRef<str>,
+{
+    let name = name.as_ref();
+    let mode = bootstrap_command.mode.as_ref();
+    let commands = &bootstrap_command.commands;
+
+    if mode == "off" {
+        // If mode is 'off', just log this information.
+        info!("Bootstrap command mode for '{}' is 'off'", name);
+        return Ok(());
+    }
+
+    info!("Processing bootstrap command '{}' ... ", name);
+
+    for command in commands.iter() {
+        let (cmd, args) = command.get_command_and_args();
+        run_command(cmd, args)?;
+    }
+
+    if mode == "once" {
+        let formatted = format!("settings.bootstrap-commands.{}.mode=off", name);
+        info!("Turning off bootstrap command '{}'", name);
+        run_command("apiclient", ["set", formatted.as_str()])?;
+    }
+
+    info!("Successfully ran bootstrap command '{}'", name);
+
+    Ok(())
+}
+
+/// Read our config file for the set of defined bootstrap commands
+fn get_bootstrap_commands<P>(config_path: P) -> Result<BootstrapCommandConfig>
+where
+    P: AsRef<Path>,
+{
+    let config_str = fs::read_to_string(config_path.as_ref()).context(error::ReadConfigSnafu {
+        config_path: config_path.as_ref(),
+    })?;
+
+    let config: BootstrapCommandConfig =
+        toml::from_str(&config_str).context(error::DeserializationSnafu {
+            config_path: config_path.as_ref(),
+        })?;
+
+    Ok(config)
+}
+
+/// Parse the args to the program and return an Args struct
+fn parse_args(args: env::Args) -> Result<Args> {
+    let mut global_args = Args::default();
+
+    let mut iter = args.skip(1);
+    while let Some(arg) = iter.next() {
+        match arg.as_ref() {
+            // Global args
+            "--log-level" => {
+                let log_level = iter.next().context(error::UsageSnafu {
+                    message: "Did not give argument to --log-level",
+                })?;
+                global_args.log_level = LevelFilter::from_str(&log_level)
+                    .context(error::LogLevelSnafu { log_level })?;
+            }
+
+            "-c" | "--config-path" => {
+                let config_str = iter.next().context(error::UsageSnafu {
+                    message: "Did not give argument to --config-path",
+                })?;
+                global_args.config_path = PathBuf::from(config_str.as_str());
+            }
+
+            _ => (),
+        }
+    }
+
+    Ok(global_args)
+}
+
+fn run() -> Result<()> {
+    let args = parse_args(env::args())?;
+
+    // SimpleLogger will send errors to stderr and anything less to stdout.
+    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
+
+    let bootstrap_commands = get_bootstrap_commands(args.config_path)?;
+
+    for (bootstrap_command_name, bootstrap_command) in bootstrap_commands.iter() {
+        let name = bootstrap_command_name.as_ref();
+        let essential = bootstrap_command.essential;
+        let status = handle_bootstrap_command(name, bootstrap_command);
+
+        ensure!(
+            !essential || status.is_ok(),
+            error::BootstrapCommandExecutionSnafu { name }
+        )
+    }
+
+    Ok(())
+}
+
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+mod error {
+    use snafu::Snafu;
+    use std::path::PathBuf;
+    use std::process::Command;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(super) enum Error {
+        #[snafu(display("Failed to read settings from config at {}: {}", config_path.display(), source))]
+        ReadConfig {
+            config_path: PathBuf,
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Failed to deserialize settings from config at {}: {}", config_path.display(), source))]
+        Deserialization {
+            config_path: PathBuf,
+            source: toml::de::Error,
+        },
+
+        #[snafu(display("Bootstrap command '{}' failed.", name))]
+        BootstrapCommandExecution { name: String },
+
+        #[snafu(display("Failed to execute '{:?}': {}", command, source))]
+        ExecutionFailure {
+            command: Command,
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Logger setup error: {}", source))]
+        Logger { source: log::SetLoggerError },
+
+        #[snafu(display("Invalid log level '{}'", log_level))]
+        LogLevel {
+            log_level: String,
+            source: log::ParseLevelError,
+        },
+
+        #[snafu(display("{}", message))]
+        Usage { message: String },
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_get_bootstrap_commands() {
+        let config_toml = r#"[bootstrap-commands."002-test-bootstrap-commands"]
+        commands = [["apiclient", "set", "motd=helloworld2"], ["apiclient", "set", "motd=helloworld3"]]
+        mode = "once"
+        essential = true
+
+        [bootstrap-commands."001-test-bootstrap-commands"]
+        commands = [["apiclient", "set", "motd=helloworld1"]]
+        mode = "always"
+        essential = true
+        "#;
+
+        let temp_dir = tempfile::TempDir::new().unwrap();
+        let temp_config = Path::join(temp_dir.path(), "bootstrap-commands.toml");
+        let _ = std::fs::write(&temp_config, config_toml).unwrap();
+
+        let bootstrap_command_config = get_bootstrap_commands(&temp_config).unwrap();
+        let bootstrap_commands = bootstrap_command_config.bootstrap_commands;
+
+        let mut expected_bootstrap_commands = BTreeMap::new();
+        let testcmd_1 = ApiclientCommand::try_from(vec![
+            "apiclient".to_string(),
+            "set".to_string(),
+            "motd=helloworld1".to_string(),
+        ])
+        .unwrap();
+        let testcmd_2 = ApiclientCommand::try_from(vec![
+            "apiclient".to_string(),
+            "set".to_string(),
+            "motd=helloworld2".to_string(),
+        ])
+        .unwrap();
+        let testcmd_3 = ApiclientCommand::try_from(vec![
+            "apiclient".to_string(),
+            "set".to_string(),
+            "motd=helloworld3".to_string(),
+        ])
+        .unwrap();
+        expected_bootstrap_commands.insert(
+            Identifier::try_from("001-test-bootstrap-commands").unwrap(),
+            BootstrapCommand {
+                commands: vec![testcmd_1],
+                mode: BootstrapMode::try_from("always").unwrap(),
+                essential: true,
+            },
+        );
+        expected_bootstrap_commands.insert(
+            Identifier::try_from("002-test-bootstrap-commands").unwrap(),
+            BootstrapCommand {
+                commands: vec![testcmd_2, testcmd_3],
+                mode: BootstrapMode::try_from("once").unwrap(),
+                essential: true,
+            },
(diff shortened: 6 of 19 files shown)


packages/os/bootstrap-commands.service
```diff
@@ -0,0 +1,18 @@
...
+Description=Bootstrap Commands
+After=settings-applier.service apiserver.service
+Requires=settings-applier.service apiserver.service
+Before=preconfigured.target
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+# Run the bootstrap commands
```
arnaldo2792: nit: this comment is kind of extra, I think what's happening is self-explanatory.

sources/api/bootstrap-commands/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+itertools = "0.13"
+log = "0.4"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1"
+simplelog = "0.12"
+snafu = "0.8"
+toml = "0.8"
+
+[dependencies.bottlerocket-modeled-types]
+git = "https://github.com/piyush-jena/bottlerocket-settings-sdk"
```
arnaldo2792: I think you can split this dependency update to a separate commit, with title `hack` so that we know it has to be changed after you release a new version of the settings SDK that includes your change.

sources/api/bootstrap-commands/README.md
```diff
@@ -0,0 +1,28 @@
+# bootstrap-commands
```
arnaldo2792: You will have to have a "website" task to update the documentation for both bootstrap commands and the `apiclient` actions.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
+/*
+# Bootstrap commands
+
+Bootstrap Commands in Bottlerocket is a feature that empowers users to execute predefined commands
+through the Bottlerocket API. It introduces a new systemd service that facilitates the execution of
+these commands, expanding the functionality of the existing Bottlerocket API.
+
+Currently, Bottlerocket provides a set of actions (such as refresh-updates and reboot) that can be
```
arnaldo2792: I wouldn't list `reboot` as an action that you would like to run a bootstrap command. You could just leave `refresh-updates`.

arnaldo2792: Also, this kind of doesn't match with the example you provide. In here, you say that bootstrap commands allow one to run "actions". However, in your example you run `apiclient set` which isn't in itself an action.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+/// Wrapper around process::Command that adds error checking.
+fn command<I, S>(bin_path: &str, args: I) -> Result<()>
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    let mut command = Command::new(bin_path);
+    command.args(args);
+    let output = command
+        .output()
```
arnaldo2792: Rather than creating new pipes for STDIN/STDOUT, we should use the parent's process pipes with `status` so that the child process inherits the pipes, but the parent process waits for it to complete.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+        .context(error::ExecutionFailureSnafu { command })?;
+
+    trace!("stdout: {}", String::from_utf8_lossy(&output.stdout));
+    trace!("stderr: {}", String::from_utf8_lossy(&output.stderr));
```
arnaldo2792: If you do what I suggested above, you wouldn't need this.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+            .with_extension("ran");
+
+        // Check if bootstrap command is already run in this boot.
+        if temp_path.exists() {
+            info!("Skipping Bootstrap command '{}' because its already run", name);
+            return Ok(());
+        }
+
+        // Create a file in /run/bootstrap-commands to indicate an attempt to run this bootstrap-command
+        command(constants::TOUCH_BIN, [temp_path])?;
```
arnaldo2792: I don't think you need to create a sentry file for bootstrap commands. If they are `once` in the API you assume they haven't run. And the `bootstrap-commands` systemd service won't be re-executed in later stages of the boot (systemd guarantees that). In bootstrap containers, we needed that file because for failed bootstrap containers systemd will attempt to re-run them as it moved to other targets. You don't have that problem.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    Ok(())
+}
+
+/// Handles running of the api commands in a Bootstrap command
+fn handle_bootstrap_command<S>(name: S, bootstrap_command: &BootstrapCommand) -> Result<()>
+where
+    S: AsRef<str>,
+{
+    let binding = Vec::new();
```
arnaldo2792: Why is this called `binding`?

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+        // Create a file in /run/bootstrap-commands to indicate an attempt to run this bootstrap-command
+        command(constants::TOUCH_BIN, [temp_path])?;
+
+        info!("Processing Bootstrap command '{}' ... ", name);
+        for cmd in commands.iter() {
+            // this condition is already taken care of
+            let args = cmd.get_args();
+            match args {
+                Ok(arguments) => {
+                    if let Err(_e) = command(API_CLIENT_PATH, arguments) {
```
arnaldo2792: Is this how we are enforcing that only `apiclient` commands are executed? If that's the case, remove the `apiclient` from your example, and just use the actual command executed by it.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+            return Ok(());
+        }
+
+        // Create a file in /run/bootstrap-commands to indicate an attempt to run this bootstrap-command
+        command(constants::TOUCH_BIN, [temp_path])?;
+
+        info!("Processing Bootstrap command '{}' ... ", name);
+        for cmd in commands.iter() {
+            // this condition is already taken care of
+            let args = cmd.get_args();
```
arnaldo2792: You can do:

```
let args = cmd.get_args().context(error::BootstrapCommandFailureSnafu { name })?;
```

And avoid the `match` arms, which is also incorrect since the way it is written doesn't actually trigger an error.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+    let args = parse_args(env::args())?;
+
+    // SimpleLogger will send errors to stderr and anything less to stdout.
+    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
+
+    let bootstrap_commands = get_bootstrap_commands(args.config_path)?;
+
+    // We want to run bootstrap commands in sorted order of their names.
+    for name in bootstrap_commands.keys().sorted() {
+        let bootstrap_command = &bootstrap_commands[name];
```
arnaldo2792: I wonder if we should turn on the same clippy rule we used in the Bottlerocket Settings SDK, and avoid accessing elements in a `HashMap` like this, even though it is safe since you are getting the keys from the actual hash_map.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+        if let Err(e) = handle_bootstrap_command(name, bootstrap_command) {
```
arnaldo2792: It is common practice in this repo to use Snafu to provide context for the errors, instead of evaluating if the function call returned an error.

packages/os/bootstrap-commands-toml
```diff
@@ -0,0 +1,12 @@
+[required-extensions]
+bootstrap-commands= "v1"
+std = { version = "v1", helpers = ["if_not_null", "join_nested_array"]}
++++
+{{#if_not_null settings.bootstrap-commands}}
+{{#each settings.bootstrap-commands}}
+[bootstrap-commands."{{@key}}"]
+commands = [{{join_nested_array this.commands}}]
```
arnaldo2792: Why do we need the helper? Can't we just pass `this.commands`?

arnaldo2792 (changes requested):

sources/api/bootstrap-commands/src/main.rs
piyush-jena: I couldn't find a better way to get key, value where keys are in sorted order.

sources/api/bootstrap-commands/Cargo.toml
piyush-jena: https://github.com/bottlerocket-os/bottlerocket-project-website/issues/473

sources/api/bootstrap-commands/README.md
piyush-jena: https://github.com/bottlerocket-os/bottlerocket-project-website/issues/473

sources/api/bootstrap-commands/src/main.rs
piyush-jena: updated

sources/api/bootstrap-commands/src/main.rs
piyush-jena: Updated.

sources/api/bootstrap-commands/src/main.rs
piyush-jena: removed.

packages/os/bootstrap-commands-toml
piyush-jena: I attempted. It didn't work.

sources/api/bootstrap-commands/src/main.rs
piyush-jena: Updated.

sources/api/bootstrap-commands/src/main.rs
piyush-jena: Updated.

sources/api/bootstrap-commands/Cargo.toml
piyush-jena: updated.

sources/api/bootstrap-commands/src/main.rs
piyush-jena: The discussed `get_args` has a default value which makes the function infallible which is why I decided not to use snafu in this case.

sources/api/bootstrap-commands/src/main.rs
piyush-jena: Updated HashMap to BTreeMap which stores keys in BST (sorted order). Verified it works with end to end testing.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+[settings.bootstrap-commands.001-test-bootstrap-commands]
+commands = [[ "apiclient", "set", "motd=helloworld"]]
+essential = true
+mode = "always"
+
+This would update motd from "Welcome to Bottlerocket!" to "helloworld".
+*/
+
+#[macro_use]
+extern crate log;
```
cbgbt: You are no longer required to use `macro_use` as of Rust 2018 edition.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+
+//use itertools::Itertools;
```
cbgbt:
```suggestion
```

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+use serde::{Deserialize, Serialize};
+use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
+use snafu::{ensure, OptionExt, ResultExt};
+use std::collections::BTreeMap;
+
+use std::env;
```
cbgbt: nit: Seems like the whitespace may not actually be separating dependencies like it was intended.

In general (to my knowledge), Bottlerocket coding style has dependencies existing without whitespace separation and allows `cargo fmt` to sort them alphabetically.

The exception would be when using `pub use` to re-export symbols -- those are often kept separate.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+            BootstrapCommand {
+                commands: Some(vec![testcmd_2, testcmd_3]),
+                mode: Some(BootstrapMode::try_from("once").unwrap()),
+                essential: Some(true),
+            },
+        );
+
+        assert_eq!(bootstrap_commands, expected_bootstrap_commands)
+    }
+}
```
cbgbt: Do you mind also testing some negative cases?

e.g. what happens with an empty input? One with an invalid command?

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+        },
+
+        #[snafu(display("Failed to deserialize settings from config at {}: {}", config_path.display(), source))]
+        Deserialization {
+            config_path: PathBuf,
+            source: toml::de::Error,
+        },
+
+        #[snafu(display("Essential Bootstrap command with name: {} failed.", name))]
+        BootstrapCommandFailure { name: String },
```
cbgbt: It seems counter-intuitive, but with `snafu` it tends to make sense to drop words like "Failure" and "Error" for enum variants, since `snafu` creates the `context` types like `BoostrapCommandSnafu`.

```suggestion
        BootstrapCommand { name: String },
```

`BootstrapCommandFailureSnafu` is a bit redundant 😄

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    let mut command = Command::new(bin_path);
+
+    command
+        .args(args)
+        .status()
+        .context(error::ExecutionFailureSnafu { command })?;
```
cbgbt: I wonder if we should capture the `stdout/stderr` streams somewhere?

Forwarding `stderr` to this process' `stderr` in a way that makes it clear that it has been passed through might be a good start, e.g. something like:

```rust
info!("[command {} stderr] {}", command_id, line);
```

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct BootstrapCommandConfig {
+    bootstrap_commands: Option<BTreeMap<Identifier, BootstrapCommand>>,
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct BootstrapCommand {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
cbgbt: Does the `default` specifier have an effect for `Option` types? If you just plan to `unwrap_or_default` these, you could keep the `default` specifier and drop the `Option`. This would tell `serde` to use `Default::default` to choose the value if the input does not have a value present.

Do we ever serialize this? Since `bootstrap-commands` isn't a library and doesn't have cause to serialize `BootstrapCommands`, maybe we should just remove the `Serialize` derive and the serialization-related attributes.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+
+use bottlerocket_modeled_types::{BootstrapMode, Identifier};
+use bottlerocket_settings_models::ValidCommand;
+
+const DEFAULT_CONFIG_PATH: &str = "/etc/bootstrap-commands/bootstrap-commands.toml";
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct BootstrapCommandConfig {
+    bootstrap_commands: Option<BTreeMap<Identifier, BootstrapCommand>>,
```
cbgbt: This could be a good case to use `#[serde(default)]` if you wanted to avoid dealing with an additional `Option` type, since it's semantically equivalent to the empty list.

I don't have a strong preference though, this is just a suggestion.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+
+/// Handles running of the api commands in a Bootstrap command
+fn handle_bootstrap_command<S>(name: S, bootstrap_command: &BootstrapCommand) -> Result<()>
+where
+    S: AsRef<str>,
+{
+    let name = name.as_ref();
+    let mode = bootstrap_command.mode.clone().unwrap_or_default();
+    let commands = bootstrap_command.commands.clone().unwrap_or_default();
```
cbgbt:
```suggestion
    let mode = bootstrap_command.mode.clone().unwrap_or_default();
    let commands = bootstrap_command.commands.as_deref().unwrap_or_default();
```

If you drop the `Option` from this struct and use `#[serde(default)]`, you could use `as_ref()` instead of `clone()` for the `mode` as well.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+    info!("Running bootstrap command with name: {}", name);
+
+    if mode == "off" {
+        // If mode is 'off', just log this information.
+        info!("Bootstrap command mode for '{}' is 'off'", name);
+    } else {
+        info!("Processing Bootstrap command '{}' ... ", name);
+
+        for command in commands.iter() {
+            // this condition is already taken care of
```
cbgbt: It's not clear to me what this comment is referring to.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+            run_command("apiclient", ["set", formatted.as_str()])?;
+        }
+
+        info!("Succesfully ran bootstrap command '{}'", name);
+    }
+
+    Ok(())
+}
+
+/// Query the API for the currently defined bootstrap commands
```
cbgbt:
```suggestion
/// Read our config file for the set of defined bootstrap commands
```

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+fn get_bootstrap_commands<P>(config_path: P) -> Result<BTreeMap<Identifier, BootstrapCommand>>
```
cbgbt: The ordering enforced by `BTreeMap` is important to the correctness of this program, but it's also easy to miss.

Suggestion: have this function return the `BootstrapCommandConfig` struct wholesale. Add a method on it like,

```rust
fn iter(&self) -> impl Iterator<Item = (Identifier, BootstrapCommand)> {
}
```

...and ensure that the docstring mentions that the iterator is sorted lexicographically by ID. This hides the `BTreeMap` implementation from callers while also making it clear what the intent is.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+fn run() -> Result<()> {
+    let args = parse_args(env::args())?;
+
+    // SimpleLogger will send errors to stderr and anything less to stdout.
+    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
+
+    let bootstrap_commands = get_bootstrap_commands(args.config_path)?;
+
+    // We want to run bootstrap commands in sorted order of their names.
+    for (name, bootstrap_command) in bootstrap_commands.iter() {
```
cbgbt: Just to follow up about the `BTree` implementation -- this comment about sorted order seems almost erroneous in isolation unless I go read the function signature and become aware that the return type is `BTreeMap`.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -975,6 +975,86 @@ pub fn join_array(
...
+            .with_context(|| error::JoinStringsWrongTypeSnafu {
+                expected_type: "array",
+                value: nested_array_param.to_owned(),
+                template: template_name,
+            })?;
+
+    let mut result = String::new();
+
+    for (i, array) in nested_array.iter().enumerate() {
+        if i > 0 {
```
cbgbt: This is quite efficient, plus the tests give me confidence. As a future note, I think a quite "rust-y" way to accomplish this may have been to use iterators + [`Vec::join`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.join), e.g. something like:


```rust
let nested_array = nested_array_param.as_array().with_context(/* etc */)?;
let nested_array: Vec<String> = nested_array.into_iter()
    .map(|inner_array| {
        let inner_array = inner_array.as_array().with_context(/* etc */)?;
        let inner_array: Vec<String> = inner_array.into_iter()
            .map(|value| format!("\"{}\""), value.as_str().context(/* etc */))
            .collect::<Result<Vec<_>>>()?;
        Ok(inner_array.join(", "))
    })
    .collect::<Result<Vec<_>>>()?;
let result = nested_array.join(", ");

File: packages/os/bootstrap-commands-toml (hunk 13 above)
Comment (cbgbt) [reply packages/os/bootstrap-commands-toml]: I think this has the capacity to output invalid `toml` -- if the inputs have any characters in `[]"`, then there may be problems.

I believe you can use `'` for raw strings in TOML, and triple-single-quotes for raw strings which allow `'` as well.

Might be a good idea to write a test case in the `bootstrap-commands` binary which checks against input like this.

Comment (cbgbt) [reply packages/os/bootstrap-commands-toml]: As an alternative to `join_nested_array`, I wonder if you could kill two birds with one stone and implement it in the template like this:

```
{{#each settings.bootstrap-commands}}
[bootstrap-commands."{{@key}}"]
commands = [
{{#each this.commands}}
    [{{#each this}}'''{{ this }}''',{{/each}}],
{{/each}}
]
{{/each}}
```

It seems to [work on the handlebars playground](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22commands%20%3D%20%5B%5Cn%7B%7B%23each%20commands%7D%7D%5Cn%20%20%5B%7B%7B%23each%20this%7D%7D'''%7B%7B%20this%20%7D%7D'''%2C%7B%7B%2Feach%7D%7D%5D%2C%5Cn%7B%7B%2Feach%7D%7D%5Cn%5D%5Cn%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%20commands%3A%20%5B%5B%5C%22apiclient%5C%22%2C%20%5C%22get%5C%22%2C%20%5C%22settings.motd%5C%22%5D%2C%20%5B%5C%22apiclient%5C%22%2C%20%5C%22set%5C%22%2C%20%5C%22settings.motd%3Dhi%5C%22%5D%5D%20%7D%5Cn%22%2C%22output%22%3A%22commands%20%3D%20%5B%5Cn%20%20%5B'''apiclient'''%2C'''get'''%2C'''settings.motd'''%2C%5D%2C%5Cn%20%20%5B'''apiclient'''%2C'''set'''%2C'''settings.motd%26%23x3D%3Bhi'''%2C%5D%2C%5Cn%5D%5Cn%22%2C%22preparationScript%22%3A%22Handlebars.registerHelper('loud'%2C%20function%20(aString)%20%7B%5Cn%20%20%20%20return%20aString.toUpperCase()%5Cn%7D)%5Cn%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D), other than the unusual URL encoding. Also seems to parse properly per [this site](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22commands%20%3D%20%5B%5Cn%7B%7B%23each%20commands%7D%7D%5Cn%20%20%5B%7B%7B%23each%20this%7D%7D'''%7B%7B%20this%20%7D%7D'''%2C%7B%7B%2Feach%7D%7D%5D%2C%5Cn%7B%7B%2Feach%7D%7D%5Cn%5D%5Cn%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%20commands%3A%20%5B%5B%5C%22apiclient%5C%22%2C%20%5C%22get%5C%22%2C%20%5C%22settings.motd%5C%22%5D%2C%20%5B%5C%22apiclient%5C%22%2C%20%5C%22set%5C%22%2C%20%5C%22settings.motd%3Dhi%5C%22%5D%5D%20%7D%5Cn%22%2C%22output%22%3A%22commands%20%3D%20%5B%5Cn%20%20%5B'''apiclient'''%2C'''get'''%2C'''settings.motd'''%2C%5D%2C%5Cn%20%20%5B'''apiclient'''%2C'''set'''%2C'''settings.motd%26%23x3D%3Bhi'''%2C%5D%2C%5Cn%5D%5Cn%22%2C%22preparationScript%22%3A%22Handlebars.registerHelper('loud'%2C%20function%20(aString)%20%7B%5Cn%20%20%20%20return%20aString.toUpperCase()%5Cn%7D)%5Cn%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D).

File: sources/api/bootstrap-commands/src/main.rs (hunk 19 above)
Comment (piyush-jena) [reply sources/api/bootstrap-commands/src/main.rs]: [Link](https://doc.rust-lang.org/std/process/struct.Command.html#method.status) status() inherits stdout and stderr of the parent.

Comment (cbgbt) [reply sources/api/bootstrap-commands/src/main.rs]: Oh, I didn't know! Nice.

**Comment by cbgbt** [comment]
>It seems to [work on the handlebars playground](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22commands%20%3D%20%5B%5Cn%7B%7B%23each%20commands%7D%7D%5Cn%20%20%5B%7B%7B%23each%20this%7D%7D'''%7B%7B%20this%20%7D%7D'''%2C%7B%7B%2Feach%7D%7D%5D%2C%5Cn%7B%7B%2Feach%7D%7D%5Cn%5D%5Cn%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%20commands%3A%20%5B%5B%5C%22apiclient%5C%22%2C%20%5C%22get%5C%22%2C%20%5C%22settings.motd%5C%22%5D%2C%20%5B%5C%22apiclient%5C%22%2C%20%5C%22set%5C%22%2C%20%5C%22settings.motd%3Dhi%5C%22%5D%5D%20%7D%5Cn%22%2C%22output%22%3A%22commands%20%3D%20%5B%5Cn%20%20%5B'''apiclient'''%2C'''get'''%2C'''settings.motd'''%2C%5D%2C%5Cn%20%20%5B'''apiclient'''%2C'''set'''%2C'''settings.motd%26%23x3D%3Bhi'''%2C%5D%2C%5Cn%5D%5Cn%22%2C%22preparationScript%22%3A%22Handlebars.registerHelper('loud'%2C%20function%20(aString)%20%7B%5Cn%20%20%20%20return%20aString.toUpperCase()%5Cn%7D)%5Cn%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D), other than the unusual URL encoding. Also seems to parse properly per [this site](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22commands%20%3D%20%5B%5Cn%7B%7B%23each%20commands%7D%7D%5Cn%20%20%5B%7B%7B%23each%20this%7D%7D'''%7B%7B%20this%20%7D%7D'''%2C%7B%7B%2Feach%7D%7D%5D%2C%5Cn%7B%7B%2Feach%7D%7D%5Cn%5D%5Cn%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%20commands%3A%20%5B%5B%5C%22apiclient%5C%22%2C%20%5C%22get%5C%22%2C%20%5C%22settings.motd%5C%22%5D%2C%20%5B%5C%22apiclient%5C%22%2C%20%5C%22set%5C%22%2C%20%5C%22settings.motd%3Dhi%5C%22%5D%5D%20%7D%5Cn%22%2C%22output%22%3A%22commands%20%3D%20%5B%5Cn%20%20%5B'''apiclient'''%2C'''get'''%2C'''settings.motd'''%2C%5D%2C%5Cn%20%20%5B'''apiclient'''%2C'''set'''%2C'''settings.motd%26%23x3D%3Bhi'''%2C%5D%2C%5Cn%5D%5Cn%22%2C%22preparationScript%22%3A%22Handlebars.registerHelper('loud'%2C%20function%20(aString)%20%7B%5Cn%20%20%20%20return%20aString.toUpperCase()%5Cn%7D)%5Cn%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D).


I figured out the encoding bit. Apparently handlebars HTML-encodes data by default with the "double-stache". [If you "triple-stache", it embeds it directly](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22commands%20%3D%20%5B%5Cn%7B%7B%23each%20commands%7D%7D%5Cn%20%20%5B%7B%7B%23each%20this%7D%7D'''%7B%7B%7B%20this%20%7D%7D%7D'''%2C%7B%7B%2Feach%7D%7D%5D%2C%5Cn%7B%7B%2Feach%7D%7D%5Cn%5D%5Cn%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%20commands%3A%20%5B%5B%5C%22apiclient%5C%22%2C%20%5C%22get%5C%22%2C%20%5C%22settings.motd%5C%22%5D%2C%20%5B%5C%22apiclient%5C%22%2C%20%5C%22set%5C%22%2C%20%5C%22settings.motd%3Dhi%5C%22%5D%5D%20%7D%5Cn%22%2C%22output%22%3A%22commands%20%3D%20%5B%5Cn%20%20%5B'''apiclient'''%2C'''get'''%2C'''settings.motd'''%2C%5D%2C%5Cn%20%20%5B'''apiclient'''%2C'''set'''%2C'''settings.motd%3Dhi'''%2C%5D%2C%5Cn%5D%5Cn%22%2C%22preparationScript%22%3A%22Handlebars.registerHelper('loud'%2C%20function%20(aString)%20%7B%5Cn%20%20%20%20return%20aString.toUpperCase()%5Cn%7D)%5Cn%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D).

File: packages/os/bootstrap-commands.service [hunk 28]
```diff
@@ -0,0 +1,17 @@
...
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/bootstrap-commands
+RemainAfterExit=true
+StandardError=journal+console
+SyslogIdentifier=bootstrap-commands
+
+[Install]
+WantedBy=preconfigured.target
```
bcressey: This needs to be `RequiredBy` or else failures will not stop the system from reaching `preconfigured.target`.
```suggestion
[Install]
RequiredBy=preconfigured.target
```

packages/os/os.spec
```diff
@@ -89,6 +92,7 @@ Requires: %{_cross_os}sundog
 Requires: %{_cross_os}xfscli
 Requires: %{_cross_os}thar-be-settings
 
+Requires: %{_cross_os}bootstrap-commands
```
bcressey: nit: this can go in the almost alphabetized list above of unconditional dependencies

sources/Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "1"
 members = [
     "api/apiserver",
     "api/apiclient",
+    "api/bootstrap-commands",
```
bcressey: I'd prefer not to put this under "api"; historically, that was the collection of crates that depended directly on API access, which isn't the case for this program.

(Several of the other crates here need to be moved out, but that cleanup can happen separately.)

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
+/*
```
bcressey: This needs to be a doc comment, or it won't populate the template used to render `README.md`
```suggestion
/*!
```

packages/os/bootstrap-commands-toml
```diff
@@ -0,0 +1,12 @@
...
+mode = "{{{this.mode}}}"
+essential = {{this.essential}}
```
bcressey: `commands`, `mode`, and `essential` are not guaranteed to be set - it's possible that only one is set. You need to guard each individual field with `{{#if_not_null ... }}`, in addition to the top-level one.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+Bootstrap Commands in Bottlerocket is a feature that empowers users to execute predefined commands
+through the Bottlerocket API. It introduces a new systemd service that facilitates the execution of
+these commands, expanding the functionality of the existing Bottlerocket API.
+
+Currently, Bottlerocket provides a set of actions (such as refresh-updates) that can be accessed
+through the apiserver, typically by running the apiclient in the control container. However, this
+new feature aims to extend the accessibility of these actions by allowing Bottlerocket consumers
+to run them prior to the bootstrap container execution.
+
+#Example:
```
bcressey:
```suggestion
# Example
```

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+# Bootstrap commands
+
+Bootstrap Commands in Bottlerocket is a feature that empowers users to execute predefined commands
+through the Bottlerocket API. It introduces a new systemd service that facilitates the execution of
+these commands, expanding the functionality of the existing Bottlerocket API.
+
+Currently, Bottlerocket provides a set of actions (such as refresh-updates) that can be accessed
+through the apiserver, typically by running the apiclient in the control container. However, this
+new feature aims to extend the accessibility of these actions by allowing Bottlerocket consumers
+to run them prior to the bootstrap container execution.
```
bcressey: Style-wise, I would mimic the style of the `bootstrap-containers/src/main.rs` - describe what the program does at a high-level, in a way that's useful for a developer working on the project and needs context on what this program is supposed to do.

Save the larger context for the website. This program is part of implementing the feature for Bottlerocket, but it is not the feature.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+
+#Example:
+Given a bootstrap command called `001-test-bootstrap-commands` with the following configuration:
+
+[settings.bootstrap-commands.001-test-bootstrap-commands]
+commands = [[ "apiclient", "set", "motd=helloworld"]]
+essential = true
+mode = "always"
+
+This would update motd from "Welcome to Bottlerocket!" to "helloworld".
```
bcressey: nit: avoid referring to the default motd value in case it changes:
```suggestion
This would set `/etc/motd` to "helloworld".
```

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+/// Handles running of the api commands in a Bootstrap command
+fn handle_bootstrap_command<S>(name: S, bootstrap_command: &BootstrapCommand) -> Result<()>
+where
+    S: AsRef<str>,
+{
+    let name = name.as_ref();
+    let mode = bootstrap_command.mode.clone().unwrap_or_default();
+    let commands = bootstrap_command.commands.clone().unwrap_or_default();
+
+    info!("Running bootstrap command with name: {}", name);
```
bcressey: I don't think we should emit this message unless we're actually running the command. Otherwise the logs may be misleading if someone just greps for "Running bootstrap command".

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+            run_command(cmd.as_str(), arg_str)?;
+        }
+
+        if mode == "once" {
+            let formatted = format!("settings.bootstrap-commands.{}.mode=off", name);
+            info!("Turning off bootstrap-command '{}'", name);
+            run_command("apiclient", ["set", formatted.as_str()])?;
+        }
+
+        info!("Succesfully ran bootstrap command '{}'", name);
```
bcressey:
```suggestion
        info!("Successfully ran bootstrap command '{}'", name);
```

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    command
+        .args(args)
+        .status()
+        .context(error::ExecutionFailureSnafu { command })?;
+
+    Ok(())
+}
+
+/// Handles running of the api commands in a Bootstrap command
```
bcressey: nit: this comment doesn't add much value, it just restates the function name

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+            let (cmd, args) = command.get_args();
+            let arg_str: Vec<&str> = args.iter().map(String::as_str).collect();
+            run_command(cmd.as_str(), arg_str)?;
+        }
+
+        if mode == "once" {
+            let formatted = format!("settings.bootstrap-commands.{}.mode=off", name);
+            info!("Turning off bootstrap-command '{}'", name);
+            run_command("apiclient", ["set", formatted.as_str()])?;
+        }
```
bcressey: The problem with this approach is that if one of the commands is `apiclient reboot`, then we may not finish changing the mode before the reboot kicks in. Same (but more subtly) if the command is `apiclient update apply -c -r` for a one-time update at boot.

To fix this, you can do something like:
1. Check for `/var/lib/bottlerocket/bootstrap-commands@<name>.ran` prior to executing a "once" command. If found, skip the command.
2. Create `/var/lib/bottlerocket/bootstrap-commands@<name>.ran` prior to executing a "once" command. This records that the command has been initiated.
3. For any "off" command, remove the marker file if it exists, as the user might later re-enable it as a "once" command and expect it to run again.

This will satisfy the first part of the contract - making sure a command only runs once.

For the second part, updating the API settings to "off" to reflect that the commands have run, you could maintain a set of bootstrap commands to disable, and disable them all at once prior to returning a value from the overall program execution.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+    for (name, bootstrap_command) in bootstrap_commands.iter() {
+        let essential = bootstrap_command.essential.unwrap_or(false);
+        let status = handle_bootstrap_command(name, bootstrap_command);
+
+        ensure!(
+            !essential || status.is_ok(),
+            error::BootstrapCommandFailureSnafu {
+                name: name.as_ref()
+            }
+        )
```
bcressey: This intersects with the "once" behavior in a strange way: if the first set of commands is marked as "essential" and "once", and it fails, then the boot should fail.

But, we have to decide what to do on a subsequent boot:
1. We already ran the command once, so we don't have to run it again, even though it wasn't successful. This exposes later commands in the list to an unexpected state - maybe they expect previous commands to have all been executed successfully, but on the next boot we'll skip over the failed ones.
2. We ran the command already, but it failed, so we need to run it again. This implies "once" means "one successful execution". This is easier for later commands in the list to reason about - they can be sure that previous essential commands were all executed successfully.

I lean toward (2) as it's consistent with how bootstrap containers are handled; we don't disable a "once" container unless it returns success.

Relative to my earlier comment, you can implement this by creating `/var/lib/bottlerocket/bootstrap-commands@<name>.ran` after the command returns success, rather than before.

sources/api/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Failed to deserialize settings from config at {}: {}", config_path.display(), source))]
+        Deserialization {
+            config_path: PathBuf,
+            source: toml::de::Error,
+        },
+
+        #[snafu(display("Essential Bootstrap command with name: {} failed.", name))]
```
bcressey: nit: avoid saying "name" in error messages, and just give the quoted identifier:
```suggestion
        #[snafu(display("Essential bootstrap command '{}' failed.", name))]
```

sources/api/bootstrap-commands/src/main.rs
piyush-jena: Updated it to `BootstrapCommandExecutionSnafu`

packages/os/bootstrap-commands-toml
piyush-jena: I solved it with triple curly braces for now but I would like to spend some more time on this to think about a better way to do this to tackle corner cases that we discussed.

sources/api/bootstrap-commands/src/main.rs
bcressey: > Relative to my earlier comment, you can implement this by creating /var/lib/bottlerocket/bootstrap-commands@<name>.ran after the command returns success, rather than before.

Ugh, but this re-introduces the problem of "what if the command causes a reboot".

I would look at [systemd-inhibit](https://www.freedesktop.org/software/systemd/man/latest/systemd-inhibit.html) instead - if you wrap the entire `bootstrap-commands` program call with this, it should inhibit any reboots that might be initiated until you exit.

That should avoid any need to deal with marker files and the like.

packages/os/bootstrap-commands-toml
```diff
@@ -0,0 +1,22 @@
...
+std = { version = "v1", helpers = ["if_not_null", "join_nested_array"]}
++++
+{{#if_not_null settings.bootstrap-commands}}
+{{#each settings.bootstrap-commands}}
+[bootstrap-commands."{{@key}}"]
+{{#if_not_null this.commands}}
+commands = [
+{{#each this.commands}}
+    [{{#each this}}'''{{{ this }}}''',{{/each}}],
+{{/each}}
```
cbgbt: I think this will work for most cases, unless a user uses `'''` for some reason. That would be unusual but not inconceivable.

There are two other ideas I had:
* Write a helper which encodes an input `serde_json::Value` as an inline toml value using [the toml crate](https://docs.rs/toml/).
* Embrace HTML escaping, go back to using "double stache", and then have the `bootstrap-commands` binary decode the HTML-escaped input using [the `html-escape` crate](https://crates.io/crates/html-escape)

sources/api/schnauzer/src/helpers.rs
piyush-jena: Replaced join_nested_array with toml_encode as discussed.

sources/api/bootstrap-commands/src/main.rs
piyush-jena: We opted on solving this with systemd-inhibit.

sources/api/bootstrap-commands/src/main.rs
piyush-jena: Resolved.

packages/os/bootstrap-commands-toml
```diff
@@ -0,0 +1,22 @@
...
++++
+{{#if_not_null settings.bootstrap-commands}}
+{{#each settings.bootstrap-commands}}
+[bootstrap-commands."{{@key}}"]
+{{#if_not_null this.commands}}
+commands = [
+{{#each this.commands}}
+    {{ toml_encode this }},
+{{/each}}
+]
```
cbgbt: No need to iterate anymore, now that you have `toml_encode`!

```suggestion
commands = {{ toml_encode this.commands }}
```

packages/os/os.spec
```diff
@@ -246,6 +250,12 @@ Requires: %{_cross_os}binutils
 %description -n %{_cross_os}driverdog
 %{summary}.
 
+
```
bcressey: nit: unnecessary extra newline
```suggestion
```

packages/os/os.spec
```diff
@@ -640,6 +652,13 @@ install -p -m 0644 %{S:400} %{S:401} %{S:402} %{buildroot}%{_cross_licensedir}
 %{_cross_bindir}/certdog
 %{_cross_templatedir}/certdog-toml
 
+
```
```suggestion
```

packages/systemd/systemd.spec
```diff
@@ -14,6 +14,7 @@ Source3: journald.conf
 Source4: issue
 Source5: systemd-journald.conf
 Source6: systemd-sysusers.conf
+Source7: systemd-logind.conf
```
bcressey: nit: I would prefer to put this in `release` where we've collected other Bottlerocket-specific configuration files for `systemd`.

packages/systemd/systemd-logind.conf
```diff
@@ -0,0 +1,2 @@
+[Login]
+InhibitDelayMaxSec=300
```
bcressey: Can you add a comment on why this is being set to 5 minutes?

sources/api/schnauzer/src/helpers/stdlib/mod.rs
```diff
@@ -1130,3 +1130,127 @@ mod test_negate_or_else {
         });
     }
 }
+
+/// `toml_encode` accepts arbitrary input and encodes it as a toml string
```
bcressey, line 1134: Can you split the new `toml_encode` helper into a different commit?

packages/os/bootstrap-commands.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Bootstrap Commands
+# We depend on systemd-logind.service for running systemd-inhibit.
+After=systemd-logind.service settings-applier.service apiserver.service
+Requires=systemd-logind.service settings-applier.service apiserver.service
+Before=preconfigured.target
```
bcressey: This is a default dependency because it's required by `preconfigured.target`:
```suggestion
```

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+configures the system by going through all the bootstrap commands in lexicographical order and
+running all the commands inside it.
+
+## Example:
+Given a bootstrap command called `001-test-bootstrap-commands` with the following configuration:
+
+[settings.bootstrap-commands.001-test-bootstrap-commands]
+commands = [[ "apiclient", "set", "motd=helloworld"]]
+essential = true
+mode = "always"
```
bcressey, line 19: Can you apply a code format block to these examples?

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+
+This would set `/etc/motd` to "helloworld".
+
+# Additional Information:
+Certain valid apiclient commands that work in a session may fail in bootstrap-commands due to
```
bcressey: nit: let's code-format mentions of `apiclient`, `bootstrap-commands` and `bootstrap-commands.service` , so it's clear that we're talking about specific programs or units and not something more general.
```suggestion
Certain valid `apiclient` commands that work in a session may fail in `bootstrap-commands` due to
```

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
+/*!
+# Bootstrap commands
+
+bootstrap-commands ensures that Bootstrap Commands are executed as defined in the system
+settings. It is called by bootstrap-commands systemd service which runs prior to the
+execution of bootstrap-containers.
```
bcressey:
```suggestion
`bootstrap-commands` ensures that Bootstrap Commands are executed as defined in the system
settings. It is called by `bootstrap-commands.service` which runs prior to the
execution of `bootstrap-containers`.
```

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+
+Each bootstrap command is a set of commands. The service queries the API for the settings, then
```
bcressey: "The service queries the API for the settings" is not accurate; the settings are rendered into a config file.

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+
+## Example:
+[settings.bootstrap-commands.001-test-bootstrap-commands]
+commands = [[ "apiclient", "exec", "admin", "ls"]]
+essential = true
+mode = "always"
+
+This command fails because bootstrap-commands.service which calls this binary is launched prior
+to preconfigured.target while host-containers@.service which is a requirement for running "exec"
+commands are launched after preconfigured.target.
```
bcressey: This doesn't feel like the right place to put commentary on how the systemd units are configured, which is handled elsewhere. That said, this is also where `README.md` comes from, so I understand why the content is here and maybe there's not a better place for it.

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+where
+    S: AsRef<str>,
+{
+    let name = name.as_ref();
+    let mode = bootstrap_command.mode.as_ref();
+    let commands = &bootstrap_command.commands;
+
+    if mode == "off" {
+        // If mode is 'off', just log this information.
+        info!("Bootstrap command mode for '{}' is 'off'", name);
```
bcressey, line 121: nit: can do an early return to outdent the remaining code:
```suggestion
    if mode == "off" {
        // If mode is 'off', just log this information.
        info!("Bootstrap command mode for '{}' is 'off'", name);
        return  Ok(());
    }
```

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+    } else {
+        info!("Processing Bootstrap command '{}' ... ", name);
```
bcressey:
```suggestion
        info!("Processing bootstrap command '{}' ... ", name);
```

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+
+        for command in commands.iter() {
+            let (cmd, args) = command.get_command_and_args();
+            run_command(cmd, args)?;
+        }
+
+        if mode == "once" {
+            let formatted = format!("settings.bootstrap-commands.{}.mode=off", name);
+            info!("Turning off bootstrap-command '{}'", name);
```
bcressey:
```suggestion
            info!("Turning off bootstrap command '{}'", name);
```

sources/bootstrap-commands/src/main.rs
piyush-jena: This addition was to resolve this [comment](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/46#discussion_r1733637534). This restriction seemed like something that the user should be informed which is why I wanted to add it in README.

piyush-jena:
Forced push addresses the following:
1. Changes to `sources/bootstrap-commands/README.md`
2. nit comments in `sources/bootstrap-commands/main.rs`
3. Removes unnecessary `Before=preconfigured.target` in `packages/os/bootstrap-commands.service`

piyush-jena:
Forced push addresses the following:
1. Split the commits to separate out `schnauzer` and `bootstrap-commands` changes.
2. Moved drop-in config for logind from `systemd.spec` to `release.spec`

bcressey (approved):

sources/bootstrap-commands/src/main.rs
```diff
@@ -0,0 +1,370 @@
...
+    #[serde(default)]
+    bootstrap_commands: BTreeMap<Identifier, BootstrapCommand>,
+}
+
+impl BootstrapCommandConfig {
+    // Gets an iterator for bootstrap_commands, sorted in lexicographical order of their names.
+    fn iter(self) -> impl Iterator<Item = (Identifier, BootstrapCommand)> {
+        self.bootstrap_commands.into_iter()
+    }
+}
```
cbgbt, line 67: I think this is fine, but as a Rust tip, there's the [`IntoIter`](https://doc.rust-lang.org/stable/std/iter/trait.IntoIterator.html) in `std` that you can implement for this behavior, which can make calling other APIs that expect `IntoIter` ergonomic.

cbgbt (approved):
Nice work! It's probably worth documenting how reboots are handled -- that might be best on the website.

piyush-jena:
forced push removes the hack commit to use the official settings-sdk

arnaldo2792 (approved):
Nice job! :tada:

PR #65, opened 2024-08-02
try to figure out what breaks cached builds

bcressey:
**Issue number:**
N/A

**Description of changes:**
Not for merging - trying to see why `cargo` is unhappy reusing builds from the cache.

**Testing done:**
😬 

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,11 +44,12 @@ jobs:
       - uses: actions/checkout@v4
       - name: Preflight step to set up the runner
         uses: ./.github/actions/setup-node
-      - run: rustup component add rustfmt
-      - run: make twoliter unit-tests
-      # TODO: fixme please!
-      # Avoid running Go lint check via `cargo make check-lints` since there's a separate golangci-lint workflow
-      # - run: make twoliter check-fmt
-      # - run: make twoliter check-clippy
-      - run: make twoliter check-shell
-      - run: make ARCH="${{ matrix.arch }}"
+      - name: Inspect cache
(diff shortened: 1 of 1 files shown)


bcressey:
There are at least two problems here.

The first is that the fingerprints in the cache don't match the fingerprints that `cargo` generates, and some files seem to be missing:
```
# from cache
target/x86_64/debug/.fingerprint/microcode-203e868d1235a310
target/x86_64/debug/.fingerprint/microcode-3d5c23d8c6a2f178
target/x86_64/debug/.fingerprint/microcode-82b3ac875abbc53a

# cargo output
target/x86_64/debug/.fingerprint/microcode-203e868d1235a310/build-script-build-script-build
target/x86_64/debug/.fingerprint/microcode-06259485b1fc1ba5/run-build-script-build-script-build
target/x86_64/debug/.fingerprint/microcode-79bfebf693197603/lib-microcode
```

The second is that the mtime is reset for all files to the time of the clone:
```
  File: packages/microcode/latest-upstream-tags.sh
  Size: 355       	Blocks: 8          IO Block: 4096   regular file
Device: 801h/2049d	Inode: 9300713     Links: 1
Access: (0755/-rwxr-xr-x)  Uid: ( 1000/  runner)   Gid: ( 1000/  runner)
Access: 2024-08-02 21:43:03.524858363 +0000
Modify: 2024-08-02 21:43:03.524858363 +0000
Change: 2024-08-02 21:43:03.524858363 +0000
 Birth: 2024-08-02 21:43:03.524858363 +0000
```

PR #69, opened 2024-08-08
Re-enable `fmt` and `licenses` lints in CI

cbgbt:
**Description of changes:**
Re-enable `fmt` and `license` lints.

I'd like to re-enable clippy, but the `dead_code` lint is starting to detect our error enum variants if we haven't been careful about conditional compilation or in cases where modules are owned by both a `lib` and a `bin` but only a subset of the error variants are used in either. This will take some careful changes to do the right way, but we should really have the `licenses` lint enabled ASAP.

**Testing done:**
* The lints pass

**Terms of contribution:**

### DIFF
--- a/sources/deny.toml
+++ b/sources/deny.toml
@@ -55,6 +55,8 @@ deny = [{ name = "structopt" }, { name = "clap", wrappers = ["cargo-readme"] }]
 skip = [
     # this can be removed once settings sdk is updated to base64 0.22.1
     { name = "base64", version = "=0.21.7" },
+    # this can be removed once settings sdk is updated to env_logger 0.11
+    { name = "env_logger", version = "=0.10.2" },
 ]
 
 skip-tree = [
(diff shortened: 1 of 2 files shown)


bcressey (approved):

sources/deny.toml
```diff
@@ -55,6 +55,8 @@ deny = [{ name = "structopt" }, { name = "clap", wrappers = ["cargo-readme"] }]
 skip = [
     # this can be removed once settings sdk is updated to base64 0.22.1
     { name = "base64", version = "=0.21.7" },
+    # this can be removed once settings sdk is updated to env_logger 0.11
```
arnaldo2792, line 58: Can we have a PR with the update in the Bottlerocket Settings SDK? :D

arnaldo2792 (approved):

sources/deny.toml
cbgbt: Already done! https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/50

PR #70, opened 2024-08-08
Use workspace dependencies for all dependencies

cbgbt:
**Description of changes:**
Most of the heavy lifting here was done by a modified fork of [cargo-autoinherit](https://github.com/cbgbt/cargo-autoinherit/tree/cbgbt-v0.1.6), save for some light formatting. I submitted my patches upstream, but it's not clear yet if they'll take them.

I'm currently working on updating `cargo-deny` in the SDK, which now includes a linter that can deny `workspace-dependency` antipatterns.

This PR is split into two commits, with my goal being that the `workspace` commit does not change Cargo.lock at all.

---
    move all dependencies to workspace dependencies
---
    sources: homogenize dependencies

    This fixes a few inconsistencies in dependency requirements:
    * base64 - this is the only semantic change in the commit. Two packages
      depended on a slightly older version of base64. cargo-deny did not
      catch this because we have disabled duplicate detection for base64 to
      accommodate other 3rd party packages.
    * tempfile - This package does not have any default features. Remove the
      `default-features = false` flag.
    * schnauzer - One package had an arbitrarily more-specific version
      requirement for schnauzer.
---

**Testing done:**
* Note that `Cargo.lock` has not changed upon the switch to workspace dependencies.
* Builds succeed

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -76,5 +76,142 @@ members = [
     "xfscli",
 ]
 
+[workspace.dependencies]
+apiclient = { version = "0.1", path = "api/apiclient" }
+block-party = { version = "0.1", path = "updater/block-party" }
+bottlerocket-release = { version = "0.1", path = "bottlerocket-release" }
+constants = { version = "0.1", path = "constants" }
+datastore = { version = "0.1", path = "api/datastore" }
+dogtag = { version = "0.1", path = "dogtag" }
+early-boot-config-provider = { version = "0.1", path = "early-boot-config/early-boot-config-provider" }
+generate-readme = { version = "0.1", path = "generate-readme" }
+imdsclient = { version = "0.1", path = "imdsclient" }
+models = { version = "0.1", path = "models" }
+parse-datetime = { version = "0.1", path = "parse-datetime" }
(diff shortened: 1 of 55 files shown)


sources/Cargo.toml
```diff
@@ -76,5 +76,142 @@ members = [
...
+serde = "1"
+serde-xml-rs = "0.6"
+serde_json = "1"
+serde_plain = "1"
+serde_yaml = "0.9"
+shell-words = "1"
+shlex = "1"
+signal-hook = "0.3"
+simplelog = "0.12"
+snafu = { version = "0.8", default-features = false }
```
bcressey: If we specify it this way to enable default features, wouldn't that simplify crate level usage?
```suggestion
snafu = "0.8"
```

Seems like everything except `signpost` uses the default features, and `signpost` should be fine with that instead.

bcressey (approved):

cbgbt:
^ Force push to address @bcressey's suggestion for `snafu` (and then another to correct a mistake I made)

bcressey (approved):

yeazelm (approved):

PR #74, opened 2024-08-09
kubernetes: apply upstream patches.

larvacea:
Apply upstream patches for kubernetes-1.24, kubernetes-1.25, and kubernetes-1.26, to mitigate CVE-2024-5321.

**Description of changes:**

EKS-D published patches for versions 1.24, 1.25, and 1.26.

**Testing done:**

Built core kit (both architectures) and variants (also both architectures), and verified that the images boot and the kubernetes services attempt to start. Did not create a cluster to run the 6 AMIs.

**Terms of contribution:**


koooosh (commented):
Only nit I have is if we should split the commit into 3, one for each kubernetes version pkg.

packages/kubernetes-1.25/Cargo.toml
```diff
@@ -101,6 +101,9 @@ sha512 = "ceb1a7844e4e2ed5fa3a838949086ebf86114bba8007a3125ea4ebd6aba7e790eb19b1
 url = "https://raw.githubusercontent.com/aws/eks-distro/71aed3084474ab65027b067d45e70081144384ea/projects/kubernetes/kubernetes/1-25/patches/0022-EKS-PATCH-Update-aws-sdk-go-to-include-new-regions.patch"
 sha512 = "e226f5a70aeafa25ff010c607b1e7913b562f37674062a0aadbe77fd62532dd5bab10c3bfc38af455a00232da0548f82cd77a86d124b17149d5f9e4bcc96fa30"
 
+[[package.metadata.build-package.external-files]]
```
yeazelm, line 104: Please break these three updates into 3 commits since they are going to different packages.

PR #80, opened 2024-08-12
Update datastore serializer to expect JSON and correctly handle null values

sam-berning:
**Issue number:**

This is a more robust/comprehensive fix to the first issue seen in https://github.com/bottlerocket-os/bottlerocket/issues/4135

**Description of changes:**

Previously, the datastore serializer handled basically any serializable data type, which allowed us to use a range of input types, but also required us to be very generic in the implementation, making it difficult to special-case particular types.

As we move towards Out-of-Tree Builds and Settings Extensions, allowing this broad range of types as an input to the datastore serializer is actually becoming less helpful, since settings will be passed around as a more generic type that will be validated against a more concrete type on set.

Because of this, it makes sense to limit the input space of the Serializer to only allow this generic type, and to ensure that all edge cases for this generic type are handled correctly. Since Settings Plugins do all of their serialization and deserialization across the FFI border using JSON, it makes sense for JSON to be this generic type.

After we've limited the datastore serializer to only accept JSON, implementors of Settings Plugins and Settings Extensions can just focus on ensuring that their types serialize correctly to JSON, and they can be sure that it will serialize correctly to the datastore format for free.

_A note for discussion: now that the datastore serializer has been limited to only accept JSON, there is serialization logic for a broader range of types than necessary. Does make sense to clean up these unused types to help hint that this serializer should only be used with `serde_json::Value`?_

**Testing done:**

I added a bunch of unit tests to the datastore serializer to ensure that all of the types in `serde_json::Value` serialize correctly.

Then, I built the core-kit, and built an AMI using my core-kit to ensure that there was no strange behavior. The instance I launched booted fine, and was able to handle settings correctly.

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/controller.rs
+++ b/sources/api/apiserver/src/server/controller.rs
@@ -136,7 +136,7 @@ where
     let os = BottlerocketRelease::new().context(error::ReleaseDataSnafu)?;
 
     // Turn into a serde Value we can manipulate.
-    let val = serde_json::to_value(os).expect("struct to value can't fail");
+    let val = serde_json::to_value(os).context(error::SettingsToJsonSnafu)?;
 
     // Structs are Objects in serde_json, which have a map of field -> value inside.  We
     // destructure to get it by value, instead of as_object() which gives references.
@@ -366,7 +366,8 @@ pub(crate) fn set_settings<D: DataStore>(
     transaction: &str,
 ) -> Result<()> {
     trace!("Serializing Settings to write to data store");
-    let pairs = to_pairs_with_prefix("settings", settings)
+    let settings_json = serde_json::to_value(settings).expect("struct to value can't fail");
+    let pairs = to_pairs_with_prefix("settings", &settings_json)
         .context(error::DataStoreSerializationSnafu { given: "Settings" })?;
     let pending = Committed::Pending {
         tx: transaction.into(),
--- a/sources/api/datastore/src/serialization/pairs.rs
+++ b/sources/api/datastore/src/serialization/pairs.rs
@@ -21,7 +21,7 @@ use crate::{serialize_scalar, Key, KeyType, ScalarError};
 ///    Settings -> DockerSettings -> bridge_ip = u64
 /// would turn into a key of "settings.docker-settings.bridge-ip" and a serialized String
 /// representing the u64 data.
-pub fn to_pairs<T: Serialize>(value: &T) -> Result<HashMap<Key, String>> {
+pub fn to_pairs(value: &serde_json::Value) -> Result<HashMap<Key, String>> {
     let mut output = HashMap::new();
     let serializer = Serializer::new(&mut output, None);
     value.serialize(serializer)?;
@@ -30,10 +30,9 @@ pub fn to_pairs<T: Serialize>(value: &T) -> Result<HashMap<Key, String>> {
 
 /// Like to_pairs, but lets you add an arbitrary prefix to the resulting keys.  A separator will
 /// automatically be added after the prefix.
-pub fn to_pairs_with_prefix<S, T>(prefix: S, value: &T) -> Result<HashMap<Key, String>>
+pub fn to_pairs_with_prefix<S>(prefix: S, value: &serde_json::Value) -> Result<HashMap<Key, String>>
 where
     S: AsRef<str>,
-    T: Serialize,
 {
     let prefix = prefix.as_ref();
     let prefix_key = Key::new(KeyType::Data, prefix).map_err(|e| {
@@ -516,12 +515,13 @@ mod test {
             list: vec![3, 4, 5],
             boolean: true,
         };
-        let keys = to_pairs(&b).unwrap();
+        let j = serde_json::to_value(b).unwrap();
+        let keys = to_pairs(&j).unwrap();
         assert_eq!(
             keys,
             hashmap!(
-                key!("B.list") => "[3,4,5]".to_string(),
-                key!("B.boolean") => "true".to_string(),
+                key!("list") => "[3,4,5]".to_string(),
+                key!("boolean") => "true".to_string(),
             )
         );
     }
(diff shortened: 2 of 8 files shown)


sources/api/sundog/src/main.rs
```diff
@@ -218,15 +218,10 @@ where
         .context(error::GetPrefixSnafu { prefixes })?;
     debug!("API response model: {}", response.to_string());
 
-    // Build a Settings struct from the response.
-    let settings = serde_json::from_value::<model::Model>(response)
-        .context(error::InterpretModelSnafu)?
-        .settings;
```
sam-berning, line 224: I am assuming here that because the settings JSON comes from apiclient that it should contain only valid settings, so there's no need to convert to and from Settings just to verify this.

sam-berning:
^ fix fmt

sources/api/sundog/src/main.rs
```diff
@@ -218,15 +218,10 @@ where
         .context(error::GetPrefixSnafu { prefixes })?;
     debug!("API response model: {}", response.to_string());
 
-    // Build a Settings struct from the response.
-    let settings = serde_json::from_value::<model::Model>(response)
-        .context(error::InterpretModelSnafu)?
```
bcressey, line 223: We can also get rid of the `InterpretModelSnafu` error.

sources/api/apiserver/src/server/controller.rs
```diff
@@ -366,7 +366,8 @@ pub(crate) fn set_settings<D: DataStore>(
     transaction: &str,
 ) -> Result<()> {
     trace!("Serializing Settings to write to data store");
-    let pairs = to_pairs_with_prefix("settings", settings)
+    let settings_json = serde_json::to_value(settings).expect("struct to value can't fail");
```
bcressey, line 369: Why is this infallible?

bcressey (approved):
Nice.

sources/api/datastore/src/serialization/pairs.rs
```diff
@@ -516,20 +515,22 @@ mod test {
...
-        let keys = to_pairs(&b).unwrap();
+        let j = serde_json::to_value(b).unwrap();
+        let keys = to_pairs(&j).unwrap();
         assert_eq!(
             keys,
             hashmap!(
-                key!("B.list") => "[3,4,5]".to_string(),
-                key!("B.boolean") => "true".to_string(),
+                key!("list") => "[3,4,5]".to_string(),
+                key!("boolean") => "true".to_string(),
```
cbgbt, line 528: The prefix dropping here in the test felt like an alarm bell for backwards incompatibility, so I dug into it a bit and think that it's safe, though I think it does indicate a change to make.

Basically, the serializer keeps track of a "key prefix" for all elements when it is converting them to pairs. In the case where no prefix is present and the element is a struct, it adds the struct's name as a prefix. In the past, this *should* only be triggered when dealing with a struct that represents the top-level of some stored object.

I believe this prefix hoisting will never be triggered if the input domain is `serde_json::Value`, since that type is an `enum`. We could potentially remove it -- what's important is to check that `configuration-files`, `metadata`, and `services` are all serialized to the appropriate pair names.

sources/api/storewolf/src/main.rs
```diff
@@ -385,7 +393,8 @@ fn populate_default_datastore<P: AsRef<Path>>(
     // If any other defaults remain (configuration files, services, etc),
     // write them to the datastore in Live state
     debug!("Serializing other defaults and writing new ones to datastore");
-    let defaults = to_pairs(&defaults_val).context(error::SerializationSnafu {
+    let defaults_json = serde_json::to_value(defaults_val).context(error::JsonConversionSnafu)?;
+    let defaults = to_pairs(&defaults_json).context(error::SerializationSnafu {
```
cbgbt, line 397: The `configuration-files` and `services` I mentioned should be handled here.

I think something that would be extremely helpful would be some integration-level checks of storewolf that the resulting datastore from a given set of defaults is as-expected. This would provide a ton of validation to the datastore serializer.

We could work on that in a followup though.

sources/api/datastore/src/serialization/pairs.rs
```diff
@@ -540,13 +541,14 @@ mod test {
             boolean: true,
         };
         let a = A { id: 42, b: Some(b) };
-        let keys = to_pairs(&a).unwrap();
+        let j = serde_json::to_value(a).unwrap();
```
cbgbt, line 548: nit: could be a good idea to define these test structures using the `json!` macro, but it isn't necessary.

sources/api/apiserver/src/server/controller.rs
sam-berning: my rationale on this was that because Settings Plugins are serialized and deserialized through JSON, in order for anything to be assigned to a Settings type it must be serializable to JSON. I guess it's probably safer to handle the error gracefully if it does occur though, so I'll add a snafu error here.

cbgbt (approved):

cbgbt:
Nice work! None of my feedback is blocking.

sources/api/datastore/src/serialization/pairs.rs
sam-berning: oh yeah, meant to leave a note on this in the PR. we had removed the cases where the serializer depended on the struct name as part of the settings plugins work because I couldn't find a way to rename `serde_json::Value`, so this shouldn't actually cause a change in behavior in practice

sources/api/datastore/src/serialization/pairs.rs
sam-berning: I intentionally left the existing tests as-is with just a conversion to `serde_json::Value` to confirm that the way we handled non-json inputs was backwards compatible (apart from the prefix behavior mentioned above)

sam-berning:
^ improved serialization error handling in apiserver, removed unused errors in sundog

cbgbt (approved):

sources/api/apiserver/src/server/controller.rs
bcressey, line 369:
```suggestion
    let settings_json = serde_json::to_value(settings).context(error::SettingsToJsonSnafu)?;
```

PR #85, opened 2024-08-13
Split ECS and VPC CNI plugins from ecs-agent

### LINKED ISSUE #90: Split ECS and VPC CNI plugins from ecs-agent into their own packages
**What I'd like:**

Currently, the `ecs-agent` pkg also includes the `amazon-ecs-cni-plugins` and `amazon-vpc-cni-plugins` submodules ([source](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/ecs-agent/Cargo.toml#L19-L22)) which are built along with it. Ideally, we want to split these plugins into their own packages for ease of building and patching.

**Any alternatives you've considered:** N/a

koooosh:
**Issue number:** #90 

Closes #90 

**Description of changes:**

Currently, the `ecs-agent` pkg also includes the `amazon-ecs-cni-plugins` and `amazon-vpc-cni-plugins` submodules ([source](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/ecs-agent/Cargo.toml#L19-L22)) which are built along with it. Ideally, we want to split these plugins into their own packages for ease of building and patching.

This PR splits the `amazon-ecs-cni-plugins` and `amazon-vpc-cni-plugins` submodules from the `ecs-agent` package into their own separate packages.

The plugin binaries are installed in the `/usr/libexec/cni/` path and have symlinks in `/usr/libexec/amazon-ecs-agent/` for the ecs-agent to access.

**Testing done:**

- Both `aws-ecs-1` and `aws-ecs-2` variants build and boot successfully.

- Binaries and symlinks are installed correctly:
```bash
bash-5.1# ls -l /usr/libexec/amazon-ecs-agent/
total 4
lrwxrwxrwx. 1 root root   22 Aug 15 09:06 aws-appmesh -> ../cni/vpc/aws-appmesh
lrwxrwxrwx. 1 root root   21 Aug 15 09:06 ecs-bridge -> ../cni/ecs/ecs-bridge
lrwxrwxrwx. 1 root root   18 Aug 15 09:06 ecs-eni -> ../cni/ecs/ecs-eni
lrwxrwxrwx. 1 root root   19 Aug 15 09:06 ecs-ipam -> ../cni/ecs/ecs-ipam
lrwxrwxrwx. 1 root root   29 Aug 15 09:06 ecs-serviceconnect -> ../cni/vpc/ecs-serviceconnect
drwxr-xr-x. 3 root root 4096 Aug 15 09:08 managed-agents
lrwxrwxrwx. 1 root root   25 Aug 15 09:06 vpc-branch-eni -> ../cni/vpc/vpc-branch-eni
lrwxrwxrwx. 1 root root   21 Aug 15 09:06 vpc-bridge -> ../cni/vpc/vpc-bridge
lrwxrwxrwx. 1 root root   18 Aug 15 09:06 vpc-eni -> ../cni/vpc/vpc-eni
lrwxrwxrwx. 1 root root   21 Aug 15 09:06 vpc-tunnel -> ../cni/vpc/vpc-tunnel
```

- ECS service is running successfully:
```bash
bash-5.1# systemctl status ecs
● ecs.service - Amazon Elastic Container Service - container agent
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/ecs.service; enabled; preset: enabled)
    Drop-In: /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/ecs.service.d
             └─00-defaults.conf
             /etc/systemd/system/ecs.service.d
             └─10-base.conf
     Active: active (running) since Thu 2024-08-15 09:10:06 UTC; 25min ago
       Docs: https://aws.amazon.com/documentation/ecs/
   Main PID: 2080 (amazon-ecs-agen)
      Tasks: 10 (limit: 18928)
     Memory: 81.9M
     CGroup: /system.slice/ecs.service
             └─2080 /usr/bin/amazon-ecs-agent
``` 
- Internal testing for `aws-ecs-1` variant succeeds.

**Terms of contribution:**


packages/amazon-vpc-cni-plugins/amazon-vpc-cni-plugins.spec
```diff
@@ -0,0 +1,100 @@
...
+LD_VPC_CNI_VERSION="-X github.com/aws/amazon-vpc-cni-plugins/version.Version=%{vpccni_gover}"
+VPC_CNI_HASH="%{vpccni_gitrev}"
+LD_VPC_CNI_SHORT_HASH="-X github.com/aws/amazon-vpc-cni-plugins/version.GitShortHash=${VPC_CNI_HASH::8}"
+
+declare -a VPC_CNI_BUILD_ARGS
+VPC_CNI_BUILD_ARGS=(
+  -ldflags "${GOLDFLAGS} ${LD_VPC_CNI_VERSION} ${LD_VPC_CNI_SHORT_HASH} ${LD_VPC_CNI_PORCELAIN}"
+)
+
+for p in \
```
arnaldo2792, line 77: Lets build all the plugins under `plugins`:

https://github.com/aws/amazon-vpc-cni-plugins/tree/master/plugins

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -0,0 +1,103 @@
...
+Name: %{_cross_os}amazon-ecs-cni-plugins
+Version: %{ecscni_gitrev}
+Release: 1%{?dist}
+Summary: Networking Plugins for ECS Task Networking
+License: Apache-2.0
+URL: https://%{ecscni_goimport}
+Source0: https://%{ecscni_goimport}/archive/%{ecscni_gitrev}/%{ecscni_gorepo}.tar.gz
+
+# Bottlerocket-specific - filesystem location for ECS CNI plugins
+Patch0001: 1001-bottlerocket-default-filesystem-locations.patch
```
arnaldo2792: Ideally, the patch number should match with the number in the file name.

packages/ecs-agent/ecs-agent.spec
```diff
@@ -76,15 +57,14 @@ Patch0005: 0005-bottlerocket-change-execcmd-directories-for-Bottlero.patch
...
-Patch1001: 1001-bottlerocket-default-filesystem-locations.patch
-
 BuildRequires: %{_cross_os}glibc-devel
 
 Requires: %{_cross_os}docker-engine
 Requires: %{_cross_os}iptables
 Requires: %{_cross_os}amazon-ssm-agent-plugin
 Requires: %{name}(binaries)
+Requires: %{_cross_os}amazon-ecs-cni-plugins
+Requires: %{_cross_os}amazon-vpc-cni-plugins
```
arnaldo2792, line 66: Nit: for consistency, lets move them after `amazon-ssm-agent-plugin`

packages/ecs-agent/ecs-agent.spec
```diff
@@ -339,44 +192,38 @@ ln -rs %{buildroot}%{_cross_localstatedir}/ecs/managed-agents/execute-command/co
...
-mv %{vpccni_gorepo}-%{vpccni_gitrev}/vendor go-vendor/%{vpccni_gorepo}
-%cross_scan_attribution go-vendor go-vendor
+install -d %{buildroot}{%{_cross_libexecdir},%{_cross_fips_libexecdir}}/amazon-ecs-agent
+# Create symlinks to ECS CNI plugin binaries
+for p in \
+  ecs-bridge \
+  ecs-eni \
+  ecs-ipam \
+; do
+  ln -rs %{buildroot}%{_cross_libexecdir}/cni/ecs/${p} %{buildroot}%{_cross_libexecdir}/amazon-ecs-agent/${p}
```
arnaldo2792: I think in this case we want the symlink to be relative:

```
ln -rs ../../cni/ecs/${p}
```

packages/ecs-agent/ecs-agent.spec
```diff
@@ -339,44 +192,38 @@ ln -rs %{buildroot}%{_cross_localstatedir}/ecs/managed-agents/execute-command/co
...
 %{_cross_attribution_file}
 %{_cross_attribution_vendor_dir}
-%license %{agent_gorepo}-%{agent_gover}/LICENSE
-%license %{agent_gorepo}-%{agent_gover}/NOTICE
-%license %{agent_gorepo}-%{agent_gover}/ecs-agent/THIRD_PARTY.md
-%license %{ecscni_gorepo}-%{ecscni_gitrev}/LICENSE.%{ecscni_gorepo}
-%license %{vpccni_gorepo}-%{vpccni_gitrev}/LICENSE.%{vpccni_gorepo}
+%license LICENSE
+%license NOTICE
+%license ecs-agent/THIRD_PARTY.md
```
arnaldo2792: You can have one line, instead of three. Probably we did the multi-line `%license` because it was too long. Now it isn't.

packages/amazon-vpc-cni-plugins/amazon-vpc-cni-plugins.spec
koooosh: Got it, will add the `ecs-serviceconnect`, `vpc-bridge`, and `vpc-tunnel` plugins here and also add symlinks to them in `ecs-agent`

koooosh:
Force push addresses @arnaldo2792 comments:
- Builds all VPC CNI plugins
- Updates patch number in ecs cni spec
- Few nit fixes in ecs agent spec

packages/ecs-agent/ecs-agent.spec
koooosh: Discussed offline, the `r` flag for `ln` command already makes it relative. We prefer that over:
```
ln -s ../cni/ecs/${p} %{buildroot}%{_cross_libexecdir}/amazon-ecs-agent/${p}
```

packages/ecs-agent/ecs-agent.spec
```diff
@@ -51,12 +38,6 @@ Source200: etc-systemd-system-ecs.service.d.mount
...
 Source300: logdog.ecs.conf
 
-# Patches are numbered according to which source they apply to
-# Patches 0000 - 0999 apply to Source0
-# Patches 1000 - 1999 apply to Source1
-# Patches 2000 - 2999 apply to Source2
-# See the %prep section for the implementation of this logic
-
 # Bottlerocket-specific - filesystem location of the pause image
 Patch0001: 0001-bottlerocket-default-filesystem-locations.patch
```
larvacea, line 42: I am probably confused, but GitHub tells me that this file was renamed out of this directory and into amazon-ecs-cni-plugins (and the spec file there incorporates it). Does this patch file appear in both packages?

larvacea: Oh! I _am_ confused. The very-similar-sounding patch 1001-bottlerocket-default-filesystem-locations.patch moved.

larvacea (approved):
Despite my confusion (1001-something.patch is not 0001-something.patch), looks fine to me.

arnaldo2792 (approved):

packages/ecs-agent/ecs-agent.spec
```diff
@@ -339,44 +192,39 @@ ln -rs %{buildroot}%{_cross_localstatedir}/ecs/managed-agents/execute-command/co
...
+  aws-appmesh \
+  ecs-serviceconnect \
+  vpc-branch-eni \
+  vpc-bridge \
+  vpc-eni \
+  vpc-tunnel \
+; do
+  ln -rs %{buildroot}%{_cross_libexecdir}/cni/vpc/${p} %{buildroot}%{_cross_libexecdir}/amazon-ecs-agent/${p}
+  ln -rs %{buildroot}%{_cross_libexecdir}/cni/vpc/${p} %{buildroot}%{_cross_fips_libexecdir}/amazon-ecs-agent/${p}
+done
```
bcressey: It's very strange for a package to symlink another package's binaries like this, and to vend those symlinks. I acknowledge that it was done for the SSM agent plugin but it's not a great pattern to emulate. It would be better to create these links in the corresponding CNI package. You could put them in an "ecs" subpackage, but that's not as important as keeping the links and binaries together.

Some concerns with this approach:
* If a new VPC or ECS plugin is added, this package definition must also change in order to create symlinks. Where if the plugin package also owns the symlink, then it will be more obvious to the packager that this dependency exists.
* This will create broken symlinks, if the plugin package drops a plugin or changes the path where plugins are installed.

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -0,0 +1,103 @@
+%global ecscni_goproject github.com/aws
+%global ecscni_gorepo amazon-ecs-cni-plugins
+%global ecscni_goimport %{ecscni_goproject}/%{ecscni_gorepo}
+%global ecscni_gitrev 53a8481891251e66e35847554d52a13fc7c4fd03
+
+# Construct reproducible tar archives
+# See https://reproducible-builds.org/docs/archives/
+%global source_date_epoch 1234567890
+%global tar_cf tar --sort=name --mtime="@%{source_date_epoch}" --owner=0 --group=0 --numeric-owner -cf
```
bcressey: You aren't using these macros.

```suggestion
```

packages/amazon-vpc-cni-plugins/amazon-vpc-cni-plugins.spec
```diff
@@ -0,0 +1,115 @@
+%global vpccni_goproject github.com/aws
+%global vpccni_gorepo amazon-vpc-cni-plugins
+%global vpccni_goimport %{vpccni_goproject}/%{vpccni_gorepo}
+%global vpccni_gitrev be5214353252f8315a1341f4df9ffbd8cf69000c
+%global vpccni_gover 1.3
+
+# Construct reproducible tar archives
+# See https://reproducible-builds.org/docs/archives/
+%global source_date_epoch 1234567890
+%global tar_cf tar --sort=name --mtime="@%{source_date_epoch}" --owner=0 --group=0 --numeric-owner -cf
```
bcressey:
```suggestion
```

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -0,0 +1,103 @@
...
+
+Name: %{_cross_os}amazon-ecs-cni-plugins
+Version: %{ecscni_gitrev}
+Release: 1%{?dist}
+Summary: Networking Plugins for ECS Task Networking
```
bcressey: nit: no need to title case the summary
```suggestion
Summary: Networking plugins for ECS task networking
```

packages/amazon-vpc-cni-plugins/amazon-vpc-cni-plugins.spec
```diff
@@ -0,0 +1,115 @@
...
+install -D -p -m 0755 aws-appmesh %{buildroot}%{_cross_libexecdir}/cni/vpc/aws-appmesh
+install -D -p -m 0755 ecs-serviceconnect %{buildroot}%{_cross_libexecdir}/cni/vpc/ecs-serviceconnect
+install -D -p -m 0755 vpc-branch-eni %{buildroot}%{_cross_libexecdir}/cni/vpc/vpc-branch-eni
+install -D -p -m 0755 vpc-bridge %{buildroot}%{_cross_libexecdir}/cni/vpc/vpc-bridge
+install -D -p -m 0755 vpc-eni %{buildroot}%{_cross_libexecdir}/cni/vpc/vpc-eni
+install -D -p -m 0755 vpc-tunnel %{buildroot}%{_cross_libexecdir}/cni/vpc/vpc-tunnel
+
+install -d %{buildroot}%{_cross_fips_libexecdir}
+install -D -p -m 0755 fips/aws-appmesh %{buildroot}%{_cross_fips_libexecdir}/cni/vpc/aws-appmesh
+install -D -p -m 0755 fips/ecs-serviceconnect %{buildroot}%{_cross_fips_libexecdir}/cni/vpc/ecs-serviceconnect
```
bcressey, line 100: @arnaldo2792 does adding these new plugins change the ECS agent's behavior? Service Connect still won't be supported - will the node register with that feature only to fail task launches that use it, or will the plugin be ignored?

bcressey (changes requested):
nit: you can drop `pkg` and `pkgs` from the commit messages; the "packages:" prefix implies that you're working with packages.

Overall this seems like a reasonable change - generally sources with different upstreams should be packaged separately - but there's no rationale provided, either in the commit messages or in the pull request summary. It's obvious what you're doing (splitting up a package) but not why.

packages/amazon-vpc-cni-plugins/amazon-vpc-cni-plugins.spec
arnaldo2792: The plugin will be ignored, the ECS agent looks for a Service Connect container image:


https://github.com/aws/amazon-ecs-agent/blob/4e8af8ac81a9e7dfafd96964eda8c1ff3c88f478/agent/app/agent_capability.go#L481

packages/ecs-agent/ecs-agent.spec
koooosh: Ack regarding dropping `pkg`* from commit messages - will fix those. Regarding rationale, I put that in the issue #90 : 

```
Ideally, we want to split these plugins into their own packages for ease of building and patching.
```

But I'll copy that over to this PR description!

koooosh: Good callouts regarding the symlinks. My (naive) reasoning was that since they are being installed in the `/amazon-ecs-agent` directory, they should be added to the `amazon-ecs-agent` spec, but I agree this will cause more maintenance issues in the future. Will create these symlinks in their respective pkg as suggested.

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
koooosh: Ack, will remove

koooosh:
Force push makes the following relevant changes:
- Moves the creation of ECS and VPC CNI plugin binary symlinks to their own respective pkgs instead of the `ecs-agent` pkg

koooosh:
Force push changes symlink subpackaging to better match repo convention like [here](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/8d610d4d87c15e57a4db9d958d7e1a8895de2073/packages/os/os.spec#L104)

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -0,0 +1,147 @@
...
+
+%description fips-bin
+%{summary}.
+
+%package ecs-agent-symlinks
+Summary: Symlinks to ECS networking plugins binaries for ECS agent
+Requires: %{name}(ecs-agent-symlinks)
+
+%description ecs-agent-symlinks
+%{summary}.
```
koooosh: Need to remove as this is already Provided by below subpkgs

packages/amazon-vpc-cni-plugins/amazon-vpc-cni-plugins.spec
```diff
@@ -0,0 +1,168 @@
...
+
+%description fips-bin
+%{summary}.
+
+%package ecs-agent-symlinks
+Summary: Symlinks to VPC networking plugins binaries for ECS agent
+Requires: %{name}(ecs-agent-symlinks)
+
+%description ecs-agent-symlinks
+%{summary}.
```
koooosh: Need to remove as this is already Provided by below subpkgs

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -0,0 +1,147 @@
...
+URL: https://%{ecscni_goimport}
+Source0: https://%{ecscni_goimport}/archive/%{ecscni_gitrev}/%{ecscni_gorepo}.tar.gz
+
+# Bottlerocket-specific - filesystem location for ECS CNI plugins
+Patch0001: 0001-bottlerocket-default-filesystem-locations.patch
+
+BuildRequires: %{_cross_os}glibc-devel
+
+Requires: %{name}(binaries)
+Requires: (%{name}(ecs-agent-symlinks) if %{_cross_os}variant-family(aws-ecs))
```
arnaldo2792: In an offline conversation with @koooosh , we discussed what's a good name for the subpackage that will provide the symlinks. We concluded that the package that provides the symlinks for the ECS agent could be named `ecs-agent-extras` instead of `symlinks`. This allows us to keep "extras" specific to the ECS variants in its own subpackage in case we have more "extras" for other variants in the future.

That said, this line will change to be:

```
Requires: (%{name}(ecs-agent-extras) if %{_cross_os}variant-family(aws-ecs))
```

And the subpackages that provide it will be:

```
%package ecs-agent-ext
Provides: %{name}(ecs-agent-extras)

%package ecs-agent-fips-ext
Provides: %{name}(ecs-agent-extras)
```

@bcressey , thoughts on the subpackages names? I suggested `-ext` following the existing convention `binaries/bin`.

bcressey: > thoughts on the subpackages names?

I would probably spell it out as `ecs-agent-extras` and `ecs-agent-fips-extras` since `ext` can be ambiguous (could mean "extension" or "external").

koooosh:
Force push makes the following changes:
- Removes unnecessary subpkgs
- Renames symlink subpkgs

arnaldo2792 (approved):

packages/amazon-vpc-cni-plugins/amazon-vpc-cni-plugins.spec
```diff
@@ -0,0 +1,159 @@
...
+for p in \
+  aws-appmesh \
+  ecs-serviceconnect \
+  vpc-branch-eni \
+  vpc-bridge \
+  vpc-eni \
+  vpc-tunnel \
+; do
+  ln -rs %{buildroot}%{_cross_libexecdir}/cni/vpc/${p} %{buildroot}%{_cross_libexecdir}/amazon-ecs-agent/${p}
+  ln -rs %{buildroot}%{_cross_libexecdir}/cni/vpc/${p} %{buildroot}%{_cross_fips_libexecdir}/amazon-ecs-agent/${p}
```
bcressey:
```suggestion
  ln -rs %{buildroot}%{_cross_fips_libexecdir}/cni/vpc/${p} %{buildroot}%{_cross_fips_libexecdir}/amazon-ecs-agent/${p}
```

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -0,0 +1,138 @@
...
+
+# Create symlinks to ECS CNI plugin binaries for amazon-ecs-agent
+install -d %{buildroot}{%{_cross_libexecdir},%{_cross_fips_libexecdir}}/amazon-ecs-agent
+for p in \
+  ecs-bridge \
+  ecs-eni \
+  ecs-ipam \
+; do
+  ln -rs %{buildroot}%{_cross_libexecdir}/cni/ecs/${p} %{buildroot}%{_cross_libexecdir}/amazon-ecs-agent/${p}
+  ln -rs %{buildroot}%{_cross_libexecdir}/cni/ecs/${p} %{buildroot}%{_cross_fips_libexecdir}/amazon-ecs-agent/${p}
```
bcressey: We should create symlinks that stay inside `/usr/fips/libexec`, so that they work whether or not the overlay mount is in place at runtime.

```suggestion
  ln -rs %{buildroot}%{_cross_fips_libexecdir}/cni/ecs/${p} %{buildroot}%{_cross_fips_libexecdir}/amazon-ecs-agent/${p}
```

However, if that causes the symlinks to be broken for the overlay mount, then we should only create the ones for `/usr/libexec`, and rely on the FIPS overlay mount to make them point to the right binaries. In that case, we wouldn't have separate FIPS or non-FIPS versions, that just invites confusion.

koooosh: Got it, let me test this

koooosh: Suggested change works with the `aws-ecs-2` variant!

koooosh:
Force push updates FIPS symlink paths to point to binaries in `/usr/fips/libexec` rather than `usr/libexec`

bcressey (approved):

PR #89, opened 2024-08-13
Pluto settings models

cbgbt:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/4135

**Description of changes:**
Pluto requests an initial set of settings from the Bottlerocket API as dependencies needed to generate additional `kubernetes` settings.

There was a bug in pluto in which `no-proxy` settings were attempted to be deserialized into a `String` when the value in the model is stored as a `Vec`. This updates pluto to use the settings models from the settings SDK to deserialize the settings returned from the API.

**Testing done:**
* [x] unit tests pass
* [x] test with proxy enabled
* [x] test cluster dns IP properly set in a cluster
* [x] test max pods properly set
* [x] test provider_id properly set
* [x] test node_name properly set

I enabled `https-proxy` pointing to a `tinyproxy` that I configured, then tested instance startup with various `no-proxy` settings:
* Unset `no-proxy`: Noted traffic in my proxy logs and also that the Cloudtrail noted the IP address of my proxy as the source for EKS `DescribeCluster` traffic
* `no-proxy: ["*"]` bypassed the proxy when communicating with the EKS API
* No `proxy` settings: all EKS settings still set as usual

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/api.rs
+++ b/sources/api/pluto/src/api.rs
@@ -1,4 +1,5 @@
-use serde::{Deserialize, Serialize};
+use bottlerocket_settings_models::{AwsSettingsV1, KubernetesSettingsV1, NetworkSettingsV1};
+use serde::Deserialize;
 use snafu::{ensure, ResultExt, Snafu};
 use std::ffi::OsStr;
 use tokio::process::Command;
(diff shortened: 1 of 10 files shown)


cbgbt:
^ Force push minor readability fix.

sources/api/pluto/src/api.rs
```diff
@@ -1,87 +1,116 @@
...
+    pub fn from_api_response(readonly: SettingsView) -> Self {
+        Self {
+            readonly,
+            delta: SettingsView::default(),
+        }
+    }
+
+    pub fn read(&self) -> &SettingsView {
+        &self.readonly
+    }
```
bcressey: nit: maybe this? `read` sounds like it will read all settings, and someone not using the macros might call it by mistake:
```suggestion
    pub fn initial(&self) -> &SettingsView {
        &self.initial
    }
```

bcressey (approved):

cbgbt:
Sorry for churn. Forgot to push after rebasing to make the inter-diffs cleaner.

^ Rebases onto develop

sam-berning (approved):

cbgbt:
^ updated to include metadata for new settings models release.

sam-berning (approved):

Issue #97, opened 2024-08-19
Device symlinks created at the wrong path


arnaldo2792:
**What I expected to happen:**
I expect that a symlink under `/dev` is created using the device mapping configured for my EC2 EBS volume.

**What actually happened:**
In certain workflows, the symlinks to the devices are created under `/dev/dev/<device>` instead of just `/dev/device`. Additionally, creating devices under `/dev` with an actual ID that could be used by the kernel is a bad practice, so we should create the symlinks for EBS volume mappings under `/dev/by-ebs-id`, to prevent collisions with kernel device names. See https://github.com/amazonlinux/amazon-ec2-utils/issues/37 for reference.

arnaldo2792:
I forgot to close this in #98

PR #98, opened 2024-08-19
Fix udev rule for EBS mappings

arnaldo2792:
**Issue number:**
#97 

**Description of changes:**

With this change, symlinks to the underlying EBS volumes will be created under `/dev/by-ebs-id` instead of the root of `/dev`. This will help prevent collisions with actual devices created by the kernel (as described in this https://github.com/amazonlinux/amazon-ec2-utils/issues/37).

As part of this change, `/dev/` will be removed from the device name if it is found after querying the device identifier, so that only the device id is returned to udev.

**Testing done:**

I ran my change in an EC2 instance, and confirmed that the devices are listed under `/dev/by-ebs-id` when the mapping is added:

```
ls -la /dev/disk/by-ebs-id
total 0
drwxr-xr-x.  2 root root 240 Aug 19 20:50 .
drwxr-xr-x. 10 root root 200 Aug 19 20:50 ..
lrwxrwxrwx.  1 root root  13 Aug 19 20:50 xvda -> ../../nvme0n1
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda1 -> ../../nvme0n1p1
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda2 -> ../../nvme0n1p2
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda3 -> ../../nvme0n1p3
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda4 -> ../../nvme0n1p4
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda5 -> ../../nvme0n1p5
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda6 -> ../../nvme0n1p6
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda7 -> ../../nvme0n1p7
lrwxrwxrwx.  1 root root  15 Aug 19 20:50 xvda8 -> ../../nvme0n1p8
lrwxrwxrwx.  1 root root  13 Aug 19 20:50 xvdcz -> ../../nvme1n1
```

I confirmed that in the workflow that prepends the device ID with `/dev/`, this prefix is removed and the symlink is created under `/dev/by-ebs-id/`

**Terms of contribution:**

### DIFF
--- a/sources/ghostdog/src/main.rs
+++ b/sources/ghostdog/src/main.rs
@@ -117,7 +117,12 @@ fn parse_device_name(device_info: &[u8], path: String) -> Result<String> {
     let offset = NVME_IDENTIFY_DATA_SIZE - 1024;
     let device_name = &device_info[offset..offset + 32];
 
-    Ok(String::from_utf8_lossy(device_name).trim_end().to_string())
+    // Remove `/dev` in case the returned device name includes it, the udev
+    // rule already includes that prefix
+    Ok(String::from_utf8_lossy(device_name)
+        .trim_start_matches("/dev/")
+        .trim_end()
+        .to_string())
 }
 
 /// Print the device type in the environment key format udev expects.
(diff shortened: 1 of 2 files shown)


yeazelm (approved):

sources/ghostdog/src/main.rs
```diff
@@ -117,7 +117,12 @@ fn parse_device_name(device_info: &[u8], path: String) -> Result<String> {
     let offset = NVME_IDENTIFY_DATA_SIZE - 1024;
     let device_name = &device_info[offset..offset + 32];
 
-    Ok(String::from_utf8_lossy(device_name).trim_end().to_string())
+    // Remove `/dev` in case the returned device name includes it, the udev
+    // rule already includes that prefix
+    Ok(String::from_utf8_lossy(device_name)
+        .replace("/dev/", "")
+        .trim_end()
+        .to_string())
```
bcressey: nit: it feels a bit loose not to anchor this to the start of the string:
```suggestion
    Ok(String::from_utf8_lossy(device_name)
        .trim_start_matches("/dev/")
        .trim_end()
        .to_string())
```

sources/ghostdog/src/main.rs
```diff
@@ -235,4 +240,30 @@ mod test {
...
         assert_eq!(find_device_type(&mut Cursor::new(&data)).unwrap(), "system");
     }
+
+    #[test]
+    fn test_valid_device_info() {
+        let mut device_name = "xvdcz".as_bytes().to_vec();
+        let mut device_info: Vec<u8> = vec![0; NVME_IDENTIFY_DATA_SIZE - 1024];
+        let mut padding = vec![32; NVME_IDENTIFY_DATA_SIZE - device_info.len() - device_name.len()];
+        device_info.append(&mut device_name);
+        device_info.append(&mut padding);
```
bcressey: nit: could factor this into a helper function now that it's repeated

bcressey (approved):

vigh-m (approved):

arnaldo2792:
(addressed nits on ghostdog)

bcressey (approved):

PR #100, opened 2024-08-20
Backfill kernel advisories for v2.3.0 and v2.3.2 releases

ginglis13:
**Issue number:**

n/a

**Description of changes:**

Add advisories for CVEs patched in kernel released in the v2.3.0 and v2.3.2 Bottlerocket core kits.

**Testing done:**

There were some duplicate CVEs I identified; I verified that these advisories are accurate and required. These duplicates were CVEs patched in the a kernel series (say, the 5.10 series) included in the v2.3.2 core kit release that were also patched by kernel series included in the v2.3.0 core kit (say, the 5.15 and 6.1 series).

**Terms of contribution:**

### DIFF
--- a/advisories/2.3.0/BRSA-tfps4zoe7xte.toml
+++ b/advisories/2.3.0/BRSA-tfps4zoe7xte.toml
@@ -1,0 +1,17 @@
+[advisory]
+id = "BRSA-tfps4zoe7xte"
+title = "kernel CVE-2024-41009"
+cve = "CVE-2024-41009"
+severity = "moderate"
+description = "In the Linux kernel, the following vulnerability has been resolved: bpf: Fix overrunning reservations in ringbuf"
+
+[[advisory.products]]
+package-name = "kernel-6.1"
+patched-version = "kernel-6.1.97-104.177.amzn2023"
+
+
+[updateinfo]
+author = "giinglis"
+issue-date = 2024-07-24T20:00:00Z
+arches = ["x86_64", "aarch64"]
+version = "2.3.0"
--- a/advisories/2.3.0/BRSA-trxsnksws0mo.toml
+++ b/advisories/2.3.0/BRSA-trxsnksws0mo.toml
@@ -1,0 +1,22 @@
+[advisory]
+id = "BRSA-trxsnksws0mo"
+title = "kernel CVE-2024-41092"
+cve = "CVE-2024-41092"
+severity = "moderate"
+description = "In the Linux kernel, the following vulnerability has been resolved: drm/i915/gt: Fix potential UAF by revoke of fence registers"
+
+[[advisory.products]]
+package-name = "kernel-5.15"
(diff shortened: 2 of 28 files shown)


ytsssun (approved):

larvacea (approved):

advisories/2.3.0/BRSA-tfps4zoe7xte.toml
```diff
@@ -0,0 +1,17 @@
+[advisory]
+id = "BRSA-tfps4zoe7xte"
+title = "kernel CVE-2024-41009"
+cve = "CVE-2024-41009"
+severity = "moderate"
+description = "In the Linux kernel, the following vulnerability has been resolved: -  - bpf: Fix overrunning reservations in ringbuf"
```
yeazelm: nit: I'm not sure I like the string of characters here `resolved: -  - bpf: Fix`. Is there a rationale for it? If it renders a certain way, that would be good to call out but this feels like bad input into the advisory and I'd like to trim this if we can?

advisories/2.3.0/BRSA-trxsnksws0mo.toml
```diff
@@ -0,0 +1,22 @@
+[advisory]
+id = "BRSA-trxsnksws0mo"
+title = "kernel CVE-2024-41092"
+cve = "CVE-2024-41092"
+severity = "moderate"
+description = "In the Linux kernel, the following vulnerability has been resolved: -  - drm/i915/gt: Fix potential UAF by revoke of fence registers"
```
yeazelm: These are coming from the Linux Kernel CNA, can we add a reference to them for folks to follow up if they have more questions? For example, including `https://lore.kernel.org/linux-cve-announce/2024072953-CVE-2024-41092-8bd7@gregkh/` for this description would help get folks one step further to following up if they have questions.

advisories/2.3.0/BRSA-tfps4zoe7xte.toml
ginglis13: This is straight from the ALAS wesbite:  https://alas.aws.amazon.com/cve/json/v1/CVE-2024-41009.json

If you'd like, I'm okay with omitting either all of `In the Linux kernel, the following vulnerability has been resolved:` for advisories, or reducing the `-` and `:`. I'm not exactly sure where the newline `\n` is converted to `-`, must be in the conversion to TOML format, but another option could be to remove the `-` entirely? This was rendering the same for our GHSAs when we started pulling wording for kernel advisories directly from ALAS: https://github.com/bottlerocket-os/bottlerocket/security/advisories/GHSA-mw65-44mw-v359

advisories/2.3.0/BRSA-trxsnksws0mo.toml
ginglis13: Sure, I can do that going forward

yeazelm (approved):

PR #102, opened 2024-08-21
Add helper functions for ipcidr

Issacwww:
**Description of changes:**
Add new helper function for EKS to config the template

**Testing done:**
Added unit test and `cargo test` passed 

**Terms of contribution:**

### DIFF
--- a/sources/deny.toml
+++ b/sources/deny.toml
@@ -69,6 +69,11 @@ skip-tree = [
     { name = "darling", version = "=0.14" },
 ]
 
+[bans.workspace-dependencies]
+duplicates = "deny"
+include-path-dependencies = true
+unused = "deny"
+
 [advisories]
 # generational-arena is currently unmaintained.
 ignore = [


sources/api/schnauzer/src/helpers.rs
```diff
@@ -907,6 +910,128 @@ pub fn goarch(
     Ok(())
 }
 
+handlebars_helper!(is_ipv4: |cidr: str| {
```
cbgbt: Nice! It would be awesome to also add an `is_ipv6` for the API, but not a blocker.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -907,6 +910,128 @@ pub fn goarch(
     Ok(())
 }
```
cbgbt: For the helpers that you created, you will need to add them to the template registries that are used when templates are rendered. There are currently two codepaths that set up registries, and adding the functions to both is fine:
* [schnauzer-v1](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/8b0be519b7e3e64e1fa5a9eaa7873a6571d01565/sources/api/schnauzer/src/v1.rs#L111)
* [schnauzer-v2](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/8b0be519b7e3e64e1fa5a9eaa7873a6571d01565/sources/api/schnauzer/src/v2/import/helpers.rs#L67)

sources/api/schnauzer/src/helpers.rs
```diff
@@ -907,6 +910,128 @@ pub fn goarch(
...
 
+handlebars_helper!(is_ipv4: |cidr: str| {
+    let ip_part = validate_and_parse_cidr(cidr)?;
+    let ip = IpAddr::from_str(ip_part);
+
+    match ip {
+        Ok(IpAddr::V4(_)) => true,
+        Ok(IpAddr::V6(_)) => false,
+        Err(_) => panic!("Invalid CIDR or IP address: {}", cidr),
```
cbgbt: We prefer to avoid explicit panics (`panic`, `expect`, `unwrap`) and instead use `Result` values. You could try something like this?

```suggestion
    let ip = IpAddr::from_str(ip_part).context(error::InvalidIPAddressSnafu)?;
    
    matches!(ip, IpAddr::V4(_))
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -239,6 +239,9 @@ mod error {
         #[snafu(display("Unknown architecture '{}' given to goarch helper", given))]
         UnknownArch { given: String },
 
+        #[snafu(display("Invalid Cidr format '{}'", cidr))]
+        InvalidCidr { cidr: String },
+
```
cbgbt: It seems like we also throw this on IP address parse errors. I'd suggest splitting it into two:

```suggestion
        #[snafu(display("Invalid Cidr format '{}'", cidr))]
        InvalidCidr { cidr: String },
        
        #[snafu(display("Invalid IP Address format '{}': {}", cidr, source))]
        InvalidIPAddress { ip_address: String, source: std::net::AddrParseError },

```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -907,6 +910,128 @@ pub fn goarch(
...
+    let new_ip = Ipv4Addr::from(octets).to_string();
+    out.write(&new_ip).with_context(|_| error::TemplateWriteSnafu {
+        template: template_name.to_owned(),
+    })?;
+    
+    Ok(())
+}
+
+fn validate_and_parse_cidr(cidr: &str) -> Result<&str, RenderError> {
```
cbgbt: Rather than using String return types, could we add a dependency on the [`cidr`](https://crates.io/crates/cidr) crate and then use the [`AnyIpCidr`](https://docs.rs/cidr/latest/cidr/enum.AnyIpCidr.html) enum for parsing?

`AnyIpCidr` implements [`FromStr`](https://docs.rs/cidr/latest/cidr/enum.AnyIpCidr.html#impl-FromStr-for-AnyIpCidr), so you should be able to do something like:

```rust
    let parsed_cidr: AnyIpCidr = cidr.parse()
        .context(error::InvalidCidrSnafu { cidr: cidr.to_string() })?;
```

For this to work, you would also need to modify the `InvalidCidr` error to have a `source` of type `cidr::errors::NetworkParseError`.

I believe you could then just call [`first_address`](https://docs.rs/cidr/latest/cidr/enum.AnyIpCidr.html#method.first_address) to get the IP part of the CIDR range, and return that as an `IpAddr`.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -3155,3 +3280,185 @@ mod test_ecs_metadata_service_limits {
...
+}
+
+#[cfg(test)]
+mod test_combined_template_for_ip_cidr {
+    use crate::helpers::{is_ipv4, cidr_to_ipaddr, replace_ipv4_octet};
+    use handlebars::{Handlebars, RenderError};
+    use serde::Serialize;
+    use serde_json::json;
+
+    const TEMPLATE: &str = r#"
```
cbgbt: Nice test!

cbgbt (commented):
Thanks so much for your contribution! The test cases are great.

I have a few style nits around error handling and such. I think the big change I'd really like to see is to move to using the `cidr` crate for CIDR range parsing, then for your helper functions, internally lean on the types exposed by that and the standard library before finally converting to Strings at the end of the helper functions.

Issacwww:
Open a new one #116 that resolved merge conflict and comments

PR #105, opened 2024-08-22
schnauzer: add reflective template helpers

### LINKED ISSUE #104: schnauzer: Add reflective helpers to support polymorphic settings
**What I'd like:**
In some cases, settings values can take many forms. Consider the [`KubernetesClusterDnsIp`](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/cebbd4cdd4cf86b24b52554b9db7bdc21aa4e67e/bottlerocket-settings-models/modeled-types/src/kubernetes.rs#L1189) settings model, which can either be a single IP address string, or else a list of them.

When processing these settings in a template, it would be useful to be able to branch based on what JSON type represents the setting. As such, I'd like to see the following helpers added to `std`:
* `is-bool`
* `is-number`
* `is-string`
* `is-array`
* `is-object`

Additionally, the following helpers would be useful for guarding against `null` in short-circuited `#if` blocks, but I'm actually not sure what the right interface would be for them yet:
* `is-set`
* `is-null`

Our handlebars renderer is configured to error if you try to access an unset value, and we also try to ensure that empty optionals are not included in the JSON object, so `null` should be pretty rare. Possibly the correct interface would be to implement `is-null` and make the result truthy if the value is explicitly `null` or if it is unset.

One alternative would be to add helpers on a case-by-case basis that are specific to the domain that the setting is used. Being able to generically move configuration logic into the template seems more robust to me.

cbgbt:
**Issue number:**

Closes #104

**Description of changes:**
```
    schnauzer: refactor helpers into dir module
```

```
    schnauzer: refactor stdlib helpers to new module
```

```
    schnauzer: add reflective template helpers

    These helpers allow you to identify the type of variables passed into
    templates by schnauzer. This is useful for properly handling polymorphic
    settings (e.g. can be a string or a list)
```

**Testing done:**
Provided unit tests pass

**Terms of contribution:**


cbgbt:
^ push to rebase

arnaldo2792 (approved):

mgsharm (approved):

PR #110, opened 2024-08-27
deny: add workspace dependency lints

cbgbt:
**Description of changes:**
This adds [workspace dependency checks](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-workspace-dependencies-field-optional) with `cargo-deny`.

The options enforce the following:
* `duplicates = deny`: emit an error for each dependency declaration that does not use `workspace = true`
* `include-path-dependencies = true`: Include path dependencies in duplication check
* `unused = "deny"`: emit an error for each dependency not used in the workspace

**Testing done:**
I rewrote a `Cargo.toml` under sources to express its own dependency, and noted that `make twoliter check-licenses` caught the error.

**Terms of contribution:**

### DIFF
--- a/sources/deny.toml
+++ b/sources/deny.toml
@@ -69,6 +69,11 @@ skip-tree = [
     { name = "darling", version = "=0.14" },
 ]
 
+[bans.workspace-dependencies]
+duplicates = "deny"
+include-path-dependencies = true
+unused = "deny"
+
 [advisories]
 # generational-arena is currently unmaintained.
 ignore = [


sam-berning (approved):
Could be good to include the explanation for each config in the commit message so someone doesn't have to hunt down this PR to find them

yeazelm (approved):

cbgbt:
Added config option overview to commit message per @sam-berning

sam-berning (approved):

PR #116, opened 2024-08-30
Add helper functions for ipcidr

Issacwww:
**Description of changes:**
Add new helper function for EKS to config the template
have to reopen new PR due to #102 uses develop branch and need to discard original commit to sync with latest.

**Testing done:**
Added unit test and `cargo test` passed 

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/Cargo.toml
+++ b/sources/api/schnauzer/Cargo.toml
@@ -43,7 +43,7 @@ toml.workspace = true
 url.workspace = true
 bottlerocket-modeled-types.workspace = true
 settings-extension-oci-defaults.workspace = true
-cidr = "0.2.3"
+cidr.workspace = true
 
 [dev-dependencies]
 # Workaround to enable a feature during integration tests.
--- a/sources/api/schnauzer/Cargo.toml
+++ b/sources/api/schnauzer/Cargo.toml
@@ -43,6 +43,7 @@ toml.workspace = true
 url.workspace = true
 bottlerocket-modeled-types.workspace = true
 settings-extension-oci-defaults.workspace = true
+cidr = "0.2.3"
 
 [dev-dependencies]
 # Workaround to enable a feature during integration tests.
--- a/sources/api/schnauzer/src/helpers/mod.rs
+++ b/sources/api/schnauzer/src/helpers/mod.rs
@@ -6,7 +6,9 @@ use bottlerocket_modeled_types::{OciDefaultsCapability, OciDefaultsResourceLimit
 use cidr::AnyIpCidr;
 use dns_lookup::lookup_host;
 use error::TemplateHelperError;
-use handlebars::{handlebars_helper,Context, Handlebars, Helper, Output, RenderContext, RenderError};
+use handlebars::{
+    handlebars_helper, Context, Handlebars, Helper, Output, RenderContext, RenderError,
+};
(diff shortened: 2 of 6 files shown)


sources/api/schnauzer/Cargo.toml
```diff
@@ -43,6 +43,7 @@ toml.workspace = true
 url.workspace = true
 bottlerocket-modeled-types.workspace = true
 settings-extension-oci-defaults.workspace = true
+cidr = "0.2.3"
```
cbgbt: The CI should catch this when it runs, but you'll want to use workspace dependencies for this. Basically you'll want to:
* Add `cidr = 0.2` to the `[workspace.dependencies]` list in `sources/Cargo.toml`
* Change this to `cidr.workspace = true`.

`make twoliter check` should catch this too, if you want to try and replicate locally.

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -563,6 +570,123 @@ pub fn host(
...
+    let new_ip = Ipv4Addr::from(octets).to_string();
+    out.write(&new_ip).with_context(|_| error::TemplateWriteSnafu {
+        template: template_name.to_owned(),
+    })?;
+    
+    Ok(())
+}
+
+fn validate_and_parse_cidr(cidr: &str) -> Result<IpAddr, RenderError> {
+    let parsed_cidr: AnyIpCidr = cidr.parse().map_err(|_| error::InvalidCidrSnafu { cidr: cidr.to_string() }.build())?;
```
cbgbt: If you mark `source` as [`NetworkParseError`](https://docs.rs/cidr/latest/cidr/errors/enum.NetworkParseError.html) in `InvalidCidrSnafu` you could just use `context` here, which would add additional information to the error message about why the parse failed.

cbgbt (commented):
This looks great! There are only a few small changes I'd request.

Seems like the CI is failing due to unformatted sources. Mind running `cargo fmt` on `schnauzer`?

You can check run all of the CI lints, etc, ahead of time by running `make twoliter check`.

sources/api/schnauzer/Cargo.toml
Issacwww: Thanks for sharing this!
encountered in local run. Any clue what might be missed? I simply run the `make twoliter check` and testing looks good
```
[cargo-make] INFO - Running Task: check-migrations
grep: Release.toml: No such file or directory
[cargo-make] ERROR - Error while executing command, exit code: 2
[cargo-make] WARN - Build Failed.
Error: Command was unsuccessful, exit code 1
make: *** [twoliter] Error 1
```

cbgbt: This actually seems to be a bug twoliter, sorry about that. I opened https://github.com/bottlerocket-os/twoliter/issues/359 to track it.

We'll provide a fix for that, but in the meantime you'll have to separately run the individual `check` tasks:  `make twoliter unit-tests check-fmt check-lints`l.

Issacwww: Ah, the individual check looks good to me
`[cargo-make] INFO - Build Done in 52.05 seconds.`

cbgbt (approved):
Looks great! 🚀

yeazelm (approved):
LGTM! Thanks for the contribution!

PR #118, opened 2024-09-03
Build open source NVIDIA kernel modules 

yeazelm:
**Issue number:**

Relates to https://github.com/bottlerocket-os/bottlerocket/issues/4172

**Description of changes:**
This PR updates the 535 branch drivers to the latest version from NVIDIA. This also adds the open source driver to the `kmod-5.15-nvidia` and `kmod-6.1-nvidia` packages but doesn't make them immediately loadable. It is related to https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/114 and helps break out the changes in a smaller change. The intention is to rebase #114 on this change once merged so that the focus of that PR can be on the choosing and loading, not on the packaging. This doesn't affect any functionality as it is right now and will come with the other related PRs as part of https://github.com/bottlerocket-os/bottlerocket/issues/4172

**Testing done:**
Build `aws-k8s-1.30-nvidia` and launched on g5.2xlarge and g3.4xlarge to ensure they boot and load the proprietary driver:
```
[root@admin]# sheltie
bash-5.1# cat /proc/driver/nvidia/version
NVRM version: NVIDIA UNIX x86_64 Kernel Module  535.183.06  Wed Jun 26 06:46:07 UTC 2024
GCC version:  gcc version 11.3.0 (Buildroot 2022.11.1)
bash-5.1# find / -name "open-gpu"
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/nvidia/open-gpu
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/factory/nvidia/open-gpu
bash-5.1# ls -al /x86_64-bottlerocket-linux-gnu/sys-root/usr/share/nvidia/open-gpu/drivers/
total 85780
drwxr-xr-x. 2 root root     4096 Sep  2 23:35 .
drwxr-xr-x. 3 root root     4096 Sep  2 23:35 ..
-rwxr-xr-x. 1 root root  4347040 Aug 31 00:39 nvidia-drm.ko
-rwxr-xr-x. 1 root root  3319056 Aug 31 00:39 nvidia-modeset.ko
-rwxr-xr-x. 1 root root   315704 Aug 31 00:39 nvidia-peermem.ko
-rwxr-xr-x. 1 root root 58070824 Aug 31 00:39 nvidia-uvm.ko
-rwxr-xr-x. 1 root root 22592744 Aug 31 00:39 nvidia.ko
[...]

PR #119, opened 2024-09-03
Add the ability for driverdog to copy modules

yeazelm:
**Issue number:**

Related to  # https://github.com/bottlerocket-os/bottlerocket/issues/4172

**Description of changes:**
This PR fixes a bug in driverdog (see the first commit) and adds the ability for driverdog to read a configuration file that doesn't link. With that new logic it also supports a link option to only copy the drivers to the right place since some modules may already be linked, but still need to be loaded by driverdog. See https://github.com/bottlerocket-os/bottlerocket/issues/4172 for more background. 

**Testing done:**
Added unit tests to ensure the new loading of the file works for linking vs copying files. I also built an `aws-k8s-1.30-nvidia` AMI and ensured things still loaded correctly. More testing will come under the PR that combines all the work together to ensure the two drivers can be chosen at runtime.

```
bash-5.1# cat /proc/driver/nvidia/version
NVRM version: NVIDIA UNIX x86_64 Kernel Module  535.183.06  Wed Jun 26 06:46:07 UTC 2024
GCC version:  gcc version 11.3.0 (Buildroot 2022.11.1)
bash-5.1# exit
exit
[root@admin]# sheltie
bash-5.1# systemctl -u link-kernel-modules.service
systemctl: invalid option -- 'u'
bash-5.1# journalctl -u link-kernel-modules.service
Sep 03 03:21:25 localhost systemd[1]: Starting Link additional kernel modules...
Sep 03 03:21:25 localhost driverdog[2219]: 03:21:25 [INFO] Linked object 'nvidia-modeset.o'
Sep 03 03:21:26 localhost driverdog[2219]: 03:21:26 [INFO] Stripped object 'nvidia-modeset.o'
Sep 03 03:21:34 ip-192-168-92-157.us-west-2.compute.internal driverdog[2219]: 03:21:34 [INFO] Linked object 'nvidia.o'
Sep 03 03:21:34 ip-192-168-92-157.us-west-2.compute.internal driverdog[2219]: 03:21:34 [INFO] Stripped object 'nvidia.o'
Sep 03 03:21:34 ip-192-168-92-157.us-west-2.compute.internal driverdog[2219]: 03:21:34 [INFO] Linked nvidia-uvm.ko
Sep 03 03:21:35 ip-192-168-92-157.us-west-2.compute.internal driverdog[2219]: 03:21:35 [INFO] Linked nvidia.ko
Sep 03 03:21:35 ip-192-168-92-157.us-west-2.compute.internal driverdog[2219]: 03:21:35 [INFO] Linked nvidia-modeset.ko
Sep 03 03:21:35 ip-192-168-92-157.us-west-2.compute.internal systemd[1]: Finished Link additional kernel modules.
bash-5.1# journalctl -u load-kernel-modules.service
Sep 03 03:21:35 ip-192-168-92-157.us-west-2.compute.internal systemd[1]: Starting Load additional kernel modules...
Sep 03 03:21:37 ip-192-168-92-157.us-west-2.compute.internal driverdog[2960]: 03:21:37 [INFO] Updated modules dependencies
Sep 03 03:21:37 ip-192-168-92-157.us-west-2.compute.internal driverdog[2960]: 03:21:37 [INFO] Loaded kernel modules
Sep 03 03:21:37 ip-192-168-92-157.us-west-2.compute.internal systemd[1]: Finished Load additional kernel modules.
```

I also built on top of https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/118 for including both drivers and provided the following config file to make sure it works:
```
[nvidia-open-gpu]
lib-modules-path = "kernel/drivers/extra/video/nvidia/open-gpu"

[nvidia-open-gpu.kernel-modules."nvidia.ko"]
copy-source = "/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/nvidia/open-gpu/drivers/"

[nvidia-open-gpu.kernel-modules."nvidia-modeset.ko"]
copy-source = "/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/nvidia/open-gpu/drivers/"

[nvidia-open-gpu.kernel-modules."nvidia-uvm.ko"]
copy-source = "/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/nvidia/open-gpu/drivers/"
```
Which resulted in:
```
bash-5.1# mkdir -p /lib/modules/6.1.102/kernel/drivers/extra/video/nvidia/open-gpu/
bash-5.1# driverdog --modules-set nvidia-open-gpu link-modules
23:53:20 [INFO] Copied nvidia.ko
23:53:20 [INFO] Copied nvidia-modeset.ko
23:53:21 [INFO] Copied nvidia-uvm.ko
```
Note, to make this work we would need  `nvidia-modeset.ko` and `nvidia-peermem.ko` in the config file for it to fully work as intended.

**Terms of contribution:**

### DIFF
--- a/sources/driverdog/README.md
+++ b/sources/driverdog/README.md
@@ -9,6 +9,11 @@ driverdog is a tool to link kernel modules at runtime. It uses a toml configurat
 `object-files`: hash with the object files to be linked, each object in the map should include the files used to link it
 `kernel-modules`: hash with the kernel modules to be linked, each kernel module in the map should include the files used to link it
 
+There are two modes for driverdog: link then load and copy then load. Link then load takes unlinked files found in `objects-source`
+and matched in `object-files` and `kernel-modules` to link together these files then copy them to `lib-modules-path`. Copy then load
+finds the modules specified in `kernel-modules` and copies them to `lib-modules-path` from the source specified in `copy-source`. Both
+modes iterate over the `kernel-modules` and load them from that path with `modprobe`.
+
 ## Colophon
 
 This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
--- a/sources/driverdog/src/main.rs
+++ b/sources/driverdog/src/main.rs
@@ -292,10 +292,11 @@ fn load_modules_sets(
             .get(&target)
             .context(error::MissingModuleSetSnafu { target })?;
         load_modules(driver_config)?
-    }
-    // Load all the modules sets if no target module was given
-    for driver_config in modules_sets.values() {
-        load_modules(driver_config)?;
+    } else {
+        // Load all the modules sets if no target module was given
+        for driver_config in modules_sets.values() {
+            load_modules(driver_config)?;
+        }
     }
 
     Ok(())
--- a/sources/driverdog/src/main.rs
+++ b/sources/driverdog/src/main.rs
@@ -5,6 +5,11 @@ driverdog is a tool to link kernel modules at runtime. It uses a toml configurat
 `objects-source`: path where the objects used to link the kernel module are
 `object-files`: hash with the object files to be linked, each object in the map should include the files used to link it
 `kernel-modules`: hash with the kernel modules to be linked, each kernel module in the map should include the files used to link it
+
+There are two modes for driverdog: link then load and copy then load. Link then load takes unlinked files found in `objects-source`
+and matched in `object-files` and `kernel-modules` to link together these files then copy them to `lib-modules-path`. Copy then load
+finds the modules specified in `kernel-modules` and copies them to `lib-modules-path` from the source specified in `copy-source`. Both
+modes iterate over the `kernel-modules` and load them from that path with `modprobe`.
 */
 
 #[macro_use]
@@ -17,7 +22,7 @@ use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::{self, Command};
 
 /// Path to the drivers configuration to use
@@ -62,8 +67,8 @@ struct Args {
 #[derive(FromArgs, Debug, PartialEq)]
 #[argh(subcommand)]
 enum Subcommand {
-    LinkModules(LinkModulesArgs),
-    LoadModules(LoadModulesArgs),
+    Link(LinkModulesArgs),
+    Load(LoadModulesArgs),
 }
 
 /// Links the kernel modules
@@ -76,17 +81,31 @@ struct LinkModulesArgs {}
 #[argh(subcommand, name = "load-modules")]
 struct LoadModulesArgs {}
 
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+/// Enum to hold the two types of configurations supported
+enum DriverType {
+    Linking(LinkingDriverConfig),
+    Copying(CopyingDriverConfig),
+}
+
 /// Holds the configuration used to link kernel modules
 #[derive(Deserialize, Debug)]
-struct DriverConfig {
+#[serde(rename_all = "kebab-case")]
+struct LinkingDriverConfig {
+    lib_modules_path: String,
+    objects_source: String,
+    kernel_modules: HashMap<String, Linkable>,
+    object_files: HashMap<String, Linkable>,
+}
+
+/// Holds the configuration used to copy kernel modules
+#[derive(Deserialize, Debug)]
+struct CopyingDriverConfig {
     #[serde(rename(deserialize = "lib-modules-path"))]
     lib_modules_path: String,
-    #[serde(rename(deserialize = "objects-source"))]
-    objects_source: String,
     #[serde(rename(deserialize = "kernel-modules"))]
-    kernel_modules: HashMap<String, Linkable>,
-    #[serde(rename(deserialize = "object-files"))]
-    object_files: HashMap<String, Linkable>,
+    kernel_modules: HashMap<String, NonLinkable>,
 }
 
 /// Holds the objects to be linked for the object/kernel module
@@ -96,9 +115,16 @@ struct Linkable {
     link_objects: Vec<String>,
 }
 
+/// Holds the modules to be copied and loaded
+#[derive(Deserialize, Debug)]
+struct NonLinkable {
+    #[serde(rename(deserialize = "copy-source"))]
+    copy_source: PathBuf,
+}
+
 // Links the modules in the modules sets
 fn link_modules_sets(
-    modules_sets: &HashMap<String, DriverConfig>,
+    modules_sets: &HashMap<String, DriverType>,
     target: Option<String>,
 ) -> Result<()> {
     // Get current kernel version
@@ -109,11 +135,17 @@ fn link_modules_sets(
         let driver_config = modules_sets
             .get(&target)
             .context(error::MissingModuleSetSnafu { target })?;
-        link_modules(driver_config, &kernel_version)?;
+        match driver_config {
+            DriverType::Copying(config) => copy_modules(config, &kernel_version)?,
+            DriverType::Linking(config) => link_modules(config, &kernel_version)?,
+        }
     } else {
         // Link all the modules sets if no target module was given
         for driver_config in modules_sets.values() {
-            link_modules(driver_config, &kernel_version)?;
+            match driver_config {
+                DriverType::Copying(config) => copy_modules(config, &kernel_version)?,
+                DriverType::Linking(config) => link_modules(config, &kernel_version)?,
+            }
         }
     }
 
@@ -121,7 +153,7 @@ fn link_modules_sets(
 }
 
 // Links the kernel modules for the given configuration, and for the given kernel version
-fn link_modules<S>(driver_config: &DriverConfig, kernel_version: S) -> Result<()>
+fn link_modules<S>(driver_config: &LinkingDriverConfig, kernel_version: S) -> Result<()>
 where
     S: AsRef<str>,
 {
@@ -144,7 +176,6 @@ where
         link_object_file(name, object_file, &build_dir, &driver_path)?;
     }
 
-    // Next, link the kernel modules
     for (name, kernel_module) in driver_config.kernel_modules.iter() {
         link_kernel_module(
             name,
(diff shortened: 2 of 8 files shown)


yeazelm:
^ Fix clippy for Enum naming.

sources/driverdog/src/main.rs
```diff
@@ -281,21 +372,32 @@ fn load_modules_sets(
...
+        command(DEPMOD_BIN_PATH, args)?;
+        info!("Updated modules dependencies");
+
         load_modules(driver_config)?
-    }
-    // Load all the modules sets if no target module was given
-    for driver_config in modules_sets.values() {
-        load_modules(driver_config)?;
+    } else {
+        // Update the modules.dep before we attempt to load kernel modules
```
larvacea: This comment applies to both branches of the if-else. We run `depmod` unconditionally, and in the target-not-empty case, whether or not there is a `depmod-file`.

sources/driverdog/src/main.rs
```diff
@@ -2,9 +2,15 @@
 driverdog is a tool to link kernel modules at runtime. It uses a toml configuration file with the following shape:
 
 `lib-modules-path`: destination path for the .ko linked files
-`objects-source`: path where the objects used to link the kernel module are
+`copy-modules-source`: path where the linked files can be found to copy to the lib-modules-path
```
larvacea: `depmod-file` might be included here, too.

sources/driverdog/README.md
```diff
@@ -5,10 +5,16 @@ Current version: 0.1.0
 driverdog is a tool to link kernel modules at runtime. It uses a toml configuration file with the following shape:
 
 `lib-modules-path`: destination path for the .ko linked files
-`objects-source`: path where the objects used to link the kernel module are
+`copy-modules-source`: path where the linked files can be found to copy to the lib-modules-path
```
larvacea: Should this document `depmod-file`? It is part of the toml schema.

sources/driverdog/src/tests/linking.conf
```diff
@@ -0,0 +1,19 @@
...
+
+[linking-driver.object-files."secondary.o"]
+link-objects = ["third.o", "fourth.o"]
+
+[linking-driver.kernel-modules."secondary.ko"]
+link-objects = ["secondary.o", "secondary.mod.o"]
+
+[linking-driver.kernel-modules."other.ko"]
+link-objects = ["other.o", "other.mod.o"]
+
```
bcressey:
```suggestion
```

sources/driverdog/src/tests/copying.conf
```diff
@@ -0,0 +1,14 @@
+[copying-driver]
+lib-modules-path = "kernel/drivers/extra/copying"
+copy-modules-source = "/usr/share/factory/copying-driver"
+
+[copying-driver.kernel-modules."main.ko"]
+link-objects = []
```
bcressey: I would probably structure it like this in order to end up with different types:
```suggestion
[copying-driver]
lib-modules-path = "kernel/drivers/extra/copying"

[copying-driver.kernel-modules."main.ko"]
copy-source = "/usr/share/factory/copying-driver"
```

Corresponding Rust code might look like:
```
enum DriverType {
    Linking(LinkingDriver),
    Copying(CopyingDriver),
}

struct LinkingDriver {
    name: String,
    link_objects: Vec<String>,
}

struct CopyingDriver {
    name: String,
    copy_source: PathBuf
}
```

That way, you can match on the different types and code just the behavior you need for that type, and you can avoid the need for an empty link-objects list.

sources/driverdog/src/main.rs
```diff
@@ -281,21 +372,32 @@ fn load_modules_sets(
...
     if let Some(target) = target {
         let driver_config = modules_sets
             .get(&target)
             .context(error::MissingModuleSetSnafu { target })?;
+
+        // If the configuration specifies a depmod file, include it in the args
+        // Only valid if targetting a specific module set
+        let mut args: Vec<String> = Vec::new();
+        if let Some(depmod_file) = &driver_config.depmod_file {
+            args.push(format!("--config={}", depmod_file));
```
bcressey: Do we need to include the custom depmod config change here? Is it needed for the eventual solution?

sources/driverdog/src/main.rs
```diff
@@ -125,35 +141,37 @@ fn link_modules<S>(driver_config: &DriverConfig, kernel_version: S) -> Result<()
...
-    // This script is used to link the kernel module
-    let common_module_script = Path::new(KERNEL_SOURCES)
-        .join(kernel_version)
-        .join("scripts/module.lds");
-
-    // First, link the object files, and store them in the temp directory
-    for (name, object_file) in driver_config.object_files.iter() {
-        link_object_file(name, object_file, &build_dir, &driver_path)?;
-    }
+    if let Some(objects_source) = &driver_config.objects_source {
```
bcressey: I'd prefer to do this test before calling the function, ideally by matching on the type (if we have different `Copyable` and `Linkable` types).

sources/driverdog/src/main.rs
```diff
@@ -196,24 +214,26 @@ where
...
-            })?
-            .to_string(),
-    ];
-    args.append(&mut dependencies_paths);
-
-    command(LD_BIN_PATH, &args)?;
-    info!("Linked {}", name);
+    if !dependencies_paths.is_empty() {
+        // Link the kernel module
+        let mut args = vec![
```
bcressey: Having a different type would help here, then you'd know that `link_kernel_module` could only be called for a `Linkable` and could keep this logic unchanged.

sources/driverdog/src/main.rs
```diff
@@ -374,8 +476,9 @@ fn run() -> Result<()> {
     }
 
     match args.subcommand {
-        Subcommand::LinkModules(_) => link_modules_sets(&all_modules_sets, args.modules_set),
-        Subcommand::LoadModules(_) => load_modules_sets(&all_modules_sets, args.modules_set),
+        Subcommand::Link(_) => link_modules_sets(&all_modules_sets, args.modules_set),
+        Subcommand::Copy(_) => copy_modules_sets(&all_modules_sets, args.modules_set),
```
bcressey: This could be consolidated into the existing `Subcommand::Link` operation, where you skip the linker invocation and just copy the `.ko` to its destination.

sources/driverdog/src/main.rs
yeazelm: This could be removed. I kept the logic here because I do see value in being able to provide a configuration file to either exclude paths or prefer other paths in case there are more complex interactions between drivers. But it is not immediately needed so I can leave it out.

sources/driverdog/src/main.rs
yeazelm: I see what you mean but the comment doesn't need to be duplicated. Specifically, with the intention of the targeted call, it is more specifically about calling `depmod` with the intention of using specific configuration for priority over finding the modules in a general case. I think that as @bcressey points out, we probably don't need this so I can also just put it back to how it was.

sources/driverdog/src/main.rs
yeazelm: I think that is actually a nice improvement. It is a little strange to only copy the files rather than link them but combining the logic based upon the config provided does simplify the codebase, even if naming is a bit loose on what `link` really means here.

sources/driverdog/src/tests/copying.conf
yeazelm: I'll give this a shot. I think I'll need to update a bit on where `objects-source` is located to make this really consistent but should make the types more helpful for tracking what work is expected to be done.

yeazelm: I've moved to an Enum to provide this type of switching.

yeazelm:
Updated the commit to reflect the two different configuration types: `LinkingDriverConfig` and  `CopyingDriverConfig` to make this logic a bit more robust for choosing between link or copy paths.

sources/driverdog/src/main.rs
```diff
@@ -89,16 +102,32 @@ struct DriverConfig {
     object_files: HashMap<String, Linkable>,
 }
 
+/// Holds the configuration used to copy kernel modules
+#[derive(Deserialize, Debug)]
+struct CopyingDriverConfig {
+    #[serde(rename(deserialize = "lib-modules-path"))]
```
cbgbt, line 105: nit: I think almost all of these can be replaced with [`#[serde(rename_all = kebab-case)]`](https://serde.rs/container-attrs.html#rename_all) on the container. I realize you are just following existing style, though.

sources/driverdog/src/main.rs
```diff
@@ -276,9 +311,49 @@ where
     Ok(())
 }
 
+// Loads the kernel modules for the given configuration, and for the given kernel version
```
cbgbt:
```suggestion
/// Loads the kernel modules for the given configuration, and for the given kernel version
```

sources/driverdog/src/main.rs
```diff
@@ -276,9 +311,49 @@ where
...
+
+    // Next, copy the kernel modules
+    for (name, module) in driver_config.kernel_modules.iter() {
+        copy_kernel_module(name, &modules_path, &module.copy_source)?;
+    }
+
+    Ok(())
+}
+
+// Copy the module to the modules path provided
```
cbgbt:
```suggestion
/// Copy the module to the modules path provided
Comment
```

sources/driverdog/src/main.rs
```diff
@@ -276,9 +311,49 @@ where
...
+fn copy_kernel_module<S, P>(name: S, modules_path: P, driver_source_path: P) -> Result<()>
```
cbgbt: nit: It's silly but these `module_path` and `driver_source_path` can't be arbitrary and different "path-like" types unless you do this.

```suggestion
fn copy_kernel_module<S, P1, P2>(name: S, modules_path: P1, driver_source_path: P2) -> Result<()>
```

sources/driverdog/src/main.rs
```diff
@@ -76,9 +81,17 @@ struct LinkModulesArgs {}
 #[argh(subcommand, name = "load-modules")]
 struct LoadModulesArgs {}
 
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+/// Enum to hold the two types of configurations supported
+enum DriverType {
+    Linking(LinkingDriverConfig),
+    Copying(CopyingDriverConfig),
+}
```
cbgbt, line 90: You have to be a bit careful with `untagged` enum parsing, since Serde just tries to parse the enums in order.

Can you add a test where you try to parse a `LinkingDriverConfig` that doesn't have any `objects_source` or `objects_files`? My guess is that it will load as a `CopyingDriverConfig`.

(Is that even a possible configuration in this domain?)

cbgbt:
Overall I think the config driven by the enums seems quite clean and easy to follow. Nice work!

sources/driverdog/src/main.rs
cbgbt: Hmm, after reading the docstring again, I don't think that's a valid `LinkingDriverConfig`.

yeazelm: This one is tricky. Like you said, Serde will try to match them all, so you must specify `copy-source` for `CopyingDriverConfig` or `link-objects` for `LinkingDriverConfig` since the underlying module configurations are different types one layer below. I added some tests to assert this but it does make this a little loose and additional types would need to be completely distinct to ensure this doesn't unexpectedly fall through to a different configuration.

yeazelm:
^ updated based upon @cbgbt comments

cbgbt (approved):

bcressey (approved):
Very nice!

sources/driverdog/src/main.rs
```diff
@@ -276,9 +308,50 @@ where
...
+    S: AsRef<str>,
+{
+    let kernel_version = kernel_version.as_ref();
+
+    // Destination for the kernel modules
+    let modules_path = Path::new(LIB_MODULES_PATH)
+        .join(kernel_version)
+        .join(&driver_config.lib_modules_path);
+
+    // Next, copy the kernel modules
```
arnaldo2792, line 322: nit: the code below is self-explanatory, you could remove the comment.

sources/driverdog/src/main.rs
```diff
@@ -291,22 +364,36 @@ fn load_modules_sets(
...
+                .collect();
+        }
+        DriverType::Linking(config) => {
+            kernel_modules = config
+                .kernel_modules
+                .keys()
+                .map(|k| k.split('.').collect::<Vec<&str>>()[0].to_string())
+                .collect();
+        }
+    }
```
arnaldo2792: nit:


```suggestion
    let mut kernel_modules: Vec<String> = match driver_config {
        DriverType::Copying(config) => {
           config
                .kernel_modules
                .keys()
                .map(|k| k.split('.').collect::<Vec<&str>>()[0].to_string())
                .collect()
        }
        DriverType::Linking(config) => {
            config
                .kernel_modules
                .keys()
                .map(|k| k.split('.').collect::<Vec<&str>>()[0].to_string())
                .collect()
        }
    }
```

arnaldo2792 (approved):

yeazelm:
^ Fix the comments from @arnaldo2792

cbgbt (approved):

PR #141, opened 2024-09-10
EFA Support - Automated host configurations

### LINKED ISSUE #140: Automated host configurations for instance that has EFA cards attached
**What I'd like:**
I want to support Elastic Fabric Adapter(EFA) on Bottlerocket. The node should be able to configure the hugepages and resource limit for optimal performance with EFA and Libfabric. As recommended in https://github.com/ofiwg/libfabric/blob/main/prov/efa/docs/overview.md:
> Ensure you have configured your instance to increase the locked memory limits (unlimited is fine) and set aside [huge pages](https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt) for the best performance. The EFA Libfabric provider will utilize these for the bounce buffers used for sends and matching receives.

The hugepages are allocatable via `settings.kernel.sysctl` with entry `vm/nr_hugepages = <number_of_hugepages>`. 

The MEMLOCK limit is configurable via `oci-defaults.resource-limits.max-locked-memory`.

The configuration should be conditionally so that they are only applied when EFA card is actually attached.

Manually configure the host is possible, but it is less ideal and more likely to introduce errors in the configuration.

ytsssun:
**Issue number:**
Related to: https://github.com/bottlerocket-os/bottlerocket/issues/1031

Closes #140

**Description of changes:**

Add EFA support which would automatically apply host configurations for hugepages allocation and rlimit bump based on the EFA Device attachment.

- Extended `oci_defaults` helper in schnauzer to apply higher MEMLOCK limits when EFA is attached.
- Add subcommand in `corndog` with a settings generator for hugepages, which will dynamically generate settings for `settings.kernel.sysctl."vm/nr_hugepages"`.
- Add subcommand `efa-present` in `ghostdog` for EFA device detection. It can be used by systemd unit for `ExecCondition` so that we can conditionally register systemd services based on EFA Card attachment.

**Testing done:**
1. Built a custom AMI that: 
- Has `pciutils` installed.
- Is configured with settings generator for `settings.kernel.sysctl."vm/nr_hugepages"` in https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/shared-defaults/defaults.toml#L143
```
[metadata.settings.kernel.sysctl."vm/nr_hugepages"]
setting-generator = "corndog generate-hugepages-setting"
```

2. Launched a EKS nodegroup with the custom AMI on `g4dn.8xlarge` instances with spec:
```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: my-efa-cluster-v2
  region: us-west-2
  version: '1.29'

iam:
  withOIDC: true

nodeGroups:
  - name: my-efa-ng-g4dn-bottlerocket
    instanceType: g4dn.8xlarge
    minSize: 0
    desiredCapacity: 2
    maxSize: 3
    availabilityZones: ["us-west-2a"]
    amiFamily: Bottlerocket
    ami: ami-xxxxxx
    volumeSize: 400
    privateNetworking: true
    efaEnabled: true
``` 
3. Connect to the node and inspect the hugepage settings.
```
[ssm-user@control]$ apiclient get settings.kernel.sysctl
{
  "settings": {
    "kernel": {
      "sysctl": {
        "vm/nr_hugepages": "3520"
      }
    }
  }
}
```
4. Verified that the hugepage settings were applied to the actual `sysctl` and `sysfs` settings.
```
bash-5.1# cat /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
3520

bash-5.1# cat /proc/sys/vm/nr_hugepages
3520
```

5. The describe node output indicates that the hugepages are pre-allocated and can be scheduled for the pods.
```
  cpu:                    32
  ...
  hugepages-1Gi:          0
  hugepages-2Mi:          7040Mi
  ...
  vpc.amazonaws.com/efa:  1
Allocatable:
  cpu:                    31850m
  ...
  hugepages-1Gi:          0
  hugepages-2Mi:          7040Mi
  ...
  vpc.amazonaws.com/efa:  1
```

6. Ran NCCL tests with MPIJobs which requests hugepages
```
 ...
resources:
  limits:
    nvidia.com/gpu: 1
    hugepages-2Mi: 5120Mi
    vpc.amazonaws.com/efa: 1
    memory: 8000Mi
  requests:
    nvidia.com/gpu: 1
    hugepages-2Mi: 5120Mi
    vpc.amazonaws.com/efa: 1
    memory: 8000Mi
``` 

7. Verified that the pod has locked memory limit configured as `unlimited` where the default in Bottlerocket would be 8192KB.
```
# ulimit -l

unlimited
```

8. The NCCL tests ran successfully with expected test result.
```
[1,0]<stdout>:#                                                              out-of-place                       in-place          
[1,0]<stdout>:#       size         count      type   redop    root     time   algbw   busbw #wrong     time   algbw   busbw #wrong
[1,0]<stdout>:#        (B)    (elements)                               (us)  (GB/s)  (GB/s)            (us)  (GB/s)  (GB/s)       
[1,0]<stdout>:           8             2     float     sum      -1    72.39    0.00    0.00      0    71.98    0.00    0.00      0
[1,0]<stdout>:          16             4     float     sum      -1    72.42    0.00    0.00      0    72.05    0.00    0.00      0
[1,0]<stdout>:          32             8     float     sum      -1    72.59    0.00    0.00      0    72.46    0.00    0.00      0
[1,0]<stdout>:          64            16     float     sum      -1    72.55    0.00    0.00      0    72.39    0.00    0.00      0
[1,0]<stdout>:         128            32     float     sum      -1    72.28    0.00    0.00      0    72.66    0.00    0.00      0
[1,0]<stdout>:         256            64     float     sum      -1    73.06    0.00    0.00      0    72.40    0.00    0.00      0
[1,0]<stdout>:         512           128     float     sum      -1    72.92    0.01    0.01      0    72.97    0.01    0.01      0
[1,0]<stdout>:        1024           256     float     sum      -1    74.28    0.01    0.01      0    75.09    0.01    0.01      0
[1,0]<stdout>:        2048           512     float     sum      -1    75.79    0.03    0.03      0    75.62    0.03    0.03      0
[1,0]<stdout>:        4096          1024     float     sum      -1    77.81    0.05    0.05      0    77.94    0.05    0.05      0
[1,0]<stdout>:        8192          2048     float     sum      -1    83.01    0.10    0.10      0    82.59    0.10    0.10      0
[1,0]<stdout>:       16384          4096     float     sum      -1    96.81    0.17    0.17      0    96.54    0.17    0.17      0
[1,0]<stdout>:       32768          8192     float     sum      -1    107.6    0.30    0.30      0    106.7    0.31    0.31      0
[1,0]<stdout>:       65536         16384     float     sum      -1    123.0    0.53    0.53      0    124.4    0.53    0.53      0
[1,0]<stdout>:      131072         32768     float     sum      -1    155.3    0.84    0.84      0    155.7    0.84    0.84      0
[1,0]<stdout>:      262144         65536     float     sum      -1    208.0    1.26    1.26      0    194.3    1.35    1.35      0
[1,0]<stdout>:      524288        131072     float     sum      -1    290.8    1.80    1.80      0    290.4    1.81    1.81      0
[1,0]<stdout>:     1048576        262144     float     sum      -1    604.2    1.74    1.74      0    602.8    1.74    1.74      0
[1,0]<stdout>:     2097152        524288     float     sum      -1   1006.6    2.08    2.08      0   1013.7    2.07    2.07      0
[1,0]<stdout>:     4194304       1048576     float     sum      -1   1731.7    2.42    2.42      0   1732.0    2.42    2.42      0
[1,0]<stdout>:     8388608       2097152     float     sum      -1   3117.2    2.69    2.69      0   3174.8    2.64    2.64      0
[1,0]<stdout>:    16777216       4194304     float     sum      -1   6117.1    2.74    2.74      0   6113.3    2.74    2.74      0
[1,0]<stdout>:    33554432       8388608     float     sum      -1    11653    2.88    2.88      0    11722    2.86    2.86      0
[1,0]<stdout>:    67108864      16777216     float     sum      -1    22642    2.96    2.96      0    22367    3.00    3.00      0
[1,0]<stdout>:   134217728      33554432     float     sum      -1    44380    3.02    3.02      0    43808    3.06    3.06      0
[1,0]<stdout>:   268435456      67108864     float     sum      -1    87872    3.05    3.05      0    89709    2.99    2.99      0
[1,0]<stdout>:   536870912     134217728     float     sum      -1   176409    3.04    3.04      0   177627    3.02    3.02      0
[1,0]<stdout>:  1073741824     268435456     float     sum      -1   351989    3.05    3.05      0   351789    3.05    3.05      0
[1,0]<stdout>:  2147483648     536870912     float     sum      -1   700468    3.07    3.07      0   706521    3.04    3.04      0
[1,0]<stdout>:# Out of bounds values : 0 OK
[1,0]<stdout>:# Avg bus bandwidth    : 1.30573 
```

9. Registered a dummy systemd unit:
```
[Unit]
Description=Example Service with ExecCondition for EFA detection
After=network.target sundog.service

[Service]
Type=simple
ExecCondition=/usr/bin/ghostdog efa-present
ExecStart=/bin/echo "EFA Device Detected!!! Here is the confirming message!"

[Install]
WantedBy=multi-user.target
```

Verified from the g4dn.8xlarge instance that the custom systemd unit was invoked.
```
bash-5.1# systemctl status test-efa-attach
○ test-efa-attach.service - Example Service with ExecCondition
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/test-efa-attach.service; enabled; preset: enabled)
     Active: inactive (dead) since Fri 2024-09-13 06:11:17 UTC; 43min ago
   Duration: 2ms
   Main PID: 3857 (code=exited, status=0/SUCCESS)
        CPU: 4ms

Sep 13 06:11:17 ip-192-168-87-183.us-west-2.compute.internal systemd[1]: Starting Example Service with ExecCondition...
Sep 13 06:11:17 ip-192-168-87-183.us-west-2.compute.internal systemd[1]: Started Example Service with ExecCondition.
Sep 13 06:11:17 ip-192-168-87-183.us-west-2.compute.internal echo[3857]: EFA Device Detected!!! Here is the confirming message!
Sep 13 06:11:17 ip-192-168-87-183.us-west-2.compute.internal systemd[1]: test-efa-attach.service: Deactivated successfully.
```

10. Launched a g4dn instance with EFA disabled. Verified that the hugepages were set to 0 by default
```
bash-5.1# apiclient get settings.kernel
{
  "settings": {
    "kernel": {
      "lockdown": "none",
      "sysctl": {
        "vm/nr_hugepages": "0"
      }
    }
  }
}
```

Also verified that the locked memory limit is 8096KB (as before).
```
max locked memory (kb)          (-l) 8192
```

And the systemd unit was not enabled due to ExecCondition was not met.
```
bash-5.1# systemctl status test-efa-attach
○ test-efa-attach.service - Example Service with ExecCondition
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/test-efa-attach.service; enabled; preset: enabled)
     Active: inactive (dead) (Result: exec-condition) since Wed 2024-09-18 04:38:00 UTC; 19min ago
  Condition: start condition failed at Wed 2024-09-18 04:38:00 UTC; 19min ago
        CPU: 4ms

Sep 18 04:38:00 ip-192-168-72-59.us-west-2.compute.internal systemd[1]: Starting Example Service with ExecCondition...
Sep 18 04:38:00 ip-192-168-72-59.us-west-2.compute.internal ghostdog[1476]: Does not detect EFA
Sep 18 04:38:00 ip-192-168-72-59.us-west-2.compute.internal systemd[1]: test-efa-attach.service: Skipped due to 'exec-condition'.
Sep 18 04:38:00 ip-192-168-72-59.us-west-2.compute.internal systemd[1]: Condition check resulted in Example Service with ExecCondition being skipped.
```

**Terms of contribution:**

### DIFF
--- a/sources/api/corndog/src/main.rs
+++ b/sources/api/corndog/src/main.rs
@@ -3,6 +3,8 @@ corndog is a delicious way to get at the meat inside the kernels.
 It sets kernel-related settings, for example:
 * sysctl values, based on key/value pairs in `settings.kernel.sysctl`
 * lockdown mode, based on the value of `settings.kernel.lockdown`
+
+corndog also provides a settings generator for hugepages, subcommand "generate-hugepages-setting".
 */
 
 use bottlerocket_modeled_types::{Lockdown, SysctlKey};
@@ -20,6 +22,10 @@ use std::{env, process};
 const SYSCTL_PATH_PREFIX: &str = "/proc/sys";
 const LOCKDOWN_PATH: &str = "/sys/kernel/security/lockdown";
 const DEFAULT_CONFIG_PATH: &str = "/etc/corndog.toml";
+const NR_HUGEPAGES_PATH_SYSCTL: &str = "/proc/sys/vm/nr_hugepages";
+/// Number of hugepages we will assign per core.
+/// See [`compute_hugepages_for_efa`] for more detail on the computation consideration.
+const HUGEPAGES_2MB_PER_CORE: u64 = 110;
 
 /// Store the args we receive on the command line.
 struct Args {
@@ -45,19 +51,29 @@ fn run() -> Result<()> {
     SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
 
     // If the user has kernel settings, apply them.
-    let kernel = get_kernel_settings(args.config_path)?;
     match args.subcommand.as_ref() {
         "sysctl" => {
+            let kernel = get_kernel_settings(args.config_path)?;
             if let Some(sysctls) = kernel.sysctl {
                 debug!("Applying sysctls: {:#?}", sysctls);
                 set_sysctls(sysctls);
             }
         }
         "lockdown" => {
+            let kernel = get_kernel_settings(args.config_path)?;
             if let Some(lockdown) = kernel.lockdown {
                 debug!("Setting lockdown: {:#?}", lockdown);
                 set_lockdown(&lockdown)?;
             }
+        }
+        "generate-hugepages-setting" => {
+            let hugepages_setting = generate_hugepages_setting()?;
+            // We will only fail if we cannot serialize the output to JSON string.
+            // sundog expects JSON-serialized output so that many types can be represented, allowing the
+            // API model to use more accurate types.
+            let output =
+                serde_json::to_string(&hugepages_setting).context(error::SerializeJsonSnafu)?;
+            println!("{}", output);
         }
         _ => usage_msg(format!("Unknown subcommand '{}'", args.subcommand)), // should be unreachable
     }
@@ -105,6 +121,55 @@ where
             error!("Failed to write sysctl value '{}': {}", key, e);
         }
     }
+}
+
+/// Generate the hugepages setting for defaults.
+fn generate_hugepages_setting() -> Result<String> {
+    // Check if customer has directly written to the nr_hugepage file.
+    let mut hugepages = fs::read_to_string(NR_HUGEPAGES_PATH_SYSCTL)
+        .map(check_for_existing_hugepages)
+        .unwrap_or("0".to_string());
+
+    // Check for EFA and compute if necessary, only when hugepages is "0".
+    if &hugepages == "0" && pciclient::is_efa_attached().unwrap_or(false) {
+        // We will use [`num_cpus`] to get the number of cores for the compute.
+        hugepages = compute_hugepages_for_efa(num_cpus::get());
+    }
+    Ok(hugepages)
+}
+
+// Check if customer has directly written to the nr_hugepage file.
+//
+// This would be a rare case to hit, as customer would normally modify the hugepages value
+// via settings API. (It could happen with a custom variant if hugepages
+// are set via a sysctl.d drop-in, for example.)
+//
+// We expect the existing_hugepages_value to be valid numeric digits. Otherwise, we will
+// use "0" as default.
+fn check_for_existing_hugepages(existing_hugepages_value: String) -> String {
+    match existing_hugepages_value.trim().parse::<u64>() {
+        Ok(value) => {
+            return value.to_string();
+        }
+        Err(err) => {
+            warn!(
+                "Failed to parse the existing hugepage value, using 0 as default. Error: {}",
+                err
+            );
+        }
+    }
+    "0".to_string()
+}
+
+/// Computation:
+/// - We need to allocate 110MB memory for each libfabric endpoint.
+/// - For optimal setup, Open MPI will open 2 libfabric endpoints each core.
+/// - The total number of hugepages will be set as (110MB * 2) * number_of_cores / hugepage_size
+/// - We will allocate default hugepage_size = 2MB.
+/// - The number of hugepage per core would be 110MB * 2 / 2MB = 110.
+fn compute_hugepages_for_efa(num_cores: usize) -> String {
+    let number_of_hugepages = num_cores as u64 * HUGEPAGES_2MB_PER_CORE;
+    number_of_hugepages.to_string()
 }
 
 /// Sets the requested lockdown mode in the kernel.
@@ -165,6 +230,7 @@ fn usage() -> ! {
     Subcommands:
         sysctl
         lockdown
+        generate-hugepages-setting
 
     Global arguments:
         --config-path PATH
@@ -207,7 +273,7 @@ fn parse_args(args: env::Args) -> Args {
                 )
             }
 
-            "sysctl" | "lockdown" => subcommand = Some(arg),
+            "sysctl" | "lockdown" | "generate-hugepages-setting" => subcommand = Some(arg),
 
             _ => usage(),
         }
@@ -251,6 +317,9 @@ mod error {
             source: Box<toml::de::Error>,
         },
 
+        #[snafu(display("Error serializing to JSON: {}", source))]
+        SerializeJson { source: serde_json::error::Error },
+
         #[snafu(display(
             "Failed to change lockdown from '{}' to '{}': {}",
             current,
@@ -271,6 +340,8 @@ type Result<T> = std::result::Result<T, error::Error>;
 
 #[cfg(test)]
 mod test {
+    use test_case::test_case;
+
     use super::*;
 
     #[test]
@@ -304,5 +375,22 @@ mod test {
             "none integrity confidentiality",
             parse_kernel_setting("none integrity confidentiality\n")
         );
+    }
+
+    #[test]
+    fn test_compute_hugepages_for_efa() {
+        let num_cores: usize = 2;
+        let computed_hugepages = compute_hugepages_for_efa(num_cores);
+        assert_eq!(computed_hugepages, "220")
+    }
+
+    #[test_case("".to_string(), "0".to_string())]
+    #[test_case("0".to_string(), "0".to_string())]
+    #[test_case("-1".to_string(), "0".to_string())]
+    #[test_case("abc".to_string(), "0".to_string())]
+    #[test_case("100".to_string(), "100".to_string())]
+    fn test_check_for_existing_hugepages(existing_value: String, expected_hugepages: String) {
+        let actual_hugepages = check_for_existing_hugepages(existing_value);
+        assert_eq!(actual_hugepages, expected_hugepages);
     }
 }
--- a/sources/api/schnauzer/src/helpers/mod.rs
+++ b/sources/api/schnauzer/src/helpers/mod.rs
@@ -125,6 +125,8 @@ const IPV6_LOCALHOST: IpAddr = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1))
 
 const DEFAULT_ECS_METADATA_SERVICE_RPS: i32 = 40;
 const DEFAULT_ECS_METADATA_SERVICE_BURST: i32 = 60;
+/// We use -1 to indicate unlimited value for resource limits.
+const RLIMIT_UNLIMITED: i64 = -1;
 
 /// Potential errors during helper execution
 mod error {
@@ -257,6 +259,9 @@ mod error {
             ip_address: String,
             source: std::net::AddrParseError,
         },
+
+        #[snafu(display("Failed to check if EFA device is attached: {}", source))]
+        CheckEfaFailure { source: pciclient::PciClientError },
 
         #[snafu(display(
             "Expected an absolute URL, got '{}' in template '{}': '{}'",
@@ -1250,7 +1255,7 @@ pub fn oci_defaults(
             runtime.get_capabilities(capabilities)
         }
         OciSpecSection::ResourceLimits => {
-            let rlimits = oci_spec_resource_limits(oci_defaults_values)?;
+            let rlimits = generate_oci_resource_limits(oci_defaults_values, EfaLspciDetector {})?;
             rlimits
                 .iter()
                 .map(|(rlimit_type, values)| runtime.get_resource_limits(rlimit_type, values))
@@ -1308,10 +1313,40 @@ fn oci_spec_capabilities(value: &Value) -> Result<String, RenderError> {
 /// This helper function generates the resource limits section of
 /// the OCI runtime spec from the provided `value` parameter, which is
 /// the settings data from the datastore (`settings.oci-defaults.resource-limits`).
+fn generate_oci_resource_limits<T: EfaDetector>(
+    value: &Value,
+    efa_detector: T,
+) -> Result<HashMap<OciDefaultsResourceLimitType, OciDefaultsResourceLimitV1>, RenderError> {
+    let mut rlimits = oci_spec_resource_limits(value)?;
+    if efa_detector.is_efa_attached()? {
+        // We need to increase the locked memory limits from the default 8096KB to unlimited
+        // to account for hugepages allocation.
+        rlimits
+            .entry(OciDefaultsResourceLimitType::MaxLockedMemory)
+            .or_insert(OciDefaultsResourceLimitV1 {
+                soft_limit: RLIMIT_UNLIMITED,
+                hard_limit: RLIMIT_UNLIMITED,
+            });
+    }
+    Ok(rlimits)
+}
+
 fn oci_spec_resource_limits(
     value: &Value,
 ) -> Result<HashMap<OciDefaultsResourceLimitType, OciDefaultsResourceLimitV1>, RenderError> {
     Ok(serde_json::from_value(value.clone())?)
+}
+
+trait EfaDetector {
+    fn is_efa_attached(&self) -> Result<bool, TemplateHelperError>;
+}
+
+struct EfaLspciDetector;
+
+impl EfaDetector for EfaLspciDetector {
+    fn is_efa_attached(&self) -> Result<bool, TemplateHelperError> {
+        pciclient::is_efa_attached().context(error::CheckEfaFailureSnafu)
+    }
 }
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
@@ -1992,6 +2027,22 @@ mod test_oci_spec {
     use serde_json::json;
     use OciDefaultsResourceLimitType::*;
 
+    // Custom struct that will always show that EFA is detected.
+    struct EfaPresentDetector;
+    impl EfaDetector for EfaPresentDetector {
+        fn is_efa_attached(&self) -> Result<bool, TemplateHelperError> {
+            Ok(true)
+        }
+    }
+
+    // Custom struct that will always show that EFA is not detected.
+    struct EfaNotPresentDetector;
+    impl EfaDetector for EfaNotPresentDetector {
+        fn is_efa_attached(&self) -> Result<bool, TemplateHelperError> {
+            Ok(false)
+        }
+    }
+
     #[test]
     fn oci_spec_capabilities_test() {
         let json = json!({
@@ -2056,6 +2107,28 @@ mod test_oci_spec {
         for rlimit in rlimits {
             check_all_rlimits(rlimit);
         }
+    }
+
+    #[test]
+    fn generate_oci_resource_limits_efa_detected() {
+        let json = json!({"max-open-files": {"hard-limit": 1, "soft-limit": 2}});
+        let rlimits = generate_oci_resource_limits(&json, EfaPresentDetector {}).unwrap();
+        let rendered = Containerd::get_resource_limits(
+            &MaxLockedMemory,
+            rlimits.get(&MaxLockedMemory).unwrap(),
+        );
+        assert_eq!(
+            rendered,
+            r#"{ "type": "RLIMIT_MEMLOCK", "hard": 18446744073709551615, "soft": 18446744073709551615 }"#
+        );
+    }
+
+    #[test]
+    fn generate_oci_resource_limits_efa_not_detected() {
+        let json = json!({"max-open-files": {"hard-limit": 1, "soft-limit": 2}});
+        let rlimits = generate_oci_resource_limits(&json, EfaNotPresentDetector {}).unwrap();
+        // If EFA is not detected, we will not set the max-locked-memory rlimit
+        assert_eq!(rlimits.get(&MaxLockedMemory), None)
     }
 
     #[test]
--- a/sources/ghostdog/src/main.rs
+++ b/sources/ghostdog/src/main.rs
@@ -1,6 +1,7 @@
 /*!
 ghostdog is a tool to manage ephemeral disks.
 It can be called as a udev helper program to identify ephemeral disks.
+It can also be called for EFA device detection which can be used for ExecCondition in systemd units.
 */
 
 use argh::FromArgs;
@@ -30,7 +31,13 @@ struct Args {
 enum SubCommand {
     Scan(ScanArgs),
     EbsDeviceName(EbsDeviceNameArgs),
+    EfaPresent(EfaPresentArgs),
 }
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "efa-present")]
+/// Detect if EFA devices are attached.
+struct EfaPresentArgs {}
 
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand, name = "scan")]
@@ -63,8 +70,19 @@ fn run() -> Result<()> {
             let device_name = find_device_name(format!("{}", path.display()))?;
             emit_device_name(&device_name);
         }
+        SubCommand::EfaPresent(_) => {
+            is_efa_attached()?;
+        }
     }
     Ok(())
+}
+
+fn is_efa_attached() -> Result<()> {
+    if pciclient::is_efa_attached().context(error::CheckEfaFailureSnafu)? {
+        Ok(())
+    } else {
+        Err(error::Error::NoEfaPresent)
+    }
 }
 
 /// Find the device type by examining the partition table, if present.
@@ -176,6 +194,10 @@ mod error {
         },
         #[snafu(display("Invalid device info for device '{}'", path.display()))]
         InvalidDeviceInfo { path: std::path::PathBuf },
+        #[snafu(display("Failed to check if EFA device is attached: {}", source))]
+        CheckEfaFailure { source: pciclient::PciClientError },
+        #[snafu(display("Did not detect EFA"))]
+        NoEfaPresent,
     }
 }
 
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -3575,6 +3575,7 @@ dependencies = [
  "maplit",
  "models",
  "num_cpus",
+ "pciclient",
  "percent-encoding",
  "pest",
  "pest_derive",
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2149,6 +2149,7 @@ dependencies = [
  "gptman",
  "hex-literal",
  "lazy_static",
+ "pciclient",
  "signpost",
  "snafu",
 ]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -1553,10 +1553,13 @@ dependencies = [
  "bottlerocket-modeled-types",
  "generate-readme",
  "log",
+ "num_cpus",
+ "pciclient",
  "serde",
  "serde_json",
  "simplelog",
  "snafu",
+ "test-case",
  "toml",
 ]
 
--- a/sources/api/corndog/Cargo.toml
+++ b/sources/api/corndog/Cargo.toml
@@ -11,6 +11,8 @@ exclude = ["README.md"]
 
 [dependencies]
 log.workspace = true
+num_cpus.workspace = true
+pciclient.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 simplelog.workspace = true
@@ -20,3 +22,6 @@ bottlerocket-modeled-types.workspace = true
 
 [build-dependencies]
 generate-readme.workspace = true
+
+[dev-dependencies]
+test-case.workspace = true
--- a/sources/api/corndog/README.md
+++ b/sources/api/corndog/README.md
@@ -7,6 +7,8 @@ It sets kernel-related settings, for example:
 * sysctl values, based on key/value pairs in `settings.kernel.sysctl`
 * lockdown mode, based on the value of `settings.kernel.lockdown`
 
+corndog also provides a settings generator for hugepages, subcommand "generate-hugepages-setting".
+
 ## Colophon
 
 This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
--- a/sources/api/schnauzer/Cargo.toml
+++ b/sources/api/schnauzer/Cargo.toml
@@ -29,6 +29,7 @@ log.workspace = true
 maplit.workspace = true
 models.workspace = true
 num_cpus.workspace = true
+pciclient.workspace = true
 percent-encoding.workspace = true
 pest.workspace = true
 pest_derive.workspace = true
--- a/sources/ghostdog/Cargo.toml
+++ b/sources/ghostdog/Cargo.toml
@@ -13,6 +13,7 @@ argh.workspace = true
 gptman.workspace = true
 hex-literal.workspace = true
 lazy_static.workspace = true
+pciclient.workspace = true
 signpost.workspace = true
 snafu.workspace = true
 
--- a/sources/ghostdog/README.md
+++ b/sources/ghostdog/README.md
@@ -4,6 +4,7 @@ Current version: 0.1.0
 
 ghostdog is a tool to manage ephemeral disks.
 It can be called as a udev helper program to identify ephemeral disks.
+It can also be called for EFA device detection which can be used for ExecCondition in systemd units.
 
 ## Colophon
 


ytsssun:
Manually cancelled workflow, we will run this after flipping the PR off DRAFT mode.

arnaldo2792:
If you want to prevent running CI workflows, use `[skip ci]` in your commit message

cbgbt:
Do you mind adding some tests in the PCI crate? It would help me understand the public interface a bit more clearly.

I have some tips to help with testing:
You already have a trait for `CommandExecutor`, which is awesome. In your tests, I'd recommend finding a way to implement this trait with some kind of test double -- you feed it pre-canned `lspci` output and make sure that your high-level API responds the right way.

In order to make ^ easier, you may need to wrap your API under a struct, which you can optionally construct with your own `CommandExecutor`. You can introduce these generics in a way that does not pollute the public API.

[Here's an example that I created on the Rust playground.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dc553c89f1086c8e821ea4acde51373d)

cbgbt:
An alternative to this suggestion would just be to write some tests that call your parsing functions!

I like wrapping APIs that do IO in a trait because it helps you write tests for a much larger portion of your code, but it does take more time.

ytsssun:
> An alternative to this suggestion would just be to write some tests that call your parsing functions!
> 
> I like wrapping APIs that do IO in a trait because it helps you write tests for a much larger portion of your code, but it does take more time.

Thanks, I was indeed adding more unit tests for that. Just updated the PR. I flipped the PR from DRAFT mode to "Ready for review".

In the unit test I implemented the mocking struct for the `lspci` command execution so that I can make sure the rest of the code is tested. Let me know what you think!

sources/ghostdog/src/main.rs
```diff
@@ -63,10 +70,21 @@ fn run() -> Result<()> {
...
+        }
     }
     Ok(())
 }
 
+fn is_efa_attached() -> Result<()> {
+    if pciclient::is_efa_attached().context(error::CheckEfaFailureSnafu)? {
+        process::exit(0);
+    } else {
+        process::exit(1);
```
yeazelm: We wrap the `main` with error handling. I think it would be fine to pass the error back and allow exit code 1 by default.

sources/ghostdog/src/main.rs
```diff
@@ -10,7 +11,7 @@ use lazy_static::lazy_static;
 use signpost::uuid_to_guid;
 use snafu::{ensure, ResultExt};
 use std::collections::HashSet;
-use std::fs;
+use std::{fs, process};
```
yeazelm: `process` is already used in `main()` without this import, we should stay consistent with usage in this file.

sources/ghostdog/src/main.rs
```diff
@@ -30,8 +31,14 @@ struct Args {
 enum SubCommand {
     Scan(ScanArgs),
     EbsDeviceName(EbsDeviceNameArgs),
+    DetectEfaDevice(DetectEfaDeviceArgs),
 }
 
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "detect-efa")]
```
yeazelm: This returns an error if EFA isn't present, I think it is more clear to make this `efa-present` so that exiting non-zero is expected.

sources/ghostdog/src/main.rs
```diff
@@ -63,10 +70,21 @@ fn run() -> Result<()> {
...
         }
+        SubCommand::DetectEfaDevice(_) => {
+            is_efa_attached()?;
+        }
     }
     Ok(())
 }
 
+fn is_efa_attached() -> Result<()> {
+    if pciclient::is_efa_attached().context(error::CheckEfaFailureSnafu)? {
```
yeazelm, line 81: I think we need to be able to emit the difference between checking the PCI bus and not finding EFA vs being unable to call `lspci` or other coding errors. Can we be a bit more particular about the errors here to ensure we treat these two classes of errors differently. Ideally we would emit a message like "No EFA detected" in the normal case and otherwise provide a more specific error if it was an issue with running the commands and getting valid output.

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+    program_interface: String,
+}
+
+impl ListDevicesParam {
+    /// Convert the ListDevicesParam into proper command line arguments so that
+    /// we can feed it to the lspci() function where it makes the actual call to the
+    /// lspci binary.
+    ///
+    fn to_command_args(self) -> Vec<OsString> {
+        let mut args: Vec<String> = vec!["-n".into(), "-m".into()];
```
yeazelm: nit: you are hard coding these arguments in this function, but you didn't explain why in the docs above. It might be worth referring to the `list_devices` function for more info that this just implements the contract in that function. Since this is the top of the file, I read this and had to get to the bottom before I found why these were the arguments you had in the call.

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+#[derive(Debug, Default, PartialEq, Eq)]
+pub struct ListDevicesOutput {
+    pci_slot: String,
+    class: String,
+    vendor: String,
+    device: String,
+    subsystem_vendor: Option<String>,
+    subsystem_device: Option<String>,
+    revision: Option<String>,
+    program_interface: Option<String>,
```
yeazelm: These are private, so we probably need to provide a way to get some of them? Or is the expectation to add functions to this library to do anything with this output?

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+            }]
+        )
+    }
+
+    #[test]
+    fn test_is_efa_attached() {
+        let mock_pci_client = MockPciClient {
+            // EFA device has class code: 0200 for ethernet controller, vendor 1d0f for amazon, device code efa0.
+            // Below is an actual output from lspci for efa device.
+            output: vec![r#"00:1d.0 "0200" "1d0f" "efa0" -p00 "1d0f" "efa0""#.to_string()],
```
yeazelm: Can you add a case that has more than one output to make sure you'll still find it even if there are multiple devices returned, or is this not expected to be the case?

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+#[derive(Debug, Default)]
+pub struct ListDevicesParamBuilder {
+    param: ListDevicesParam,
+}
+
+impl ListDevicesParamBuilder {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
```
cbgbt: Builder pattern is great! But I'd argue there's an ideal way to implement it that's time-consuming to do. The tl;dr of my suggestion here is to add a dependency on [`bon`](https://docs.rs/bon/latest/bon/#builder-for-a-struct) and use `#[derive(Builder)]` which will do the ideal implementation on your behalf and not cost you time. Another alternative is [`typed-builder`](https://crates.io/crates/typed-builder).

The longer version -- most simple implementations of the `Builder` pattern usually have a few unsavory edges. For your implementation specifically, it's not clear what arguments are Optional and which aren't just by looking at the types. Seems like they are all required? What if we want to add an optional field later?

Wouldn't it be great if we could get the compiler to catch someone not providing all of the required methods? There's a cool pattern you can use to accomplish this in Rust called the TypeState pattern.

[Check out this blog post describing how it works.](https://www.greyblake.com/blog/builder-with-typestate-in-rust/)

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+                    // s_vendor and s_device is captured together.
+                    if list_devices_output.subsystem_vendor.is_some() {
+                        // We will ignore this token since we do not recognize them.
+                        // According to the doc:
+                        // > New options can be added in future versions, but they will always have a single argument not
+                        // > separated from the option by any spaces, so they can be easily ignored if not recognized.
+                        continue;
+                    }
+                    list_devices_output.subsystem_vendor =
+                        Some(unquote(token)).filter(|s| !s.is_empty());
```
yeazelm: So `""` will become `Some("")`?

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+            (ListDevicesParamBuilder::new(), vec!["-n", "-m"]),
+            (
+                ListDevicesParamBuilder::new().vendor("1d0f"),
+                vec!["-n", "-m", "-d", "1d0f:::"],
+            ),
+            (
+                ListDevicesParamBuilder::new().device("efa0"),
+                vec!["-n", "-m", "-d", ":efa0::"],
+            ),
+            (
```
cbgbt: Nice tests!

Unlike a lot of other test runners, cargo can't run the rest of the test cases in here if one of these fails, so most people prefer to create an individual `#[test]` per test-case.

There's a crate called [`test_case`](https://crates.io/crates/test_case) which provides a macro that makes that ergonomic and easy.

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+/// > described in this section. All other formats are likely to change between versions of lspci.
+///
+/// * We will use `-m` to get the machine-readable format which ensures compatibility.
+/// * We will use `-n` to get the numeric output for the vendor code and device codes.
+/// * We will optionally insert `-d <selection_expression>` to filter the output.
+///
+/// # Output
+/// Output is [`Result<Vec<ListDevicesOutput>>`], the [`ListDevicesOutput`] is based on
+/// the output format that is decorated by "-n -m".
+///
```
yeazelm: These docs are helpful for the nitty gritty, but it doesn't tell me really anything about how you expect one to use this function, which is actually the most important part of this docstring.

yeazelm (commented):
I've only reviewed the pciclient so far but wanted to get that feedback in now.

sources/pciclient/src/lib.rs
ytsssun: I can add accessors for the fields. My original intention is to limit access to these fields within the crate so that we perform higher level output for external usage.

sources/pciclient/src/lib.rs
ytsssun: So `""` will become None. For better readability, I can add a comment here. I can also add a helper `empty_string_to_none` that does this filter.

sources/pciclient/src/lib.rs
ytsssun: Will add example for this.

sources/pciclient/src/lib.rs
ytsssun: Totally valid ask, I can extend this output to multiple devices, it should still be able to detect.

sources/pciclient/src/lib.rs
ytsssun: Yes, this is mentioned in the `list_devices` function. I will refer to the function so that it is a bit more clear on why we choose `-n` and `-m` here.

sources/ghostdog/src/main.rs
ytsssun: Sounds good, will do.

sources/ghostdog/src/main.rs
ytsssun: Will do.

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+    check_efa_attachment(PciClient {})
+}
+
+mod error {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+
+    pub enum Error {
```
cbgbt: Mind giving this error a more descriptive name, since it's part of the public API?

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+/// * We will use `-n` to get the numeric output for the vendor code and device codes.
+/// * We will optionally insert `-d <selection_expression>` to filter the output.
+///
+/// # Output
+/// Output is [`Result<Vec<ListDevicesOutput>>`], the [`ListDevicesOutput`] is based on
+/// the output format that is decorated by "-n -m".
+///
+pub fn list_devices(list_devices_param: ListDevicesParam) -> Result<Vec<ListDevicesOutput>> {
+    call_list_devices(PciClient {}, list_devices_param)
+}
```
cbgbt: It's a nit, but moving the private methods into their own submodule may be helpful, that way consumers of this crate only have to peek at one place to see the public API.

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -1250,7 +1253,10 @@ pub fn oci_defaults(
             runtime.get_capabilities(capabilities)
         }
         OciSpecSection::ResourceLimits => {
-            let rlimits = oci_spec_resource_limits(oci_defaults_values)?;
+            let mut rlimits = oci_spec_resource_limits(oci_defaults_values)?;
+            if pciclient::is_efa_attached().context(error::CheckEfaFailureSnafu)? {
+                adjust_resource_limits_for_efa(&mut rlimits);
+            }
```
cbgbt: Ah, I think these unit tests are failing. Exposing the generic `lspci` backend to consumers so they can write tests might be helpful (or implementing a similar interface-based backend here)

sources/pciclient/src/lib.rs
cbgbt: A high-level API is a good idea if you have an idea for one, it makes it easier to change the underlying implementation if we need to.

I think you have a case now where you need to choose how to expose the results -- I think `list_devices`' output is all private with no way to access the results.

sources/api/schnauzer/src/helpers/mod.rs
ytsssun: Right, I moved the detect EFA block out of the `oci_spec_resource_limits`, now the unit tests are passing. The `oci_defaults` is not included in the unit tests so putting this logic here is not breaking any tests. I also made sure the `adjust_resource_limits_for_efa` logic is tested (added a test case for it).

ytsssun: Right now the CI is not happy with the doctest, I will modify the docstring so that only the rust code example can be ran.

sources/pciclient/src/lib.rs
ytsssun: Thanks, I will update this to use `bon` for the derive(Builder).

sources/pciclient/src/lib.rs
ytsssun: Right, given that list_devices already emits the output, it is better to add accessor to the outputs or make them public.

sources/pciclient/src/lib.rs
ytsssun: Sounds good. Will move the private functions to submodule, any recommendation on the naming?

sources/pciclient/src/lib.rs
ytsssun: Will do.

sources/pciclient/src/lib.rs
ytsssun: Nice, this is neat! Thanks

sources/pciclient/src/lib.rs
ytsssun: Looks like we did not have dependency on [`bon`](https://docs.rs/bon/latest/bon/#builder-for-a-struct) in our codebase. Should I still add it? I also noticed that we implemented our own builder in other crates (e.g. in [`netdog`](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/fd5148f12d6ebf8e1000f3e2529cfd7b8fd6f66b/sources/netdog/src/networkd/config/netdev.rs#L195-L230).

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +121,19 @@ where
...
 }
 
+/// Computation:
+/// - We need to allocate 110MB memory for each libfabric endpoint.
+/// - For optimal setup, Open MPI will open 2 libfabric endpoints each core.
+/// - The total number of hugepages will be set as (110MB * 2) * number_of_cores / hugepage_size
+/// - We will allocate default hugepage_size = 2MB.
+/// - The number of hugepage per core would be 110MB * 2 / 2MB = 110.
+/// We will use [`num_cpus`] to get the number of cores for the compute.
+fn compute_hugepages() -> Result<String> {
```
yeazelm: nit: this is how to compute hugepages for EFA. Is this the only right value or only the right value for EFA? Should this include EFA in the function name?

sources/pciclient/src/lib.rs
cbgbt: The Typestate pattern isn't new, but it is becoming more popular. I think given the nice properties it has, and that `bon` makes it easy to use, we should take the dependency!

ytsssun:
Raised a PR that includes just the crate `pciclient` for easier review. https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/149

Will update this PR once that is merged.

sources/api/corndog/src/main.rs
ytsssun: Sure.

sources/ghostdog/src/main.rs
ytsssun: Updated.

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -1314,6 +1320,21 @@ fn oci_spec_resource_limits(
...
+    if !rlimits.contains_key(&OciDefaultsResourceLimitType::MaxLockedMemory) {
+        let max_locked_memory_limits = OciDefaultsResourceLimitV1 {
+            soft_limit: -1,
+            hard_limit: -1,
+        };
+        rlimits.insert(
+            OciDefaultsResourceLimitType::MaxLockedMemory,
+            max_locked_memory_limits,
+        );
+    }
```
ytsssun: Clippy is not happy with this. Will change with  
```suggestion
    rlimits.entry(OciDefaultsResourceLimitType::MaxLockedMemory).or_insert(OciDefaultsResourceLimitV1 {
        soft_limit: -1,
        hard_limit: -1,
    });
```

Will give it a last test before pushing.

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +121,19 @@ where
     }
 }
 
+/// Computation:
+/// - We need to allocate 110MB memory for each libfabric endpoint.
+/// - For optimal setup, Open MPI will open 2 libfabric endpoints each core.
+/// - The total number of hugepages will be set as (110MB * 2) * number_of_cores / hugepage_size
+/// - We will allocate default hugepage_size = 2MB.
+/// - The number of hugepage per core would be 110MB * 2 / 2MB = 110.
+/// We will use [`num_cpus`] to get the number of cores for the compute.
```
arnaldo2792: Nit: for consistency you are missing ` -`

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -1314,6 +1320,20 @@ fn oci_spec_resource_limits(
...
+/// We need to increase the locked memory limits from the default 8096KB to unlimited
+/// to account for hugepages allocation.
+fn adjust_resource_limits_for_efa(
+    rlimits: &mut HashMap<OciDefaultsResourceLimitType, OciDefaultsResourceLimitV1>,
+) {
+    rlimits
+        .entry(OciDefaultsResourceLimitType::MaxLockedMemory)
+        .or_insert(OciDefaultsResourceLimitV1 {
+            soft_limit: -1,
+            hard_limit: -1,
```
arnaldo2792: Should a constant be created as `UNLIMITED = -1`? The `-1` doesn't mean too much if we don't provide context somehow, we are just assuming that `-1` means unlimited.

sources/ghostdog/src/main.rs
```diff
@@ -1,6 +1,7 @@
 /*!
 ghostdog is a tool to manage ephemeral disks.
 It can be called as a udev helper program to identify ephemeral disks.
+It can also be called for EFA device detection which can be used for ExecCondition in systemd units.
```
arnaldo2792, line 4:
```suggestion
Also, it can be called to detect EFA devices
```

You don't have to explicitly call out the `ExecCondition` here.

sources/ghostdog/src/main.rs
```diff
@@ -30,8 +31,14 @@ struct Args {
 enum SubCommand {
     Scan(ScanArgs),
     EbsDeviceName(EbsDeviceNameArgs),
+    EfaPresent(EfaPresentArgs),
 }
 
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "efa-present")]
+/// Detect if EFA is attached.
```
arnaldo2792:
```suggestion
/// Detect if EFA devices are attached.
```

sources/api/corndog/src/main.rs
```diff
@@ -45,20 +50,29 @@ fn run() -> Result<()> {
...
         }
         "lockdown" => {
+            let kernel = get_kernel_settings(args.config_path)?;
             if let Some(lockdown) = kernel.lockdown {
                 debug!("Setting lockdown: {:#?}", lockdown);
                 set_lockdown(&lockdown)?;
             }
         }
+        "generate-hugepages-setting" => {
+            let mut hugepages: String = "0".to_string();
```
bcressey: We wouldn't expect to end up in this situation, but for safety I would read the current `nr_hugepages` value, and then return that (rather than our calculated number) if it's non-zero.

(It could happen with a custom variant if hugepages are set via a `sysctl.d` drop-in, for example.)

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +121,19 @@ where
...
+/// - For optimal setup, Open MPI will open 2 libfabric endpoints each core.
+/// - The total number of hugepages will be set as (110MB * 2) * number_of_cores / hugepage_size
+/// - We will allocate default hugepage_size = 2MB.
+/// - The number of hugepage per core would be 110MB * 2 / 2MB = 110.
+/// We will use [`num_cpus`] to get the number of cores for the compute.
+fn compute_hugepages_for_efa() -> Result<String> {
+    let num_cores = num_cpus::get();
+    let number_of_hugepages = num_cores as u64 * HUGEPAGES_2MB_PER_CORE;
+    Ok(number_of_hugepages.to_string())
+}
```
bcressey: The EFA hugepage allocation script ends up adding an additional page:
```
efa_total_huge_pages_memory=$(($efa_ep_huge_pages_memory * $number_of_cores * 2))
efa_number_of_huge_pages=$(($efa_total_huge_pages_memory / $huge_pages_size + 1))
```

Not sure if this is intentional.

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -2058,6 +2078,21 @@ mod test_oci_spec {
...
 
+    #[test]
+    fn oci_spec_resource_limits_adjust_for_efa_test() {
+        let json = json!({"max-open-files": {"hard-limit": 1, "soft-limit": 2}});
+        let mut rlimits = oci_spec_resource_limits(&json).unwrap();
+        adjust_resource_limits_for_efa(&mut rlimits);
+        let rendered = Containerd::get_resource_limits(
+            &MaxLockedMemory,
+            rlimits.get(&MaxLockedMemory).unwrap(),
+        );
```
bcressey: Can you add a test for the other case, where there's no EFA device?

sources/api/corndog/src/main.rs
ytsssun: Sure, it is probably safer to do that for edge cases. Will update.

sources/api/corndog/src/main.rs
ytsssun: That is fair. Will update that.

sources/api/schnauzer/src/helpers/mod.rs
ytsssun: Sure thing.

sources/api/corndog/src/main.rs
ytsssun: Actually that helper script accounted for other hugepage sizes where `efa_total_huge_pages_memory` may not be a multiplier of `huge_pages_size`. Where in our case it is not quite true. 

The pinned memory per core would always be 220MB, and we will only allocate 2MB hugepages, so we know that we will always allocate 110 hugepages per core. In this case, there is no need to "+1" here.

sources/ghostdog/src/main.rs
ytsssun: The above sentence mentioned `udev` for its use case, that is why I mentioned ExecCondition for this subcommand. In fact, it can only be used as `ExecCondition` since it will only emit exit code.

sources/api/schnauzer/src/helpers/mod.rs
ytsssun: I had to implement a trait that wraps the `pciclient::is_efa_attached()` call so that I can unit test this for both cases. 

Let me know if you like it or not.

sources/api/corndog/src/main.rs
ytsssun: Updated, wanted to hear your thoughts on my implementation. The implementation would try to read the `nr_hugepages` and try to parse it. It would set to "0" if it cannot recognize the value read from the file.

The hugepages is expected to be an unsigned long value, see https://github.com/torvalds/linux/blob/master/include/linux/hugetlb.h#L673.

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +117,51 @@ where
     }
 }
 
+/// Generate the hugepages setting for defaults.
+fn generate_hugepages_setting() -> Result<()> {
+    let existing_hugepages_value =
```
yeazelm: Can't this fail and then the context will stop the rest of the execution but you actually want to continue in this case right?

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +117,51 @@ where
...
+        fs::read_to_string(NR_HUGEPAGES_PATH_SYSCTL).context(error::ReadHugepagesFileSnafu)?;
+    let mut hugepages = "0".to_string();
+    // This would be a rare case to hit, as customer would be modifying the hugepage value at
+    // boot-time not via API settings. (It could happen with a custom variant if hugepages
+    // are set via a sysctl.d drop-in, for example.)
+    //
+    // We expect the existing_hugepages_value to be valid numeric digits. Otherwise, we will
+    // use "0" as default.
+    if &existing_hugepages_value != "0" {
```
yeazelm: This feels like a candidate for a helper function we can unit test.

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +117,51 @@ where
...
+                    err
+                );
+            }
+        }
+    }
+    if pciclient::is_efa_attached().context(error::CheckEfaFailureSnafu)? {
+        hugepages = compute_hugepages_for_efa()?;
+    }
+    let output = serde_json::to_string(&hugepages).context(error::SerializeJsonSnafu)?;
+    println!("{}", output);
```
yeazelm: I'm worried about all the places you could bail early and not return this value when the expectation is that you return one even if most things don't work (like reading the PCI devices can just exit with the default). Am I missing something in the expectation of how this command is used and it's safe for it to just not return _any_ value?

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +117,51 @@ where
...
+}
+
+/// Computation:
+/// - We need to allocate 110MB memory for each libfabric endpoint.
+/// - For optimal setup, Open MPI will open 2 libfabric endpoints each core.
+/// - The total number of hugepages will be set as (110MB * 2) * number_of_cores / hugepage_size
+/// - We will allocate default hugepage_size = 2MB.
+/// - The number of hugepage per core would be 110MB * 2 / 2MB = 110.
+/// - We will use [`num_cpus`] to get the number of cores for the compute.
+fn compute_hugepages_for_efa() -> Result<String> {
```
yeazelm: Nit: I don't think this can fail? Do we need to return a Result?

sources/api/corndog/src/main.rs
ytsssun: Good point, I guess my question is, are expect the boot to fail if there is failure detecting the EFA card? If not, I agree we should just provide a default value and move on.

ytsssun: I am looking at other settings generators, looks like we bail early too - https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/api/shibaken/src/admin_userdata.rs#L17-L36

Should we keep it consistent with the behaviors here?

sources/api/corndog/src/main.rs
ytsssun: Will update.

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -125,6 +125,8 @@ const IPV6_LOCALHOST: IpAddr = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1))
 
 const DEFAULT_ECS_METADATA_SERVICE_RPS: i32 = 40;
 const DEFAULT_ECS_METADATA_SERVICE_BURST: i32 = 60;
+/// We use -1 to indicate unlimited value for resource limits.
+const UNLIMITED: i64 = -1;
```
bcressey:
```suggestion
const RLIMIT_UNLIMITED: i64 = -1;
```

sources/api/corndog/src/main.rs
ytsssun: Added unit tests for it.

sources/api/corndog/src/main.rs
ytsssun: Synced offline, updated the code so that we don't bail early in other cases. We need to keep the serde error since the settings generator is expected to generate JSON string. This is the only case we should fail for EFA settings generator.

sources/api/corndog/src/main.rs
ytsssun: Updated, as we synced offline, we will use default "0" for those cases. When lspci failed or we are not able to read the `nr_hugepages` file, we should not fail.

sources/api/corndog/src/main.rs
```diff
@@ -107,6 +117,63 @@ where
...
+        fs::read_to_string(NR_HUGEPAGES_PATH_SYSCTL).context(error::ReadHugepagesFileSnafu)
+    {
+        hugepages = check_for_existing_hugepages(existing_hugepages_value);
+    }
+    if let Ok(is_efa_attached) = pciclient::is_efa_attached() {
+        if is_efa_attached {
+            // We will use [`num_cpus`] to get the number of cores for the compute.
+            hugepages = compute_hugepages_for_efa(num_cpus::get());
+        }
+    }
```
bcressey: This ends up clobbering the existing hugepages value if EFA is attached.

sources/api/schnauzer/src/helpers/mod.rs
ytsssun: Updated

sources/api/corndog/src/main.rs
ytsssun: Thanks for the nice catch. The refactoring kind of impacted the logic here. Just updated.

bcressey (approved):

yeazelm (approved):

sources/ghostdog/src/main.rs
```diff
@@ -176,6 +194,10 @@ mod error {
         },
         #[snafu(display("Invalid device info for device '{}'", path.display()))]
         InvalidDeviceInfo { path: std::path::PathBuf },
+        #[snafu(display("Failed to check if EFA device is attached: {}", source))]
+        CheckEfaFailure { source: pciclient::PciClientError },
+        #[snafu(display("Did not detect EFA"))]
```
ytsssun, line 199: Adjusted the display here to use past tense. `Does` -> `Did`.

bcressey (approved):

sources/api/corndog/src/main.rs
```diff
@@ -45,20 +51,30 @@ fn run() -> Result<()> {
...
             if let Some(lockdown) = kernel.lockdown {
                 debug!("Setting lockdown: {:#?}", lockdown);
                 set_lockdown(&lockdown)?;
             }
         }
+        "generate-hugepages-setting" => {
+            let hugepages_setting = generate_hugepages_setting()?;
+            // We will only fail if we cannot serialize the output to JSON string.
+            // sundog expects JSON-serialized output so that many types can be represented, allowing the
+            // API model to use more accurate types.
```
cbgbt, line 73:
```suggestion
            let hugepages_setting = generate_hugepages_setting();
            // We will only fail if we cannot serialize the output to JSON string.
            // sundog expects JSON-serialized output so that many types can be represented, allowing the
            // API model to use more accurate types.

```

I think this comment can't be true because `generate_hugepages_setting()` returns a `Result`. Can we either make it not return a `Result`, or else add code at this level which explicitly performs the fallback behavior we're expecting if generating hugepage settings fails?

I'd also add commentary here about what the fallback behavior is. Right now, you have to dive into `generate_hugepages_setting()` to find out.

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -1250,7 +1255,7 @@ pub fn oci_defaults(
             runtime.get_capabilities(capabilities)
         }
         OciSpecSection::ResourceLimits => {
-            let rlimits = oci_spec_resource_limits(oci_defaults_values)?;
+            let rlimits = generate_oci_resource_limits(oci_defaults_values, EfaLspciDetector {})?;
```
cbgbt, line 1258: nit:

```suggestion
            let rlimits = generate_oci_resource_limits(oci_defaults_values, EfaLspciDetector)?;
```

cbgbt (commented):
I think I just missed the boat, but the commit message on your second commit has a subject line that is a bit too long -- it should be 72 characters or fewer.

You can tell if you've broken the limit on the GitHub UI because it wraps the subject line with ellipses.

PR #142, opened 2024-09-10
Add pciutils to packages in release

yeazelm:
**Description of changes:**
This adds the `pciutils` package to the release package dependencies so that lspci is available for debugging devices on Bottlerocket images.

**Testing done:**
Booted an image and checked `lspci` works on x86_64 and aarch64
### aarch t4g.large
```
[root@admin]# sheltie
bash-5.1# lspci
00:00.0 Host bridge: Amazon.com, Inc. Device 0200
00:01.0 Serial controller: Amazon.com, Inc. Device 8250
00:04.0 Non-Volatile memory controller: Amazon.com, Inc. NVMe EBS Controller
00:05.0 Ethernet controller: Amazon.com, Inc. Elastic Network Adapter (ENA)
00:1f.0 Non-Volatile memory controller: Amazon.com, Inc. NVMe EBS Controller
```

### NVIDIA x86_64
```
[root@admin]# sheltie
bash-5.1# lspci
00:00.0 Host bridge: Intel Corporation 440FX - 82441FX PMC [Natoma]
00:01.0 ISA bridge: Intel Corporation 82371SB PIIX3 ISA [Natoma/Triton II]
00:01.3 Non-VGA unclassified device: Intel Corporation 82371AB/EB/MB PIIX4 ACPI (rev 08)
00:03.0 VGA compatible controller: Amazon.com, Inc. Device 1111
00:04.0 Non-Volatile memory controller: Amazon.com, Inc. NVMe EBS Controller
00:05.0 Ethernet controller: Amazon.com, Inc. Elastic Network Adapter (ENA)
00:1d.0 Non-Volatile memory controller: Amazon.com, Inc. NVMe EBS Controller
00:1e.0 3D controller: NVIDIA Corporation TU104GL [Tesla T4] (rev a1)
00:1f.0 Non-Volatile memory controller: Amazon.com, Inc. NVMe SSD Controller
```

**Terms of contribution:**

### DIFF
--- a/packages/pciutils/Cargo.toml
+++ b/packages/pciutils/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "pciutils"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://www.kernel.org/pub/software/utils/pciutils/"
+
+[[package.metadata.build-package.external-files]]
+url = "https://www.kernel.org/pub/software/utils/pciutils/pciutils-3.13.0.tar.gz"
+sha512 = "39cf4141c87c9a39fb42ec7a2412525f4283c62a1d1aa3533eb92bae4c59d3beb9a2ab1a9fcfe89b1f6cb8f0a011ef2f32fbed3111d357ce43d9569b3d0253d9"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
--- a/packages/pciutils/pciutils.spec
+++ b/packages/pciutils/pciutils.spec
@@ -1,0 +1,48 @@
+Name: %{_cross_os}pciutils
+Version: 3.13.0
+Release: 1%{?dist}
+Summary: PCI bus related utilities
+License: GPL-2.0-only
+URL: https://www.kernel.org/pub/software/utils/pciutils/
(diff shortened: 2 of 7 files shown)


packages/pciutils/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://mj.ucw.cz/download/linux/pci/"
```
ytsssun: Nit: pciutils is also available at  https://www.kernel.org/pub/software/utils/pciutils/. Should we use that for consistency? We have a good number of packages that consume from kernel.org.

ytsssun (approved):

packages/pciutils/Cargo.toml
yeazelm: Good catch, I had switched in the spec file and not here.

ytsssun (approved):

packages/pciutils/pciutils.spec
```diff
@@ -0,0 +1,51 @@
+%global debug_package %{nil}
+%global __strip %{_bindir}/true
```
bcressey: We shouldn't need these.

```suggestion
```

yeazelm: Nice call! I added these when trying things but they are not needed!

yeazelm:
^ removed extra lines per comments.

bcressey (approved):

PR #149, opened 2024-09-13
source: add pciclient crate for high level access to lspci

ytsssun:
**Description of changes:**
Added crate pciclient to provide a high level helper to query the PCI information and get the device information.

**Testing done:**
Unit tests.

More tests are done in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/141 which uses the crate to call lspci and detect EFA.

**Terms of contribution:**

### DIFF
--- a/sources/pciclient/src/lib.rs
+++ b/sources/pciclient/src/lib.rs
@@ -1,0 +1,146 @@
+/*!
+`pciclient` provides high-level methods to invoke `lspci` and query for attached devices information.
+
+pciclient provides util functions that can:
+- List the devices attached with some filtering option.
+- Detect specifically if EFA device is attached.
+*/
+mod private;
+
+use private::{call_list_devices, check_efa_attachment, PciClient};
+
+use bon::Builder;
+use derive_getters::Getters;
+use std::ffi::OsString;
+
+/// This [`ListDevicesParam`] is based on the "-d" flag for `lspci`, which
+/// allows selection/filtering of the output. Spec for "-d":
+/// ```sh
+/// -d [<vendor>]:[<device>][:<class>[:<prog-if>]]
+///
+/// Show only devices with specified vendor, device, class ID,
+/// and programming interface.  The ID's are given in
+/// hexadecimal and may be omitted or given as "*", both
+/// meaning "any value". The class ID can contain "x"
+/// characters which stand for "any digit".
+/// ```
+#[derive(Debug, Default, PartialEq, Builder)]
+pub struct ListDevicesParam {
+    #[builder(into)]
+    vendor: Option<String>,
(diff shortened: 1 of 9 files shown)


sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Debug, Default, PartialEq, Eq, Getters)]
+pub struct ListDevicesOutput {
+    pub pci_slot: String,
+    pub class: String,
+    pub vendor: String,
+    pub device: String,
+    pub subsystem_vendor: Option<String>,
+    pub subsystem_device: Option<String>,
+    pub revision: Option<String>,
+    pub program_interface: Option<String>,
```
yeazelm: Can we make these non `pub` now with the Getter?

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+/// the output format that is decorated by "-n -m".
+///
+/// # Example
+/// ```rust,ignore
+/// use pciclient::{ListDevicesParam, list_devices};
+///
+/// let list_devices_param = ListDevicesParam::builder().vendor("1d0f").build();
+/// match list_devices(list_devices_param) {
+///     Ok(list_devices_output) => {
+///         println!("{:?}", list_devices_output);
```
yeazelm: nit: I'd like to actually show off how to use the getters here rather than just rely upon debug output.

sources/pciclient/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+            required_field
+        ))]
+        MissingRequiredField { required_field: String },
+    }
+}
+
+use bon::Builder;
+use derive_getters::Getters;
+pub use error::PciClientError;
+use private::{call_list_devices, check_efa_attachment, PciClient};
```
yeazelm: nit: We usually include these at the top.

ytsssun: I have moved the `bon` and `derive_getters` and private crate functions deps to top. Looks like for Error and Result related statements, we keep them after the definition of the Error mod. See https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/api/storewolf/src/main.rs#L138-L141

sources/pciclient/src/lib.rs
ytsssun: Added.

sources/pciclient/src/lib.rs
ytsssun: Good catch! Updated, thanks!

cbgbt (approved):
Nice!

yeazelm (approved):
LGTM!

PR #155, opened 2024-09-19
Change Go version for k8s 1.23-1.29

larvacea:
Go 1.23 leaks file descriptors if os/exec fails.

**Issue number:**

Closes issue https://github.com/bottlerocket-os/bottlerocket/issues/4201 once consumed.

**Description of changes:**

Add GO_MAJOR=1.22 for kubernetes packages 1.23 through 1.29. kubernetes 1.30 already had GO_MAJOR=1.22, and kubernetes 1.31 will require 1.23.

**Testing done:**

Built nodegroups, added to cluster with the Calico configuration that rapidly reproduces the failure.

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.26/kubernetes-1.26.spec
+++ b/packages/kubernetes-1.26/kubernetes-1.26.spec
@@ -156,6 +156,7 @@ cp third_party/forked/golang/PATENTS PATENTS.golang
 
 %build
 export FORCE_HOST_GO=1
+export GO_MAJOR="1.22"
 # Build codegen programs with the host toolchain.
 make hack/update-codegen.sh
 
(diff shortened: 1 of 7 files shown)


packages/kubernetes-1.26/kubernetes-1.26.spec
```diff
@@ -164,6 +164,7 @@ make hack/update-codegen.sh
 unset CC
 export KUBE_BUILD_PLATFORMS="linux/%{_cross_go_arch}"
 export %{kube_cc}
+export GO_MAJOR="1.22"
```
bcressey: This should be set up near `FORCE_HOST_GO` so that codegen happens with the same version of Go that we use to build kubelet.

bcressey (approved):

arnaldo2792 (approved):

PR #158, opened 2024-09-20
add default security settings

tzneal:
**Issue number:**

Closes #

**Description of changes:**
Adds a default security setting to block writable/executable memory for all services.

**Testing done:**

Before change:
```sh
# systemctl show containerd | grep -i MemoryDenyWriteExecute
MemoryDenyWriteExecute=no
```

After change:
```sh
bash-5.1# systemctl show containerd | grep -i MemoryDenyWriteExecute
MemoryDenyWriteExecute=yes
```

**Terms of contribution:**

### DIFF
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -86,6 +86,7 @@ Source1100: systemd-tmpfiles-setup-service-debug.conf
 Source1101: systemd-resolved-service-env.conf
 Source1102: systemd-networkd-service-env.conf
 Source1103: systemd-logind-inhibit-maxdelay.conf
+Source1104: systemd-service-security.conf
 
 # network link rules
 Source1200: 80-release.link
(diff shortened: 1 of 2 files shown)


bcressey (approved):

cbgbt (commented):
I'm not very familiar with this systemd setting, but one thing I'm not so sure of is how it will interact with the migrations performed by the update system.

Migrations are written to sealed anonymous files, which I could see being included here.

I'm happy to test this and come back, but I'd be curious to test building an image with this change, then attempting to in- place update it.

arnaldo2792 (approved):

cbgbt (approved):
Pardon my latency, my original test plan involved multiple migration iterations but I realized that it wasn't necessary if I just started my chain from a build from this commit.

The testing looks good. Thanks!

PR #159, opened 2024-09-23
add package-level support for erofs as root filesystem

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/twoliter/pull/379

**Description of changes:**
This implements the core kit side of erofs support.

Require a 6.1 kernel for using erofs, and add the necessary kernel config and SELinux policy statements to make it work.

The ugly parts come from my goal of avoiding multiple forms of compression; since erofs includes its own compression, we can avoid compressing license and attribution data, kernel modules, and kernel development files.

Kernel module decompression is handled by `twoliter`: if we're building an erofs root filesystem, it decompresses any kernel modules. Uncompressed license and attribution data is *mostly* handled by `twoliter`, which skips generating the squashfs when using the erofs feature. However, we need to tell the OS not to try to mount the missing squashfs.

Handling the kernel development files is more complicated, because there's a lot of logic in the `release` package to handle setting up `/usr/src/kernels` as a writable mount. I couldn't see a great way to prevent the `kernel-6.1` package from depending on some of the specific units in `release`, but I tried to minimize it. We now create two `kernel-devel` packages, one compressed with squashfs and the other left unpacked, and we use a drop-in to replace [the squashfs mount](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/47f92e4f61696716ed0703a13d9af4ba6eb9e6c5/packages/release/var-lib-kernel-devel-lower.mount.in) in `release` with a regular bind mount.

**Testing done:**
Built `aws-k8s-1.30` with and without the image feature enabled.

Without:
```
bash-5.1# findmnt / -o FSTYPE -n
ext4

bash-5.1# df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       904M  700M  143M  84% /

# kernel-devel lower is a squashfs mount
bash-5.1# findmnt /var/lib/kernel-devel/.overlay/lower/ -o FSTYPE -n
squashfs

# /usr/share/licenses is a squashfs mount
bash-5.1# findmnt /usr/share/licenses -o FSTYPE -n
squashfs

# all licenses are present
bash-5.1# find /usr/share/licenses -type f | wc -l
2907

# mount has the correct label
bash-5.1# ls -latrZ /usr/bin/mount
-rwsr-xr-x. 1 root root system_u:object_r:mount_exec_t:s0 65608 Sep 22 00:56 /usr/bin/mount
```

With:
```
bash-5.1# findmnt / -o FSTYPE -n
erofs

bash-5.1# df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       337M  337M     0 100% /

# kernel-devel lower is a bind mount
bash-5.1# findmnt /var/lib/kernel-devel/.overlay/lower/ -o FSTYPE -n
erofs

# /usr/share/licenses is not a mount
bash-5.1# findmnt /usr/share/licenses -o FSTYPE -n
<no output>

# all licenses are present
bash-5.1# find /usr/share/licenses -type f |wc -l
2907

# mount has the correct label
bash-5.1# ls -latrZ /usr/bin/mount
-rwsr-xr-x. 1 root root system_u:object_r:mount_exec_t:s0 65608 Sep 23 23:05 /usr/bin/mount
```

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/kernel-5.10.spec
+++ b/packages/kernel-5.10/kernel-5.10.spec
@@ -45,6 +45,9 @@ Requires: %{name}-devel = %{version}-%{release}
 # The 5.10 kernel is not FIPS certified.
 Conflicts: %{_cross_os}image-feature(fips)
 
+# Using EROFS for the root partition requires a 6.1+ kernel.
+Conflicts: %{_cross_os}image-feature(erofs-root-partition)
+
 %global kernel_sourcedir %{_cross_usrsrc}/kernels
 %global kernel_libdir %{_cross_libdir}/modules/%{version}
 
--- a/packages/kernel-5.15/kernel-5.15.spec
+++ b/packages/kernel-5.15/kernel-5.15.spec
@@ -40,6 +40,9 @@ Requires: %{name}-devel = %{version}-%{release}
 # The 5.15 kernel is not FIPS certified.
 Conflicts: %{_cross_os}image-feature(fips)
 
+# Using EROFS for the root partition requires a 6.1+ kernel.
+Conflicts: %{_cross_os}image-feature(erofs-root-partition)
+
 %global kernel_sourcedir %{_cross_usrsrc}/kernels
 %global kernel_libdir %{_cross_libdir}/modules/%{version}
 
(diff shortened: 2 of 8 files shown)


arnaldo2792 (approved):
Nice!

yeazelm (approved):

hsiangkao:
@bcressey Hi Ben, I wonder if it's possible to enable EROFS kernel configs (=m) for Amazon Linux 2023 too?

PR #173, opened 2024-10-01
Compile FIPS rust binaries

arnaldo2792:
**Issue number:**

**Description of changes:**

The first commit in the series adds `rustls` and `aws-smithy-experimental` to the workspace dependencies, so that crates that depend on these libraries can enforce the FIPS flags each expose. 

The subsequent commits add a `fips` flag to the binaries flagged by the `check-fips` script in the Bottlerocket SDK. In https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/191, the `check-fips` was disabled in Rust programs that required the FIPS treatment but in this series the checks are re-enabled.

The spec files for `os` was updated to provide the `fips-bin` package for any binary that needs it. This resembles  what's already done for `Go` packages so that the correct binaries are installed in the image when the `fips` image flag is used while building a variant.

`pluto` was modified to split the HTTP Proxy support, given that `aws-smithy-experimental` doesn't expose the same APIs as the "stable" `aws-smithy`. 

**Testing done:**
- I enabled the `fips` image flag in `aws-k8s-1.30`, and confirmed the `fips-bin` packages were installed:

```bash
[root@admin]# cat /.bottlerocket/rootfs/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/bottlerocket/application-inventory.json | jq '.Content[].Name' -r | grep fips-bin
apiclient-fips-bin
aws-iam-authenticator-fips-bin
aws-signing-helper-fips-bin
cfsignal-fips-bin
cni-plugins-fips-bin
containerd-fips-bin
early-boot-config-aws-fips-bin
ecr-credential-provider-1.30-fips-bin
host-ctr-fips-bin
hostname-imds-fips-bin
kubelet-1.30-fips-bin
logdog-fips-bin
metricdog-fips-bin
migration-fips-bin
pluto-fips-bin
runc-fips-bin
shibaken-fips-bin
updog-fips-bin
```
- The instance joined the cluster:
 
```bash
└─> ❯ k get nodes
NAME                                          STATUS   ROLES    AGE     VERSION
ip-192-168-81-52.us-west-2.compute.internal   Ready    <none>   3h18m   v1.30.1-eks-e564799
```

- I checked the FIPS binaries in `/usr/bin/ ` and `/usr/libexec` are actually the FIPS binaries and not their non-FIPS counterpart. 

**Terms of contribution:**

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -17,3 +17,9 @@ operating system images.
 Contains modified hyper-proxy files [mod.rs, stream.rs, tunnel.rs] from
 https://github.com/tafia/hyper-proxy 2021-09-20.
 Copyright (c) 2017 Johann Tuffe. Licensed under the MIT License.
+
+=^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+Contains aws-smithy-experimental from
+https://github.com/smithy-lang/smithy-rs/tree/release-2024-10-09.
+Licensed under the Apache-2.0 License.
--- a/packages/early-boot-config/early-boot-config.spec
+++ b/packages/early-boot-config/early-boot-config.spec
@@ -1,7 +1,5 @@
 %global _cross_first_party 1
 %undefine _debugsource_packages
-%undefine cross_check_fips
-%undefine cross_check_fips
 
 Name: %{_cross_os}early-boot-config
 Version: 0.1
--- a/packages/netdog/netdog.spec
+++ b/packages/netdog/netdog.spec
@@ -1,7 +1,5 @@
 %global _cross_first_party 1
 %undefine _debugsource_packages
-%undefine cross_check_fips
-%undefine cross_check_fips
 
 Name: %{_cross_os}netdog
 Version: 0.1.1
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -1,7 +1,5 @@
 %global _cross_first_party 1
 %undefine _debugsource_packages
-%undefine cross_check_fips
-%undefine cross_check_fips
 
 Name: %{_cross_os}os
 Version: 0.0
@@ -121,7 +119,24 @@ Requires: %{_cross_os}settings-plugins
 
 %package -n %{_cross_os}apiclient
 Summary: Bottlerocket API client
+Requires: %{_cross_os}apiclient(binaries)
 %description -n %{_cross_os}apiclient
+%{summary}.
+
+%package -n %{_cross_os}apiclient-bin
+Summary: Bottlerocket API client binaries
+Provides: %{_cross_os}apiclient(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}apiclient)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}apiclient-fips-bin)
+%description -n %{_cross_os}apiclient-bin
+%{summary}.
+
+%package -n %{_cross_os}apiclient-fips-bin
+Summary: Bottlerocket API client binaries, FIPS edition
+Provides: %{_cross_os}apiclient(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}apiclient)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}apiclient-bin)
+%description -n %{_cross_os}apiclient-fips-bin
 %{summary}.
 
 %package -n %{_cross_os}sundog
@@ -172,7 +187,26 @@ Requires: %{_cross_os}settings-defaults
 %package -n %{_cross_os}migration
 Summary: Tools to migrate version formats
 Conflicts: %{_cross_os}image-feature(no-in-place-updates)
+Requires: %{_cross_os}migration(binaries)
 %description -n %{_cross_os}migration
+
+%package -n %{_cross_os}migration-bin
+Summary: Binaries to migrate version formats
+Provides: %{_cross_os}migration(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}migration)
+Conflicts: %{_cross_os}image-feature(no-in-place-updates)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}migration-fips-bin)
+%description -n %{_cross_os}migration-bin
+%{summary}.
+
+%package -n %{_cross_os}migration-fips-bin
+Summary: Binaries to migrate version formats, FIPS edition
+Provides: %{_cross_os}migration(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}migration)
+Conflicts: %{_cross_os}image-feature(no-in-place-updates)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}migration-bin)
+%description -n %{_cross_os}migration-fips-bin
+%{summary}.
 
 %package -n %{_cross_os}settings-committer
 Summary: Commits settings from user data, defaults, and generators at boot
@@ -192,19 +226,72 @@ Summary: Bottlerocket GPT priority querier/switcher
 
 %package -n %{_cross_os}updog
 Summary: Bottlerocket updater CLI
+Requires: %{_cross_os}updog(binaries)
 Conflicts: %{_cross_os}image-feature(no-in-place-updates)
 %description -n %{_cross_os}updog
 not much what's up with you
 
+%package -n %{_cross_os}updog-bin
+Summary: Bottlerocket updater CLI binaries
+Provides: %{_cross_os}updog(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}updog)
+Conflicts: %{_cross_os}image-feature(no-in-place-updates)
+Conflicts: %{_cross_os}image-feature(fips) or %{_cross_os}updog-fips-bin
+%description -n %{_cross_os}updog-bin
+%{summary}.
+
+%package -n %{_cross_os}updog-fips-bin
+Summary: Bottlerocket updater CLI binaries, FIPS edition
+Provides: %{_cross_os}updog(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}updog)
+Conflicts: %{_cross_os}image-feature(no-in-place-updates)
+Conflicts: %{_cross_os}image-feature(no-fips) or %{_cross_os}updog-bin
+%description -n %{_cross_os}updog-fips-bin
+%{summary}.
+
 %package -n %{_cross_os}metricdog
 Summary: Bottlerocket health metrics sender
+Requires: %{_cross_os}metricdog(binaries)
 %description -n %{_cross_os}metricdog
+%{summary}.
+
+%package -n %{_cross_os}metricdog-bin
+Summary: Bottlerocket health metrics sender binaries
+Provides: %{_cross_os}metricdog(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}metricdog)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}metricdog-fips-bin)
+%description -n %{_cross_os}metricdog-bin
+%{summary}.
+
+%package -n %{_cross_os}metricdog-fips-bin
+Summary: Bottlerocket health metrics sender binaries, FIPS edition
+Provides: %{_cross_os}metricdog(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}metricdog)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}metricdog-bin)
+%description -n %{_cross_os}metricdog-fips-bin
 %{summary}.
 
 %package -n %{_cross_os}logdog
 Summary: Bottlerocket log extractor
+Requires: %{_cross_os}logdog(binaries)
 %description -n %{_cross_os}logdog
-use logdog to extract logs from the Bottlerocket host
+%{summary}.
+
+%package -n %{_cross_os}logdog-bin
+Summary: Bottlerocket log extractor binaries
+Provides: %{_cross_os}logdog(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}logdog)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}logdog-fips-bin)
+%description -n %{_cross_os}logdog-bin
+%{summary}.
+
+%package -n %{_cross_os}logdog-fips-bin
+Summary: Bottlerocket log extractor binaries, FIPS edition
+Provides: %{_cross_os}logdog(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}logdog)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}logdog-bin)
+%description -n %{_cross_os}logdog-fips-bin
+%{summary}.
 
 %package -n %{_cross_os}prairiedog
 Summary: Tools for kdump support
@@ -220,12 +307,46 @@ Summary: Bottlerocket certificates handler
 
 %package -n %{_cross_os}pluto
 Summary: Dynamic setting generator for kubernetes
+Requires: %{_cross_os}pluto(binaries)
 %description -n %{_cross_os}pluto
 %{summary}.
 
+%package -n %{_cross_os}pluto-bin
+Summary: Dynamic setting generator for kubernetes binaries
+Provides: %{_cross_os}pluto(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}pluto)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}pluto-fips-bin)
+%description -n %{_cross_os}pluto-bin
+%{summary}.
+
+%package -n %{_cross_os}pluto-fips-bin
+Summary: Dynamic setting generator for kubernetes binaries, FIPS edition
+Provides: %{_cross_os}pluto(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}pluto)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}pluto-bin)
+%description -n %{_cross_os}pluto-fips-bin
+%{summary}.
+
 %package -n %{_cross_os}shibaken
-Summary: Run tasks reliant on IMDS
+Summary: IMDS client and settings generator
+Requires: %{_cross_os}shibaken(binaries)
 %description -n %{_cross_os}shibaken
+%{summary}.
+
+%package -n %{_cross_os}shibaken-bin
+Summary: IMDS client and settings generator binaries
+Provides: %{_cross_os}shibaken(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}shibaken)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}shibaken-fips-bin)
+%description -n %{_cross_os}shibaken-bin
+%{summary}.
+
+%package -n %{_cross_os}shibaken-fips-bin
+Summary: IMDS client and settings generator binaries, FIPS edition
+Provides: %{_cross_os}shibaken(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}shibaken)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}shibaken-bin)
+%description -n %{_cross_os}shibaken-fips-bin
 %{summary}.
 
 %package -n %{_cross_os}warm-pool-wait
@@ -236,7 +357,24 @@ Requires: %{_cross_os}shibaken
 
 %package -n %{_cross_os}cfsignal
 Summary: Bottlerocket CloudFormation Stack signaler
+Requires: %{_cross_os}cfsignal(binaries)
 %description -n %{_cross_os}cfsignal
+%{summary}.
+
+%package -n %{_cross_os}cfsignal-bin
+Summary: Bottlerocket CloudFormation Stack signaler binaries
+Provides: %{_cross_os}cfsignal(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}cfsignal)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}cfsignal-fips-bin)
+%description -n %{_cross_os}cfsignal-bin
+%{summary}.
+
+%package -n %{_cross_os}cfsignal-fips-bin
+Summary: Bottlerocket CloudFormation Stack signaler binaries, FIPS edition
+Provides: %{_cross_os}cfsignal(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}cfsignal)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}cfsignal-bin)
+%description -n %{_cross_os}cfsignal-fips-bin
 %{summary}.
 
 %package -n %{_cross_os}shimpei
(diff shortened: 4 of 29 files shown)


arnaldo2792:
Forced push includes:

- Add feature flags to the crates that require `fips` treatment
- Compile additional `fips` binaries for the crates that require them

sources/api/migration/migrator/Cargo.toml
```diff
@@ -9,7 +9,11 @@ build = "build.rs"
 # Don't rebuild crate just because of changes to README.
 exclude = ["README.md"]
 
+[features]
+fips = ["tough/fips"]
```
bcressey, line 13: Why does migrator depend on `tough`? Can we break that dependency somehow?

sources/api/apiclient/Cargo.toml
```diff
@@ -9,7 +9,11 @@ build = "build.rs"
 # Don't rebuild crate just because of changes to README.
 exclude = ["README.md"]
 
+[features]
+fips = ["aws-lc-rs/fips", "rustls/fips"]
+
 [dependencies]
+aws-lc-rs = { workspace = true, features = ["bindgen"] }
```
bcressey: Why we need `aws-lc-rs` as a direct dependency? Is that so we're allowed to refer to its features?

sources/imdsclient/Cargo.toml
```diff
@@ -9,10 +9,15 @@ build = "build.rs"
...
 
+[features]
+fips = ["rustls/fips", "aws-lc-rs/fips"]
+
 [dependencies]
+aws-lc-rs = { workspace = true, features = ["bindgen"] }
 http.workspace = true
 log.workspace = true
 reqwest.workspace = true
+rustls.workspace = true
```
bcressey: Can we disable TLS for this version of `reqwest`?

Feature unification may make this not so helpful, though it could be useful for `early-boot-config`.

We could also look at using `ureq` or a different library for `imdsclient`, though that's a larger scope than this change since it'd affect the retry handling.

sources/logdog/Cargo.toml
```diff
@@ -8,10 +8,15 @@ publish = false
...
 
+[features]
+fips = ["rustls/fips", "aws-lc-rs/fips"]
+
 [dependencies]
+aws-lc-rs = { workspace = true, features = ["bindgen"] }
 flate2 = { workspace = true, features = ["default"] }
 glob.workspace = true
 reqwest = { workspace = true, features = ["blocking", "rustls-tls-native-roots"] }
+rustls.workspace = true
```
bcressey, line 19: `logdog` is another one where we don't really need TLS, right?

sources/api/pluto/src/eks.rs
```diff
@@ -46,8 +46,14 @@ where
...
-        let http_connector = proxy::setup_http_client(https_proxy, no_proxy)?;
-        let http_client = HyperClientBuilder::new().build(http_connector);
+        #[cfg(feature = "fips")]
+        let crypto_mode = CryptoMode::AwsLcFips;
+        #[cfg(not(feature = "fips"))]
+        let crypto_mode = CryptoMode::AwsLc;
+
+        let http_client = HyperClientBuilder::new()
+            .crypto_mode(crypto_mode)
+            .build_https();
```
bcressey: Don't we need this for the EC2 client also?

sources/api/pluto/src/eks.rs
```diff
@@ -46,8 +46,14 @@ where
     let config = sdk_config(region).await;
 
     let client = if let Some(https_proxy) = https_proxy {
-        let http_connector = proxy::setup_http_client(https_proxy, no_proxy)?;
```
bcressey, line 49: Would be good to flag this with a prominent `FIXME` since we can't merge without proxy support.

packages/os/os.spec
```diff
@@ -169,8 +186,25 @@ Requires: %{_cross_os}settings-defaults
...
 Summary: Tools to migrate version formats
 Conflicts: %{_cross_os}image-feature(no-in-place-updates)
+Requires: %{_cross_os}migration(binaries)
 %description -n %{_cross_os}migration
 
+%package -n %{_cross_os}migration-bin
+Summary: Binaries to migrate version formats
+Provides: %{_cross_os}migration(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}migration)
+Conflicts: (%{_cross_os}image-feature(no-inplace-updates) or %{_cross_os}image-feature(fips) or %{_cross_os}migration-fips-bin)
```
```suggestion
Conflicts: (%{_cross_os}image-feature(no-in-place-updates) or %{_cross_os}image-feature(fips) or %{_cross_os}migration-fips-bin)
```

This could also just be a separate `Conflicts`.

packages/os/os.spec
```diff
@@ -169,8 +186,25 @@ Requires: %{_cross_os}settings-defaults
...
+%description -n %{_cross_os}migration-bin
+%{summary}.
+
+%package -n %{_cross_os}migration-fips-bin
+Summary: Binaries to migrate version formats, FIPS edition
+Provides: %{_cross_os}migration(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}migration)
+Conflicts: (%{_cross_os}image-feature(no-inplace-updates) or %{_cross_os}image-feature(no-fips) or %{_cross_os}migration-bin)
```
```suggestion
Conflicts: (%{_cross_os}image-feature(no-in-place-updates) or %{_cross_os}image-feature(no-fips) or %{_cross_os}migration-bin)
```

bcressey (commented):
Nice. The `os.spec` changes are large but look good.

Curious to see what this does to build times 😟 but overall it's clean and lines up with what we've done for Go.

sources/api/apiclient/Cargo.toml
arnaldo2792: Correct, otherwise `cargo` isn't happy if we try to enable a feature of a crate that isn't actually a dependency.

sources/api/migration/migrator/Cargo.toml
arnaldo2792: I'll try and come back with an answer.

sources/api/pluto/src/eks.rs
arnaldo2792: Of course! I'll add it as a comment (I used `WIP` the commit message to signal that this isn't ready)

sources/api/pluto/src/eks.rs
arnaldo2792: I missed that, I'll update in a subsequent revision.

sources/logdog/Cargo.toml
arnaldo2792: Actually, `logdog` "supports" https:

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/62450eb4c60a52ecbcb60b4abfdfea63e63e5ebf/sources/logdog/src/log_request.rs#L153

sources/api/migration/migrator/Cargo.toml
arnaldo2792: I read the code, and `migrator` depends on `tough` to find the local TUF repo, its manifest, and the migrations:

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/3f328ac47a8178fedac8989fefb6a4fdda3f0791/sources/api/migration/migrator/src/main.rs#L163

We could break the dependency if instead we have some structs to handle repositories, similar to `RepositoryLoader`. I tried removing the `http` feature in `tough` (since that is actually not needed) to see if the dependencies on `ring`/`aws-lc-rs` went away, but they didn't, so the `check-fips` script will still flag `migrator`.

arnaldo2792:
(Forced push includes a rebase)

arnaldo2792:
(Forced push includes a rebase onto develop and #197 )

arnaldo2792:
Forced push includes:
- For `pluto`, split the Proxy support since `aws-smithy-experimental` doesn't expose a property to override the `HttpConnector` in the client as we do today
- For `cfsignal`, migrate to `aws-smithy-experimental` but unlike `pluto` we don't support Proxies in it. There is a tracking issue for that (https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/187)
- Provide `*-fips-bin` packages for `early-boot-config` and `netdog`
- Addressed comments in previous revision

I'll keep it as a draft until I remove my `hack` commit to increase the Twoliter version.

arnaldo2792:
(forced update removes hack commit to bump twoliter's version)

packages/netdog/netdog.spec
```diff
@@ -53,21 +51,46 @@ Summary: Reverse DNS Hostname detector
...
+Provides: %{_cross_os}hostname-imds(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}hostname-imds)
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}hostname-imds-fips-bin)
+%description -n %{_cross_os}hostname-imds-bin
+%{summary}
+
+%package -n %{_cross_os}hostname-imds-fips-bin
+Summary: IMDS Hostname detector binaries, FIPS edition
+Provides: %{_cross_os}hostname-imds(binaries)
+Requires: %{_cross_os}hostname-imds(binaries)
```
bcressey: Copy/paste error; this shouldn't require the same capability that it provides:
```suggestion
Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}hostname-imds)
```

packages/early-boot-config/early-boot-config.spec
```diff
@@ -72,8 +98,27 @@ Requires: %{name}-local
 %endif
     %{nil}
 
-%global early_boot_config_bindir %{_cross_libexecdir}/early-boot-config/bin
-%global early_boot_config_provider_dir %{_cross_libexecdir}/early-boot-config/data-providers.d
+# Store the output so we can print it after waiting for the backgrounded job.
```
bcressey: Why do this for `early-boot-config` but not for `netdog`? Does it make a big difference in build times?

sources/api/pluto/src/ec2.rs
```diff
@@ -1,5 +1,8 @@
 use crate::aws::sdk_config;
 use crate::proxy;
+#[cfg(feature = "fips")]
+use aws_smithy_experimental::hyper_1_0::{CryptoMode, HyperClientBuilder};
```
bcressey:
```suggestion
#[cfg(feature = "fips")]
use aws_smithy_experimental::hyper_1_0::{CryptoMode, HyperClientBuilder as Hyper10ClientBuilder};
```

sources/api/pluto/src/eks.rs
```diff
@@ -1,6 +1,9 @@
 use crate::aws::sdk_config;
 use crate::proxy;
 use aws_sdk_eks::types::KubernetesNetworkConfigResponse;
+#[cfg(feature = "fips")]
+use aws_smithy_experimental::hyper_1_0::{CryptoMode, HyperClientBuilder};
```
bcressey:
```suggestion
#[cfg(feature = "fips")]
use aws_smithy_experimental::hyper_1_0::{CryptoMode, HyperClientBuilder as Hyper10ClientBuilder};
```

packages/os/os.spec
```diff
@@ -192,19 +226,68 @@ Summary: Bottlerocket GPT priority querier/switcher
...
+Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}logdog-fips-bin)
+%description -n %{_cross_os}logdog-bin
+%{summary}.
+
+%package -n %{_cross_os}logdog-fips-bin
+Summary: Bottlerocket log extractor binaries, FIPS edition
+Provides: %{_cross_os}logdog(binaries)
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}logdog-bin)
+%description -n %{_cross_os}logdog-fips-bin
+%{summary}.
```
bcressey: For overall robustness, these should require their parent `logdog` package. Otherwise a variant could request `bottlerocket-logdog-fips-bin` and not also end up with `bottlerocket-logdog`.

```suggestion
%package -n %{_cross_os}logdog-bin
Summary: Bottlerocket log extractor binaries
Provides: %{_cross_os}logdog(binaries)
Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}logdog)
Conflicts: (%{_cross_os}image-feature(fips) or %{_cross_os}logdog-fips-bin)
%description -n %{_cross_os}logdog-bin
%{summary}.

%package -n %{_cross_os}logdog-fips-bin
Summary: Bottlerocket log extractor binaries, FIPS edition
Provides: %{_cross_os}logdog(binaries)
Requires: (%{_cross_os}image-feature(fips) and %{_cross_os}logdog)
Conflicts: (%{_cross_os}image-feature(no-fips) or %{_cross_os}logdog-bin)
%description -n %{_cross_os}logdog-fips-bin
%{summary}.
```

arnaldo2792:
Forced push includes:

- Drop `early-boot-config` and `netdog` FIPS binaries
- Add additional `tls` flags for `imdsclient` and `apiclient` to turn on and off TLS support whenever needed
- Rename `HyperClientBuilder` to `Hyper10ClientBuilder`
- Add missing `Requires` to `logdog` subpackages

packages/os/os.spec
```diff
@@ -172,8 +187,27 @@ Requires: %{_cross_os}settings-defaults
...
 Summary: Tools to migrate version formats
 Conflicts: %{_cross_os}image-feature(no-in-place-updates)
+Requires: %{_cross_os}migration(binaries)
 %description -n %{_cross_os}migration
 
+%package -n %{_cross_os}migration-bin
+Summary: Binaries to migrate version formats
+Provides: %{_cross_os}migration(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{_cross_os}migration)
+Conflicts: %{_cross_os}image-feature(no-in-place-updates)
```
ginglis13, line 197: nit: is this redundant since the super-package `migration` already defines this conflict? I also see some inconsistencies in the conflict statements, like later in `updog`, these two conflicts are combined like:

```
Conflicts: (%{_cross_os}image-feature(no-in-place-updates) or %{_cross_os}image-feature(no-fips) or %{_cross_os}updog-bin)
```

arnaldo2792: I should have separated the two `Conflicts` in `updog`, what I did here is what was suggested in a previous revision: separate the `Conflicts` statements.

The "redundancy" is needed given that at some point there could be a `%{_cross_os}migration` package provided by an external kit, that doesn't mark `%{_cross_os}image-feature(no-in-place-updates)` as a conflict. In such case, we still don't want to install `%{_cross_os}migration-bin` if the `in-place-updates` image feature is enabled.

ginglis13 (approved):

sources/api/pluto/src/eks.rs
```diff
@@ -70,3 +69,40 @@ where
         field: "kubernetes_network_config",
     })
 }
+
+#[cfg(not(feature = "fips"))]
+fn build_client<H, N>(
+    https_proxy: Option<H>,
+    no_proxy: Option<&[N]>,
+    config: aws_config::SdkConfig,
+) -> Result<aws_sdk_eks::Client>
```
bcressey, line 78: `pluto` needs to arrange to call this before either the EC2 or EKS client is initialized:
```
let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```

It's not OK to rely on `imdsclient` to do it; it's essentially a coincidence that all the settings that `pluto` currently deals with have some interaction with IMDS.

Arguably, `imdsclient` shouldn't do it at all and any caller that needs TLS and FIPS should arrange to do it. I'd like to see that change attempted since it's weird to depend on an HTTP-only client to set up something for HTTPS.

arnaldo2792:
Forced push includes:
- Separate `Conflicts` for `updog`
- Drop `fips` feature flag for `imdsclient` since for the time being it is an HTTP-only client
- Install `rustls` crypto provider in `pluto`, `cfsignal` and `shibaken`. This last one doesn't really requires it but due to feature unification, the `check-fips` script flags `shibaken` as required to be built with `aws-lc-rs`

arnaldo2792:
(Forced push updates commit messages to drop extra `Signed-off-by` comment)

arnaldo2792:
(Forced push updates commit message for `shibaken` changes)

sources/Cargo.toml
```diff
@@ -81,7 +81,7 @@ members = [
 ]
 
 [workspace.dependencies]
-apiclient = { version = "0.1", path = "api/apiclient" }
+apiclient = { version = "0.1", path = "api/apiclient", default-features = false }
```
bcressey, line 85: nit: I don't understand the "enforce TLS when FIPS mode is on" part of the commit message, that should be reworded.

sources/logdog/Cargo.toml
```diff
@@ -8,10 +8,15 @@ publish = false
 # Don't rebuild crate just because of changes to README.
 exclude = ["README.md"]
 
+[features]
```
bcressey, line 11: nit: many of these commit messages say `rustl's` instead of `rustls`, and `FIPs` instead of `FIPS`.

The sentences should all end with a `.` also.

packages/os/os.spec
```diff
@@ -121,9 +119,26 @@ Requires: %{_cross_os}settings-plugins
 
 %package -n %{_cross_os}apiclient
 Summary: Bottlerocket API client
+Requires: %{_cross_os}apiclient(binaries)
```
bcressey, line 122: nit: typo ("traetment") in commit message, missing period

bcressey (approved):

arnaldo2792:
(Forced push fixes nits in commit messages)

arnaldo2792:
Forced push includes:
- Vend `aws-smithy-experimental`
- Fix commit messages as suggested

sources/aws-smithy-experimental/external-types.toml
```diff
@@ -0,0 +1,4 @@
+allowed_external_types = [
+    "aws_smithy_runtime_api::*",
+    "aws_smithy_async::*"
+]
```
bcressey: nit: we can exclude this

COPYRIGHT
```diff
@@ -17,3 +17,9 @@ operating system images.
 Contains modified hyper-proxy files [mod.rs, stream.rs, tunnel.rs] from
 https://github.com/tafia/hyper-proxy 2021-09-20.
 Copyright (c) 2017 Johann Tuffe. Licensed under the MIT License.
+
+=^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+Contains modified aws-smithy-experimental files [README.md, Cargo.toml] from
```
bcressey: I would drop this specific list and update this in a PR that actually adds source code changes:
```
Contains aws-smithy-experimental files from
```

COPYRIGHT
```diff
@@ -17,3 +17,9 @@ operating system images.
...
+https://github.com/smithy-lang/smithy-rs/tree/HEAD/rust-runtime/aws-smithy-experimental 2024-10-22.
```
bcressey: This should point to a stable commit or release tag of some kind.

bcressey (commented):
CI is failing and has a suggested fix for `Cargo.toml`:
```
error: unexpected `cfg` condition name: `crypto_unstable`
   --> aws-smithy-experimental/src/hyper_1_0.rs:687:11
    |
687 |     #[cfg(crypto_unstable)]
    |           ^^^^^^^^^^^^^^^
    |
    = help: expected names are: `clippy`, `debug_assertions`, `doc`, `docsrs`, `doctest`, `feature`, `miri`, `overflow_checks`, `panic`, `proc_macro`, `relocation_model`, `rustfmt`, `sanitize`, `sanitizer_cfi_generalize_pointers`, `sanitizer_cfi_normalize_integers`, `target_abi`, `target_arch`, `target_endian`, `target_env`, `target_family`, `target_feature`, `target_has_atomic`, `target_has_atomic_equal_alignment`, `target_has_atomic_load_store`, `target_os`, `target_pointer_width`, `target_thread_local`, `target_vendor`, `test`, `ub_checks`, `unix`, and `windows`
    = help: consider using a Cargo feature instead
    = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
             [lints.rust]
             unexpected_cfgs = { level = "warn", check-cfg = ['cfg(crypto_unstable)'] }
```

arnaldo2792:
Forced push includes:
- Use `release-2024-10-09` in `COPYRIGHT` for `aws-smithy-experimental`
- Remove `external-types.toml`
- Add linting rule to warn instead of fail for the missing `crypto_unstable` flag

packages/os/os.spec
```diff
@@ -367,24 +548,53 @@ echo "** Output from non-static builds:"
...
+   exit "${fips_aws_sdk_rc}"
+fi
+
+if [ "${static_fips_rc}" -ne 0 ]; then
+   exit "${static_fips_rc}"
+fi
+
+if [ "${static_rc}" -ne 0 ]; then
+   exit "${static_rc}"
+fi
```
ginglis13, line 587: is there any reason to change the behavior from failing fast to waiting for each background build to complete, then checking for build success/failure?

packages/os/os.spec
```diff
@@ -317,24 +461,61 @@ exec 1>"${static_output}" 2>&1
 %cargo_build_static --manifest-path %{_builddir}/sources/Cargo.toml \
     -p apiclient \
     &
+
```
ginglis13: nit: newline

sources/api/pluto/Cargo.toml
```diff
@@ -19,13 +25,16 @@ headers.workspace = true
 http.workspace = true
 hyper = { workspace = true, features = ["default"] }
 hyper-rustls = { workspace = true, features = ["http2", "logging", "native-tokio", "tls12"] }
-imdsclient.workspace = true
+imdsclient = { workspace = true }
```
ginglis13: nit:
```suggestion
imdsclient.workspace = true
```

sources/logdog/src/log_request.rs
```diff
@@ -286,6 +286,8 @@ where
 
 /// Uses the reqwest library to send a GET request to `URL` and returns the response.
 fn send_get_request(url: &str) -> Result<Response> {
+    let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```
ginglis13, line 289: do we want to `install_default()` in respective `main` / `run` functions for our sources ? 

https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#using-the-per-process-default-cryptoprovider
> applications should call [CryptoProvider::install_default()](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default) early in their fn main(). If applications uses a custom provider based on the one built-in, they can activate the custom-provider feature to ensure its usage.

I see that's done for other cases of `install_default()` in this PR

packages/os/os.spec
arnaldo2792: This guarantees that if multiple jobs failed we will get all the outputs. If bail in the first fail, we will only get the output for that failure.

sources/logdog/src/log_request.rs
arnaldo2792: I didn't want to install the default provider unless really necessary. For `logdog`, that's only when the request type is `http` or `https`.

sources/aws-smithy-experimental/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "aws-smithy-experimental"
+version = "0.1.4"
+authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>"]
+description = "Experiments for the smithy-rs ecosystem"
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/smithy-lang/smithy-rs"
```
bcressey: nit: we should set `publish = false` here, and remove the repo URL, since we don't want to ever publish this fork of the crate

bcressey (approved):

arnaldo2792:
Forced push includes:
- Fix format in `pluto/Cargo.toml` for `imdsclient` dependency
- Prevent `aws-smithy-experimental` copy from being published
- Address other nits

ginglis13 (approved):

sources/api/migration/migrator/Cargo.toml
jpculp, line 13: Does this need to be:
```
fips = ["tough/fips", "aws-lc-rs/fips"]
```

arnaldo2792: Nope, `tough/fips` enables `aws-lc-rs/fips` feature flag.

bcressey (approved):

Issue #193, opened 2024-10-11
move kernels into separate kit


bcressey:
**What I'd like:**
I'd like to move kernels into their own kit, separate from the core kit. They don't require any of the userspace libraries or programs built in the rest of the core kit, and they're a key driver of lengthy build times.

#### Pros:
* Faster builds, both in the core kit and the new kernel kit repo. This helps CI and local development.
* Decoupled release cycle: core kit releases don't have to come out as often as kernel kits.

#### Cons:
* Two repositories to update to newer SDK releases.
* Two kits to release and update in downstream projects.

#### What's in scope:
* kernel-5.10, kernel-5.15, kernel-6.1
* kmod-5.10-nvidia, kmod-5.15-nvidia, kmod-6.1-nvidia
* kmod-6.1-neuron
* microcode
* linux-firmware
* grub
* shim

I'm on the fence about selinux-policy, but I think it's OK for it to stay in the core kit.

**Any alternatives you've considered:**
Status quo.

cbgbt:
Presumably the dependency graph will be like this?

![kit dependency graph](https://github.com/user-attachments/assets/86973bec-ee6a-4d8f-a716-234175fa407f)

This would be great for compile times in core-kit!

bryantbiggs:
efa kmod?

arnaldo2792:
The EFA kernel module is baked in the kernel, we don't consume the kmod from [here]

[here]: https://github.com/amzn/amzn-drivers/releases

ytsssun:
We have completed the migration for kernel kit - https://github.com/bottlerocket-os/bottlerocket-kernel-kit. Closing this issue.

PR #194, opened 2024-10-11
drop dependency on glibc for nvidia kmods

bcressey:
**Issue number:**
#193

**Description of changes:**
The NVIDIA kernel modules don't depend on the userspace C library. All of the userspace libraries are already compiled and don't need it either.

**Testing done:**
Builds!

**Terms of contribution:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -473,7 +473,6 @@ dependencies = [
 name = "kmod-5_10-nvidia"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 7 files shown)


yeazelm (approved):

larvacea (approved):
Not only Builds! but also Boots!, Joins Cluster!, and Runs nvidia_smi!

It occurred to me to do this just in case the link step might fail during boot.

PR #197, opened 2024-10-15
Update tough to 0.19

arnaldo2792:
**Issue number:**

N / A

**Description of changes:**

Update `tough` to 0.19. As part of this update, drop the `http` feature flag for `tough` in migrator, since `migrator` doesn't use the `http` client provided by `tough`. `reqwest` was bumped to `0.12`  but a clarification was needed since the AWS SDK depends on older libraries shared with `reqwest` (e.g. `hyper-rustls`). We could remove the `clarification` once the AWS SDK moves to a newer libraries.  Due to the `reqwest` update,  the `tcp` feature flag for `hyper` had to be explicitly enabled in `apiclient` otherwise it panics. The `tcp` flag was likely set by one of the older dependencies but dropped in newer versions.

A small change was required in `imdsclient` to use the `StatusCode` provided by `reqwest` instead of the `http` crate. `indexmap`, used in `netdog`, was bumped to version `2` and I confirmed that  `toml 0.8`  (the version used in the sources) [already] uses this version.

The CI environment now installs `clang` given that the Rust bindings for `aws-lc` are generated with `bindgen`, which uses `libclang` [under the hood].

[already]: https://github.com/toml-rs/toml/blob/b05e8c489be8ebfc0acacc1ec3556d95cd8d2198/crates/toml/Cargo.toml#L44

[under the hood]: https://rust-lang.github.io/rust-bindgen/requirements.html

**Testing done:**
- Build successfully
- Built `aws-k8s-1.30` x86_64, and the instance joined the cluster:

```bash
NAME                                         STATUS   ROLES    AGE     VERSION
ip-192-168-71-2.us-west-2.compute.internal   Ready    <none>   3h36m   v1.30.1-eks-e564799
```

- Upgrade/downgrade testing with a dummy migration:

```rust
use migration_helpers::common_migrations::ReplaceStringMigration;
use migration_helpers::{migrate, Result};
use std::process;

fn run() -> Result<()> {
    migrate(ReplaceStringMigration {
        setting: "settings.motd",
        old_val: "Welcome to Bottlerocket!",
        new_val: "New",
    })
}

fn main() {
    if let Err(e) = run() {
        eprintln!("{}", e);
        process::exit(1);
    }
}
```
Confirmed updates were found and the migrations ran:

```bash
[ssm-user@control]$ apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "3f5e65be-dirty",
    "pretty_name": "Bottlerocket OS 1.26.0 (aws-k8s-1.30)",
    "variant_id": "aws-k8s-1.30",
    "version_id": "1.26.0"
  }
}
[ssm-user@control]$ apiclient update check
03:35:25 [INFO] Refreshing updates...
{
  "active_partition": {
    "image": {
      "arch": "x86_64",
      "variant": "aws-k8s-1.30",
      "version": "1.26.0"
    },
    "next_to_boot": true
  },
  "available_updates": [
    "1.26.1"
  ],
  "chosen_update": {
    "arch": "x86_64",
    "variant": "aws-k8s-1.30",
    "version": "1.26.1"
  },
  "most_recent_command": {
    "cmd_status": "Success",
    "cmd_type": "refresh",
    "exit_status": 0,
    "stderr": "",
    "timestamp": "2024-10-15T03:35:26.915246473Z"
  },
  "staging_partition": null,
  "update_state": "Available"
}
[ssm-user@control]$ apiclient update apply
03:35:30 [INFO] Downloading and applying update to disk...
03:35:35 [INFO] Still waiting for updated status, will wait up to 594.5s longer...
03:35:40 [INFO] Still waiting for updated status, will wait up to 589.5s longer...
03:35:45 [INFO] Still waiting for updated status, will wait up to 584.5s longer...
03:35:50 [INFO] Setting the update active so it will apply on the next reboot...
03:35:51 [INFO] Update has been applied and will take effect on next reboot.
```

The instance came back up after the reboot:

```bash
[ssm-user@control]$ apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "ed026279-dirty",
    "pretty_name": "Bottlerocket OS 1.26.1 (aws-k8s-1.30)",
    "variant_id": "aws-k8s-1.30",
    "version_id": "1.26.1"
  }
}
```

**Terms of contribution:**

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -3,7 +3,7 @@ description: "Performs setup for caching and other common needs."
 runs:
   using: "composite"
   steps:
-    - run: sudo apt -y install build-essential openssl libssl-dev pkg-config liblz4-tool
+    - run: sudo apt -y install build-essential openssl libssl-dev pkg-config liblz4-tool clang
       shell: bash
     - uses: actions/cache/restore@v4
       # Restore most recent cache if available.
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -110,6 +110,7 @@ actix-web-actors = { version = "4", default-features = false }
 argh = "0.1"
 async-trait = "0.1"
 aws-config = "1"
+aws-lc-rs = "1"
 aws-sdk-cloudformation = "1"
 aws-sdk-ec2 = "1"
 aws-sdk-eks = "1"
@@ -145,7 +146,7 @@ httptest = "0.15"
 hyper = { version = "0.14", default-features = false }
 hyper-rustls = { version = "0.24", default-features = false }
 hyper-unix-connector = "0.2"
-indexmap = "1"
+indexmap = "2"
 ipnet = "2"
 itertools = "0.13"
 lazy_static = "1"
@@ -167,7 +168,7 @@ quick-xml = "0.26"
 quote = "1"
 rand = { version = "0.8", default-features = false }
 regex = "1"
-reqwest = { version = "0.11", default-features = false }
+reqwest = { version = "0.12", default-features = false }
 semver = "1"
 serde = "1"
 serde-xml-rs = "0.6"
@@ -190,7 +191,7 @@ tokio-test = "0.4"
 tokio-tungstenite = { version = "0.20", default-features = false }
 tokio-util = "0.7"
 toml = "0.8"
-tough = "0.18"
+tough = "0.19"
 unindent = "0.2"
 url = "2"
 walkdir = "2.4"
--- a/sources/api/apiclient/Cargo.toml
+++ b/sources/api/apiclient/Cargo.toml
@@ -17,7 +17,7 @@ futures.workspace = true
 futures-channel.workspace = true
 http.workspace = true
 httparse.workspace = true
-hyper = { workspace = true, features = ["client", "http1", "http2"] }
+hyper = { workspace = true, features = ["client", "http1", "http2", "tcp"] }
 hyper-unix-connector.workspace = true
 libc.workspace = true
 log.workspace = true
--- a/sources/api/migration/migrator/Cargo.toml
+++ b/sources/api/migration/migrator/Cargo.toml
@@ -24,7 +24,7 @@ simplelog.workspace = true
 snafu.workspace = true
 tokio = { workspace = true, features = ["fs", "macros", "rt-multi-thread"] }
 tokio-util = { workspace = true, features = ["compat", "io-util"] }
-tough = { workspace = true, features = ["http"] }
+tough = { workspace = true }
 update_metadata.workspace = true
 url.workspace = true
 
--- a/sources/clarify.toml
+++ b/sources/clarify.toml
@@ -201,3 +201,25 @@ license-files = [
     { path = "LICENSE-MIT", hash = 0xb9f15462 },
     { path = "LICENSE-APACHE", hash = 0x91d5a0a7 },
 ]
+
+[clarify.aws-lc-sys]
+expression = "ISC AND (Apache-2.0 OR ISC) AND OpenSSL AND MIT"
+license-files = [
+    { path = "LICENSE", hash = 0xf308ccd7 },
+    { path = "aws-lc/LICENSE", hash = 0xb6d14686 },
+    { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
+]
+
+[clarify.aws-lc-fips-sys]
+# jitterentropy is licensed under either GPL-2.0-only or BSD-3-Clause, we pick BSD-3-Clause
+expression = "ISC AND (Apache-2.0 OR ISC) AND OpenSSL AND MIT AND BSD-3-Clause"
+license-files = [
+    { path = "LICENSE", hash = 0xf308ccd7 },
+    { path = "aws-lc/LICENSE", hash = 0xf9c0df6a },
+    { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
+    { path = "aws-lc/third_party/jitterentropy/LICENSE", hash = 0x86ca01d4},
+    { path = "aws-lc/third_party/jitterentropy/LICENSE.bsd", hash = 0x7a297032 },
+]
+skip-files = [
+    "aws-lc/third_party/jitterentropy/LICENSE.gplv2"
+]
--- a/sources/deny.toml
+++ b/sources/deny.toml
@@ -67,7 +67,7 @@ skip-tree = [
     { name = "httptest", version = "=0.15" },
     # schnauzer uses cached, which uses an older version of darling
     { name = "darling", version = "=0.14" },
-    # reqwest brings a set of new dependencies that the aws-sdk-doesn't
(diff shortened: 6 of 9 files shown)


arnaldo2792:
Forced push includes:

- Don't change `sources/Cargo.lock` version

.github/actions/setup-node/action.yml
```diff
@@ -3,7 +3,7 @@ description: "Performs setup for caching and other common needs."
 runs:
   using: "composite"
   steps:
-    - run: sudo apt -y install build-essential openssl libssl-dev pkg-config liblz4-tool
+    - run: sudo apt -y install build-essential openssl libssl-dev pkg-config liblz4-tool clang
```
bcressey, line 6: the commit message says "install cmake" but you are installing clang?

sources/deny.toml
```diff
@@ -67,6 +67,9 @@ skip-tree = [
     { name = "httptest", version = "=0.15" },
     # schnauzer uses cached, which uses an older version of darling
     { name = "darling", version = "=0.14" },
+    # reqwest brings a set of new dependencies that the aws-sdk-doesn't
```
bcressey:
```suggestion
    # reqwest brings a set of new dependencies that the aws-sdk doesn't
```

sources/clarify.toml
```diff
@@ -201,3 +201,23 @@ license-files = [
...
+[clarify.aws-lc-sys]
+expression = "ISC AND (Apache-2.0 OR ISC) AND OpenSSL AND MIT"
+license-files = [
+    { path = "LICENSE", hash = 0xf308ccd7 },
+    { path = "aws-lc/LICENSE", hash = 0xb6d14686 },
+    { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
+]
+
+[clarify.aws-lc-fips-sys]
+expression = "BSD-3-Clause AND MIT AND GPL-2.0-only"
```
bcressey: I would expect all the license statements for `aws-lc-sys` would also apply to `aws-lc-fips-sys`, with the addition of the `jitterentropy` license.

`jitterentropy` says we can pick either `BSD-3-Clause` or `GPL-2.0-only` so just pick `BSD-3-Clause` and add a comment to that effect.

sources/imdsclient/src/lib.rs
```diff
@@ -19,9 +19,8 @@ The result is returned as a `String` _(ex. m5.large)_.
 
 use std::sync::RwLock;
 
-use http::StatusCode;
 use log::{debug, info, trace, warn};
-use reqwest::Client;
+use reqwest::{Client, StatusCode};
```
bcressey, line 23: It'd be good to add a short explanation of why this is changing (rather than just "what" is changing).

sources/api/migration/migrator/Cargo.toml
```diff
@@ -24,7 +24,7 @@ simplelog.workspace = true
 snafu.workspace = true
 tokio = { workspace = true, features = ["fs", "macros", "rt-multi-thread"] }
 tokio-util = { workspace = true, features = ["compat", "io-util"] }
-tough = { workspace = true, features = ["http"] }
+tough = { workspace = true }
```
bcressey, line 27: nit: explain "why not" in the commit message (although it is somewhat obvious)

arnaldo2792:
Forced push includes:

- Fix commit message that mentioned `cmake` instead of `clang`
- Remove extra dash from comment
- Fix licenses for `aws-lc-rs`
- Mentioned why `reqwest::StatusCode` is used
- Mentioned why the `http` feature in `migrator` isn't needed

arnaldo2792:
Forced push includes:
- Fix typo in commit message

ginglis13 (approved):

sources/clarify.toml
```diff
@@ -201,3 +201,22 @@ license-files = [
...
+
+[clarify.aws-lc-fips-sys]
+expression = "ISC AND (Apache-2.0 OR ISC) AND OpenSSL AND MIT AND BSD-3-Clause"
+license-files = [
+    { path = "LICENSE", hash = 0xf308ccd7 },
+    { path = "aws-lc/LICENSE", hash = 0xf9c0df6a },
+    { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
+    # jitterentropy is licensed under either GPL-2.0-only or BSD-3-Clause
+    { path = "aws-lc/third_party/jitterentropy/LICENSE.bsd", hash = 0x7a297032 },
+]
```
bcressey:
```suggestion
[clarify.aws-lc-fips-sys]
# jitterentropy is licensed under either GPL-2.0-only or BSD-3-Clause, we pick BSD-3-Clause
expression = "ISC AND (Apache-2.0 OR ISC) AND OpenSSL AND MIT AND BSD-3-Clause"
license-files = [
    { path = "LICENSE", hash = 0xf308ccd7 },
    { path = "aws-lc/LICENSE", hash = 0xf9c0df6a },
    { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
    { path = "aws-lc/third_party/jitterentropy/LICENSE.bsd", hash = 0x7a297032 },
]
skip-files = ["aws-lc/third_party/jitterentropy/LICENSE.gplv2"]
```

arnaldo2792:
(Forced pushes include fixes for CI, to skip the GPL license for aws-lc-fips-sys)

bcressey (approved):

sources/api/apiclient/Cargo.toml
```diff
@@ -17,7 +17,7 @@ futures.workspace = true
 futures-channel.workspace = true
 http.workspace = true
 httparse.workspace = true
-hyper = { workspace = true, features = ["client", "http1", "http2"] }
+hyper = { workspace = true, features = ["client", "http1", "http2", "tcp"] }
```
cbgbt, line 20: I'm surprised that `http` doesn't imply `tcp`... but I guess Unix sockets and all that.

sources/Cargo.toml
```diff
@@ -110,6 +110,7 @@ actix-web-actors = { version = "4", default-features = false }
 argh = "0.1"
 async-trait = "0.1"
 aws-config = "1"
+aws-lc-rs = "1"
 aws-sdk-cloudformation = "1"
```
cbgbt, line 114: Does it make sense to enable bindgen at the workspace level? Do we ever want `lc` without binding generation?

cbgbt (approved):
I have a question about `lc` enablement at the workspace level, but I don't think it's a blocker here. By default we try to give individual crates the ability to pick and choose features -- this also makes it more clear what the crate actually needs at an individual level.

sources/Cargo.toml
arnaldo2792: I thought so too, but I tried to followed what we do in other crates where we enable the features only when the crates need them. I can go either way, but currently we don't have any crate in `sources/Cargo.toml` that enables features and that's why I chose against this.

arnaldo2792:
(forced push fixes an error in the clarification)

ginglis13 (approved):

cbgbt (approved):

PR #204, opened 2024-10-17
host-ctr: support FIPS ECR service endpoints

ginglis13:
**Issue number:**

**Description of changes:**

Add support to host-ctr for FIPS ECR endpoints.

* add support for `-fips` service endpoints in host-ctr
* modify `ecr-prefix` to set the `fips` service endpoint if the variant is in FIPS mode and the region is FIPS supported

**Testing done:**

On an `aws-dev` instance in `us-west-2` with FIPS mode enabled:
```
[ssm-user@control]$ apiclient get settings.host-containers.control.source
{
  "settings": {
    "host-containers": {
      "control": {
        "source": "328549459982.dkr.ecr-fips.us-west-2.amazonaws.com/bottlerocket-control:v0.7.17"
      }
    }
  }
}
[ssm-user@control]$ cat /proc/sys/crypto/fips_enabled
1
```

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/src/helpers/mod.rs
+++ b/sources/api/schnauzer/src/helpers/mod.rs
@@ -16,7 +16,7 @@ use serde_plain::derive_fromstr_from_deserialize;
 use settings_extension_oci_defaults::OciDefaultsResourceLimitV1;
 use snafu::{OptionExt, ResultExt};
 use std::borrow::Borrow;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::str::FromStr;
@@ -70,6 +70,19 @@ lazy_static! {
         m.insert("us-west-2", "328549459982");
         m
     };
+
+    /// A set to tell us which regions have FIPS ECR endpoints.
+    /// https://docs.aws.amazon.com/general/latest/gr/ecr.html
+    static ref ECR_FIPS_REGIONS: HashSet<&'static str> = {
+        let mut h = HashSet::new();
+        h.insert("us-east-1");
+        h.insert("us-east-2");
+        h.insert("us-gov-east-1");
+        h.insert("us-gov-west-1");
+        h.insert("us-west-1");
+        h.insert("us-west-2");
+        h
+    };
 }
 
 /// But if there is a region that does not exist in our map (for example a new
@@ -77,6 +90,9 @@ lazy_static! {
 /// containers from here.
 const ECR_FALLBACK_REGION: &str = "us-east-1";
 const ECR_FALLBACK_REGISTRY: &str = "328549459982";
+
+/// Path to FIPS sysctl file.
+const FIPS_ENABLED_SYSCTL_PATH: &str = "/proc/sys/crypto/fips_enabled";
 
 lazy_static! {
     /// A map to tell us which endpoint to pull updates from for a given region.
@@ -517,6 +533,14 @@ pub fn tuf_prefix(
         })?;
 
     Ok(())
+}
+
+/// Utility function to determine if a variant is in FIPS mode based
+/// on /proc/sys/crypto/fips_enabled.
+fn fips_enabled() -> bool {
+    std::fs::read_to_string(FIPS_ENABLED_SYSCTL_PATH)
+        .map(|s| s.trim() == "1")
+        .unwrap_or(false)
 }
 
 /// The `metadata-prefix` helper is used to map an AWS region to the correct
@@ -1426,7 +1450,16 @@ fn ecr_registry<S: AsRef<str>>(region: S) -> String {
     match partition {
         "aws-cn" => format!("{}.dkr.ecr.{}.amazonaws.com.cn", registry_id, region),
         "aws-iso-e" => format!("{}.dkr.ecr.{}.cloud.adc-e.uk", registry_id, region),
-        _ => format!("{}.dkr.ecr.{}.amazonaws.com", registry_id, region),
+        _ => {
+            // Only inject the FIPS service endpoint if the variant is in FIPS mode and the
+            // region supports FIPS.
+            let suffix = if fips_enabled() && ECR_FIPS_REGIONS.contains(region) {
+                "-fips"
+            } else {
+                ""
+            };
+            format!("{}.dkr.ecr{}.{}.amazonaws.com", registry_id, suffix, region)
+        }
     }
 }
 
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -38,7 +38,8 @@ import (
 // Example 1: 777777777777.dkr.ecr.us-west-2.amazonaws.com/my_image:latest
 // Example 2: 777777777777.dkr.ecr.cn-north-1.amazonaws.com.cn/my_image:latest
 // Example 3: 777777777777.dkr.ecr.eu-isoe-west-1.cloud.adc-e.uk/my_image:latest
-var ecrRegex = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.(amazonaws\.com(\.cn)?|cloud\.adc-e\.uk).*`)
+// Example 4: 777777777777.dkr.ecr-fips.us-west-2.amazonaws.com/my_image:latest
+var ecrRegex = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr(-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.(amazonaws\.com(\.cn)?|cloud\.adc-e\.uk).*`)
 
 const (
 	// The maximum size of an	image label.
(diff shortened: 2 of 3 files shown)


sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -1426,7 +1524,18 @@ fn ecr_registry<S: AsRef<str>>(region: S) -> String {
     match partition {
         "aws-cn" => format!("{}.dkr.ecr.{}.amazonaws.com.cn", registry_id, region),
         "aws-iso-e" => format!("{}.dkr.ecr.{}.cloud.adc-e.uk", registry_id, region),
-        _ => format!("{}.dkr.ecr.{}.amazonaws.com", registry_id, region),
+        _ => format!(
+            "{}.dkr.ecr{}.{}.amazonaws.com",
+            registry_id,
+            // Only inject the fips service endpoint if the variant is in FIPS mode and the
+            // region supports FIPS.
+            if fips_enabled() && region.starts_with("us-") {
```
ginglis13: Open to feedback that this is not robust enough and instead we should be checking against a set of the known FIPS-supported regions in all partitions.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -627,6 +626,19 @@ func parseImageURISpecialRegions(input string) (ecr.ECRSpec, error) {
...
 
+	// Return early if the FIPS endpoint is being used. amazon-ecr-containerd-resolver doesn't yet support FIPS urls:
+	// https://github.com/awslabs/amazon-ecr-containerd-resolver/blob/7b72333e780f5a5168936eae79fb89448e2f2a8f/ecr/ref.go#L43
+	// The ecr-prefix helper for admin and control host containers will have already accounted for setting this endpoint
+	// if the region has FIPS support.
+	if matches[2] == "-fips" {
+		region := matches[3]
+		ecrRefPrefix := fmt.Sprintf("ecr.aws/arn:aws:ecr-fips:%s:", region)
+		return ecr.ParseRef(fmt.Sprintf("%s%s:repository/%s", ecrRefPrefix, account, fullRepoPath))
+	}
```
ginglis13: Similar to above comment, open to feedback that this should have region-specific logic (or belong in its own method, not `parseImageURISpecialRegions`. `ecr-prefix` should have taken care of this already, but if one were to use `host-ctr` manually / override host container sources to a FIPS URI to an image in a non-FIPS region, this would fail.

ginglis13:
^ specify `toml` for a codeblock in doc comment so it doesn't get run as a doc test

ginglis13:
^ recent force pushes trim this PR's scope down to just FIPS ECR support in `ecr-prefix` and `host-ctr`. I took my own advice and decided to make the FIPS supported regions a set in both cases.

ginglis13:
^ latest force pushes rebase, renamed variable

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -594,21 +595,28 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
...
 // constructs the canonical ECR references while skipping certain checks.
-// We only do this for special regions that are not yet supported by the aws-go-sdk.
+// We only do this for special regions that are not yet supported by the aws-go-sdk and for ECR fips endpoints.
 // Referenced source: https://github.com/awslabs/amazon-ecr-containerd-resolver/blob/a5058cf091f4fc573813a032db37a9820952f1f9/ecr/ref.go#L70-L71
 func parseImageURISpecialRegions(input string) (ecr.ECRSpec, error) {
 	ecrRefPrefixMapping := map[string]string{
 		"ap-southeast-5": "ecr.aws/arn:aws:ecr:ap-southeast-5:",
 		"eu-isoe-west-1": "ecr.aws/arn:aws-iso-e:ecr:eu-isoe-west-1:",
 	}
+	// https://docs.aws.amazon.com/general/latest/gr/ecr.html
```
arnaldo2792: nit: a comment above the URL would be helpful, since the link doesn't point out the problem solved by the map of regions. It will be good to call out in the comment that ECR has FIPS endpoints in GovCloud different than the "normal" endpoints.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -594,21 +595,28 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
...
+	fipsSupportedRegionMapping := map[string]string{
+		"us-east-1":     "",
+		"us-east-2":     "",
+		"us-west-1":     "",
+		"us-west-2":     "",
+		"us-gov-east-1": "",
+		"us-gov-west-1": "",
+	}
```
arnaldo2792: Are you using a map for fast look ups? Since there aren't that many regions, maybe using a list would be OK. Alternatively, we could follow what others did here and use a map of booleans:

https://github.com/aws/amazon-ecs-agent/blob/ababaac1a77ac7f05f069859b246afeb2d99012b/agent/s3/factory/factory.go#L53

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -594,21 +595,28 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
 
 // parseImageURISpecialRegions mimics the parsing in ecr.ParseImageURI but
 // constructs the canonical ECR references while skipping certain checks.
-// We only do this for special regions that are not yet supported by the aws-go-sdk.
+// We only do this for special regions that are not yet supported by the aws-go-sdk and for ECR fips endpoints.
```
arnaldo2792: nit: for consistency use `s/fips/FIPS` in comments

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -627,6 +635,23 @@ func parseImageURISpecialRegions(input string) (ecr.ECRSpec, error) {
...
 
+	// Return early if the FIPS endpoint is being used. amazon-ecr-containerd-resolver doesn't yet support FIPS urls:
+	// https://github.com/awslabs/amazon-ecr-containerd-resolver/blob/7b72333e780f5a5168936eae79fb89448e2f2a8f/ecr/ref.go#L43
+	// The ecr-prefix helper for admin and control host containers will have already accounted for setting this endpoint
+	// if the region has FIPS support.
+	if matches[2] == "-fips" {
+		region := matches[3]
+		_, isFips := fipsSupportedRegionMapping[region]
+		if !isFips {
+			return ecr.ECRSpec{}, fmt.Errorf("%s: %s", "invalid FIPS region in internal mapping", region)
```
arnaldo2792: Shouldn't the error be `invalid FIPS region` or `unsupported FIPS region`? Otherwise it is confusing and makes me think that the region was invalid in the internal mapping, but in reality this error will occur when an invalid FIPS region is used.

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -70,6 +70,19 @@ lazy_static! {
         m.insert("us-west-2", "328549459982");
         m
     };
+
+    /// A set to tell us which regions have FIPS support.
+    /// https://docs.aws.amazon.com/general/latest/gr/ecr.html
+    static ref FIPS_REGION_SET: HashSet<&'static str> = {
```
arnaldo2792:
```suggestion
    static ref FIPS_REGIONS: HashSet<&'static str> = {
````

File: sources/api/schnauzer/src/helpers/mod.rs [hunk 8]
```diff
@@ -78,6 +91,9 @@ lazy_static! {
 const ECR_FALLBACK_REGION: &str = "us-east-1";
 const ECR_FALLBACK_REGISTRY: &str = "328549459982";
 
+/// Filepath to FIPS configuration.
+const PROC_SYS_CRYPTO_FIPS_ENABLED: &str = "/proc/sys/crypto/fips_enabled";
```
arnaldo2792:
```suggestion
/// Path to FIPS sysctl file
const FIPS_ENABLED_SYSCTL_PATH: &str = "/proc/sys/crypto/fips_enabled";
```

sources/api/schnauzer/src/helpers/mod.rs
```diff
@@ -1426,7 +1450,18 @@ fn ecr_registry<S: AsRef<str>>(region: S) -> String {
...
+            registry_id,
+            // Only inject the fips service endpoint if the variant is in FIPS mode and the
+            // region supports FIPS.
+            if fips_enabled() && FIPS_REGION_SET.contains(region) {
+                "-fips"
+            } else {
+                ""
+            },
+            region
+        ),
```
arnaldo2792:
```suggestion
        _ => {
          let suffix = if fips_enabled() && FIPS_REGION_SET.contains(region) {
                "-fips"
            } else {
                ""
            };
            
            format!(
            "{}.dkr.ecr{}.{}.amazonaws.com",
            registry_id,
            suffix,
            region
        }),
```

sources/host-ctr/cmd/host-ctr/main.go
ginglis13: Sounds good. as for 

> It will be good to call out in the comment that ECR has FIPS endpoints in GovCloud different than the "normal" endpoints.

What do you mean by this? https://docs.aws.amazon.com/general/latest/gr/ecr.html shows both GovCloud regions follow the same endpoint format as US commercial regions. Are you asking to call out that GovCloud provides both non-FIPS and FIPS endpoints in ECR?

sources/host-ctr/cmd/host-ctr/main.go
ginglis13: I'm using this map like a set, since Go std lib does not provide a set type. That's https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/204/files#diff-e5dea204d3b274d07b5340673eeaa9b44e00cdf41d6fe9d25db80b4638ab9352R644

```
_, isFips := fipsSupportedRegionMapping[region]
```


I'll clarify that in both the comment and variable name itself (rename to fipsSupportedRegionSet)

ginglis13:
^ force push addressed feedback, mostly style.

bcressey (approved):

arnaldo2792 (approved):

PR #205, opened 2024-10-21
selinux-policy: adjust kernel permissions for NFS

### LINKED ISSUE #4116: Cannot execute binaries stored in an NFS Server running on a Bottlerocket node
**Image I'm using:**
`Bottlerocket OS 1.20.4 (aws-k8s-1.30)`

**Context:**
We have some software that runs multiple pods for multiple stages in a pipeline. To be able to complete this dynamically and allow retries on specific steps, we spawn short-lived pods that connect to an NFS server running in-cluster for its ephemeral data. A typical installation would have the orchestrator and the NFS server to begin with. When the orchestrator receives a piece of work, it will:
* Create a subfolder in the NFS server, and download any required executables to it
* Spawn a pod, which runs the executables in the subfolder
* Once the pod has finished running, the subfolder gets removed

The NFS server is a simple variant of [this alpine server](https://github.com/OctopusDeploy/nfs-server-alpine).

When running an NFS Server in a container in bottlerocket, you are able to execute files on the share from a mount in a different container.

**What actually happened:**
The `NFSD` process is denied `execute` access. This is exhibited in this AVC denial log:
```
Jul 26 01:20:06 ip-10-0-19-55.ap-southeast-2.compute.internal audit[2830356]: AVC avc:  denied  { execute } for  pid=2830356 comm="nfsd" name="bootstrapRunner" dev="nvme1n1p1" ino=151427631 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:data_t:s0:c432,c649 tclass=file permissive=0
```

From what I can tell, this is because the process is running as a `kernel` task, even though it's actually exposing data from a share from a container. My current line of thinking is that this is because it's a privileged container and actually hooking into the kernel-level support.  The `nfsd` processes have the `system_u:system_r:kernel_t:s0` SELinux context, and are _not_ children of the NFS server pod.

**What I've tried to do to work around the problem:**
I've attempted to work around this problem by using EFS rather than locally hosting, but when using access points and dynamically provisioned volumes, `chmod` commands get `permission denied`, which fails many scripts (and even `tar` in some cases).

To reproduce the problem, you can create the resources I've added below in a Kubernetes cluster that is running Bottlerocket OS `1.20.4`. I have been doing this in an AWS EKS cluster.

You will be able to see the logs after running `logdog` from the admin container **in the node running the NFS server, not the nfs-client pod**. To run this reproduction, you will also need the NFS CSI driver, which you can install using helm:
```bash
helm upgrade --install --atomic \
--repo https://raw.githubusercontent.com/kubernetes-csi/csi-driver-nfs/master/charts \
--namespace kube-system \
--version v4.6.0 \
csi-driver-nfs \
csi-driver-nfs
```

If you deploy this outside of the `default` namespace, please adjust the server URL to instead point to the namespace you're deploying to - replace `nfs.default.svc.cluster.local` with `nfs.<your-namespace>.svc.cluster.local`.

### Resources:
**NFS Server**
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: nfs-server
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: nfs
  serviceName: "nfs"
  template:
    metadata:
      labels:
        app.kubernetes.io/name: nfs
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: kubernetes.io/os
                operator: In
                values:
                - linux
              - key: kubernetes.io/arch
                operator: In
                values:
                - arm64
                - amd64
      containers:
      - env:
        - name: SHARED_DIRECTORY
          value: /octopus
        - name: SYNC
          value: "true"
        image: octopusdeploy/nfs-server:1.0.1
        imagePullPolicy: IfNotPresent
        name: nfs-server
        ports:
        - containerPort: 2049
          protocol: TCP
        resources:
          requests:
            cpu: 50m
            memory: 50Mi
        securityContext:
          privileged: true
        volumeMounts:
        - mountPath: /octopus
          name: octopus-volume
      restartPolicy: Always
      volumes:
      - emptyDir:
          sizeLimit: 10Gi
        name: octopus-volume
  updateStrategy:
    type: RollingUpdate
---
apiVersion: v1
kind: Service
metadata:
  name: nfs
spec:
  clusterIP: None
  ports:
  - name: nfs
    port: 2049
    protocol: TCP
    targetPort: 2049
  selector:
    app.kubernetes.io/name: nfs
  sessionAffinity: None
  type: ClusterIP
```

**PV/PVC**
```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: nfs-pv-10gi
spec:
  accessModes:
  - ReadWriteMany
  capacity:
    storage: 10Gi
  csi:
    driver: nfs.csi.k8s.io
    volumeAttributes:
      server: nfs.default.svc.cluster.local
      share: /
    volumeHandle: nfs.default.svc.cluster.local/octopus##
  mountOptions:
  - nfsvers=4.1
  - lookupcache=none
  - soft
  - timeo=50
  - retrans=4
  persistentVolumeReclaimPolicy: Retain
  storageClassName: nfs-csi
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: nfs-pvc-10gi
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 10Gi
  storageClassName: nfs-csi
```

**Client Pod**
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nfs-client
spec:
  selector:
    matchLabels:
      app: nfs-client
  template:
    metadata:
      labels:
        app: nfs-client
    spec:
      containers:
      - name: nfs-client
        image: alpine
        command: ["sh"]
        args: 
        - -c
        - 'echo "echo \"hello world\"" > /octopus/runme.sh && chmod +x /octopus/runme.sh && sh -c "/octopus/runme.sh"'
        resources:
          limits:
            memory: "128Mi"
            cpu: "500m"
        volumeMounts:
        - mountPath: /octopus
          name: mount
      volumes:
      - name: mount
        persistentVolumeClaim:
          claimName: nfs-pvc-10gi```

### PR DESCRIPTION
**Pull request by bcressey**
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/4116

**Description of changes:**
When serving files over NFS, the kernel's permissions to access the inode are checked when determining whether the file can be executed by clients. If the kernel lacks permission, then the client will see a permission error and won't be able to execute the program.

Work around this by allowing the kernel access to execute mutable files, while denying it permission to execute a new program without transitioning to a new SELinux context.

To close off other potential paths to execution, define a transition to a "forbidden" type that is explicitly denied, and deny the use of any object as an entry point to that domain.

**Testing done:**
Verified that the test case in the related issue now passes.

Added a test case to attempt to trigger kernel execution of untrusted code via one of the well-known forms of container escape - installing a new core pattern handler. Confirmed that execution was blocked:
```
[ 1535.724944] audit: type=1400 audit(1729528842.857:147): avc:  denied  { execute_no_trans } for  pid=26787 comm="kworker/u8:4" path="/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/172/fs/tmp/core-pattern-escape.sh" dev="nvme1n1p1" ino=54842222 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:data_t:s0:c346,c677 tclass=file permissive=0
```

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/mcs.cil
+++ b/packages/selinux-policy/mcs.cil
@@ -86,9 +86,11 @@
 ; Restrict process transitions unless one of these conditions is met:
 ; * the new label exactly matches the old label
 ; * the source context is for a trusted subject
+; * the target context is not forbidden
 
 (mlsconstrain (processes (transform))
-  (or (eq t1 trusted_s)
+  (or (and (eq t1 trusted_s)
+           (neq t2 forbidden_t))
       (and (and (and (and
            (eq u1 u2)
            (eq r1 r2))
--- a/packages/selinux-policy/rules.cil
+++ b/packages/selinux-policy/rules.cil
@@ -166,9 +166,27 @@
 ; Subjects that must run verified code can execute immutable objects, since
 ; those are all protected by dm-verity.
 (allow verified_s immutable_o (files (execute)))
+(allow kernel_t immutable_o (files (execute)))
 
 ; Subjects that must run verified code cannot execute mutable objects.
 (neverallow verified_s mutable_o (files (execute)))
+
+; Ideally the kernel would also be denied permission to execute mutable
+; objects. However, this breaks certain scenarios such as serving files
+; over NFS, where the kernel's permissions are checked.
+(allow kernel_t mutable_o (file (execute)))
+
+; Prevent the kernel from executing mutable objects by blocking execution
+; unless there's a defined transition.
+(neverallow kernel_t mutable_o (file (execute_no_trans)))
+
+; Backstop against kernel execution of mutable objects by defining a type
+; transition, which is then explicitly disallowed.
+(typetransition kernel_t mutable_o process forbidden_t)
+(neverallow kernel_t forbidden_t (processes (transform)))
+
+; Block the use of any object as an entry point to the forbidden type.
+(neverallow forbidden_t all_o (files (enter)))
 
 ; All subjects are allowed to write to objects with their own label.
 ; This includes files like the ones under /proc/self.
--- a/packages/selinux-policy/subject.cil
+++ b/packages/selinux-policy/subject.cil
@@ -68,12 +68,18 @@
 (roletype system_r super_t)
 (context admin (system_u system_r super_t s0))
 
+; Processes that should never exist.
+(type forbidden_t)
+(roletype system_r forbidden_t)
+(context forbidden (system_u system_r forbidden_t s0))
+
 ; The set of all subjects.
 (typeattribute all_s)
 (typeattributeset all_s (
   kernel_t init_t system_t mount_t api_t
   network_t clock_t bus_t runtime_t
-  container_t control_t super_t))
+  container_t control_t super_t
+  forbidden_t))
 
 ; Subjects that are treated as a privileged part of the OS.
 (typeattribute privileged_s)
@@ -121,7 +127,8 @@
 
 ; Subjects shipped with the OS that should only execute verified code.
 (typeattribute verified_s)
-(typeattributeset verified_s (xor (host_s) (runtime_t mount_t api_t init_t)))
+(typeattributeset verified_s (xor (host_s) (
+  runtime_t mount_t api_t init_t kernel_t)))
 
 ; Subjects that are allowed to manage the system clock.
 (typeattribute clock_s)

### CODE REVIEW

File: packages/selinux-policy/mcs.cil [hunk 1]
```diff
@@ -86,9 +86,11 @@
 ; Restrict process transitions unless one of these conditions is met:
 ; * the new label exactly matches the old label
 ; * the source context is for a trusted subject
+; * the target context is not forbidden
 
 (mlsconstrain (processes (transform))
-  (or (eq t1 trusted_s)
+  (or (and (eq t1 trusted_s)
+           (neq t2 forbidden_t))
```
bcressey, line 93: I added this so that any rules related to process transitions wouldn't be affected by the presence of MLS categories on the label. Previously it was a blanket "allow this for trusted subjects" and now it is "allow this for trusted subjects, unless the target context is forbidden".

packages/selinux-policy/rules.cil
```diff
@@ -166,10 +166,28 @@
 ; Subjects that must run verified code can execute immutable objects, since
 ; those are all protected by dm-verity.
 (allow verified_s immutable_o (files (execute)))
+(allow kernel_t immutable_o (files (execute)))
```
bcressey, line 169: This is necessary because `kernel_t` dropped out of the `verified_s` set.

packages/selinux-policy/rules.cil
```diff
@@ -166,10 +166,28 @@
...
 
 ; Subjects that must run verified code cannot execute mutable objects.
 (neverallow verified_s mutable_o (files (execute)))
 
+; Ideally the kernel would also be denied permission to execute mutable
+; objects. However, this breaks certain scenarios such as serving files
+; over NFS, where the kernel's permissions are checked.
+(allow kernel_t mutable_o (file (execute)))
```
bcressey, line 177: This is the actual fix to the NFS serving issue.

Note that `files (execute)` grants both "execute" and "execute_no_trans"; I've moved to the more specific `file` here in order to split up those two permissions.

packages/selinux-policy/rules.cil
```diff
@@ -166,10 +166,28 @@
...
+
+; Prevent the kernel from executing mutable objects by blocking execution
+; unless there's a defined transition.
+(neverallow kernel_t mutable_o (file (execute_no_trans)))
```
bcressey, line 181: This prevents execution because of [this check](https://elixir.bootlin.com/linux/v6.1.112/source/security/selinux/hooks.c#L2334) in the "no transition" branch.

packages/selinux-policy/rules.cil
```diff
@@ -166,10 +166,28 @@
...
+
+; Backstop against kernel execution of mutable objects by defining a type
+; transition, which is then explicitly disallowed.
+(typetransition kernel_t mutable_o process forbidden_t)
+(neverallow kernel_t forbidden_t (processes (transform)))
```
bcressey, line 186: This prevents execution because of [this check](https://elixir.bootlin.com/linux/v6.1.112/source/security/selinux/hooks.c#L2341) in the "transition" branch.

packages/selinux-policy/rules.cil
```diff
@@ -166,10 +166,28 @@
...
+
+; Block the use of any object as an entry point to the forbidden type.
+(neverallow forbidden_t all_o (files (enter)))
```
bcressey, line 189: This causes the [next check](https://elixir.bootlin.com/linux/v6.1.112/source/security/selinux/hooks.c#L2347) in the "transition" branch to fail.

packages/selinux-policy/subject.cil
```diff
@@ -68,12 +68,18 @@
 (roletype system_r super_t)
 (context admin (system_u system_r super_t s0))
 
+; Processes that should never exist.
+(type forbidden_t)
+(roletype system_r forbidden_t)
+(context forbidden (system_u system_r forbidden_t s0))
```
bcressey, line 74: This is boilerplate to define the `forbidden_t` type.

packages/selinux-policy/subject.cil
```diff
@@ -68,12 +68,18 @@
...
+
 ; The set of all subjects.
 (typeattribute all_s)
 (typeattributeset all_s (
   kernel_t init_t system_t mount_t api_t
   network_t clock_t bus_t runtime_t
-  container_t control_t super_t))
+  container_t control_t super_t
+  forbidden_t))
```
bcressey, line 82: Add `forbidden_t` to the set of all subjects (`all_s`) since there are a couple "neverallow" rules that should apply to it also.

packages/selinux-policy/subject.cil
```diff
@@ -121,7 +127,8 @@
 
 ; Subjects shipped with the OS that should only execute verified code.
 (typeattribute verified_s)
-(typeattributeset verified_s (xor (host_s) (runtime_t mount_t api_t init_t)))
+(typeattributeset verified_s (xor (host_s) (
+  runtime_t mount_t api_t init_t kernel_t)))
```
bcressey, line 131: This drops `kernel_t` from the `verified_s` set, so that we can retain the more restrictive rule:
```
(neverallow verified_s mutable_o (files (execute)))
```

larvacea (approved):

arnaldo2792 (approved):

PR #206, opened 2024-10-21
mount binfmt_misc by default and restrict writes

### LINKED ISSUE #4215: multiarch/qemu-user-static not working on 1.23
This is a pretty strange one, but I wanted to raise it in case someone else was hitting it.
In a CICD context, we build multiarch images on x64 by executing the `multiarch/qemu-user-static` image before trying to build arm64 and amd64 docker images. 

We use this command to do this : 
```
podman run --authfile /run/containers/0/auth-ecr.json --rm --privileged multiarch/qemu-user-static --reset -p yes
```
As far as I know this registers alternative executable types in the kernel, so QEMU will know when to step in to emulate the commands. This has been working for quite a while, but when 1.23 was pushed out, this process broke. 

Expected results : 
Image to build correctly 

Actual results:
Image build fails with exec format error
```
[2024-09-26T06:40:10.212Z] + buildah build --authfile /run/containers/0/auth-ecr.json --layers --format oci --pull --network=host --ulimit nofile=24000:24000 --squash --jobs 2 --platform=linux/amd64,linux/arm64 --manifest jobs-job-service:0.0.550-test-builds-gbucknel-b4-g2b0d41d .
[202[2024-09-26T06:40:23.636Z] process failed to start with error: fork/exec /bin/sh: exec format errorprocess exited with error: exec: not startedsubprocess exited with status 1

[2024-09-26T06:40:30.904Z] Error: [linux/arm64]: building at STEP "RUN mkdir -p /usr/apps/service-config": exit status 1

script returned exit code 1
```
We were able to get everything working again by going back to 1.22. 

Also , running the [register script](https://github.com/qemu/qemu/blob/master/scripts/qemu-binfmt-conf.sh) by hand in a sheltie session on the host seems to make arm64 builds work in 1.23 again, so trying to do the same thing in a bootstrap container is something I'd like to try.  

Looking at the changelog, I'm a bit puzzled as to where the change could've been. Will continue to try to figure it out. If anyone else has seen this, it would be really interesting.

bcressey:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/4215

**Description of changes:**
This ensures that the binfmt_misc filesystem is mounted at boot, like other kernel filesystems.

Exclude the systemd-binfmt service and the automount unit, since the expectation is that privileged containers will register any handlers that the system requires.

Restrict writes to privileged containers, since `binfmt_misc` operations are not otherwise constrained by capabilities.

**Testing done:**
Verified that the binfmt_misc filesystem was mounted at boot.

```
bash-5.1# findmnt /proc/sys/fs/binfmt_misc/
TARGET                   SOURCE      FSTYPE      OPTIONS
/proc/sys/fs/binfmt_misc binfmt_misc binfmt_misc rw,nosuid,nodev,noexec,relatime
```

Added a test case to ensure that unprivileged containers running as UID with access to a mounted binfmt_misc filesystem could not write to it:
```
[ 1540.521904] audit: type=1400 audit(1729528847.657:148): avc:  denied  { write } for  pid=26856 comm="unprivileged-co" name="register" dev="binfmt_misc" ino=3 scontext=system_u:system_r:container_t:s0:c172,c178 tcontext=system_u:object_r:binfmt_misc_fs_t:s0 tclass=file permissive=0
```

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/fs.cil
+++ b/packages/selinux-policy/fs.cil
@@ -20,7 +20,7 @@
 ; Specify a context for filesystems that do not support other ways to label.
 (genfscon autofs / any)
 (genfscon bdev / any)
-(genfscon binfmt_misc / any)
+(genfscon binfmt_misc / binfmt_misc_fs)
 (genfscon bpf / any)
 (genfscon cgroup / any)
 (genfscon cgroup2 / any)
--- a/packages/selinux-policy/object.cil
+++ b/packages/selinux-policy/object.cil
@@ -59,6 +59,11 @@
 (roletype object_r proc_t)
 (context proc (system_u object_r proc_t s0))
 
+; Files in the binfmt_misc filesystem
+(type binfmt_misc_fs_t)
+(roletype object_r binfmt_misc_fs_t)
+(context binfmt_misc_fs (system_u object_r binfmt_misc_fs_t s0))
+
 ; Files where we have no specific policy objectives, such as
 ; tmpfs mounts and various kernel filesystems.
 (type any_t)
@@ -135,7 +140,7 @@
 
 ; Dynamic objects are files on temporary storage with special rules.
 (typeattribute dynamic_o)
-(typeattributeset dynamic_o (etc_t))
+(typeattributeset dynamic_o (etc_t binfmt_misc_fs_t))
 
 ; Shared objects are files on local storage for containers.
 (typeattribute shared_o)
@@ -204,7 +209,7 @@
   os_t init_exec_t api_exec_t clock_exec_t
   network_exec_t bus_exec_t runtime_exec_t
   mount_exec_t cni_exec_t csi_exec_t
-  any_t etc_t proc_t
+  any_t etc_t proc_t binfmt_misc_fs_t
   local_t data_t private_t secret_t cache_t
   lease_t measure_t state_t
   api_socket_t))
--- a/packages/selinux-policy/rules.cil
+++ b/packages/selinux-policy/rules.cil
@@ -179,11 +179,15 @@
 (allow all_s ephemeral_o (files (mutate mount)))
 
 ; Trusted subjects are allowed to write to and manage mounts
-; for "dynamic" files in /etc.
-(allow trusted_s dynamic_o (files (mutate mount)))
+; for files in /etc.
+(allow trusted_s etc_t (files (mutate mount)))
 
 ; wicked calls netdog which writes /etc/resolv.conf.
 (allow network_t etc_t (files (mutate)))
+
+; Privileged subjects are allowed to write to and manage mounts for
+; the binfmt_misc filesystem.
+(allow privileged_s binfmt_misc_fs_t (files (mutate mount)))
 
 ; Other subjects cannot modify these "dynamic" files.
 (neverallow other_s dynamic_o (files (mutate mount)))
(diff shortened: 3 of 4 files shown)


packages/systemd/systemd.spec
```diff
@@ -161,7 +161,7 @@ CONFIGURE_OPTS=(
  -Defi=true
  -Dtpm=false
  -Denvironment-d=false
- -Dbinfmt=false
+ -Dbinfmt=true
```
bcressey, line 164: The only part of this feature we want is `proc-sys-fs-binfmt_misc.mount`, the mount unit for the filesystem. I opted to enable it and exclude all the other files since I felt that would be more resilient against subsequent behavior changes, if this feature ended up turned on.

(This is why there are so many new excludes.)

packages/selinux-policy/rules.cil
```diff
@@ -179,12 +179,16 @@
 (allow all_s ephemeral_o (files (mutate mount)))
 
 ; Trusted subjects are allowed to write to and manage mounts
-; for "dynamic" files in /etc.
-(allow trusted_s dynamic_o (files (mutate mount)))
+; for files in /etc.
+(allow trusted_s etc_t (files (mutate mount)))
```
bcressey, line 183: `binfmt_misc_t` is now part of `dynamic_o`, so I made this rule more specific since the trusted software has no reason to define binfmt handlers.

packages/selinux-policy/rules.cil
```diff
@@ -179,12 +179,16 @@
...
 
 ; wicked calls netdog which writes /etc/resolv.conf.
 (allow network_t etc_t (files (mutate)))
 
+; Privileged subjects are allowed to write to and manage mounts for
+; the binfmt_misc filesystem.
+(allow privileged_s binfmt_misc_fs_t (files (mutate mount)))
```
bcressey, line 190: This is the main change; allowing privileged containers write access to binfmt_misc files.

(Unprivileged containers are restricted by the rule immediately below this one.)

packages/selinux-policy/object.cil
```diff
@@ -59,6 +59,11 @@
 (roletype object_r proc_t)
 (context proc (system_u object_r proc_t s0))
 
+; Files in the binfmt_misc filesystem
+(type binfmt_misc_fs_t)
+(roletype object_r binfmt_misc_fs_t)
+(context binfmt_misc_fs (system_u object_r binfmt_misc_fs_t s0))
```
bcressey, line 65: This is boilerplate to define the `binfmt_misc_fs_t` type.

packages/selinux-policy/fs.cil
```diff
@@ -20,7 +20,7 @@
 ; Specify a context for filesystems that do not support other ways to label.
 (genfscon autofs / any)
 (genfscon bdev / any)
-(genfscon binfmt_misc / any)
+(genfscon binfmt_misc / binfmt_misc_fs)
```
bcressey, line 23: This changes the default label for the binfmt misc filesystem to the new type.

packages/selinux-policy/object.cil
```diff
@@ -135,7 +140,7 @@
 
 ; Dynamic objects are files on temporary storage with special rules.
 (typeattribute dynamic_o)
-(typeattributeset dynamic_o (etc_t))
+(typeattributeset dynamic_o (etc_t binfmt_misc_fs_t))
```
bcressey, line 143: `dynamic_o` might be one of the worst names in the policy, but it has the intended set of "never allow" rules already.

larvacea (approved):

arnaldo2792 (approved):

PR #207, opened 2024-10-21
build neuron kmod in kernel packages

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket/issues/4218

**Description of changes:**
Build the Neuron kmod as part of each kernel build, so that it can be signed with the ephemeral module signing key.

**Testing done:**
Verified that the Neuron module was auto-loaded on `inf1`, `inf2`, and `trn1` instance types, for each of the three kernels (5.10, 5.15, 6.1).
```
bash-5.1# systemctl status modprobe@neuron.service
● modprobe@neuron.service - Load Kernel Module neuron
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/modprobe@.service; static)
    Drop-In: /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/modprobe@neuron.service.d
             └─neuron.conf
     Active: active (exited) since Thu 2024-10-17 19:50:18 UTC; 2min 59s ago
       Docs: man:modprobe(8)
   Main PID: 1066 (code=exited, status=0/SUCCESS)
        CPU: 38ms
 
Oct 17 19:50:18 localhost systemd[1]: Finished Load Kernel Module neuron.
```

On instance types without Neuron hardware, the module load was skipped and subsequent service starts were suppressed:
```
bash-5.1# systemctl status modprobe@neuron.service
○ modprobe@neuron.service - Load Kernel Module neuron
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/modprobe@.service; static)
    Drop-In: /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/modprobe@neuron.service.d
             └─neuron.conf
     Active: inactive (dead) (Result: exec-condition) since Thu 2024-10-17 19:36:23 UTC; 4min 2s ago
  Condition: start condition failed at Thu 2024-10-17 19:36:37 UTC; 3min 48s ago
             └─ ConditionPathExists=!/etc/.neuron-modprobe-done was not met
       Docs: man:modprobe(8)
        CPU: 5ms
 
Oct 17 19:36:23 localhost ghostdog[974]: Did not detect Neuron
Oct 17 19:36:23 localhost systemd[1]: modprobe@neuron.service: Skipped due to 'exec-condition'.
Oct 17 19:36:23 localhost systemd[1]: Condition check resulted in Load Kernel Module neuron being skipped.
Oct 17 19:36:35 i-03852ed99640923e2.us-west-2.compute.internal systemd[1]: Load Kernel Module neuron was skipped because of an unmet condition check (ConditionPathExists=!/etc/.neuron-modprobe-done).
Oct 17 19:36:37 i-03852ed99640923e2.us-west-2.compute.internal systemd[1]: Load Kernel Module neuron was skipped because of an unmet condition check (ConditionPathExists=!/etc/.neuron-modprobe-done).
Notice: journal has been rotated since unit was started, output may be incomplete.
```

**Terms of contribution:**


sources/pciclient/src/private.rs
```diff
@@ -148,6 +158,20 @@ pub(crate) fn check_efa_attachment<T: CommandExecutor>(command_executor: T) -> R
         .any(|device_info| device_info.device.contains(EFA_KEYWORD)))
 }
 
+/// Call `lspci` and check if there is any Neuron device attached.
+/// Internal usage, adding command_executor as parameter allows us to better unit test.
+/// For external usage, check [`crate::is_neuron_attached`].
+pub(crate) fn check_neuron_attachment<T: CommandExecutor>(command_executor: T) -> Result<bool> {
+    let neuron_devices: HashSet<_> = NEURON_DEVICES.iter().map(|s| s.to_string()).collect();
```
ytsssun: Nit: can we use lazy_static and make this HastSet static? Althought, It is likely not much of a performance improvement given the size of the array and the number of times we invoke this.

packages/kernel-6.1/kernel-6.1.spec
```diff
@@ -187,10 +220,18 @@ make -s\\\
...
 
+%if "%{_cross_arch}" == "x86_64"
+%kmake %{?_smp_mflags} M=%{_builddir}/neuron
+%endif
+
 %install
 %kmake %{?_smp_mflags} headers_install
 %kmake %{?_smp_mflags} modules_install
 
+%if "%{_cross_arch}" == "x86_64"
```
ytsssun, line 234: Should we track somewhere when Neuron supports ARM?

ytsssun (approved):

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -42,6 +48,11 @@ Requires: %{_cross_os}microcode-licenses
 Requires: %{name}-modules = %{version}-%{release}
 Requires: %{name}-devel = %{version}-%{release}
 
+# Pull in platform-dependent modules.
+%if "%{_cross_arch}" == "x86_64"
+Requires: (%{name}-modules-neuron if (%{_cross_os}variant-platform(aws) without %{_cross_os}variant-flavor(nvidia)))
```
arnaldo2792, line 53: Should this be more restrictive, or is it OK to get the Neuron kernel module in FIPS variants?

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -237,6 +279,18 @@ rm -f %{buildroot}%{kernel_libdir}/build %{buildroot}%{kernel_libdir}/source
 ln -sf %{_usrsrc}/kernels/%{version} %{buildroot}%{kernel_libdir}/build
 ln -sf %{_usrsrc}/kernels/%{version} %{buildroot}%{kernel_libdir}/source
 
+# Install a copy of System.map so that module dependencies can be regenerated.
+install -p -m 0600 System.map %{buildroot}%{kernel_libdir}
```
ginglis13, line 283: Why is this needed for 5.10 and 5.15, and not 6.1? I don't see anything depmod-specific across our kernels, but may be missing something.

ginglis13: Disregard, I see we already have this installed for 6.1 :)

packages/kernel-6.1/kernel-6.1.spec
bcressey: > Should we track somewhere when Neuron supports ARM?

I don't think so. If EC2 ever launches an instance type that combines Neuron and Graviton, we can remove the conditional.

packages/kernel-5.10/kernel-5.10.spec
bcressey: > Should this be more restrictive, or is it OK to get the Neuron kernel module in FIPS variants?

Unlike NVIDIA, Neuron doesn't provide a PKCS#11 interface to userspace, so it seems fine to include it on FIPS variants.

arnaldo2792 (approved):

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -72,6 +83,18 @@ Summary: Modules for the Linux kernel
 %description modules
 %{summary}.
 
+%if "%{_cross_arch}" == "x86_64"
+%package modules-neuron
+Summary: Modules for the Linux kernel with Neuron hardware
+Requires: %{name}
+Requires: %{_cross_os}ghostdog
+Requires: %{_cross_os}variant-platform(aws)
+Conflicts: %{_cross_os}variant-flavor(nvidia)
```
yeazelm, line 92: Is there an actual conflict with these two or is this just using the `Conflicts` to get the right experience where `-nvidia` variants won't include this driver but the rest will?

sources/pciclient/src/private.rs
```diff
@@ -12,6 +13,15 @@ use crate::{
...
 const AMAZON_VENDOR_CODE: &str = "1d0f";
 const EFA_KEYWORD: &str = "efa";
+const NEURON_DEVICES: [&str; 7] = [
+    "7064", // inf1 device id 0
+    "7065", // inf1 device id 1
+    "7066", // inf1 device id 2
+    "7067", // inf1 device id 3
+    "7164", // trn1 device id 0
+    "7264", // inf2 device id 0
+    "7364", // trn2 device id 0
```
yeazelm: This is unfortunate that we might have to update this list as new devices are launched. I mostly wanted to call it out that we will need to watch for new devices as they launch to enable support here in the future.

yeazelm (approved):

packages/kernel-5.10/kernel-5.10.spec
bcressey: > Is there an actual conflict with these two or is this just using the Conflicts to get the right experience where -nvidia variants won't include this driver but the rest will?

I'm using the `Conflicts` here to indicate a license level conflict; the intent is to enforce that images only end up with one or the other. It's not required to have the right thing happen for the downstream variants in `bottlerocket-os/bottlerocket`.

bcressey:
⬆️ force push:
* rebases for recent kernel package updates
* uses a lazy static hash set per @ytsssun

ytsssun (approved):

arnaldo2792 (approved):

PR #211, opened 2024-10-23
Advisory updates for 3.1.0

KCSesh:
**Description of changes:**
 * Adding in advisories for 3.1.0 release. 
 * Updated 3.1.0 changelog to include the Kubernetes updates. 

**Terms of contribution:**

### DIFF
--- a/advisories/3.1.0/BRSA-0h4z1xmewjhf.toml
+++ b/advisories/3.1.0/BRSA-0h4z1xmewjhf.toml
@@ -1,0 +1,30 @@
+[advisory]
+id = "BRSA-0h4z1xmewjhf"
+title = "kmod-nvidia CVE-2024-0126"
+cve = "CVE-2024-0126"
+severity = "high"
+description = "A flaw was found in NVIDIA GPU software for Windows and Linux which could lead to code execution, denial of service, escalation of privileges, information disclosure, and data tampering"
+
+[[advisory.products]]
+package-name = "kmod-5.10-nvidia"
+patched-version = "1.0.0"
+patched-release = "0"
+patched-epoch = "1"
+
+[[advisory.products]]
+package-name = "kmod-5.15-nvidia"
+patched-version = "1.0.0"
+patched-release = "0"
+patched-epoch = "1"
+
+[[advisory.products]]
+package-name = "kmod-6.1-nvidia"
+patched-version = "1.0.0"
+patched-release = "0"
+patched-epoch = "1"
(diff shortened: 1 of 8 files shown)


advisories/3.1.0/BRSA-0h4z1xmewjhf.toml
```diff
@@ -0,0 +1,30 @@
+[advisory]
+id = "BRSA-0h4z1xmewjhf"
+title = "kmod-nvidia CVE-2024-0126"
+cve = "CVE-2024-0126"
+severity = "high"
+description = "A flaw was found in NVIDIA GPU software for Windows and Linux which could lead to code execution, denial of service, escalation of privileges, information disclosure, and data tampering"
+
+[[advisory.products]]
+package-name = "kmod-5.10-nvidia"
+patched-version = "535.216.01"
```
ginglis13: This applies for each of the products - the `Version` in this package's spec is actually `1.0.0`: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/38b66be93f21c8666fff5ed909605c5d6ffdf940/packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec#L19

ginglis13 (approved):

larvacea (approved):

advisories/3.1.0/BRSA-0h4z1xmewjhf.toml
koooosh: That's correct. To verify, I launched an instance using the latest twoliter and I see this in the app-inv:
```
{
      "Name": "kernel-6.1",
      "Publisher": "Bottlerocket",
      "Version": "6.1.112",
      "Release": "1.1729189882.4798750d.br1",
      "Epoch": "0",
      "InstalledTime": "2024-10-22T23:12:55Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://www.kernel.org/",
      "Summary": "The Linux kernel"
    }
```

This matches what's in [the Version field of the spec file](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/38b66be93f21c8666fff5ed909605c5d6ffdf940/packages/kernel-6.1/kernel-6.1.spec#L4) rather than the actual entire field which includes build id: `6.1.112-122.189`

koooosh (approved):
LGTM, I verified that the `patched-version` now matches the `Version` field in the specs!

PR #213, opened 2024-10-23
pluto: add proxy support for FIPS

jpculp:
**Issue number:**

**Description of changes:**

Replaces `hyper-proxy` with `hyper-http-proxy`.

Adds support for hyper client creation using a custom HTTP connector in `aws-smithy-experimental`.

**Testing done:**

Launched two instances (one built fips and one not) with the following network configuration:
 
```toml
[settings.network]
https-proxy = "<my-proxy>:8888"
```

Checked proxy server logs and found all traffic from my private IP being tunneled, including the EKS API calls for `DescribeCluster`.

In CloudTrail, I could see that the `DescribeCluster` API call source IP was my proxy server:

```
    "eventTime": "2024-10-23T23:15:56Z",
    "eventSource": "eks.amazonaws.com",
    "eventName": "DescribeCluster",
    "awsRegion": "us-west-2",
    "sourceIPAddress": "<my-proxy>",
    "userAgent": "aws-sdk-rust/1.3.3 os/linux lang/rust/1.81.0",
```

**Testing notes:**

If you want to test this on your own, here are a couple of hints:
- I used `tinyproxy` for this PR, but `squidproxy` will also work as I used it for testing https://github.com/bottlerocket-os/bottlerocket/pull/3439.
- In the `tinyproxy` config, uncomment the `LogFile` line and add your EKS cluster's IP range to `Allow`.
- It makes things a little easier to put the proxy in the same VPC/securitygroups as your EKS nodegroups. Then you can set `settings.network.https-proxy` to your EC2 host's private IP DNS name and proxy port.

**Terms of contribution:**


jpculp:
Based on https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/173.

jpculp:
Rebased and ready for review.

sources/api/apiclient/Cargo.toml
```diff
@@ -23,7 +23,7 @@ futures.workspace = true
 futures-channel.workspace = true
 http.workspace = true
 httparse.workspace = true
-hyper = { workspace = true, features = ["client", "http1", "http2", "tcp"] }
+hyper = { version = "0.14", features = ["client", "http1", "http2", "tcp"] }
```
ginglis13, line 27: should we add a comment here as to why we're pinning `hyper` to a different version for apiclient?

ginglis13 (approved):

sources/api/apiclient/Cargo.toml
arnaldo2792: I think we want to stay (at least for now) on hyper 0.14 for `apiclient` because we use it for the `apiclient exec` feature and we don't want to migrate to a newer hyper version for the FIPS support.

sources/api/pluto/src/proxy.rs
```diff
@@ -90,11 +91,19 @@ where
         ));
     }
 
-    let https_connector = HttpsConnectorBuilder::new()
-        .with_native_roots()
+    let mut base_connector = HttpConnector::new_with_resolver(GaiResolver::new());
+    base_connector.enforce_http(false);
```
bcressey: I think it would be cleaner to put the hyper proxy dependency into the smithy client, rather than duplicating this code or exposing more of its methods:
https://gist.github.com/bcressey/d52538b4f7cb13dcbbdf984a0c0f713e

Apart from that, my only misgiving is where to put the `proxy::setup_proxy` interface so that both `pluto` and `cfsignal` can access it.

bcressey: > Apart from that, my only misgiving is where to put the proxy::setup_proxy interface so that both pluto and cfsignal can access it.

Maybe add this as a separate module to the smithy client fork for now, since it has the right dependencies already.

jpculp:
Moved proxy logic from `pluto::proxy` to `aws-smithy-experimental::hyper_1_0`.

bcressey (approved):

ginglis13 (approved):

sources/aws-smithy-experimental/src/hyper_1_0.rs
```diff
@@ -209,6 +213,60 @@ mod build_connector {
...
+        if proxy_uri.scheme().is_none() {
+            proxy_uri = format!("http://{}", https_proxy)
+                .parse::<Uri>()
+                .expect("Unable to parse proxy URI as HTTPS");
+        }
+        let mut proxy = Proxy::new(intercept, proxy_uri);
+        // Parse https_proxy as URL to extract out auth information if any
+        let proxy_url = Url::parse(https_proxy).expect("Unable to parse HTTPS proxy as URL");
+
+        if !proxy_url.username().is_empty() || proxy_url.password().is_some() {
```
ginglis13, line 261: nit: I was thinking this might read more clearly if it were
```suggestion
        if !(proxy_url.username().is_empty() && proxy_url.password().is_none()) {
```

since the only case where we don't set auth is when both username and password aren't provided

COPYRIGHT
```diff
@@ -14,12 +14,6 @@ operating system images.
 
 =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
-Contains modified hyper-proxy files [mod.rs, stream.rs, tunnel.rs] from
-https://github.com/tafia/hyper-proxy 2021-09-20.
-Copyright (c) 2017 Johann Tuffe. Licensed under the MIT License.
-
-=^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
-
 Contains aws-smithy-experimental from
```
arnaldo2792: Don't we need to add the list of files we modified?

jpculp:
- Added list of modified `.rs` files to `COPYRIGHT`.
- Rebased.

sources/aws-smithy-experimental/src/hyper_1_0.rs
jpculp: This seems fine, but since this has been working as is for a bit I'd like to keep this particular PR to just the necessities.

bcressey (approved):

ginglis13 (approved):

arnaldo2792 (approved):

PR #221, opened 2024-10-25
Update list of required packages to build Bottlerocket

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**

Add `clang` to the list of packages required to build Bottlerocket crates

**Testing done:**

:grimacing: 

**Terms of contribution:**

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -13,11 +13,11 @@ The build system requires certain operating system packages to be installed.
 Ensure the following packages are installed:
 ##### Ubuntu
 ```shell
-apt install build-essential openssl libssl-dev pkg-config liblz4-tool
+apt install build-essential openssl libssl-dev pkg-config liblz4-tool cmake clang libclang-dev
 ```
 ##### Fedora
 ```shell
-yum install make automake gcc openssl openssl-devel pkg-config lz4 perl-FindBin perl-lib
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -13,11 +13,11 @@ The build system requires certain operating system packages to be installed.
 Ensure the following packages are installed:
 ##### Ubuntu
 ```shell
-apt install build-essential openssl libssl-dev pkg-config liblz4-tool
+apt install build-essential openssl libssl-dev pkg-config liblz4-tool clang
 ```
 ##### Fedora
 ```shell
-yum install make automake gcc openssl openssl-devel pkg-config lz4 perl-FindBin perl-lib
+dnf install make automake gcc openssl openssl-devel pkg-config lz4 perl-FindBin perl-lib clang-tools-extra
```
piyush-jena: It seems like we need `clang` here too. :(

larvacea (changes requested):
If we want to do `cargo build` then `cmake` should be in the `dnf install` list.

BUILDING.md
ytsssun: I had to install `cmake`, `clang` and `clang-tools-extra`. Can you include three of them? 

This [BUILDING.md](https://github.com/bottlerocket-os/bottlerocket/blob/develop/BUILDING.md) could use an update as well.

ytsssun (approved):

vigh-m (approved):

larvacea (approved):

PR #224, opened 2024-10-28
switch to AL23 as upstream for libkcapi

bcressey:
**Issue number:**

**Description of changes:**
This aligns libkcapi with the version specified in [the FIPS 140-3 security policy](https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp4808.pdf) for the Amazon Linux 2023 Kernel Cryptographic API.

Bump the epoch to ensure that the "older" 1.4.0 version is preferred over the "newer" 1.5.0 version from past core kit releases.

Trim the set of installed files down to just the ones referenced by the security policy: `sha512hmac` and `libkcapi.so.1.4.0`.

**Testing done:**
Enabled the FIPS feature for `aws-dev` and verified that the kernel integrity check still works:
```
bash-5.2# cd /boot
bash-5.2# sha512hmac -c .vmlinuz.hmac
vmlinuz: OK
```

This operation implicitly checks the SHA-512 HMAC of `/usr/bin/sha512hmac` and the SHA-256 HMAC of `/usr/lib64/libkcapi.so.1.4.0`.

**Terms of contribution:**

### DIFF
--- a/packages/libkcapi/libkcapi.spec
+++ b/packages/libkcapi/libkcapi.spec
@@ -1,7 +1,6 @@
-# libkcapi since 85bce6035b (1.5.0) uses sha512hmac with the same key for all
-# self-checks. Earlier versions used sha256hmac with a different key to check
-# the shared library.
+# Helper functions that use OpenSSL to compute an HMAC with specified keys.
 %global openssl_sha512_hmac openssl sha512 -hmac FIPS-FTW-RHT2009 -hex
+%global openssl_sha256_hmac openssl sha256 -hmac orboDeJITITejsirpADONivirpUkvarP -hex
 
 # We need to compute the HMAC after the binaries have been stripped.
 %define __spec_install_post\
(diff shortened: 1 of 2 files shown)


packages/libkcapi/libkcapi.spec
```diff
@@ -1,31 +1,29 @@
-# libkcapi since 85bce6035b (1.5.0) uses sha512hmac with the same key for all
-# self-checks. Earlier versions used sha256hmac with a different key to check
-# the shared library.
+# Helper functions that use OpenSSL to compute an HMAC with specified keys.
 %global openssl_sha512_hmac openssl sha512 -hmac FIPS-FTW-RHT2009 -hex
+%global openssl_sha256_hmac openssl sha256 -hmac orboDeJITITejsirpADONivirpUkvarP -hex
```
arnaldo2792, line 3: Just for my own understanding, why this key?

arnaldo2792 (approved):

ginglis13 (approved):

packages/libkcapi/libkcapi.spec
bcressey: > Just for my own understanding, why this key?

Not sure of the origin; it's [the well-known key](https://github.com/search?q=orboDeJITITejsirpADONivirpUkvarP&type=code&p=1) for this.

PR #228, opened 2024-10-30
nvidia-k8s-device-plugin: wait for kubelet.sock before starting

### LINKED ISSUE #4250: Bottlerocket 1.25 is preventing nvidia driver to start
### Description
We have been using Bottlerocket 1.25 for the past 3 days. Since the upgrade, some of our GPU nodes (`g5.xlarge`) are failing to initialize the NVIDIA driver on Kubernetes 1.30. The affected image is `bottlerocket-aws-k8s-1.30-nvidia-x86_64-v1.25.0-388e1050`.

### Issue
When the driver fails to initialize, the node does not advertise its `nvidia.com/gpu` capacity, causing the corresponding pods to remain in a Pending state indefinitely. This prevents any pods scheduled on the affected node from starting. 
The kubelet log on the admin container of an affected node contains the following error:
  
  ```
  Oct 17 12:05:33 ip-xx-xx-xx-xx.ec2.internal kubelet[1655]: E1017 12:05:33.335816    1655 client.go:90] "ListAndWatch ended unexpectedly for device plugin" err="rpc error: code = Unavailable desc = error reading from server: EOF" resource="nvidia.com/gpu"
  ```

### Additional Information
- We are using Karpenter to provision our nodes, but I don't think the issue is related to it. The NodeClaim remain in an `Unknown` state indefinitely and shows the following condition message:
```
    Message:               Resource "nvidia.com/gpu" was requested but not registered
```
- The issue occurs on approximately 1 out of every 6 nodes, suggesting a potential race condition during node startup that interferes with NVIDIA driver initialization.
- This problem was not observed with Bottlerocket 1.24.

### Expected Behavior
The NVIDIA driver should initialize correctly on all GPU nodes, and the node should advertise its GPU capacity to ensure proper scheduling.

### Steps to Reproduce
1. Deploy Bottlerocket 1.25 on 6 (or more) `g5.xlarge` instances in a Kubernetes 1.30 cluster.
2. Observe the node initialization process and monitor whether the NVIDIA driver is initialized correctly.
3. Check the node’s GPU capacity and NodeClaim status.

cbgbt:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket/issues/4250

**Description of changes:**
It is possible for `nvidia-k8s-device-plugin` and `kubelet` to race, causing graphics nodes to fail to expose gpus via kubelet.

Specifically, `nvidia-k8s-device-plugin` starts after `kubelet`; however, it depends on the `kubelet` device-plugin management socket to be available in order to register itself.

The `kubelet` service does not synchronize its start of the device-plugin management socket with its systemd "notify" signal, which means that `kubelet` may start before the socket is ready.

If the socket is created *after* `nvidia-k8s-device-pluging` starts watching the socket for inotify events, it may trigger the device-plugin's restart logic (the device plugin assumes that kubelet has restarted in this case).

Unfortunately, device plugin restarts seem to be somewhat flaky due to issues discussed in https://github.com/bottlerocket-os/bottlerocket/issues/4250.

This change causes the `nvidia-k8s-device-plugin` to require `kubelet.sock` to exist as a socket. The unit will fail to start, and subsequently retry every 2 seconds until the socket is available. We perform an initial sleep, because it turns out that `kubelet.sock` usually does not exist by the time that systemd tries to start `nvidia-k8s-device-plugin`.

**Testing done:**
I created [this patch](https://github.com/cbgbt/bottlerocket-core-kit/commit/88aab0868b018a186c6bd691b5ba75e427ab85fd) which causes the inotify race to always occur, which massively increased the incidence of the failure case.

After hundreds of instance launches, I have not witnessed a single instance with missing GPU resources (whereas the failure incidence is ~40% on Bottlerocket 1.25.0 with my faulty patch added).

* [x] basic node readiness tests
* [x] cycle over 1000 instance launches without triggering the bug

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
@@ -4,6 +4,10 @@ After=kubelet.service
 Wants=kubelet.service
 
 [Service]
+# Ensure that the kubelet device plugin socket exists before we start
+# A brief sleep is needed to avoid the `test` failing its first check
+ExecStartPre=/usr/bin/sleep 0.1
+ExecStartPre=/usr/bin/test -S /var/lib/kubelet/device-plugins/kubelet.sock
 ExecStart=/usr/bin/nvidia-device-plugin --device-list-strategy volume-mounts --device-id-strategy index --pass-device-specs=true
 Type=simple
 TimeoutSec=0


yeazelm (approved):

bcressey (approved):

zestrells:
We just encountered the same issue on the newly released AMI. Do we know if this 100% fixed the issue, or just reduced the times it occurs? 

`osImage: Bottlerocket OS 1.26.2 (aws-k8s-1.29-nvidia)`

PR #229, opened 2024-10-31
Clarify license for open gpu driver

yeazelm:
**Description of changes:**

The license was slight vague for the Open GPU driver in the kmod-*-nvidia packages. This adds punctuation to clarify the order.

**Testing done:**
Built the kit. No functional change.

**Terms of contribution:**

### DIFF
--- a/packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
+++ b/packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
@@ -71,7 +71,7 @@ Requires: %{name}-tesla(fabricmanager)
 %package open-gpu-%{tesla_major}
 Summary: NVIDIA %{tesla_major} Open GPU driver
 Version: %{tesla_ver}
-License: MIT OR GPL-2.0-only
+License: MIT AND GPL-2.0-only
 Requires: %{_cross_os}variant-platform(aws)
 
 %description open-gpu-%{tesla_major}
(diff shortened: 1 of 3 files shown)


packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -71,7 +71,7 @@ Requires: %{name}-tesla(fabricmanager)
 %package open-gpu-%{tesla_major}
 Summary: NVIDIA %{tesla_major} Open GPU driver
 Version: %{tesla_ver}
-License: MIT OR GPL-2.0-only
+License: MIT OR (MIT AND GPLv2)
```
bcressey: I don't think the MIT part is accurate given this statement in the license:
> However, when linked together to form a Linux kernel module, the resulting Linux kernel module is dual licensed as MIT/GPLv2.

We are definitely linking it together to form a Linux kernel module here, and the license should reflect that only the dual licensed form applies.

bcressey: `GPLv2` is also not an SPDX identifier, you have to pick one of `GPL-2.0-or-later` or `GPL-2.0-only`.

`GPL-2.0-or-later` is what the open driver license says but in practice - given that it's being linked to the Linux kernel - `GPL-2.0-only` is what you end up with.

So I would change this to `MIT AND GPL-2.0-only`.

yeazelm:
^ Updated from comments

bcressey (approved):

cbgbt (approved):

PR #251, opened 2024-11-09
fix: changed the default value of nrext64 to 0 to avoid unstable feature of nrext64

Sparksssj:
**Issue number:4281**

Closes #

**Description of changes:**
Set nrext64 default value to 0 in xfsprogs to ensure XFS filesystem compatibility. 
For kernel version lower than 6.1:

it avoids the error 
which reason is nrext64 is not supported by older kernel version
For kernel version newer than or equals to 6.1:

This change prevents the "EXPERIMENTAL Large extent counts feature in use. Use at your own risk!" warning.

**Testing done:**
Launched instances on kernel version 5.10, 5.15, 6.1 and tested as mentioned in the issue, no problem now and default value is 0.

**Terms of contribution:**


bcressey:
I really don't want to patch xfsprogs for this. We should use this feature on kernels where it's supported, and not use it on kernels where it's not.

Please look into ways to influence the config file used by `mkfs.xfs` by default. Then we can package different defaults in the different kernel packages.

Sparksssj:
OK I can try another approach. I noticed there are several configuration files in `xfsprogs/mkfs` directory, but they are not used by default. What do you think if we edit our call in `ephemeral_storage` to make config file mandatory and apply automatically. Then we maybe make a patch for every version of kernel to set the default value of nrext64.

bcressey:
> What do you think if we edit our call in ephemeral_storage to make config file mandatory and apply automatically.

Sounds like the right direction.

I would suggest combining the Debian approach described [here](https://lwn.net/Articles/924480/) - a different config file for each kernel, along with a tmpfiles.conf snippet that sets up a `/etc/mkfs.xfs.conf` symlink, packaged with that kernel - with a code change on the `apiserver` side to pass `/etc/mkfs.xfs.conf` as the config.

cbgbt:
One of the motivators for patching `xfsprogs` itself here is that calls to `mkfs.xfs` outside of the API could result in unusable filesystems with the default settings. Even with a usable filesystem on 6.1, the kernel emits a message warning that the FS is using experimental features.

~Likewise, the default config shipped by `mkfs.xfs` for kernel 6.6 enables `nrext64`; however, that kernel also emits a warning that the feature is experimental when you mount the filesystem. We may wish to disable it on 6.6 even though it is enabled by default in the associated xfsprogs package.~

Perhaps getting "correct" behavior from the API is enough, since that's the supported interface on Bottlerocket anyways.

Sparksssj:
Accidentally deleted first commit, and made commit tree irreversible. Have to create a new PR

PR #252, opened 2024-11-11
Enable rdma core in release

ytsssun:
**Issue number:**

**Description of changes:**
- Add `rdma-core` to all "aws-*" variants.
- Fixed a typo in kernel-6.1 where the `%{name}-bootconfig-vmware` was loaded for metal variant.

**Testing done:**
Test building an (aws-k8s-1.27-nvidia) variant with this change in [BoB repo](https://github.com/bottlerocket-os/bottlerocket/pull/4290), consuming from a personal kit.

Also verified that the EFA drivers are available in the 5.15 and 5.10 kernels, so it checks the compatibility box of the rdma-core package.

RDMA helpers are working as expected.

**Terms of contribution:**

### DIFF
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -129,6 +129,7 @@ Requires: %{_cross_os}os
 Requires: %{_cross_os}pciutils
 Requires: %{_cross_os}policycoreutils
 Requires: %{_cross_os}procps
+Requires: (%{_cross_os}rdma-core if %{_cross_os}variant-platform(aws))
 Requires: %{_cross_os}selinux-policy
 Requires: %{_cross_os}shim
 Requires: %{_cross_os}systemd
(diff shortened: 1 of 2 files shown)


bcressey (approved):

larvacea (approved):

koooosh (approved):

packages/release/release.spec
```diff
@@ -129,6 +129,7 @@ Requires: %{_cross_os}os
 Requires: %{_cross_os}pciutils
 Requires: %{_cross_os}policycoreutils
 Requires: %{_cross_os}procps
+Requires: (%{_cross_os}rdma-core if %{_cross_os}variant-platform(aws))
```
arnaldo2792, line 132: nit: for consistency, you could move the conditionals to the end

arnaldo2792 (approved):

packages/release/release.spec
ytsssun: I have considered that too. But I added rdma-core package here to follow lexical ordering.

PR #254, opened 2024-11-11
packages: update libnvidia-container and nvidia-container-toolkit to v1.17.1

piyush-jena:
**Testing done:**
* Ensured that GPU resources were exposed to k8s pods

**Terms of contribution:**

### DIFF
--- a/Twoliter.toml
+++ b/Twoliter.toml
@@ -1,5 +1,5 @@
 schema-version = 1
-release-version = "3.2.0"
+release-version = "3.2.1"
 
 [vendor.bottlerocket]
 registry = "public.ecr.aws/bottlerocket"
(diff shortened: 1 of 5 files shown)


cbgbt (commented):
Looks good to me, assuming we get the contents in the lookaside cache!

Twoliter.toml
```diff
@@ -1,5 +1,5 @@
 schema-version = 1
-release-version = "3.2.0"
+release-version = "3.2.1"
```
koooosh, line 2: Do we want this version bump in this PR?

cbgbt: Nice catch. We at least want this as a separate commit, even if it lives in this PR.

piyush-jena: Thank you! Its updated now.

ytsssun (approved):
LGTM, assuming the build passes.

cbgbt (approved):

PR #257, opened 2024-11-13
fix: changed the default value of nrext64 to 0 to avoid unstable feature of nrext64

### LINKED ISSUE #4281: XFS Ephemeral Storage fails to mount with: Superblock has unknown incompatible features (0x20) enabled.
**Image I'm using:** `bottlerocket-aws-k8s-1.25-x86_64-v1.26.1-943d9a41`
This is on a `i3en.2xlarge` node with local NVMe flash I'm hoping to use as Ephemeral storage. 
If I set the type to ext4 it works.

**What I expected to happen:** The local flash storage would be formatted and mounted as XFS storage

**What actually happened:** The storage isn't mounted. 

I'm guessing because the xfs tools are newer than the kernel?

Error in dmesg:

```
```

Put this in the config:
```
    commands = [
      ["apiclient", "ephemeral-storage", "init", "-t", "xfs"],
      ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/var/lib/kubelet"]
    ] 
    essential = true
    mode = "always"
```

Sparksssj:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket/issues/4281

**Description of changes:**
A new PR for https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/251, see discussion there.

**Testing done:**

**Terms of contribution:**


Sparksssj:
@bcressey Is this logic correct this time?

packages/kernel-6.1/lts_6.1.conf
```diff
@@ -0,0 +1,14 @@
+# V5 features that were the mkfs defaults when the upstream Linux 6.1 LTS
```
yeazelm, line 1: This will solve the issue for variants that don't use XFS for the BOTTLEROCKET-DATA partition where the only XFS usage is through the ephemeral storage calls to create the filesystem. We use [systemd-mkfs](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/2a673d64daaf19bee95d8565cbbb09d8815588cc/packages/release/prepare-local-fs.service#L15) which doesn't appear to take arguments besides the two we provide. We might need to see if there is a way to provide this configuration file to systemd-makefs or else variants using XFS for the data partition will continue to have this error.

Sparksssj: Well. I'm not so sure if we can use a configuration file for systemd-makefs. But if we cannot, the situation just goes back to what we discussed initially: we may not want to just set the default value for nrext64 to 0 for all kernel versions. But this indeed need more discussions.

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -16,6 +16,9 @@ Source100: config-bottlerocket
 Source220: neuron-sysinit.target.drop-in.conf
 Source221: modprobe@neuron.service.drop-in.conf
 
+# XFS configuration
+Source300: lts_5.10.conf
```
arnaldo2792: You should name this file `mkfs.xfs.conf` since the current name right now doesn't tell us much about what it is or what's it used for. The same for the other kernels.

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -83,6 +86,12 @@ Summary: Modules for the Linux kernel
 %description modules
 %{summary}.
 
+%package mkfs-confs
+Summary: mkfs confiigurations for different file systems
```
arnaldo2792:
```suggestion
Summary: mkfs configurations for different file systems
```

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,46 @@
+From 46605e1f1d67bf2344ffab9447e763ebc13f6387 Mon Sep 17 00:00:00 2001
+From: Sparks Song <shijiao@amazon.com>
+Date: Thu, 14 Nov 2024 00:17:10 +0000
+Subject: [PATCH] mkfs: source defaults from config file
+
```
yeazelm, line 6: We should have a commit message that explains what you are trying to do with a bit more explanation on why we are adding it.

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,46 @@
...
+diff --git a/mkfs/xfs_mkfs.c b/mkfs/xfs_mkfs.c
+index 6d2469c..bc76b80 100644
+--- a/mkfs/xfs_mkfs.c
++++ b/mkfs/xfs_mkfs.c
+@@ -14,6 +14,7 @@
+ #include "libfrog/dahashselftest.h"
+ #include "proto.h"
+ #include <ini.h>
++#include <unistd.h>
```
yeazelm: I'd really like to avoid adding a dependency in this patch. Can we see if <ini.h> has a function we can use for this purpose of checking if the file exists? We should consider updating `cfgfile_parse()` or adding a new, specialized function that safely checks if the file is present and moves on if not present instead of adding our own file checking logic.

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,46 @@
...
+ 
+ #define TERABYTES(count, blog)	((uint64_t)(count) << (40 - (blog)))
+ #define GIGABYTES(count, blog)	((uint64_t)(count) << (30 - (blog)))
+@@ -4369,6 +4370,21 @@ main(
+ 	 * printf(_("Default configuration sourced from %s\n"), dft.source);
+ 	 */
+ 
++	char *default_config_path = "/usr/share/xfs/mkfs.xfs.conf";
```
yeazelm: The more idiomatic way to do this would be to use `#define` to set a constant like this. It helps signal to the reader this never changes.

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,46 @@
...
++	char *original_cfgfile = cli.cfgfile;  // Save the original value (from -c if specified)
```
yeazelm: This isn't the only thing overwritten by reading in a configuration file. If a user specifies other options, won't this logic smash over that too?

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,46 @@
...
++
++	// First try to load the default config if it exists
++	if (access(default_config_path, R_OK) == 0) {
```
yeazelm: See my above comment but I'd rather stay consistent with the codebase to use the libinih library for this check.

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,46 @@
...
++		cli.cfgfile = default_config_path;
++		cfgfile_parse(&cli);
++		printf(_("Default configuration sourced from %s\n"), default_config_path);
++	} else {
++		fprintf(stderr, _("Warning: Default config file %s not found or not readable\n"), default_config_path);
```
yeazelm: I don't think we should print this. We are adding in this logic and the way we know it is working is because it prints the other message, the absence of this line would indicate we didn't use a default file (and is the expected default behavior)

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -326,6 +327,12 @@ pub fn format_device<S: AsRef<OsStr>>(device: S, format: &Filesystem) -> Result<
     mkfs.arg("-L");
     mkfs.arg(RAID_DEVICE_NAME);
 
+    // Apply XFS config file if formatting with XFS
+    if matches!(format, Filesystem::Xfs) {
+        mkfs.arg("-c");
+        mkfs.arg(format!("options={}", XFS_MKFS_CONF));
```
yeazelm, line 333: With the change to `xfsprogs`, do we need this now?

packages/kernel-6.1/kernel-6.1.spec
```diff
@@ -139,6 +142,12 @@ Summary: Modules for the Linux kernel on bare metal
 %description modules-metal
 %{summary}.
 
+%package mkfs-confs
+Summary: mkfs confiigurations for different file systems
+
+%description mkfs-confs
```
yeazelm, line 147: nit:
```suggestion
%package xfs-mkfs-confs
```

packages/os/os.spec
```diff
@@ -114,6 +114,7 @@ Requires: (%{_cross_os}driverdog if %{_cross_os}variant-flavor(nvidia))
 %package -n %{_cross_os}apiserver
 Summary: Bottlerocket API server
 Requires: %{_cross_os}settings-plugins
+Requires: %{_cross_datadir}/xfs/mkfs.xfs.conf
```
yeazelm, line 117: I think we might make this a dependency of xfsprogs rather than in os.

packages/kernel-6.1/kernel-6.1.spec
Sparksssj: Why would we change this? I see all other packages have a description there

arnaldo2792: I think it is a type, but rather the suggestion is to change the subpackage name to be `xfs`. I suggested the more generic `mkfs-confgs` since we could use the package to include more configurations for other filesystems if needed.

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
Sparksssj: Then I will remove this line, and only print a warning when default file found but not readable

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
Sparksssj: I took another look at the code and realized the user specific config file is parsed actually after this code change, which starts at line 4391: 	while ((c = getopt_long(argc, argv, "b:c:d:i:l:L:m:n:KNp:qr:s:CfV",
So in this case we don't need to save the original value at all because it's empty at the moment. We can just simply apply the default file here, and safely overwrite it with user specified config file if there is one using this line: 	cfgfile_parse(&cli);

packages/kernel-6.1/kernel-6.1.spec
yeazelm: The diff wasn't intended to be the description, sorry about that confusion. I see the intention is to make this package generic. I'm fine with this but then we should probably call out that in a comment that this package is intended to hold future other configuration files as well, rather than specifically grabbing the package for a particular filesystem. My mental model was that we would only include these files for supported filesystems, so we would group them for each one as we add them, but this works as well.

PR #258, opened 2024-11-13
feat: add nvidia MIG

piyush-jena:
**Issue number:**

Related: 
* https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/63
* https://github.com/bottlerocket-os/bottlerocket/issues/4252

**Description of changes:**
Adding nvidia-migmanager service and binary that configures the instance with nvidia mig.  

**Testing done:**

1. Instance joined the cluster

```
NAME                                           STATUS   ROLES    AGE   VERSION
ip-XXXX.us-west-2.compute.internal   Ready    <none>   15h   v1.29.5-eks-1109419
```

2. Model Default:

```
bash-5.1# apiclient get settings.kubelet-device-plugin
{
  "settings": {
    "kubelet-device-plugins": {
      "nvidia": {
        "device-id-strategy": "index",
        "device-list-strategy": "volume-mounts",
        "device-partitioning-strategy": "none",
        "device-sharing-strategy": "none",
        "pass-device-specs": true
      }
    }
  }
}
```

3. Model Updates:
```
bash-5.1#: apiclient set settings.kubelet-device-plugins.nvidia.device-partitioning-strategy="mig"
bash-5.1#: apiclient set settings.kubelet-device-plugins.nvidia.mig.profile."a100-40gb"="1g.5gb"
bash-5.1# apiclient get settings.kubelet-device-plugin
{
  "settings": {
    "kubelet-device-plugins": {
      "nvidia": {
        "device-id-strategy": "index",
        "device-list-strategy": "volume-mounts",
        "device-partitioning-strategy": "mig",
        "device-sharing-strategy": "none",
        "mig": {
          "profile": {
            "a100-40gb": "1g.5gb"
          }
        },
        "pass-device-specs": true
      }
    }
  }
}
```

`kubectl describe node` shows 56 gpus post instance reboot.

4. Bounded check:

```
bash-5.1# apiclient apply <<EOF
> [settings.kubelet-device-plugins.nvidia.mig.profile]
> "hello"="1g.5gb"
> EOF
Failed to apply settings: Failed to PATCH settings from '-' to '/settings?tx=apiclient-apply-7NsnlaurtHEacSYL': Status 400 when PATCHing /settings?tx=apiclient-apply-7NsnlaurtHEacSYL: Json deserialize error: Unable to deserialize into NvidiaGPUModel: NVIDIA GPU Model must match '^([a-z])(\d+)\.(\d+)gb$', given: hello at line 1 column 62
bash-5.1# apiclient apply <<EOF
> [settings.kubelet-device-plugins.nvidia.mig.profile]
> "a100.40gb"="2"
> EOF
bash-5.1# apiclient apply <<EOF
> [settings.kubelet-device-plugins.nvidia.mig.profile]
> "a100.40gb"="5"
> EOF
Failed to apply settings: Failed to PATCH settings from '-' to '/settings?tx=apiclient-apply-GzUHB0axGlWNPzGw': Status 400 when PATCHing /settings?tx=apiclient-apply-GzUHB0axGlWNPzGw: Json deserialize error: Unable to deserialize into MIGProfile: MIG Profile must match '^[0-9]g\.\d+gb$', given: 5 at line 1 column 71
```

5. Files generated:

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
@@ -5,7 +5,11 @@ std = { version = "v1", helpers = ["default"] }
 +++
 version: v1
 flags:
+  {{#if (eq settings.kubelet-device-plugins.nvidia.device-partitioning-strategy "mig")}}
+  migStrategy: "single"
+  {{else}}
   migStrategy: "none"
+  {{/if}}
   failOnInitError: true
   plugin:
     passDeviceSpecs: {{default true settings.kubelet-device-plugins.nvidia.pass-device-specs}}
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-mig-conf
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-mig-conf
@@ -1,0 +1,10 @@
+[required-extensions]
+kubelet-device-plugins = "v1"
+std = { version = "v1", helpers = ["if_not_null", "toml_encode"]}
++++
+{{#if_not_null settings.kubelet-device-plugins.nvidia.device-partitioning-strategy}}
+device-partitioning-strategy = "{{{settings.kubelet-device-plugins.nvidia.device-partitioning-strategy}}}"
+{{/if_not_null}}
+{{#if_not_null settings.kubelet-device-plugins.nvidia.mig.profile}}
+profile = {{ toml_encode settings.kubelet-device-plugins.nvidia.mig.profile }}
+{{/if_not_null}}
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
@@ -16,6 +16,7 @@ Source0: https://%{goimport}/archive/v%{gover}/v%{gover}.tar.gz#/k8s-device-plug
 Source1: nvidia-k8s-device-plugin.service
 Source2: nvidia-k8s-device-plugin-conf
 Source3: nvidia-k8s-device-plugin-exec-start-conf
+Source4: nvidia-k8s-device-plugin-mig-conf
 
 
 BuildRequires: %{_cross_os}glibc-devel
@@ -69,6 +70,7 @@ install -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}
 install -d %{buildroot}%{_cross_unitdir}/nvidia-k8s-device-plugin.service.d
 install -D -m 0644 %{S:2} %{buildroot}%{_cross_templatedir}/nvidia-k8s-device-plugin-conf
 install -D -m 0644 %{S:3} %{buildroot}%{_cross_templatedir}/nvidia-k8s-device-plugin-exec-start-conf
+install -D -m 0644 %{S:4} %{buildroot}%{_cross_templatedir}/nvidia-k8s-device-plugin-mig-conf
 
 
 %files
@@ -78,6 +80,7 @@ install -D -m 0644 %{S:3} %{buildroot}%{_cross_templatedir}/nvidia-k8s-device-pl
 %dir %{_cross_unitdir}/nvidia-k8s-device-plugin.service.d
 %{_cross_templatedir}/nvidia-k8s-device-plugin-conf
 %{_cross_templatedir}/nvidia-k8s-device-plugin-exec-start-conf
+%{_cross_templatedir}/nvidia-k8s-device-plugin-mig-conf
 
 %files bin
 %{_cross_bindir}/nvidia-device-plugin
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "bottlerocket-model-derive"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "darling 0.20.10",
  "quote",
@@ -1255,8 +1255,8 @@ dependencies = [
 
 [[package]]
 name = "bottlerocket-modeled-types"
-version = "0.7.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+version = "0.8.0"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "base64 0.22.1",
  "bottlerocket-model-derive",
@@ -1291,7 +1291,7 @@ dependencies = [
 [[package]]
 name = "bottlerocket-scalar"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "serde",
  "serde_plain",
@@ -1300,7 +1300,7 @@ dependencies = [
 [[package]]
 name = "bottlerocket-scalar-derive"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-scalar",
  "darling 0.20.10",
@@ -1324,8 +1324,8 @@ dependencies = [
 
 [[package]]
 name = "bottlerocket-settings-models"
-version = "0.7.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+version = "0.8.0"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -1375,7 +1375,7 @@ dependencies = [
 [[package]]
 name = "bottlerocket-settings-sdk"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "argh",
  "bottlerocket-template-helper",
@@ -1388,7 +1388,7 @@ dependencies = [
 [[package]]
 name = "bottlerocket-string-impls-for"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "serde",
 ]
@@ -1396,7 +1396,7 @@ dependencies = [
 [[package]]
 name = "bottlerocket-template-helper"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "darling 0.20.10",
  "proc-macro2",
@@ -4559,7 +4559,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-autoscaling"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4572,7 +4572,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-aws"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4585,7 +4585,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-bootstrap-commands"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4599,7 +4599,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-bootstrap-containers"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4612,7 +4612,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-cloudformation"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4625,7 +4625,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-container-registry"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4638,7 +4638,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-container-runtime"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4651,7 +4651,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-dns"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4664,7 +4664,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-ecs"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4677,7 +4677,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-host-containers"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4690,7 +4690,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-kernel"
 version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4702,8 +4702,8 @@ dependencies = [
 
 [[package]]
 name = "settings-extension-kubelet-device-plugins"
-version = "0.1.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+version = "0.2.0"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
@@ -4716,7 +4716,7 @@ dependencies = [
 [[package]]
 name = "settings-extension-kubernetes"
 version = "0.2.0"
-source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.7.0#2a7c0986846eb98122e6812634d0526988a72c64"
+source = "git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-settings-models-v0.8.0#c9e5f989c24bea2006527c6f1a4708225ab70cc4"
 dependencies = [
  "bottlerocket-model-derive",
  "bottlerocket-modeled-types",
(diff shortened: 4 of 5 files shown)


packages/systemd/systemd-tmpfiles.conf
```diff
@@ -1,5 +1,6 @@
 d /run/cache 0755 root root -
 d /run/lock 0755 root root -
+d /run/prairiedog 0755 root root -
```
bcressey: The `systemd` package's tmpfiles snippet isn't the right place for this.

I'm also not convinced `migmanager` should be using `prairiedog`'s temp directory, or that we want to route it all through `prairiedog`.

We could just add another program to `reboot-if-required.service` (like this, or via drop-in):
```
[Service]
Type=oneshot
ExecStart=/usr/bin/prairiedog reboot-if-required
ExecStart=-/usr/bin/nvidia-migmanager reboot-if-required
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+
+fn stop_fabric_manager(mig_settings: NvidiaMIGConfig, gpu_model: NvidiaGPU) -> Result<()> {
+    ensure!(
+        !(mig_settings.device_partitioning_strategy == NvidiaDevicePartitioningStrategy::MIG
+            && (gpu_model == NvidiaGPU::A100_40GB || gpu_model == NvidiaGPU::A100_80GB)),
+        error::FabricManagerSnafu {}
+    );
+
+    Ok(())
+}
```
bcressey: This should have a comment to indicate why it "works", since `ensure!` just looks like error-handling.

Instead I'd log a warning and use `process::exit` to exit non-zero.

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -13,6 +13,7 @@ use std::{fs, io};
 // Boot config related consts
 const BOOTCONFIG_INITRD_PATH: &str = "/var/lib/bottlerocket/bootconfig.data";
 const PROC_BOOTCONFIG: &str = "/proc/bootconfig";
+const REBOOT_REQD_MARKER_DIR: &str = "/run/prairiedog";
```
bcressey:
```suggestion
const REBOOT_REQUIRED_MARKER_DIR: &str = "/run/prairiedog";
```

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -295,6 +296,30 @@ fn boot_config_to_boot_settings_json(bootconfig_str: &str) -> Result<String> {
     serde_json::to_string(&boot_settings).context(error::OutputJsonSnafu)
 }
 
+/// Given a path, check if the directory or its subdirectories contains files
+fn contains_files(dir: &Path) -> bool {
+    let mut ans: bool = false;
```
bcressey: Do you actually need to walk the full directory? You could look for a specific marker file like `/run/prairiedog/reboot-required`.

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -307,12 +332,16 @@ where
 
     let new_boot_settings = get_boot_config_settings(config_path)?.unwrap_or(DEFAULT_BOOT_SETTINGS);
 
-    let reboot_required = if new_boot_settings.reboot_to_reconcile.unwrap_or(false) {
+    let mut reboot_required = if new_boot_settings.reboot_to_reconcile.unwrap_or(false) {
```
bcressey: I'd prefer to leave this as a bootconfig-specific function, and update the caller (`reboot_if_required` in `main.rs`) to check for another reboot condition.

sources/nvidia-migmanager/README.md
```diff
@@ -0,0 +1,26 @@
...
+`settings.kubelet-device-plugins.nvidia.device-partitioning-strategy`. Then, it activates
+MIG and applies the profile according to the type of GPU present in the instance
+### Example:
+```toml
+[settings.kubelet-device-plugins.nvidia]
+device-partitioning-strategy="mig"
+[settings.kubelet-device-plugins.nvidia.mig.profile]
+"a100.40gb"="2"
+"h100.80gb"="4"
+"h200.141gb"="3"
```
bcressey: nit: whitespace for readability:
```suggestion
[settings.kubelet-device-plugins.nvidia]
device-partitioning-strategy = "mig"

[settings.kubelet-device-plugins.nvidia.mig.profile]
"a100.40gb" = "2"
"h100.80gb" = "4"
"h200.141gb" = "3"
```

sources/nvidia-migmanager/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+Current version: 0.1.0
+
+## NVIDIA MIG Manager
+`nvidia-migmanager` ensures that MIG settings are applied to an instance that supports
+it. It is called by `nvidia-migmanager.service`.
+NVIDIA MIG is only supported in select NVIDIA GPUs. The settings are first rendered
+into a config file. The binary first checks if MIG is activated by checking the value of
+`settings.kubelet-device-plugins.nvidia.device-partitioning-strategy`. Then, it activates
+MIG and applies the profile according to the type of GPU present in the instance
```
bcressey: "NVIDIA MIG is only supported in select NVIDIA GPUs" - this may or may not be the right place for that documentation note, but I'd put it in a separate paragraph, maybe at the end and replace it with a list of the GPUs that are actually supported.

You also have two "first" sentences which makes it unclear what's happening first. I would omit "the settings are rendered into a config file" if that's happening elsewhere - think of this as a generic program that could be included in another distro if its dependencies were met.

packages/os/os.spec
```diff
@@ -425,6 +429,11 @@ Summary: XFS progs cli
 %description -n %{_cross_os}xfscli
 %{summary}.
 
+%package -n %{_cross_os}nvidia-migmanager
+Summary: Manages nvidia-migmanager
```
```suggestion
Summary: Manages NVIDIA MIG configuration
```

(otherwise it sounds like it's managing itself)

packages/os/mig-nvidia-fabricmanager.service.drop-in.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecCondition=/usr/bin/nvidia-migmanager stop-fabricmanager
```
bcressey: This ends up feeling pretty unnatural - it's hard to reason about how Fabric Manager will behave just from inspecting the unit over in the kernel kit, and it's a pretty core constraint that it can't run if MIG is enabled.

Please consider adding this program and the related packaging in the kernel kit instead. You still need some of the changes here but it'll just be the relatively modest changes to the k8s device plugin template.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -5,7 +5,15 @@ std = { version = "v1", helpers = ["default"] }
 +++
 version: v1
 flags:
+  {{#if settings.kubelet-device-plugins.nvidia.device-partitioning-strategy}}
+  {{#if (eq settings.kubelet-device-plugins.nvidia.device-partitioning-strategy "mig")}}
```
bcressey: Can the built-in `eq` handler deal with a null here?

packages/os/nvidia-migmanager-tmpfiles.conf
```diff
@@ -0,0 +1,2 @@
+d /etc/nvidia-migmanager 0750 root root -
+d /run/nvidia-migmanager 0755 root root -
```
bcressey: What are these directories used for?

packages/os/nvidia-migmanager.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=NVIDIA MIG manager service
+After=nvidia-fabricmanager.service nvidia-persistenced.service
```
bcressey: Does configuring MIG create new `nvidia*` device nodes in `/dev`?

As I understand it, `nvidia-persistenced` works by keeping open a reference to the device node, so if new nodes are created then I would expect it needs to run afterwards.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+    if mig_settings.device_partitioning_strategy == NvidiaDevicePartitioningStrategy::MIG {
+        if mig_state == MigState::Disabled {
+            // Enable MIG for the GPU
+            set_mig_mode(true)?;
+
+            // If GPU is A100 create marker file for reboot to reconcile
+            // for the gpu reset and move from transitional state to enabled
+            if gpu_model == NvidiaGPU::A100_40GB || gpu_model == NvidiaGPU::A100_80GB {
+                info!("Rebooting to apply MIG Settings.");
+                command(TOUCH_PATH, ["/run/prairiedog/mig_enabled.ran"])?;
```
bcressey: As a general style note, use `std::fs::write` for this instead of executing a binary to create a file.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+use std::path::{Path, PathBuf};
+use std::process::{self, Command};
+use std::str::FromStr;
+
+use bottlerocket_modeled_types::{MIGProfile, NvidiaDevicePartitioningStrategy, NvidiaGPUModel};
+
+const NVIDIA_VENDOR_ID: &str = "10de";
+const DEFAULT_CONFIG_PATH: &str = "/etc/nvidia-migmanager/nvidia-migmanager.toml";
+const NVIDIA_SMI_PATH: &str = "/usr/libexec/nvidia/tesla/bin/nvidia-smi";
+const TOUCH_PATH: &str = "usr/bin/touch";
```
bcressey:
```suggestion
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+        !(mig_settings.device_partitioning_strategy == NvidiaDevicePartitioningStrategy::MIG
+            && (gpu_model == NvidiaGPU::A100_40GB || gpu_model == NvidiaGPU::A100_80GB)),
+        error::FabricManagerSnafu {}
+    );
+
+    Ok(())
+}
+
+/// Parses user arguments into an Args struct
+fn parse_args(args: env::Args) -> Result<(Args, Subcommand)> {
```
bcressey: Please use [argh](https://github.com/google/argh) for this rather than rolling your own parsing. I'd like to avoid custom parsers for any new code.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+            }
+
+            // Other arguments are passed to the subcommand parser
+            _ => subcommand_args.push(arg),
+        }
+    }
+
+    match subcommand.as_deref() {
+        Some("apply-mig") => Ok((global_args, Subcommand::HandleMIGManager {})),
+        Some("stop-fabricmanager") => Ok((global_args, Subcommand::StopFabricManager {})),
```
bcressey: I would name this after the purpose of the command, not the side effect it's expected to have when used in a systemd unit.

If I ran these from the command line then the exit code would be something I could reason about, and the "query" phrasing is better because I don't expect side effects:
```
# better since it's just about MIG stuff
nvidia-migmanager is-mig-active

# not as good because it isn't obvious why MIG would affect fabric manager
nvidia-migmanager should-fabric-manager-start

# implies that Fabric Manager will be stopped as a side effect
nvidia-migmanager stop-fabricmanager
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+/// Stores the subcommand to be executed
+#[derive(Debug)]
+enum Subcommand {
+    HandleMIGManager,
+    StopFabricManager,
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct NvidiaMIGConfig {
```
bcressey:
```suggestion
struct NvidiaMigConfig {
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+            log_level: LevelFilter::Info,
+            config_path: PathBuf::from_str(DEFAULT_CONFIG_PATH).unwrap(),
+        }
+    }
+}
+
+/// Stores the subcommand to be executed
+#[derive(Debug)]
+enum Subcommand {
+    HandleMIGManager,
```
bcressey:
```suggestion
    HandleMigManager,
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+        Logger { source: log::SetLoggerError },
+
+        #[snafu(display("Invalid log level '{}'", log_level))]
+        LogLevel {
+            log_level: String,
+            source: log::ParseLevelError,
+        },
+
+        #[snafu(display("MIG Profile couldn't be determined for the GPU."))]
+        MIGProfile {},
```
bcressey:
```suggestion
        MigProfile {},
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+            config_path: PathBuf,
+            source: toml::de::Error,
+        },
+
+        #[snafu(display("'{}' failed - stderr: {}",
+                        bin_path, String::from_utf8_lossy(&output.stderr)))]
+        CommandFailure { bin_path: String, output: Output },
+
+        #[snafu(display("Stopping Fabric Manager from running"))]
+        FabricManager {},
```
bcressey: This can just be a log message at the site, rather than using a structured error to print the message.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+// Runs the nvidia-smi command to apply the correct MIG profile in all the GPUs
+fn set_mig_profile(profile_string: &str) -> Result<()> {
+    info!("Activating MIG profile.");
+
+    command(NVIDIA_SMI_PATH, ["mig", "-cgi", profile_string, "-C"])?;
+    Ok(())
+}
+
+// Runs nvidia-smi command to find out the current state and the next state.
+fn get_gpu_state() -> Result<MigState> {
```
bcressey: Comment seems misleading - we're only returning the current state.

Consider using [the typestate pattern](https://cliffle.com/blog/rust-typestate/) to end up with types that only support a subset of operations depending on the device state.

The approach here puts the burden on the reader (and author) to confirm that the right state is checked before running any commands, versus having methods that are only available for the right state.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+            "--format=csv,noheader",
+        ],
+    )?;
+
+    let mut current_states = HashSet::new();
+    let mut pending_states = HashSet::new();
+    let mut gpu_state = MigState::Unknown;
+
+    for line in output.lines() {
+        let parts: Vec<_> = line.split(", ").collect();
```
bcressey: Does the CSV output format really have whitespace after the commas?
```suggestion
        let parts: Vec<_> = line.split(",").collect();
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+    command(NVIDIA_SMI_PATH, ["mig", "-cgi", profile_string, "-C"])?;
+    Ok(())
+}
+
+// Runs nvidia-smi command to find out the current state and the next state.
+fn get_gpu_state() -> Result<MigState> {
+    let output = command(
+        NVIDIA_SMI_PATH,
+        [
+            "--query-gpu=gpu_name,mig.mode.current,mig.mode.pending",
```
bcressey: Is the GPU name used for anything?

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+        .build();
+
+    let present_devices =
+        pciclient::list_devices(list_input).context(error::ListPciDevicesSnafu)?;
+
+    // If there a multiple devices with the same ID, dedup them to minimize iterations
+    let unique_ids = present_devices
+        .iter()
+        .map(|x| format!("0x{}", x.device().to_uppercase()).clone())
+        .collect::<HashSet<_>>();
```
bcressey: If the system has multiple GPUs, can each GPU be in a different state with respect to MIG?

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+    if unique_ids.contains("0x20B0") {
+        info!("Found NVIDIA A100-40GB GPU.");
+        Ok(NvidiaGPU::A100_40GB)
+    } else if unique_ids.contains("0x20B2") || unique_ids.contains("0x20B5") {
+        info!("Found NVIDIA A100-80GB GPU.");
+        Ok(NvidiaGPU::A100_80GB)
+    } else if unique_ids.contains("0x2330")
+    //        || unique_ids.contains("0x2321")
+    //        || unique_ids.contains("0x2331")
+    //        || unique_ids.contains("0x2339")
```
bcressey: Why are these commented out?

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+fn get_gpu_model() -> Result<NvidiaGPU> {
+    let list_input = pciclient::ListDevicesParam::builder()
+        .vendor(NVIDIA_VENDOR_ID)
+        .build();
+
+    let present_devices =
+        pciclient::list_devices(list_input).context(error::ListPciDevicesSnafu)?;
+
+    // If there a multiple devices with the same ID, dedup them to minimize iterations
+    let unique_ids = present_devices
```
bcressey: The code assumes that all GPUs will be the same, but it might improve it to abstract over the possibility of heterogenous GPUs, and to be precise about which GPU is being operated on.

In other words - return a `Result<Vec<NvidiaGpu>>`, ensure each `NvidiaGpu` entry has enough metadata to identify it in calls to `nvidia-smi` to query the device state for that GPU, and so on.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+            }
+        }
+
+        // Activate MIG Profile if any of them is selected else select the default profile.
+        if gpu_model == NvidiaGPU::A100_40GB {
+            let gpu_model = NvidiaGPUModel::try_from("a100.40gb").unwrap();
+            let mig_profile = mig_settings
+                .profile
+                .entry(gpu_model)
+                .or_insert(default_profile);
```
bcressey: If each GPU model was a type, it could implement a `get_mig_profile` method that took a `&str` and returned a `String` with the argument to `nvidia-smi`. Then you could have an impl on the enum that provided the same method, and did an exhaustive match across all devices to call the right variant's method.

The primary benefit is the exhaustiveness checking at build time, which is missing here.

But you could take it further and define GPU specific enums for MIG profiles, and aliases for deserialization:
```
#[derive(Deserialize)]
enum NvidiaA100_40GB_MigProfile {
    #[serde(alias = "7g40gb")]
    #[serde(alias = "1")]
    Mig_7G_40Gb,

    #[serde(alias = "3g20gb")]
    #[serde(alias = "2")]
    Mig_3G_20Gb,
    ...
}
```

Then the GPU model type could have a helper method to take a profile, parse it into an enum variant of its own type, and call a display method to format itself as a string.

As with `NvidiaGPU::get_mig_profile`, you could have a method in the top-level enum's impl that would turn that profile string into a display string, delegating to each GPU type's implementation.

sources/nvidia-migmanager/src/main.rs
piyush-jena: Updated with `warn!()` and `process::exit(1)`

packages/os/nvidia-migmanager-tmpfiles.conf
piyush-jena: I have removed `/run/nvidia-migmanager` since I ended up not using it. `/etc/nvidia-migmanager` has `nvidia-migmanager.toml` where the settings related to MIG are rendered.

sources/nvidia-migmanager/src/main.rs
piyush-jena: Updated to `is-mig-active`.

sources/nvidia-migmanager/src/main.rs
piyush-jena: I was testing it for P5 instance and the GPU's device id in P5 is `0x2330`. But there are other variants of H100 which have one of the other device ids so I ended up uncommenting it.

sources/nvidia-migmanager/src/main.rs
piyush-jena: Yes. I ran the command in g4dn.xlarge and got this output

```
Tesla T4, [N/A], [N/A]
```

sources/nvidia-migmanager/src/main.rs
piyush-jena: No. Removed in the latest commit.

sources/nvidia-migmanager/src/main.rs
piyush-jena: Updated

sources/api/prairiedog/src/bootconfig.rs
piyush-jena: Moved implementation to `nvidia-migmanager/src/main.rs` and the new implementation only checks for a particular file instead of a whole directory.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
piyush-jena: Since `settings.kubelet-device-plugins.nvidia.device-partitioning-strategy` has a default value, the check is unnecessary.

packages/systemd/systemd-tmpfiles.conf
piyush-jena: Added a new drop-in for `ExecStart=-/usr/bin/nvidia-migmanager reboot-if-required`

sources/nvidia-migmanager/src/main.rs
piyush-jena: Yes. Each GPU can be separately configured.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,685 @@
...
+        info!("Found NVIDIA H100-80GB GPU.");
+        Ok(NvidiaGpu::H100_80GB)
+    } else if pci_device_id.starts_with("0x2335")
+        || pci_device_id.starts_with("0x233B")
+        || pci_device_id.starts_with("0x2348")
+    {
+        info!("Found NVIDIA H200-141GB GPU.");
+        Ok(NvidiaGpu::H200_141GB)
+    } else {
+        info!("Found NVIDIA Device but couldn't confirm variant.");
```
KCSesh: This to me justifies a warning right? 

```suggestion
        warn!("Found NVIDIA Device but couldn't confirm variant.");
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+const GPU_MODEL_REGEX: &str = r"[A-Za-z]\d+\.(\d+)gb";
+const MIG_PROFILE_REGEX: &str = r"(\d+)g\.(\d+)gb";
+
+/// Stores arguments
+#[derive(FromArgs, PartialEq, Debug)]
+struct Args {
+    /// log-level trace|debug|info|warn|error
+    #[argh(option)]
+    log_level: Option<LevelFilter>,
+    /// configuration file with the modules to be linked
```
bcressey:
```suggestion
    /// configuration file with the desired MIG settings
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+fn get_gpu_state(current_state: &str, next_state: &str) -> MigState {
+    let mut gpu_state = MigState::Unknown;
+
+    if current_state != next_state {
+        gpu_state = MigState::Transition;
+    } else if current_state == next_state && current_state == "Enabled" {
+        gpu_state = MigState::Enabled;
+    } else if current_state == next_state && current_state == "Disabled" {
+        gpu_state = MigState::Disabled;
+    } else if current_state == next_state && current_state == "[N/A]" {
```
bcressey: `current_state == next_state` will be true for all these subsequent branches
```suggestion
    } else if current_state == "Enabled" {
        gpu_state = MigState::Enabled;
    } else if current_state == "Disabled" {
        gpu_state = MigState::Disabled;
    } else if current_state == "[N/A]" {
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    Ok(gpu_info)
+}
+
+#[derive(Deserialize)]
+enum NvidiaA100_40gbMigProfile {
+    #[serde(alias = "1g.5gb")]
+    #[serde(alias = "7")]
+    Mig1g5gb,
```
bcressey: Might be worth putting all these in a separate module, and importing at the top:
```
use crate::mig_profile::*;
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+    Ok(config)
+}
+
+fn reboot_if_required() -> Result<()> {
+    let reboot_required = Path::new(REBOOT_REQUIRED_MARKER_FILE).exists();
+
+    if reboot_required {
+        info!("GPU reset is required to apply MIG Settings. Initiating reboot...");
+        command(SYSTEMCTL_PATH, ["reboot"])?;
+    } else {
```
bcressey: `prairiedog` has an infinite sleep after this - see the comment there for why, but I expect we'd want to replicate that here

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+        for gpu in &gpu_info {
+            if gpu.model == NvidiaGpu::A100_40GB || gpu.model == NvidiaGpu::A100_80GB {
+                warn!("MIG mode is Enabled in an Ampere GPU. Disabling Fabric Manager ...");
+                process::exit(1);
+            }
+        }
+    }
+
+    Ok(())
+}
```
bcressey: If this is model-specific, and not just a binary "is MIG enabled or not?" decision, then maybe rename the command:
```
nvidia-migmanager is-fabric-manager-compatible
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
+use snafu::{ensure, ResultExt};
+use std::cmp::min;
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::fs;
+use std::path::Path;
+use std::process::{self, Command};
+
+use bottlerocket_modeled_types::{MigProfile, NvidiaDevicePartitioningStrategy, NvidiaGpuModel};
```
bcressey: I'm not sure we benefit enough from these imports to take on the dependency.

`MigProfile` gets turned immediately back into a string and then turned into one of the specialized types here, so we could just keep it as a string.

`NvidiaDevicePartitioningStrategy` - we only use the `MIG` variant, so we could just check for that string value.

`NvidiaGpuModel` could probably be kept as a string and turned into `NvidiaGpu` when needed.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+        let gpu_model = NvidiaGpuModel::try_from(model_str).unwrap();
+        let mig_profile = mig_settings
+            .profile
+            .get(&gpu_model)
+            .unwrap_or(&default_profile);
+
+        info!("MIG Profile or the number of GPU slices: {:?}", mig_profile);
+        let json_string = serde_json::to_string(mig_profile.as_ref()).unwrap();
+        let profile =
+            serde_json::from_str::<T>(&json_string).context(error::JsonDeserializationSnafu)?;
```
bcressey: I would use `serde_plain::from_str` and `serde_plain::to_string` for these.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+}
+
+fn handle_mig_manager(mig_settings: NvidiaMigConfig, gpu_info: Vec<MigGpu>) -> Result<()> {
+    if mig_settings.device_partitioning_strategy == NvidiaDevicePartitioningStrategy::MIG {
+        let (is_ampere_gpu_present, has_disabled_mig) =
+            gpu_info
+                .iter()
+                .fold((false, false), |(ampere, disabled), gpu| {
+                    let is_ampere =
+                        gpu.model == NvidiaGpu::A100_40GB || gpu.model == NvidiaGpu::A100_80GB;
```
bcressey: nit: consider methods on `NvidiaGpu` and `MigState` that match the active variant; the `is_ampere` check is repeated a couple times

```
impl NvidiaGpu {
    fn is_ampere(&self) -> bool {
        use NvidiaGpu::*;
        match self {
            A100_40GB | A100_80GB => true,
            _ => false,
        }
    }
}

...
gpu.model.is_ampere()
gpu.state.is_disabled() 
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+                std::fs::write("/run/nvidia-migmanager/reboot-required", "Enabling MIG").unwrap();
+            }
+        }
+
+        let a100_40gb_indices: Vec<usize> = gpu_info
+            .iter()
+            .enumerate()
+            .filter(|(_, gpu)| gpu.model == NvidiaGpu::A100_40GB)
+            .map(|(index, _)| index)
+            .collect();
```
bcressey: this is repeated a few times, and could be factored out into a function

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+                    .captures(mig_profile)
+                    .map(|captures| {
+                        (
+                            captures[1].parse().unwrap_or(0),
+                            captures[2].parse().unwrap_or(0),
+                        )
+                    })
+                    .unwrap();
+
+                let num_slices: usize = min(gpu_ram / slice_ram, 7 / compute_slices);
```
bcressey: Where does the magic number `7` come from?

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+            serde_json::from_str::<T>(&json_string).context(error::JsonDeserializationSnafu)?;
+
+        set_mig_profile(gpu_indices, profile.get_mig_profile())?;
+    }
+
+    Ok(())
+}
+
+fn handle_mig_manager(mig_settings: NvidiaMigConfig, gpu_info: Vec<MigGpu>) -> Result<()> {
+    if mig_settings.device_partitioning_strategy == NvidiaDevicePartitioningStrategy::MIG {
```
bcressey: Rather than have both "enable" and "disable" paths in the same function, I would split this into two functions:
```
fn handle_mig_manager(mig_settings: NvidiaMigConfig, gpu_info: Vec<MigGpu>) -> Result<()> {
    if mig_settings.device_partitioning_strategy == NvidiaDevicePartitioningStrategy::MIG {
        enable_mig(mig_settings, gpu_info)
    } else {
        disable_mig(mig_settings, gpu_info)
    }
}
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+            NvidiaGpuModel::try_from("a100.80gb").unwrap(),
+            NvidiaGpuModel::try_from("h100.80gb").unwrap(),
+            NvidiaGpuModel::try_from("h200.141gb").unwrap(),
+        ]
+        .into_iter()
+        .collect();
+
+        // The GPU in the current instance is not one of the known GPUs. We attempt using a profile that doesn't belong to one of the known GPUs.
+        for (gpu, mig_profile) in &mig_settings.profile {
+            if !known_gpus.contains(gpu) {
```
bcressey: Isn't this always true? You could make it another `ensure!` or omit the check, to help de-indent the code.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,688 @@
...
+
+        if has_enabled_mig {
+            // Disable MIG for the GPU
+            set_mig_mode(false)?;
+
+            // If GPU is A100 create marker file for reboot to reconcile
+            // for the gpu reset and move from transitional state to disabled
+            if is_ampere_gpu_present {
+                info!("Rebooting to apply MIG Settings.");
+                std::fs::write("/run/nvidia-migmanager/reboot-required", "Disabling MIG").unwrap();
```
bcressey: Please take another pass through and convert fallible `unwrap()` calls like this into snafu errors.

For `Regex::new` and other places that are nominally fallible but given known inputs at compile time, you can use either `unwrap` or `expect` - the latter lets you provide an error message which can also help explain why it should never happen.

sources/nvidia-migmanager/src/main.rs
piyush-jena: Each GPU has total 7 compute slices. I have added a comment to explain this.

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,702 @@
...
+    if has_disabled_mig {
+        // Enable MIG for all the GPU
+        set_mig_mode(true)?;
+
+        // If any GPU is A100 create marker file for reboot to reconcile
+        // for the gpu reset and move from transitional state to enabled
+        if is_ampere_gpu_present {
+            info!("Rebooting to apply MIG Settings...");
+            let marker_file_path = "/run/nvidia-migmanager/reboot-required";
+            std::fs::write(marker_file_path, "Enabling MIG").context(error::WriteMarkerSnafu {
```
KCSesh:
```suggestion
            std::fs::write(REBOOT_REQUIRED_MARKER_FILE, "Enabling MIG").context(error::WriteMarkerSnafu {
```

sources/nvidia-migmanager/src/main.rs
```diff
@@ -0,0 +1,702 @@
...
+    if has_enabled_mig {
+        // Disable MIG for the GPU
+        set_mig_mode(false)?;
+
+        // If GPU is A100 create marker file for reboot to reconcile
+        // for the gpu reset and move from transitional state to disabled
+        if is_ampere_gpu_present {
+            info!("Rebooting to apply MIG Settings...");
+            let marker_file_path = "/run/nvidia-migmanager/reboot-required";
+            std::fs::write(marker_file_path, "Disabling MIG").context(error::WriteMarkerSnafu {
```
KCSesh:
```suggestion
            std::fs::write(REBOOT_REQUIRED_MARKER_FILE, "Disabling MIG").context(error::WriteMarkerSnafu {
```

piyush-jena:
Force push fixes all the above comments.

piyush-jena:
Got rid of hack commit since we removed dependency on settings-sdk structs to simplify the code.

packages/os/mig-nvidia-fabricmanager.service.drop-in.conf
piyush-jena: I have opened a PR in kernel kit - https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/35.

I will get rid of the binary and systemd services, drop-ins and undo the changes in os.spec.

sources/nvidia-migmanager/src/main.rs
piyush-jena: mig_settings might contain some known and some unsupported (by us) GPUs (that support MIG). If the current instance has one of the "currently unsupported" GPUs, we wouldn't want to call `nvidia-smi` commands in cases we know aren't valid.

bcressey (approved):

KCSesh (approved):

PR #262, opened 2024-11-14
drop "dependencies" table for all packages

bcressey:
**Issue number:**
N/A

**Description of changes:**
`buildsys` is only capable of modeling build-time dependencies, and only for dependencies that fall within the current workspace - not anything that comes from a kit.

Using the "dependencies" table to track the runtime dependencies of the generated RPMs is superfluous, and likely to be incorrect even for packages that reside in the same workspace. Meanwhile, it cannot work properly for runtime dependencies that come from kits, since the metadata for those packages is not available.

Historically, listing these packages as a dependency caused them to be built before the dependent package was used. The "release" package in particular depended on most packages in the workspace, which meant that they would be built even if none of the variants included them as an explicit dependency.

Now all the packages in the workspace are expected to be dependencies of the kit that includes them, so this workaround is not necessary.

**Testing done:**
Built locally.

**Terms of contribution:**


jmt-lab (approved):

cbgbt (approved):

packages/release/Cargo.toml
```diff
@@ -9,47 +9,3 @@ build = "../build.rs"
 
 [lib]
 path = "../packages.rs"
-
```
arnaldo2792, line 12: Neat!

arnaldo2792 (approved):

PR #265, opened 2024-11-15
Use upstream sources for packages sourced from Amazon Linux

ginglis13:
**Issue number:**

n/a

**Description of changes:**

Build packages derived from Amazon Linux sources from upstream sources directly rather than relying on their existence in the Bottlerocket lookaside cache.

**Testing done:**

Build core kit for aarch64 and x86_64

**Terms of contribution:**


packages/kernel-5.10/Cargo.toml
```diff
@@ -15,11 +15,13 @@ path = "../packages.rs"
 # Use latest-kernel-srpm-url.sh to get this.
 url = "https://cdn.amazonlinux.com/blobstore/0af5f80d00a3d5a867d4959d74751bc7d24b1bcb0ab8a5de558ae301ae0fa52e/kernel-5.10.228-219.884.amzn2.src.rpm"
```
arnaldo2792, line 16: If we are going to do this, there are two other packages for which we use Amazon Linux's upstream:

- grub
- libkcapi

ginglis13:
^ force push adds `grub` and `libkcapi` as well, and squashes changes to one commit

bcressey (approved):

arnaldo2792 (approved):

PR #271, opened 2024-11-15
Fix source groups

arnaldo2792:
**Description of changes:**

`early-boot-config` was missing `api` and `imdsclient` from its `source-group` dependencies, so changes to either `apiclient` or `imdsclient` didn't trigger a rebuild of `early-boot-config`.

`pluto` 's `Cargo.toml` included `source-groups` by mistake, and it doesn't have any effect.

**Testing done:**

After this change, I confirmed that `early-boot-config` was re-compiled whenever I changed `apiclient` or `imdsclient` (even by just `touch`ing the files). Before this change, `touch`ing didn't trigger a rebuild.

**Terms of contribution:**

### DIFF
--- a/packages/early-boot-config/Cargo.toml
+++ b/packages/early-boot-config/Cargo.toml
@@ -10,6 +10,8 @@ path = "../packages.rs"
 
 [package.metadata.build-package]
 source-groups = [
+    "api",
+    "constants",
     "early-boot-config/early-boot-config",
     "early-boot-config/early-boot-config-provider",
 
@@ -20,6 +22,8 @@ source-groups = [
     "early-boot-config/user-data-providers/local-overrides",
     "early-boot-config/user-data-providers/vmware-cd-rom",
     "early-boot-config/user-data-providers/vmware-guestinfo",
+    "generate-readme",
+    "imdsclient",
 ]
 
 # RPM BuildRequires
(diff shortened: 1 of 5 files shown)


packages/early-boot-config/Cargo.toml
```diff
@@ -10,6 +10,8 @@ path = "../packages.rs"
 
 [package.metadata.build-package]
 source-groups = [
+    "api",
+    "imdsclient",
```
bcressey: Let's just list `early-boot-config` here instead of the individual subdirectories.

Appears to be missing `constants`, `generate-readme`, `imdsclient`,  ...

The brute force way to generate this list would be:
* copy the sources tree to another location
* delete everything from the sources tree that isn't a directory in `source-groups`
* trim the workspace `Cargo.toml` and regenerate `Cargo.lock`
* see if the package still builds

`static-pods` and `netdog` probably also need to be revisited with an eye to completing the dependency picture.

arnaldo2792:
Forced push includes:
- Changes in `os/Cargo.toml` to add missing dependencies for rebuild
- Changes in `netdog/Cargo.toml` to add missing dependencies for rebuild
- Changes in `static-pods/Cargo.toml` to add missing dependencies for rebuild

cbgbt (approved):

bcressey (approved):

PR #273, opened 2024-11-15
set LoaderTimeInitUSec and LoaderTimeExecUSec in GRUB

### LINKED ISSUE #272: make systemd-analyze report time spent in firmware and boot loader
**What I'd like:**
`systemd-analyze` can report time spent in the firmware and bootloader, per the [Boot Loader Interface](https://systemd.io/BOOT_LOADER_INTERFACE/):
> If LoaderTimeInitUSec and LoaderTimeExecUSec are set, systemd-analyze will include them in its boot-time analysis.

This doesn't work on Bottlerocket, but it would be nice if it did because it gives a more complete picture of the time it takes to boot a node.

Status quo. However, then I need to correlate logs from different systems in order to calculate the time spent before the kernel started, and I don't have an easy way to figure out where that time went - whether it's GRUB, shim, or something else.

bcressey:
**Issue number:**
Fixes #272 

**Description of changes:**
For `systemd-analyze` to report the time spent in firmware and in the boot loader, we need to set `LoaderTimeInitUSec` and `LoaderTimeExecUSec`, so the values can be read by [efi_loader_get_boot_usec()](https://github.com/systemd/systemd/blob/248eeec612d50e75c9da541721eeea8ac72e27ea/src/shared/efi-loader.c#L36) and set by [boot_timestamps()](https://github.com/systemd/systemd/blob/248eeec612d50e75c9da541721eeea8ac72e27ea/src/shared/boot-timestamps.c#L9).

This patch series teaches GRUB to do this for x86 and aarch64. It relies on backported [upstream patches](https://mail.gnu.org/archive/html/grub-devel/2023-05/msg00147.html) from the series that added the Boot Loader Interface module]. That series implemented support for a different part of BLI: the `LoaderInfo` and `LoaderDevicePartUUID` variables.

**Testing done:**
Tested this under the different EFI implementations I have access to - edk2 via QEMU/KVM, EC2 Nitro instances, and VMware ESXi VMs.

Verified that `systemd-analyze` reports firmware and loader time:
```
# systemd-analyze
Startup finished in 4.392s (firmware) + 833ms (loader) + 776ms (kernel) + 3.728s (userspace) = 9.729s
multi-user.target reached after 3.727s in userspace.
```

Confirmed that the variables aren't marked as non-volatile:
```
[fedora@admin]$ efivar -p --name 4a67b082-0a4c-41cf-b6c7-440b29bb8c4f-LoaderTimeExecUSec
GUID: 4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
Name: "LoaderTimeExecUSec"
Attributes:
	Boot Service Access
	Runtime Service Access
Value:
00000000  36 00 36 00 31 00 36 00  30 00 30 00 30 00 00 00  |6.6.1.6.0.0.0...|

[fedora@admin]$ efivar -p --name 4a67b082-0a4c-41cf-b6c7-440b29bb8c4f-LoaderTimeInitUSec
GUID: 4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
Name: "LoaderTimeInitUSec"
Attributes:
	Boot Service Access
	Runtime Service Access
Value:
00000000  31 00 37 00 30 00 34 00  30 00 30 00 30 00 00 00  |1.7.0.4.0.0.0...|
```

**Terms of contribution:**

### DIFF
--- a/packages/grub/0049-efi-Add-grub_efi_set_variable_with_attributes.patch
+++ b/packages/grub/0049-efi-Add-grub_efi_set_variable_with_attributes.patch
@@ -1,0 +1,88 @@
+From 37cd5dfd522882e1d95c4205635f9b2763256707 Mon Sep 17 00:00:00 2001
+From: Oliver Steffen <osteffen@redhat.com>
+Date: Fri, 26 May 2023 13:35:42 +0200
+Subject: [PATCH] efi: Add grub_efi_set_variable_with_attributes()
+
+Add a function to the EFI module that allows setting EFI variables
+with specific attributes.
+
+This is useful for marking variables as volatile, for example.
+
+Signed-off-by: Oliver Steffen <osteffen@redhat.com>
+Reviewed-by: Daniel Kiper <daniel.kiper@oracle.com>
+[bcressey:
+  - backport to 2.06
+  - avoid changes from bb4aa6e0 ("efi: Drop all uses of efi_call_XX() wrappers")]
+Signed-off-by: Ben Cressey <bcressey@amazon.com>
+
+(cherry picked from commit 7e4da6fb2d03ea20f7e11efc496e2e6cf360048b)
+---
+ grub-core/kern/efi/efi.c | 20 +++++++++++++-------
+ include/grub/efi/efi.h   |  6 ++++++
+ 2 files changed, 19 insertions(+), 7 deletions(-)
+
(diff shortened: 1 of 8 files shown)


larvacea (approved):

yeazelm (approved):
This is very cool! Nice set of patches!

arnaldo2792 (approved):

PR #278, opened 2024-11-18
host-ctr: added host certs to host containers

sumukhballal:
** Description of changes **

Add host certs to Host containers using a volume.

**Testing done:**

 - [x] Built core-kit.
 - [x] Build Bottlerocket.
 - [x] Booted BR and ran a yum update test.

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -851,6 +851,12 @@ func withDefaultMounts(containerID string, persistentDir string) oci.SpecOpts {
 			Destination: "/etc/bottlerocket-release",
 			Source:      "/etc/os-release",
 		},
+		// Bottlerocket host certs for the container
+		{
+			Options:     []string{"bind", "ro"},
+			Destination: "/.bottlerocket/certs",
+			Source:      "/etc/pki/tls/certs",
+		},
 		// Bottlerocket RPM inventory available to the container
 		{
 			Options:     []string{"bind", "ro"},


cbgbt:
I'm curious what happens if a host-container runs `update-ca-trust` with this volume mount.

`update-ca-trust` is the process that's run on Fedora-like distros when new ca-certificates are installed. It combines the certs in the system's cert directory into the CA bundle. In the most typical case, this is run whenever someone installs or upgrades the `ca-certificates` package. Yum is a "greedy" package manager, and it likes to update the dependencies of packages that you `yum install`, so there's a good chance that someone could run `update-ca-trust` without ever intending to via `yum` marking the package for update.

The `update-ca-trust` binary should fail during such a yum transaction because we have the certs directory mounted via a read-only volume... but what happens to the yum transaction?

You could test this by running e.g. `sudo yum -y reinstall ca-certificates` in the admin container.

cbgbt:
Something else to consider: this is technically a breaking change for existing host-containers which may have alternative or private certificates.

We may wish to be more cautious by:
* Introduce a new setting, `settings.host-containers.<container>.use-system-ca-certificates`, and only mount if this is set to `true`.
* Only default this to `true` for the admin and control containers for existing releases, but use a default of true for all host containers in subsequent variants.

Technically, defaulting this to `true` for the `admin` and `control` containers is *still* a breaking change. We could possibly generate the setting for those containers using a setting-generator that checked to see whether or not the system was using the default admin/host container (which is *still* a breaking change really, you can do anything you want in there so we can't guarantee folks aren't adding certs).

jpculp:
I think that this works as long as you have something similar to this in the entrypoint of the host container:

```bash
# Link host certs if present into container & run update-ca-trust
HOST_CERTS="/.bottlerocket/certs"
for cert in $(ls -1 "${HOST_CERTS}"); do
  ln -s "${HOST_CERTS}/${cert}" "/etc/pki/ca-trust/source/anchors/${cert}"
done
update-ca-trust
```

bcressey (approved):

cbgbt (approved):

PR #279, opened 2024-11-18
allow overriding the max-pods file with one from your variant

tzneal:
**Issue number:**

Closes #

**Description of changes:**

Files in RPMs can't conflict, so we modify pluto to read from a normally non-existent min & max file. A variant can then write this file to override an absolute min/max number of pods per node.

**Testing done:**

Built a custom variant with a eni-max-pods-max file

**Terms of contribution:**


yeazelm (approved):
LGTM!

bcressey (approved):

bcressey (approved):

PR #289, opened 2024-11-27
Add netlink debug to kernels

yeazelm:
**Issue number:** 

Closes # https://github.com/bottlerocket-os/bottlerocket/issues/4313

**Description of changes:**
CRIU requires NETLINK_DEBUG for capturing socket info for checkpoint and restore. This adds the module to the kernel config for that use case.

**Testing done:**
Built `aws-k8s-1.31` and confirmed module loads.
``` 
bash-5.1# modprobe netlink_diag
bash-5.1# lsmod | grep netlink
netlink_diag           16384  0
nf_conntrack_netlink    57344  0
nfnetlink_acct         16384  3 xt_nfacct
nfnetlink              20480  7 nfnetlink_acct,nf_conntrack_netlink,nf_tables
nf_conntrack          184320  7 xt_conntrack,nf_nat,xt_state,xt_nat,nf_conntrack_netlink,xt_connmark,xt_MASQUERADE
```
Confirmed the same works on `aws-k8s-1.24` (5.15 kernel) and `aws-ecs-1` (5.10 kernel). I also checked `aarch64` for 6.1 kernel.

**Terms of contribution:**


packages/kernel-6.1/kernel-6.1.spec
```diff
@@ -1312,6 +1312,7 @@ install -p -m 0644 %{S:302} %{buildroot}%{_cross_bootconfigdir}/05-metal.conf
 %{_cross_kmoddir}/kernel/net/netfilter/xt_TPROXY.ko.*
 %{_cross_kmoddir}/kernel/net/netfilter/xt_TRACE.ko.*
 %{_cross_kmoddir}/kernel/net/netfilter/xt_u32.ko.*
+%{_cross_kmoddir}/kernel/net/netlink/netlink_diag.ko.*
```
larvacea, line 1315: Why is this the only spec file change? All three kernel configs look as if they should create a kmod.

yeazelm: The `modules` section in the other two kernel spec files are not as explicit. We are choosing exactly which modules we want in 6.1 but 5.10 and 5.15 basically include all the modules via this: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/kernel-5.10/kernel-5.10.spec#L302. So the specific break out is required only for 6.1. This is why I tried the other kernels explicitly to make sure the module is loadable on a booted box.

larvacea (approved):

ytsssun (approved):

arnaldo2792:
- What other configurations are we getting by enabling the new `CONFIG` ?
- Why shouldn't the module be built in the kernel image rather than as a kernel module?

yeazelm:
> * What other configurations are we getting by enabling the new `CONFIG` ?

This just adds the one module.
<details>

```
diff --color /home/fedora/git/bottlerocket-core-kit/directory-for-develop/config-5.10-x86_64.config directory-for-updated-configs/config-5.10-x86_64.config
1671c1671
< # CONFIG_NETLINK_DIAG is not set
---
> CONFIG_NETLINK_DIAG=m
diff --color /home/fedora/git/bottlerocket-core-kit/directory-for-develop/config-5.15-aarch64.config directory-for-updated-configs/config-5.15-aarch64.config
1670c1670
< # CONFIG_NETLINK_DIAG is not set
---
> CONFIG_NETLINK_DIAG=m
diff --color /home/fedora/git/bottlerocket-core-kit/directory-for-develop/config-5.15-x86_64.config directory-for-updated-configs/config-5.15-x86_64.config
1664c1664
< # CONFIG_NETLINK_DIAG is not set
---
> CONFIG_NETLINK_DIAG=m
diff --color /home/fedora/git/bottlerocket-core-kit/directory-for-develop/config-6.1-aarch64.config directory-for-updated-configs/config-6.1-aarch64.config
1706c1706
< # CONFIG_NETLINK_DIAG is not set
---
> CONFIG_NETLINK_DIAG=m
diff --color /home/fedora/git/bottlerocket-core-kit/directory-for-develop/config-6.1-x86_64.config directory-for-updated-configs/config-6.1-x86_64.config
1717c1717
< # CONFIG_NETLINK_DIAG is not set
---
> CONFIG_NETLINK_DIAG=m
```

</details>

> * Why shouldn't the module be built in the kernel image rather than as a kernel module?

It isn't needed for most operations, so loading it only when you need it keeps the kernel binary smaller. We can make it built-in if we would like, but I tend to opt for loadable modules unless we expect the vast majority of use cases to need it.

bcressey (approved):

yeazelm:
This will need to be cut against the https://github.com/bottlerocket-os/bottlerocket-kernel-kit/ to be merged. I'll cut this PR against that repo shorty.

yeazelm:
I've cut https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/15 to get this merged.

PR #295, opened 2024-12-05
xfsprogs: add patch to load default config file

### LINKED ISSUE #4281: XFS Ephemeral Storage fails to mount with: Superblock has unknown incompatible features (0x20) enabled.

Sparksssj:
Issue number:

Closes https://github.com/bottlerocket-os/bottlerocket/issues/4281

Description of changes:
A new PR for https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/251, see discussion there.

Testing done:

Manual testing done using the exact setup as original issue mentioned , see the nrext64 value changed to 0 as expected.
Also, tested config files consuming behavior and made sure the program would not crash when receive a bad default config file. It would just continue with the default value given in the `xfs_mkfs.c` file.

With correct config files:
```
bash-5.1# xfs_info /dev/nvme2n1
meta-data=/dev/nvme2n1           isize=512    agcount=4, agsize=7629394 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=1
         =                       reflink=1    bigtime=1 inobtcount=1 nrext64=0
data     =                       bsize=4096   blocks=30517576, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=16384, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
bash-5.1#
bash-5.1#
bash-5.1# umount /dev/nvme2n1
bash-5.1# mkfs.xfs /dev/nvme2n1
Parameters parsed from default config file /usr/share/xfs/mkfs.xfs.conf successfully
mkfs.xfs: /dev/nvme2n1 contains a mounted filesystem
```

Without correct config files:
```
bash-5.1# mount | grep ephemeral
/dev/nvme2n1 on /mnt/.ephemeral type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota)
bash-5.1# xfs_info /dev/nvme2n1
meta-data=/dev/nvme2n1           isize=512    agcount=4, agsize=7629394 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=1
         =                       reflink=1    bigtime=1 inobtcount=1 nrext64=1
data     =                       bsize=4096   blocks=30517576, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=16384, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
bash-5.1#
bash-5.1# umount /dev/nvme2n1
bash-5.1# mkfs.xfs /dev/nvme2n1
/usr/share/xfs/mkfs.xfs.conf: Unrecognised input on line 6. Aborting.
mkfs.xfs: /dev/nvme2n1 contains a mounted filesystem
```
The program would abort the `parse default config` function and keep other functions work well.
We can always track if default config file works by using
```
bash-5.1# umount /dev/nvme2n1
bash-5.1# mkfs.xfs /dev/nvme2n1
```
And see if there's a log message showing
```
Parameters parsed from default config file /usr/share/xfs/mkfs.xfs.conf successfully
```
This log message is expected every time we call mkfs.xfs, with or without manual CLI config file provided.
Terms of contribution:

### DIFF
--- a/packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
+++ b/packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
@@ -1,0 +1,42 @@
+From eea379598278cc62a74c9932f624c89407a966f8 Mon Sep 17 00:00:00 2001
+From: Sparks Song <shijiao@amazon.com>
+Date: Mon, 3 Feb 2025 22:44:31 +0000
+Subject: [PATCH] mkfs: source defaults from config file to make nrext64
+ default off on multiple kernel versions
+
+---
+ mkfs/xfs_mkfs.c | 11 +++++++++++
+ 1 file changed, 11 insertions(+)
+
+diff --git a/mkfs/xfs_mkfs.c b/mkfs/xfs_mkfs.c
+index 6d2469c..87e7457 100644
+--- a/mkfs/xfs_mkfs.c
++++ b/mkfs/xfs_mkfs.c
+@@ -39,6 +39,9 @@
+  */
+ #define WHACK_SIZE (128 * 1024)
+
++/* Default path for the mkfs.xfs configuration file */
++#define DEFAULT_CONFIG_PATH "/usr/share/xfsprogs/mkfs/default.conf"
++
+ /*
+  * XXX: The configured block and sector sizes are defined as global variables so
+  * that they don't need to be passed to getnum/cvtnum().
+@@ -4428,6 +4431,14 @@ main(
+ 	 * the options from this file parsed, we can then proceed with parameter
+ 	 * and bounds checking and making the filesystem.
+ 	 */
++
++	if (cli.cfgfile == NULL) {
++		struct stat defcfg_statbuf;
++		if (stat(DEFAULT_CONFIG_PATH, &defcfg_statbuf) == 0) {
++			cli.cfgfile = DEFAULT_CONFIG_PATH;
++		}
++	}
++
+ 	cfgfile_parse(&cli);
+
+ 	protostring = setup_proto(cli.protofile);
+--
+2.47.0
+
--- a/packages/xfsprogs/xfsprogs.spec
+++ b/packages/xfsprogs/xfsprogs.spec
@@ -8,6 +8,8 @@ Source0: http://kernel.org/pub/linux/utils/fs/xfs/xfsprogs/xfsprogs-%{version}.t
 Source1: http://kernel.org/pub/linux/utils/fs/xfs/xfsprogs/xfsprogs-%{version}.tar.sign
 Source2: gpgkey-0C1D891C50A732E0680F7B644675A111E50B5FA6.asc
 
+Patch1000: 0001-mkfs-source-defaults-from-config-file.patch
+
 BuildRequires: %{_cross_os}glibc-devel
 BuildRequires: %{_cross_os}libuuid-devel
 BuildRequires: %{_cross_os}libinih-devel
@@ -29,7 +31,7 @@ Requires: %{name}
 
 %prep
 %{gpgverify} --data=<(xzcat %{S:0}) --signature=%{S:1} --keyring=%{S:2}
-%autosetup -n xfsprogs-%{version}
+%autosetup -n xfsprogs-%{version} -p1
 
 %build
 %cross_configure \


packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -16,6 +16,9 @@ Source100: config-bottlerocket
 Source220: neuron-sysinit.target.drop-in.conf
 Source221: modprobe@neuron.service.drop-in.conf
 
+# XFS configuration
+Source300: mkfs.xfs-5_10.conf
```
yeazelm: nit: I'd like to keep the name of these files consistent on the packages. Since they are all in different packages, its fine to name them all `mkfs.xfs.conf`

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,50 @@
...
+ 	 * the options from this file parsed, we can then proceed with parameter
+ 	 * and bounds checking and making the filesystem.
+ 	 */
+-	cfgfile_parse(&cli);
++
++	if (cli.cfgfile == NULL) {
++		/* No user config file specified, try default config */
++		FILE *default_config_file = fopen(DEFAULT_CONFIG_PATH, "r");
++		if (default_config_file != NULL) {
++			fclose(default_config_file);
```
yeazelm: This feels a bit odd to `fopen()` the file only for checking if it exists, then close it, and finally parse it. Would it make more sense to implement similar logic to `cfgfile_parse()` but not error out completely if the file isn't present/parsable?

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,89 @@
...
++			fprintf(stderr,
++		_("Memory allocation failure parsing %s. Aborting.\n"),
++				cli->cfgfile);
++		} else {
++			fprintf(stderr,
++		_("Unknown error %d opening default config file %s. Aborting.\n"),
++				error, cli->cfgfile);
++		}
++	} else {
++		printf(_("Parameters parsed from default config file %s successfully\n"),
```
yeazelm: Did you see this output in the journal when it ran successfully? Do we expect this to be seen on every boot or are the mkfs.xfs calls for the XFS backed data partitions not emitting logs?

yeazelm (commented):
This looks great! Did you confirm on `aws-k8s-1.25` that using the ephemeral storage commands no longer have issues like in https://github.com/bottlerocket-os/bottlerocket/issues/4281?

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
Sparksssj: Comment 1: I did make confirmation on `aws-k8s-1.25` with size `i3en.2xlarge` and corresponding userdata which is exactly what user mentioned in original issue.
Comment 2: Yes. We do see this output when this's succeed. And these logs will not be seen directly in EC2 console. And can be seen if you run 
```
umount /dev/nvme2n1
mkfs.xfs /dev/nvme2n1
```

yeazelm: Ok, so it sounds like you did validate that this code fixes the issue in https://github.com/bottlerocket-os/bottlerocket/issues/4281 but just on `aws-k8s-1.24` instead of `aws-k8s-1.25` which should work the same here.

yeazelm (approved):
I've done some offline discussion here around testing but this looks good to me!

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
Sparksssj: Yes. I did both 1.24 and 1.25 tests, both works.

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -83,6 +86,12 @@ Summary: Modules for the Linux kernel
 %description modules
 %{summary}.
 
+%package mkfs-confs
```
bcressey: Suggestions to improve the dependencies:
1. Require the parent package so that the installed configs match the installed kernel.
2. Provide a generic capability (using parens, by convention) that other packages can key off of
3. Conflict with any other package providing the same capability

Goal is to make it so that only one `mkfs-confs` package can be installed.

```suggestion

%package mkfs-confs
Requires: %{name}
Provides: %{_cross_os}kernel(mkfs-confs)
Conflicts: %{_cross_os}kernel(mkfs-confs)
```

However, that said - I would not make this a separate subpackage, but instead add that `Provides` and `Conflicts` to the parent kernel package.

packages/kernel-5.10/mkfs.xfs.conf
```diff
@@ -0,0 +1,5 @@
+# V5 features that were the mkfs defaults when the upstream Linux 5.10 LTS
+# kernel was released at the end of 2020.
+
+[inode]
+nrext64=0
```
bcressey: Is this really the only feature of interest that's worth pinning? It seems a bit sparse.

packages/xfsprogs/xfsprogs.spec
```diff
@@ -14,6 +16,9 @@ BuildRequires: %{_cross_os}libblkid-devel
 
 Requires: %{_cross_os}liburcu
 Requires: %{_cross_os}libinih
+Requires: (%{_cross_os}kernel-5.10-mkfs-confs if %{_cross_os}kernel-5.10)
+Requires: (%{_cross_os}kernel-5.15-mkfs-confs if %{_cross_os}kernel-5.15)
+Requires: (%{_cross_os}kernel-6.1-mkfs-confs if %{_cross_os}kernel-6.1)
```
bcressey: Please just require a generic capability that any kernel package can implement. Otherwise it will be challenging to bring a custom kernel of a different version.

```suggestion
Requires: (%{_cross_os}kernel(mkfs-confs)
```

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,89 @@
...
+  * XXX: The configured block and sector sizes are defined as global variables so
+  * that they don't need to be passed to getnum/cvtnum().
+@@ -4271,6 +4274,44 @@ cfgfile_parse(
+ 		cli->cfgfile);
+ }
+
++
++/* This function is similar to cfgfile_parse() and specifically parses
++ * default config files. It doesn’t exit on failure.
++ */
```
bcressey: This function seems almost identical to `cfgfile_parse` except the `exit(1)` has been removed and the message is changed in a trivial way.

Rather than duplicate the function, I would just leave the message unmodified (it will obviously point to the default config file) and add a check around `exit(1)` like this:
```
if (strcmp(cli->cfgfile, DEFAULT_CONFIG_PATH) != 0)
    exit(1);
```

The problem with the copy-paste-tweak approach to patching is that if the upstream version of the function ever changes, your copy-paste version won't pick up those changes, and the patch is very likely to still apply cleanly.

By modifying the existing function in a careful way, you are much more likely to see the patch fail to apply on newer versions if there's been a significant change, which is what you want as it will signal the need to revisit the downstream alteration.

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,89 @@
...
+-	cfgfile_parse(&cli);
++
++	if (cli.cfgfile == NULL) {
++		/* No user config file specified, try default config */
++		cli.cfgfile = DEFAULT_CONFIG_PATH;
++		cfgfile_parse_default(&cli);
++	} else {
++		/* User specified their own config file, use that */
++		cfgfile_parse(&cli);
++	}
```
bcressey: This should all just be:
```
if (cli.cfgfile == NULL) {
    cli.cfgfile = DEFAULT_CONFIG_PATH;
}

cfgfile_parse(&cli);
```

That minimizes the changes to this function while still providing the fallback behavior you're trying to add.

packages/kernel-5.10/mkfs.xfs.conf
Sparksssj: Yea, this is indeed the only feature we care about to fix the original issue. Unless you have more thoughts and we can sure add more in other PRs -- I still think we should only keep this one here because this is the one related to the issue.

packages/kernel-6.1/mkfs.xfs.conf
```diff
@@ -0,0 +1,5 @@
+# V5 features that were the mkfs defaults when the upstream Linux 6.1 LTS
+# kernel was released at the end of 2022.
+
+[inode]
+nrext64=0
```
cbgbt: Should we keep this as the default of `1` on 6.1? If I recall correctly, isn't that when the default of `1` was introduced?

Sparksssj: Yes, it is. But it will show `EXPERIMENTAL Large extent counts feature in use. Use at your own risk!`. Since we are not very confident about that so we just set it 0 for now. The purpose for this change is to conveniently change it to 1 when needed.

Sparksssj:
Thanks to @bcressey I made a new revision, here's the log message which looks good to me
```
bash-5.1# dmesg | grep -i xfs
[    0.195369] SGI XFS with ACLs, security attributes, quota, no debug enabled
[    1.674746] XFS (nvme1n1p1): Mounting V5 Filesystem
[    1.718697] XFS (nvme1n1p1): Ending clean mount
[    3.895366] XFS (nvme2n1): Mounting V5 Filesystem
[    3.900304] XFS (nvme2n1): Ending clean mount


bash-5.1# mount | grep ephemeral
/dev/nvme2n1 on /mnt/.ephemeral type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota)


bash-5.1# xfs_info /dev/nvme2n1
meta-data=/dev/nvme2n1           isize=512    agcount=4, agsize=7629394 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=1
         =                       reflink=1    bigtime=1 inobtcount=1 nrext64=0
data     =                       bsize=4096   blocks=30517576, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=16384, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
```

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,39 @@
...
++
+ /*
+  * XXX: The configured block and sector sizes are defined as global variables so
+  * that they don't need to be passed to getnum/cvtnum().
+@@ -4428,6 +4431,11 @@ main(
+ 	 * the options from this file parsed, we can then proceed with parameter
+ 	 * and bounds checking and making the filesystem.
+ 	 */
++	if (cli.cfgfile == NULL) {
++		//check if default config file is there
```
bcressey: You will need to add an actual check that the file exists.

If it doesn't exist, then this would result in a fatal error until `mkfs.xfs` was invoked with config file that did exist.

Since you don't have any package-level dependencies that would ensure the config file is present, you must be defensive and assume it won't be there.

packages/xfsprogs/0001-mkfs-source-defaults-from-config-file.patch
```diff
@@ -0,0 +1,39 @@
...
+diff --git a/mkfs/xfs_mkfs.c b/mkfs/xfs_mkfs.c
+index 6d2469c..91a0f37 100644
+--- a/mkfs/xfs_mkfs.c
++++ b/mkfs/xfs_mkfs.c
+@@ -39,6 +39,9 @@
+  */
+ #define WHACK_SIZE (128 * 1024)
+
++/* Default path for the mkfs.xfs configuration file */
++#define DEFAULT_CONFIG_PATH "/usr/share/xfs/mkfs.xfs.conf"
```
bcressey: This should be in the same place as the other `mkfs.xfs` config files:
```
/usr/share/xfsprogs/mkfs/default.conf
```

bcressey (dismissed):
Commit subject should be:
```
xfsprogs: add patch to load default config file
```

You can change the PR title to say the same thing, now that we don't actually have any configuration files changing here.

Sparksssj:
updated commit message

bcressey:
Revoking approval since @Sparksssj says this causes an error - need to understand why before proceeding:
```
[  402.091994] systemd-makefs[1452]: -m reflink option respecified.
```

cbgbt:
While looking into this with @Sparksssj, we discovered that `mkfs.xfs` uses a set of global static structures to keep track of what parameters it has already seen. For some reason despite `reflink` only being referenced in our config file, the "respecified" logic is being triggered and causing mkfs fail.

What's odd is that I don't see how our code is functionally different to just supplying a configfile with the `-c` option.

What might be a good next step would be to add some debugging statements in the respecification logic ([code link](https://kernel.googlesource.com/pub/scm/fs/xfs/xfsprogs-dev/+/refs/tags/v6.9.0/mkfs/xfs_mkfs.c#1445)).

---

There's also [this comment in `mkfs`](https://kernel.googlesource.com/pub/scm/fs/xfs/xfsprogs-dev/+/refs/tags/v6.9.0/mkfs/xfs_mkfs.c#4358) adding a `TODO` for the exact feature we want:

 	/*
	 * TODO: Sourcing defaults from a config file
	 *
	 * Before anything else, see if there's a config file with different
	 * defaults. If a file exists in <package location>, read in the new
	 * default values and overwrite them in the &dft structure. This way the
	 * new defaults will apply before we parse the CLI, and the CLI will
	 * still be able to override them. When more than one source is
	 * implemented, emit a message to indicate where the defaults being
	 * used came from.
	 *
	 * printf(_("Default configuration sourced from %s\n"), dft.source);
	 */

We attempted to implement this by parsing the configfile and using `memcpy()` to overwrite the defaults ([following the example of how it's already done for the defaults system](https://kernel.googlesource.com/pub/scm/fs/xfs/xfsprogs-dev/+/refs/tags/v6.9.0/mkfs/xfs_mkfs.c#4373)).

This is unfortunately where we learned that the "respecified" tracking uses global statics -- even though our configfile parse used its own memory, it still kept track of the values that it had "seen" and thus caused the respecification errors.

PR #313, opened 2024-12-11
removal of packages moving to bottlerocket-kernel-kit

jmt-lab:
**Description of changes:**
Removes packages that have now moved to kernel kit:

* grub
* kernel-5.10
* kernel-5.15
* kernel-6.1
* kmod-5.10-nvidia
* kmod-5.15-nvidia
* kmod-6.1-nvidia
* linux-firmware
* microcode
* libkcapi
* shim

**Terms of contribution:**


rpkelly (approved):

ginglis13 (approved):

Twoliter.toml
```diff
@@ -1,5 +1,5 @@
 schema-version = 1
-release-version = "4.1.0"
+release-version = "5.0.0"
```
cbgbt, line 2: nit: Version bumps are typically (though not always!) done in a separate commit.

This otherwise LGTM!

cbgbt (approved):

Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
bcressey, line 3: Will this create issues with older Rust toolchains? When was version 4 support added?

bcressey (approved):

Cargo.lock
jmt-lab: O.o have no idea, i just ran this today using cargo 
1.83.0 2024-10-29

bcressey (approved):

PR #320, opened 2024-12-16
Allow bind mounts prefixed with /mnt/ for ephemeral storage

zaheerm:
**Issue number:**

Closes #

**Description of changes:**

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/ephemeral_storage.rs
+++ b/sources/api/apiserver/src/server/ephemeral_storage.rs
@@ -23,6 +23,12 @@ static EPHEMERAL_MNT: &str = ".ephemeral";
 static RAID_DEVICE_DIR: &str = "/dev/md/";
 static RAID_DEVICE_NAME: &str = "ephemeral";
 
+pub struct BindDirs {
+    pub allowed_exact: HashSet<&'static str>,
+    pub allowed_prefixes: HashSet<&'static str>,
+    pub disallowed_contains: HashSet<&'static str>,
+}
+
 /// initialize prepares the ephemeral storage for formatting and formats it.  For multiple disks
 /// preparation is the creation of a RAID0 array, for a single disk this is a no-op. The array or disk
 /// is then formatted with the specified filesystem (default=xfs) if not formatted already.
@@ -116,8 +122,17 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
     let mount_point = Path::new(&mount_point);
     let allowed_dirs = allowed_bind_dirs(variant);
     for dir in &dirs {
+        let exact_match = allowed_dirs.allowed_exact.contains(dir.as_str());
+        let prefix_match = allowed_dirs
+            .allowed_prefixes
+            .iter()
+            .any(|prefix| dir.starts_with(prefix));
+        let disallowed_match = allowed_dirs
+            .disallowed_contains
+            .iter()
+            .any(|contains| dir.contains(contains));
         ensure!(
-            allowed_dirs.contains(dir.as_str()),
+            exact_match || (prefix_match && !disallowed_match),
             error::InvalidParameterSnafu {
                 parameter: dir,
                 reason: "specified bind directory not in allow list",
@@ -269,18 +284,26 @@ pub fn ephemeral_devices() -> Result<Vec<String>> {
 }
 
 /// allowed_bind_dirs returns a set of the directories that can be bound to ephemeral storage, which
-/// varies based on the variant
-pub fn allowed_bind_dirs(variant: &str) -> HashSet<&'static str> {
-    let mut allowed = HashSet::from(["/var/lib/containerd", "/var/lib/host-containerd"]);
+/// varies based on the variant, a set of the prefixes of directories that are allowed to be bound.
+/// and a set of substrings that are disallowed in the directory name.
+pub fn allowed_bind_dirs(variant: &str) -> BindDirs {
+    let mut allowed_exact = HashSet::from(["/var/lib/containerd", "/var/lib/host-containerd"]);
     if variant.contains("k8s") {
-        allowed.insert("/var/lib/kubelet");
-        allowed.insert("/var/log/pods");
+        allowed_exact.insert("/var/lib/kubelet");
+        allowed_exact.insert("/var/log/pods");
     }
     if variant.contains("ecs") {
-        allowed.insert("/var/lib/docker");
-        allowed.insert("/var/log/ecs");
+        allowed_exact.insert("/var/lib/docker");
+        allowed_exact.insert("/var/log/ecs");
     }
-    allowed
+    let allowed_prefixes = HashSet::from(["/mnt/"]);
+    let disallowed_contains = HashSet::from(["..", "/mnt/.ephemeral"]);
+
+    BindDirs {
+        allowed_exact,
+        allowed_prefixes,
+        disallowed_contains,
+    }
 }
 
 /// scans the raid array to identify if it has been created already
--- a/sources/api/apiserver/src/server/mod.rs
+++ b/sources/api/apiserver/src/server/mod.rs
@@ -703,18 +703,22 @@ async fn list_ephemeral_storage_dirs(
 ) -> Result<HttpResponse> {
     let os_info = controller::get_os_info()?;
 
-    let allowed = ephemeral_storage::allowed_bind_dirs(&os_info.variant_id);
+    let allowed_dirs = ephemeral_storage::allowed_bind_dirs(&os_info.variant_id);
     let mut text_response = String::new();
-    for dir in &allowed {
+    for dir in &allowed_dirs.allowed_exact {
         text_response.push_str(dir);
         text_response.push('\n');
     }
 
-    let allowed: Vec<String> = allowed.iter().map(|x| String::from(*x)).collect();
+    let allowed: Vec<String> = allowed_dirs
+        .allowed_exact
+        .iter()
+        .map(|x| String::from(*x))
+        .collect();
     list_ephemeral_response(req, query, allowed, text_response).await
 }
 
-// Responds to a list request with the text or JSON resposne depending on the query format.
+// Responds to a list request with the text or JSON response depending on the query format.
 async fn list_ephemeral_response(
     req: HttpRequest,
     query: web::Query<HashMap<String, String>>,


sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -114,10 +114,13 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
 
     let mount_point = format!("/mnt/{}", EPHEMERAL_MNT);
     let mount_point = Path::new(&mount_point);
-    let allowed_dirs = allowed_bind_dirs(variant);
+    let (allowed_exact, allowed_prefixes) = allowed_bind_dirs(variant);
```
bcressey: Two concerns:
1. We should disallow paths starting with `/mnt/.ephemeral` (the value in `mount_point`) since that directory and its subdirectories are used for keeping state and siloing content of the different mount points.
2. We should reject any paths that contain `..` as that could traverse outside `/mnt` or over-mount other paths within it.

bcressey (commented):
Thanks for picking this up!

sources/api/apiserver/src/server/ephemeral_storage.rs
zaheer-merali: @bcressey thanks for the feedback, I have added checks for presence of .. and /mnt/.ephemeral in the next commit.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -114,10 +114,17 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
...
+        let exact_match = allowed_exact.contains(dir.as_str());
+        let prefix_match = allowed_prefixes
+            .iter()
+            .any(|prefix| dir.starts_with(prefix));
+        let disallowed_match = disallowed_contains
+            .iter()
+            .any(|contains| dir.contains(contains));
         ensure!(
-            allowed_dirs.contains(dir.as_str()),
+            (exact_match || prefix_match) && !disallowed_match,
```
bcressey: nit: might be slightly more efficient if we assume exact matches will never be disallowed:
```suggestion
            (exact_match || (prefix_match && !disallowed_match),
```

sources/api/apiserver/src/server/mod.rs
```diff
@@ -703,18 +703,19 @@ async fn list_ephemeral_storage_dirs(
 ) -> Result<HttpResponse> {
     let os_info = controller::get_os_info()?;
 
-    let allowed = ephemeral_storage::allowed_bind_dirs(&os_info.variant_id);
+    let (allowed_exact, _allowed_prefixes, _disallowed_contains) =
```
bcressey: nit: it's fine to just ignore and not use these returns:
```suggestion
    let (allowed_exact, _, _) =
```

bcressey (approved):
@zaheerm left a couple of nits but LGTM overall. Can you squash these commits together?

@KCSesh or @yeazelm could one of you review and/or test?

sources/api/apiserver/src/server/ephemeral_storage.rs
zaheer-merali: Thanks, done this in most latest commit.

sources/api/apiserver/src/server/mod.rs
zaheer-merali: Thanks, done this in most latest commit.

zaheer-merali:
@yeazelm Let me know if I should make any more changes to this PR. I marked it as draft initially but I'm happy with it being merged as is if the PR code is acceptable.

yeazelm:
Hey @zaheer-merali, I just saw this was out for testing, sorry for the delay! I'll test it out now and come back as soon as I'm done.

yeazelm:
I was able to validate this works (mostly) like we expect. I can now pass a directory in `/mnt` and it works, and if I pass something outside of the allowlist, it fails. The curious part is that I marked these as `essential = true` and the failure continued to boot. That isn't a bug in this PR, but we'll need to sort out why this didn't fail the boot. Here are my test results.

For a basic test: 
This input works:
```
commands = [
    ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/var/lib/containerd", "/var/lib/kubelet", "/var/log/pods", "/mnt/testdir"]
]
essential = true
mode = "always"
```
And the directory exists
```
bash-5.1# ls -al /mnt
total 4
drwxr-xr-x.  4 root root   39 Jan  6 19:04 .
drwxr-xr-x. 19 root root 4096 Jan  6 18:45 ..
drwxr-xr-x.  6 root root  102 Jan  6 19:04 .ephemeral
drwxr-xr-x.  2 root root   22 Jan  6 19:06 testdir
```

If I try to put something in `/var`, it fails:
```
commands = [
    ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/var/lib/containerd", "/var/lib/kubelet", "/var/log/pods", "/var/lib/testdir"]
]
essential = true
mode = "always"
```
Output
```
Jan 06 19:09:54 localhost systemd[1]: Starting Bootstrap Commands...
Jan 06 19:09:54 ip-192-168-62-156.us-west-2.compute.internal bootstrap-commands[1769]: 19:09:54 [INFO] Processing bootstrap command 'k8s-ephemeral-storage' ...
Jan 06 19:09:56 ip-192-168-62-156.us-west-2.compute.internal bootstrap-commands[1785]: Failed to initialize ephemeral storage: Failed POST request to '/actions/ephemeral-storage/bind': Status 500 when POSTing /actions/ephemeral-storage/bind: Unable to bind ephemeral storage: Invalid Parameter '/var/lib/testdir', specified bind directory not in allow list
Jan 06 19:09:56 ip-192-168-62-156.us-west-2.compute.internal bootstrap-commands[1769]: 19:09:56 [INFO] Successfully ran bootstrap command 'k8s-ephemeral-storage'
Jan 06 19:09:56 ip-192-168-62-156.us-west-2.compute.internal systemd[1]: Finished Bootstrap Commands.
```

I also tried to sneak in under a path and that is blocked:
```
[root@admin]# journalctl -D /.bottlerocket/rootfs/var/log/journal/ -u bootstrap-commands
-- Logs begin at Mon 2025-01-06 20:02:19 UTC, end at Mon 2025-01-06 20:04:08 UTC. --
Jan 06 20:02:28 localhost systemd[1]: Starting Bootstrap Commands...
Jan 06 20:02:28 ip-192-168-32-137.us-west-2.compute.internal bootstrap-commands[1768]: 20:02:28 [INFO] Processing bootstrap command 'k8s-ephemeral-storage' ...
Jan 06 20:02:30 ip-192-168-32-137.us-west-2.compute.internal bootstrap-commands[1784]: Failed to initialize ephemeral storage: Failed POST request to '/actions/ephemeral-storage/bind': Status 500 when POSTing /actions/ephemeral-storage/bind: Unable to bind ephemeral storage: Invalid Parameter '/var/lib/containerd/sneaky', specified bind directory not in allow list
Jan 06 20:02:30 ip-192-168-32-137.us-west-2.compute.internal systemd[1]: Finished Bootstrap Commands.
Jan 06 20:02:30 ip-192-168-32-137.us-west-2.compute.internal bootstrap-commands[1768]: 20:02:30 [INFO] Successfully ran bootstrap command 'k8s-ephemeral-storage'
```

With it being marked essential, we should be failing but this behavior is happening with the current code so we can track that problem separately.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -269,18 +276,27 @@ pub fn ephemeral_devices() -> Result<Vec<String>> {
...
-pub fn allowed_bind_dirs(variant: &str) -> HashSet<&'static str> {
-    let mut allowed = HashSet::from(["/var/lib/containerd", "/var/lib/host-containerd"]);
+/// varies based on the variant, a set of the prefixes of directories that are allowed to be bound.
+/// and a set of substrings that are disallowed in the directory name.
+pub fn allowed_bind_dirs(
+    variant: &str,
+) -> (
+    HashSet<&'static str>,
+    &'static [&'static str],
+    &'static [&'static str],
```
yeazelm: nit: can we get these to be consistent and use `HashSet` here?
```suggestion
    HashSet<&'static str>,
    HashSet<&'static str>,
```

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -269,18 +276,27 @@ pub fn ephemeral_devices() -> Result<Vec<String>> {
...
 /// allowed_bind_dirs returns a set of the directories that can be bound to ephemeral storage, which
-/// varies based on the variant
-pub fn allowed_bind_dirs(variant: &str) -> HashSet<&'static str> {
-    let mut allowed = HashSet::from(["/var/lib/containerd", "/var/lib/host-containerd"]);
+/// varies based on the variant, a set of the prefixes of directories that are allowed to be bound.
+/// and a set of substrings that are disallowed in the directory name.
+pub fn allowed_bind_dirs(
+    variant: &str,
+) -> (
+    HashSet<&'static str>,
```
yeazelm: This is getting complex enough of an interface we may want a separate type so we track which is which. Something like `BindDirs` would work so we could reason about them via the type system.
```
struct BindDirs {
    allowed_exact: HashSet<&'static str>,
    allowed_prefixes: HashSet<&'static str>, 
    disallowed_contains: HashSet<&'static str>,
}
```
This is a bit of a bigger overhaul to the code and we can do this clean up afterwards if you feel this is beyond your time/desire to accomplish too but I figured I would mention it here. This would allow the change in mod.rs to feel more idiomatic to the rest of the codebase.

sources/api/apiserver/src/server/ephemeral_storage.rs
zaheer-merali: pushed a commit to change to this

sources/api/apiserver/src/server/ephemeral_storage.rs
zaheer-merali: thanks @yeazelm  I agree. I pushed a change that does this.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -269,18 +282,28 @@ pub fn ephemeral_devices() -> Result<Vec<String>> {
...
+        allowed_exact.insert("/var/log/pods");
     }
     if variant.contains("ecs") {
-        allowed.insert("/var/lib/docker");
-        allowed.insert("/var/log/ecs");
+        allowed_exact.insert("/var/lib/docker");
+        allowed_exact.insert("/var/log/ecs");
+    }
+    let allowed_prefixes = HashSet::from(["/mnt/"]);
+    let disallowed_contains: = HashSet::from(["..", "/mnt/.ephemeral"]);
```
yeazelm: Looks like we have a typo and it either wants a type annotation here or to remove the `:`: https://github.com/bottlerocket-os/bottlerocket-core-kit/actions/runs/12640012198/job/35262690972?pr=320
```suggestion
    let disallowed_contains = HashSet::from(["..", "/mnt/.ephemeral"]);
```

zaheer-merali: thanks @yeazelm , that was a silly thing on my part...should be fixed now!

yeazelm: Thanks! That looks better!

yeazelm:
The changes look good, can we get the commits squashed again into one?

yeazelm (approved):
LGTM! Thanks for the contribution!

zaheer-merali:
@yeazelm it missed the struct having to be made public to allow the struct to be exposed to mod.rs. So I resquashed the 4 character change.

yeazelm:
> @yeazelm it missed the struct having to be made public to allow the struct to be exposed to mod.rs. So I resquashed the 4 character change.

I caught the same thing when testing the build and you beat me to it! Nice catch!z

yeazelm (approved):

PR #321, opened 2024-12-17
add GPG verification where possible

bcressey:
**Issue number:**
N/A

**Description of changes:**
Add a GPG verification step prior to unpacking sources for all packages where a signature is provided by the vendor.

**Testing done:**
Built locally.

**Terms of contribution:**


packages/amazon-vpc-cni-plugins/Cargo.toml
```diff
@@ -8,6 +8,10 @@ build = "../build.rs"
 [lib]
 path = "../packages.rs"
 
+[package.metadata.build-package]
+# Use the Git submodule commit from aws/amazon-ecs-agent rather than an upstream release.
+releases-url = "https://github.com/aws/amazon-ecs-agent/commits/master/amazon-vpc-cni-plugins"
```
KCSesh, line 13: The commits don't currently align across https://github.com/aws/amazon-ecs-agent/commits/master/amazon-vpc-cni-plugins and https://github.com/aws/amazon-vpc-cni-plugins/archive/be5214353252f8315a1341f4df9ffbd8cf69000c.

Should we point to the source package itself?

bcressey: > Should we point to the source package itself?

I tried to find a way to link to the commit ID for the submodule itself, and as far as I can tell that's not possible with the GitHub web UI. This links to the commit history for the submodule, which felt like the next best thing since one of the changes in that commit will show the new project commit ID.

I can drop if it that's too subtle, but people keep trying to update this package so I wanted a more clear statement on where the correct version is supposed to come from.

yeazelm (approved):

packages/linux-firmware/gpgkey-4CDE8575E547BF835FE15807A31B6BD72486CFD6.asc
```diff
@@ -0,0 +1,79 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
KCSesh: Is this supposed to be here? Seems it belongs in the Kernel-Kit only?

packages/microcode/gpgkey-4CDE8575E547BF835FE15807A31B6BD72486CFD6.asc
```diff
@@ -0,0 +1,79 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
KCSesh: Same comment as linux-firmware

packages/runc/Cargo.toml
```diff
@@ -16,6 +16,11 @@ url = "https://github.com/opencontainers/runc/releases/download/v1.1.14/runc.tar
 path = "runc-v1.1.14.tar.xz"
 sha512 = "71c21c2cff82402d937163e73805dc4f73eca948d5a05c2fe2aa0b1161438adf538caad2e3b98d9e0b786d7f04c01971a494a1fc3e24fe94fb76e64bf9453ad9"
 
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/opencontainers/runc/releases/download/v1.1.14/runc.tar.xz.asc"
```
KCSesh: We just merged runc-1.1.15 :) 
https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/323

packages/shim/gpgkey-8107B101A432AAC9FE8E547CA348D61BC2713E9F.asc
```diff
@@ -0,0 +1,119 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
KCSesh: Same as linux-firmware

KCSesh (commented):
LGTM - other than the remaining comments!

bcressey:
⬆️ force push to fix the stray files.

bcressey:
⬆️ force push for rebase

bcressey:
⬆️ force pushes to add the signature file for runc 1.1.15, and then to fix the hash

packages/runc/runc.spec
```diff
@@ -14,6 +14,11 @@ Summary: CLI for running Open Containers
 License: Apache-2.0
 URL: https://%{goimport}
 Source0: https://%{goimport}/releases/download/v%{gover}/%{gorepo}.tar.xz#/%{gorepo}-v%{gover}.tar.xz
+Source1: https://%{goimport}/releases/download/v%{gover}/%{gorepo}.tar.xz.asc#/%{gorepo}-v%{gover}.tar.xz.asc
+# Aleksa Sarai
+Source2: gpgkey-B64E4955B29FA3D463F2A9062897FAD2B7E9446F.asc
+# Kir Kolyshkin
+Source3: gpgkey-C2428CD75720FACDCF76B6EA17DE5ECB75A1100E.asc
```
bcressey, line 21: I can remove the unused key, but my reasoning for keeping both is that if each release is signed by one of these two, there will be less churn overall to list both and just switch which one is used for verification.

KCSesh: Sounds fair - we can always change our mind on that as we see how it evolves!

KCSesh (approved):

PR #323, opened 2024-12-17
Third party package updates

vigh-m:
**Description of changes:**
* Update the following packages
  * containerd to v1.7.24
  * soci-snapshotter to v0.8.0
  * runc to v1.1.15
  * cni-plugins to v1.6.1
* Add support for a `UPSTREAM_SOURCE_FALLBACK` env variable when building new packages in the core-kit

  Command:
  ```
  make ARCH=aarch64 UPSTREAM_SOURCE_FALLBACK=true && make ARCH=x86_64 UPSTREAM_SOURCE_FALLBACK=true
  ```

**Testing done:**
* Testsys tests for `aws-k8s-1.30` AMI:
  ```
  NAME                                         TYPE                         STATE                                           PASSED                     FAILED                    SKIPPED   BUILD ID                          LAST UPDATE
   x86-64-aws-k8s-130-quick                     Test                         passed                                               5                          0                       7198   a8df6eb2-dirty                    2024-12-17T19:36:18Z
   x86-64-aws-k8s-130                           Resource                     completed                                                                                                      a8df6eb2-dirty                    2024-12-17T19:34:29Z
  ```
* Also manually launched a java workload via `kubectl`

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ KIT ?= bottlerocket-core-kit
 UNAME_ARCH = $(shell uname -m)
 ARCH ?= $(UNAME_ARCH)
 VENDOR ?= bottlerocket
+UPSTREAM_SOURCE_FALLBACK = false
 
 ifeq ($(UNAME_ARCH), aarch64)
 	TWOLITER_SHA256=$(TWOLITER_SHA256_AARCH64)
@@ -40,8 +41,12 @@ update: prep
 fetch: prep
 	@$(TWOLITER) fetch --arch $(ARCH)
 
-build: fetch 
+build: fetch
+ifeq ($(UPSTREAM_SOURCE_FALLBACK), "false")
 	@$(TWOLITER) build kit $(KIT) --arch $(ARCH)
+else
+	@$(TWOLITER) build kit $(KIT) --arch $(ARCH) --upstream-source-fallback
+endif
 
 publish: prep
 	@$(TWOLITER) publish kit $(KIT) $(VENDOR)
--- a/packages/amazon-ecs-cni-plugins/Cargo.toml
+++ b/packages/amazon-ecs-cni-plugins/Cargo.toml
@@ -9,6 +9,9 @@ build = "../build.rs"
 path = "../packages.rs"
 
 [[package.metadata.build-package.external-files]]
+# This is locked against the version shipped in ecs-agent
+# Verify that the ecs-agent version shipped in bottlerocket tracks the same one here
+# https://github.com/aws/amazon-ecs-agent/releases
 url = "https://github.com/aws/amazon-ecs-cni-plugins/archive/53a8481891251e66e35847554d52a13fc7c4fd03/amazon-ecs-cni-plugins.tar.gz"
 sha512 = "e819c1aae509d19461999bf717d126b3e918b73dc6049e415c4911be6cb11159404bb45bb6c92cdfa16b5b30bb174731e972e3f2be44fa0b51bbc7a969049ab7"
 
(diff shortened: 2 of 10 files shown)


packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -1,11 +1,11 @@
 %global ecscni_goproject github.com/aws
 %global ecscni_gorepo amazon-ecs-cni-plugins
 %global ecscni_goimport %{ecscni_goproject}/%{ecscni_gorepo}
-%global ecscni_gitrev 53a8481891251e66e35847554d52a13fc7c4fd03
+%global ecscni_gitrev 7b4ec6016ab221469fa3abfc00ea7c05f236c26c
 
 Name: %{_cross_os}amazon-ecs-cni-plugins
-# https://github.com/aws/amazon-ecs-cni-plugins/blob/53a8481891251e66e35847554d52a13fc7c4fd03/VERSION#L1
-Version: 2020.09.0
+# https://github.com/aws/amazon-ecs-cni-plugins/blob/7b4ec6016ab221469fa3abfc00ea7c05f236c26c/VERSION#L1
```
ginglis13: We should keep this version aligned with the ecs-agent's. It was updated to 2024.09.0 in https://github.com/aws/amazon-ecs-agent/releases/tag/v1.89.2, but the core-kit ships ecs-agent v1.88.0

(I ran into this recently and admittedly should have left a breadcrumb comment here mentioning that we keep those versions aligned 😅 )

Makefile
```diff
@@ -11,6 +11,7 @@ KIT ?= bottlerocket-core-kit
 UNAME_ARCH = $(shell uname -m)
 ARCH ?= $(UNAME_ARCH)
 VENDOR ?= bottlerocket
+UPSTREAM_SOURCE_FALLBACK="false"
```
ginglis13: We can already set `BUILDSYS_UPSTREAM_SOURCE_FALLBACK` https://github.com/bottlerocket-os/twoliter/blob/2ae29fa73ffe5f1bb7cbcb2799a1e169b3478314/twoliter/embedded/Makefile.toml#L91

like `BUILDSYS_UPSTREAM_SOURCE_FALLBACK=true make`

larvacea: Nit: the other assignments in Makefile put spaces before and after the assignment operator. They also do not quote simple strings.

larvacea (approved):

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
vigh-m: Ack. I'll add a line in the spec or `Cargo.toml` to call this out more explicitly

Makefile
vigh-m: So, running `BUILDSYS_UPSTREAM_SOURCE_FALLBACK=true make` doesn't pass the setting into twoliter. To pass options to twoliter you would need to do something like 
```
make twoliter build kit bottlerocket-core-kit --arch x86_64 -e BUILDSYS_UPSTREAM_SOURCE_FALLBACK=true
```

ginglis13: Hm, maybe my comment only applies to `make twoliter ...` workflows. e.g., I will use

```
BUILDSYS_ARCH=x86_64 PACKAGE=glibc make twoliter build-package
```

To build just one package for a specific arch.

e: I tested w/ an ecs-agent bump and 

```
BUILDSYS_UPSTREAM_SOURCE_FALLBACK=true PACKAGE=ecs-agent make twoliter build-package
```

worked. But I see the value in adding this to simply `make` 👍

vigh-m:
⬆️ Addressing comments. Removed version bump for `aws-ecs-cni-plugins`

KCSesh (approved):
LGTM. Thanks for the Makefile change!

PR #327, opened 2024-12-19
systemd-cryptsetup dependency chain

mikn:
**Description of changes:**
This introduces a chain of dependencies for enabling systemd-cryptsetup. Initially, I did not compile the cryptsetup binary itself and I did not package the systemd-cryptenroll binary as we would embed them in our admin container and in our bootstrap system, but I guess it would be more useful to have them part of the primary system? Willing to discuss though.

I am also unsure whether the devmapper-event library is needed or not.

**Testing done:**
I have made sure the binaries don't give linking errors when invoked, but I have not tested primary functionality yet. Will happen later.

**Terms of contribution:**


mikn:
I am currently maintaining this in our variant repository as well, but would appreciate to get this upstream as discussed.

cbgbt:
Thanks for your contribution, and apologies for the latency -- things are a bit slow on account of the holidays. We'll have a look!

mikn:
That's okay! I have some patches to push to this after the holidays. There is an optional dependency on libargon2 also in cryptsetup which is not yet included. IT is a key derivation algorthim which has not yet been fips certified however. We want to use it ourselves as we have no fips requirements as such and it is a provably more secure key derivation algorithm than the pbkdf2 algo which is fips approved.
I would need some help to allow the FIPS switch to work in this dependency chain.

packages/cryptsetup/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+force-upstream = true
+
+[build-dependencies]
+glibc = { path = "../glibc" }
+device-mapper = { path = "../device-mapper" }
+json-c = { path = "../json-c" }
+libblkid = { path = "../util-linux" }
+libuuid = { path = "../util-linux" }
+libpopt = { path = "../libpopt" }
+systemd = { path = "../systemd" }
```
bcressey: Since systemd also depends on cryptsetup, this would create a dependency cycle, which we'd need to break in one of two ways:
1. adding a `systemd-bootstrap` package or similar (`libsystemd`, ...) with the same sources and a trimmed down build, just enough to bootstrap cryptsetup
2. adding systemd headers and libraries to the SDK - effectively doing the bootstrap there

From a quick glance at the sources, it looks like the dependency is only used to find the path to the systemd tmpfiles.d directory, which we can specify as an argument to the configure script instead.

packages/aws-lc-fips/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "aws-lc-fips"
```
bcressey, line 2: Ideally this package would be named `libcrypto` since it provides that shared library.

packages/aws-lc-fips/aws-lc-fips.spec
```diff
@@ -0,0 +1,77 @@
...
+
+BuildRequires: %{_cross_os}glibc-devel
+
+Requires: %{_cross_os}glibc
+
+%description
+AWS-LC is a general-purpose cryptographic library maintained by the
+AWS Cryptography team for AWS and their customers. It іs based on code
+from the Google BoringSSL project and the OpenSSL project. This version
+includes FIPS support.
```
bcressey, line 20: I prefer this pattern to minimize the amount of words written in spec files that no one will ever read:
```suggestion
%description
%{summary}.
```

packages/cryptsetup/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+url = "https://www.kernel.org/pub/linux/utils/cryptsetup/v2.7/cryptsetup-2.7.5.tar.xz"
+sha512 = "13eca93cdb00a143d2ca60b6f66ede5adc4072ca0c4bfebd8454a3541e69d269fcdb4afc97ad799e87a999b2bd46c1f31fa924a3d616d72a3337970b1e718d55"
+force-upstream = true
+
+[build-dependencies]
+glibc = { path = "../glibc" }
+device-mapper = { path = "../device-mapper" }
+json-c = { path = "../json-c" }
+libblkid = { path = "../util-linux" }
+libuuid = { path = "../util-linux" }
```
bcressey: `buildsys` only knows about the crate or top-most package name, so this should just be:
```suggestion
util-linux = { path = "../util-linux" }
```

packages/aws-lc-fips/aws-lc-fips.spec
```diff
@@ -0,0 +1,77 @@
...
+%build
+%set_cross_build_flags
+
+mkdir -p aws-lc-build
+cd aws-lc-build
+
+%cross_cmake ../ \
+    -GNinja \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+    -DBUILD_SHARED_LIBS=ON \
```
bcressey, line 41: If nothing needs `libssl.so`, only `libcrypto.so`, I'd prefer to disable it to cut down on surface area for potential vulnerabilities:
```suggestion
    -DBUILD_SHARED_LIBS=ON \
    -DBUILD_LIBSSL=OFF \
```

packages/libtss2/libtss2.spec
```diff
@@ -0,0 +1,111 @@
+# Skip check-rpaths since we expect them for systemd.
+%global __brp_check_rpaths %{nil}
+
```
bcressey:
```suggestion
```

packages/libtss2/libtss2.spec
```diff
@@ -0,0 +1,111 @@
...
+ --with-runstatedir=%{_rundir}
+)
+
+%set_cross_build_flags
+
+%cross_configure "${CONFIGURE_OPTS[@]}"
+
+# Fix Rpath issues
+#sed -i 's|^hardcode_libdir_flag_spec=.*|hardcode_libdir_flag_spec=""|g' libtool
+#sed -i 's|^runpath_var=LD_RUN_PATH|runpath_var=DIE_RPATH_DIE|g' libtool
```
bcressey: There's a macro for this:
```suggestion
%force_disable_rpath
```

packages/systemd/systemd.spec
```diff
@@ -3,7 +3,7 @@
 
 Name: %{_cross_os}systemd
 Version: 252.22
-Release: 1%{?dist}
+Release: 1.%{?dist}
```
bcressey:
```suggestion
Release: 1%{?dist}
```

bcressey (commented):
Back from vacation now and taking a look!

Are you OK if I push minor fix commits to your branch for this PR? Not sure if you enabled that flag but it could save some round trips for cleanup stuff that might otherwise annoy you.

mikn:
Firstly, thank you for the review!

> Back from vacation now and taking a look!
> 
> Are you OK if I push minor fix commits to your branch for this PR? Not sure if you enabled that flag but it could save some round trips for cleanup stuff that might otherwise annoy you.

I actually don't know how to enable that! But you are welcome to!

wrt disabling building libssl.so I think that is a bit hairy, as systemd, with the `-Dopenssl=true` enables a few other things (such as DNSSEC, and thus SSL, in resolved). It is however required to pass in to enable tpm2 support.

I am unsure whether tss2 requires libssl or just libcrypto. I can try to see where things break, but it will probably be next week as we are in the middle of delivery this week.

packages/cryptsetup/Cargo.toml
mikn: Ah yes, that's true! For some reason this doesn't break the world for me, but I did build... one of them first and have probably never cleaned up my build space since.

packages/aws-lc-fips/Cargo.toml
mikn: I did name it this as we are also building libssl as per the comment below. I think for us, we do want libssl if possible, as we will probably set up remote systemd-journal with TLS soonish also (for audit logging).
We could split up the install packages though?

packages/aws-lc-fips/aws-lc-fips.spec
mikn: Okay, I tried building now and it does activate a fair few things in systemd (to have openssl available) - I spotted that udev did not build, neither did the shared library.
I do not really have time to dive into this and craft a patch to only enable the crypto dependencies in Systemd (and not the libssl ones), but I do apologise for polluting your otherwise very pristine distro with libssl. :smiling_face_with_tear:

packages/cryptsetup/Cargo.toml
mikn: I did remove it completely now as per your recommendation.

mikn:
Okay - as mentioned in previous comments, we would prefer if `libssl` also ends up in Bottlerocket together with `libcrypto` despite how much effort you spent avoiding that very fact. :face_holding_back_tears: 
I think it is up to you @bcressey and/or the rest of the team at AWS if this is something you would like to support or not.

For clarity; we would probably not spend the time to craft out the patch required soon to not enable `libssl` dependencies in `systemd` ourselves, as we most likely would need it down the road for securing our audit log system. (Currently the design is using remote `systemd-journald` as it is one of the very few integrated log storage, filtering and forwarding solutions)

Let me know how we should move forward. :pray:

bcressey:
January's been kind of a rough month but this has been on my mind to review and I expect to dig in soon. Thanks for your patience!

> For clarity; we would probably not spend the time to craft out the patch required soon to not enable libssl dependencies in systemd ourselves, as we most likely would need it down the road for securing our audit log system.

I've been mulling over this. I could see a path to including `libssl` from `aws-lc`, but there are some internal conversations that have to happen to ensure that we're brought into the relevant conversations in the event of a security issue.

If you're OK with an incremental approach, I can work on systemd patches for this PR to avoid assumptions that `libcrypto` means `libssl` is also available, and then figure out how to tackle the `libssl` and systemd remote logging pieces separately.

mikn:
> January's been kind of a rough month but this has been on my mind to review and I expect to dig in soon. Thanks for your patience!
> 
> > For clarity; we would probably not spend the time to craft out the patch required soon to not enable libssl dependencies in systemd ourselves, as we most likely would need it down the road for securing our audit log system.
> 
> I've been mulling over this. I could see a path to including `libssl` from `aws-lc`, but there are some internal conversations that have to happen to ensure that we're brought into the relevant conversations in the event of a security issue.
> 
> If you're OK with an incremental approach, I can work on systemd patches for this PR to avoid assumptions that `libcrypto` means `libssl` is also available, and then figure out how to tackle the `libssl` and systemd remote logging pieces separately.

I apologise for not replying to this before now, as January was rough for you, February has been particularly busy for us!

It should be mentioned that we also use `systemd-creds` for our secrets management (we bind the secrets to the machine by using the TPM flags for `systemd-creds`) and it does also have a dependency on `libopenssl` here; https://github.com/systemd/systemd/blob/main/src/creds/meson.build#L10 . It was sadly not as easy as opening `creds.c` to figure out if it was just on `crypto.h` or on other parts of `openssl` also.
If you could also include that in your grafting scope (and include `systemd-creds` in Bottlerocket proper), it would mean that we could drop our fork of the systemd packaging, which I guess would be the major reason to upstream this. :)

That said, I would be very grateful for an incremental approach, and if you want to take the time to carve out the patches needed for `systemd` to not have to include `libssl.so` for now, that would be very appreciated!

bcressey:
I've been digging into this over the past day and had a couple of questions.

Did you prefer the external libargon2 over the internal one for some reason? The `phc-winner-argon2` project hasn't had a release since 2019. cryptsetup has made additional fixes to its bundled copy over the years, so it seems to be more actively maintained.

[Increase Argon2 LUKS2 default](https://gitlab.com/cryptsetup/cryptsetup/-/commit/3c2f92a7afc2ae4c99fa937f1b189bc1375374ad) from seven years ago added a note that the bundled implementation was slow, but [Add optimized Argon2 SSE code](https://gitlab.com/cryptsetup/cryptsetup/-/commit/ba384d15d2c880af6a0c5555b6b47505cc281cc2) added the optimized code from the reference implementation, but didn't remove the note about the slow implementation. However, I assume the performance is similar if `--enable-internal-sse-argon2` is passed at build time.

The `--enable-fips` flag for cryptsetup turns on a runtime check for whether the system is in FIPS mode, and if so it disables   functions other than PBKDF2. If I set `--with-luks2-pbkdf=pbkdf2` so that the default function works whether or not the system is in FIPS mode, does that create problems for your usage?

I was struggling with the side effects of telling systemd that `openssl` was available and I hit on what feels like a nice alternative: turn the `cryptsetup` and `tpm2` features on but leave `openssl` off. We may still need to patch `meson.build` so that something that we want to link with libcrypto knows it's available, but that's likely less annoying than fixing random compatibility issues in `systemd-resolved` and `systemd-dissect`.

My working branch is [here](https://github.com/bcressey/bottlerocket-core-kit/commits/cryptsetup/). I have a few things to fix, like double-checking licenses and dealing with systemd packaged files, but it looks pretty good to me.

mikn:
I must admit that I did not dig into the implementation of argon2 in
cryptsetup beyond what they mentioned in their documentation, but I do
trust your judgement and see no need to have the library available
separately if you think the cryptsetup internal version is as good/better.

Wrt setting pbkdf to pbkdf2 per default, that's not a problem - we
explicitly encrypt it with argon2 flags.

That's great! Really appreciate it. Have you had time to look at
systemd-creds also? Otherwise perhaps it is an exercise I could engage in
once this lands if you're interested in enabling it in mainline?

bcressey:
> Have you had time to look at systemd-creds also?

I'll take a look at that next!

I've posted separate PRs for some prerequisites to this change. #438 will make the eventual changes to `systemd.spec` much easier to review, and #444 has just the new packages so they can be reviewed in isolation.

I kicked the tires on this over the weekend and made a couple tweaks.

I brought back the `cryptsetup` binary (and `libpopt`) since systemd doesn't seem to have a tool to turn a block device into a new LUKS2 volume, and without that it's harder to demonstrate that everything works end-to-end. `libpopt` is maintained by the RPM project now so it's in a better spot than I realized.

I also disabled FIPS for `cryptsetup`, since AWS-LC's `FIPS_mode()` function just returns a build-time constant, which causes `cryptsetup` to always assume it's in FIPS mode and to disable argon2.

mikn:
@bcressey This is amazing work and I am very happy that you are going through this! We are very excited to be part of the Bottlerocket community given how welcome you all have made us feel and we are now convinced we made the right choice building our foundations on your OS.

Please let me know if there is anything you'd like me to validate, check or implement now, otherwise I am eager to validate that this works for us once all pieces are in place. :pray:

For now our plan looks like this given the the work you are doing:
1. Wait for #444 to merge
2. update core kit to version that includes new packages
3. remove our own packages of the now core kit included packages (except `aws-lc`, as `systemd-creds` requires it also)
4. make sure all works
5. Rebase `systemd.spec` changes (which for us also includes `systemd-creds`)
6. Wait for `systemd-creds` to land
7. Remove large portions of our internal fork

@bcressey we have some other forks and packages, specifically we decided to fork the `release` package because we are dealing with the `bottlerocket-private` partition differently (to enable have data on there for mounting the `bottlerocket-data` partition), and we also forked the `selinux-policy` package as we are adding in a domain for `Firecracker` specifically (`container_kvm_t`).
Would you like me to start up community discussions or issues on how to move those modifications into mainline also?

bcressey:
@mikn that plan seems fine to me. The other dependencies should be in the next core kit, out in a day or two. For your `systemd-creds` package, you should be able to drop your `aws-lc` package too, that just turned into `libcrypto` per library-packaging naming conventions.

I'd hoped to get `systemd-creds` landed too, along with the `systemd-cryptsetup` change, but I haven't had the bandwidth to close on one of these three paths:
1. work out how to avoid `libsystemd` linking with `libcrypto`, so we don't have to ship that by default yet
2. ship `libcrypto`, but disable `libssl` and kick that can down the road until we figure out the path for `libcurl` and `systemd-journal-remote`
3. just give up and ship all of it

> we are dealing with the bottlerocket-private partition differently

@ytsssun was looking at [#4348](https://github.com/bottlerocket-os/bottlerocket/issues/4348), it's something we want to do anyhow for various reasons (improved boot time).

> we also forked the selinux-policy package as we are adding ... `container_kvm_t`

I expect parts of this can be upstreamed pretty easily - at least, the type declarations and putting it in the sets with the rest of the container types. Happy to look at a patch or a PR for that.

bcressey:
@mikn it looks like all of the remote journal programs - `systemd-journal-upload`, `systemd-journal-remote`, `systemd-journal-gatewayd` - use gnutls rather than openssl. Not sure what functionality is lost for `systemd-journal-upload` without it, or if it still builds, but I assume it can get `libssl` by way of `libcurl`.

I'm working on a hybrid of (2) and (3) above with that in mind.

mikn:
Yes, the TLS support would generally be through libcurl is my understanding
also. It isn't urgent nor a patch I'm carrying to enable remote components
yet. I guess this goes in parallel with the work to enable
systemd-cryptsetup?

bcressey:
I tried to push my commits to your branch but GitHub said I didn't have permissions.

I've opened a new pull request with the remaining pieces to actually enable `systemd-cryptsetup`. Please take a look at #518.

PR #329, opened 2024-12-21
Use kubernetes device-ownership setting

vigh-m:
*Description of changes:*

* Extends `packages/containerd/containerd-config-toml_k8s_containerd_sock` to support the new setting being added in https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/69. This is to support the same setting in the containerd CRI plugin as described [here](https://kubernetes.io/blog/2021/11/09/non-root-containers-and-devices/)

Testing done:

* The setting shows up as expected in an AMI built using this `settings-sdk`
  ```
  # apiclient get settings.kubernetes.device-ownership-from-security-context
  {
    "settings": {
      "kubernetes": {
        "device-ownership-from-security-context": false
      }
    }
  }
  ```

* Changing the setting works as defined:
  ```
  [root@admin]# apiclient set settings.kubernetes.device-ownership-from-security-context="true"
  [root@admin]# apiclient get settings.kubernetes.device-ownership-from-security-context
  {
    "settings": {
      "kubernetes": {
        "device-ownership-from-security-context": true
      }
    }
  }
  [root@admin]# sheltie cat /etc/containerd/config.toml | grep device
  device_ownership_from_security_context = true
  ```

* The setting has the desired effect on device ownership
    * When `true`:
      ``` 
      # kubectl exec -it single-node-test -- bash
      ubuntu@single-node-test:/$ ls -lah /dev/ | grep neuron
      crw-------. 1 ubuntu 2000 244, 0 Dec 20 00:25 neuron0
      ```
    * When `false`:
      ``` 
      # kubectl exec -it single-node-test -- bash
      ubuntu@single-node-test:/$ ls -lah /dev/ | grep neuron
      crw-------. 1 root root 244, 0 Dec 19 20:03 neuron0
      ```

**Terms of contribution:**


sources/Cargo.toml
```diff
@@ -210,23 +210,23 @@ x509-parser = "0.16"
...
-tag = "bottlerocket-settings-plugin-v0.1.0"
+git = "https://github.com/vigh-m/bottlerocket-settings-sdk"
+tag = "test-1"
 version = "0.1.0"
 
 [workspace.dependencies.settings-extension-oci-defaults]
-git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.6.0"
+git = "https://github.com/vigh-m/bottlerocket-settings-sdk"
+tag = "test-1"
```
ytsssun: Looks like we have not yet merged the sdk PR. Could you change the status of this core-kit PR to draft?

sources/deny.toml
```diff
@@ -92,5 +92,6 @@ allow-git = [
     # The settings SDK is currently provided as a git dependency,
     # We will allow it as an exception until the following is resolved:
     # https://github.com/bottlerocket-os/bottlerocket-settings-sdk/issues/18
-    "https://github.com/bottlerocket-os/bottlerocket-settings-sdk",
+    # "https://github.com/bottlerocket-os/bottlerocket-settings-sdk",
+    "https://github.com/vigh-m/bottlerocket-settings-sdk",
```
ytsssun: This will also be updated once sdk change is merged.

sources/Cargo.toml
vigh-m: Ack. I had these changes in a separate commit to make it easy to remove them later. I'll move to a draft as well.

yeazelm (approved):

ytsssun (approved):

PR #333, opened 2024-12-29
corndog: use systemd-sysctl for sysctl options

### LINKED ISSUE #4314: Persist userdata sysctl changes to /etc/sysctl.d
**What I'd like:**

Persist userdata-provided sysctls to `/etc/sysctl.d` in a format where they can survive through `systemd-sysctl.service` restarts

**Background**

Currently, passing in sysctls via userdata results in the apiclient running `sysctl` commands directly.

We pass in sysctls into our Bottlerocket AMI via userdata, but also run Cilium, which has a `sysctl-fixup` [container](https://github.com/cilium/cilium/blob/v1.16.2/tools/sysctlfix/main.go) that creates a config file (`/etc/sysctl.d/99-zzz-override_cilium.conf` by default) and restarts the host `sysctld` systemd unit file. When this happens, the userdata-provided sysctls are clobbered.

Bootstrap commands/containers explicitly don't have the capability to modify config files in `/etc`, so there's no good solution to this issue currently.

- Admittedly none

aetimmes:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket/issues/4314

**Description of changes:**

Change corndog's sysctl function to utilize systemd-sysctl instead of writing values ephemerally to tmpfs, so that other system functions (eg Ciliums sysctl-fixup) don't completely clobber the sysctls set by corndog.

**Testing done:**

Unit tests added inline

**Terms of contribution:**

### DIFF
--- a/sources/api/corndog/src/main.rs
+++ b/sources/api/corndog/src/main.rs
@@ -8,7 +8,7 @@ corndog also provides a settings generator for hugepages, subcommand "generate-h
 */
 
 use bottlerocket_modeled_types::{Lockdown, SysctlKey};
-use log::{debug, error, info, trace, warn};
+use log::{debug, info, trace, warn};
 use serde::{Deserialize, Serialize};
 use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
 use snafu::ResultExt;
@@ -18,10 +18,13 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::string::String;
 use std::{env, process};
+use tempfile::NamedTempFile;
 
-const SYSCTL_PATH_PREFIX: &str = "/proc/sys";
 const LOCKDOWN_PATH: &str = "/sys/kernel/security/lockdown";
 const DEFAULT_CONFIG_PATH: &str = "/etc/corndog.toml";
+const SYSCTL_CONFIG_DIR: &str = "/etc/sysctl.d";
+const SYSCTL_CONFIG_FILENAME: &str = "95-corndog.conf";
+const SYSTEMD_SYSCTL_BIN: &str = "/usr/lib/systemd/systemd-sysctl";
 const NR_HUGEPAGES_PATH_SYSCTL: &str = "/proc/sys/vm/nr_hugepages";
 /// Number of hugepages we will assign per core.
 /// See [`compute_hugepages_for_efa`] for more detail on the computation consideration.
@@ -43,6 +46,22 @@ struct KernelSettings {
     sysctl: Option<HashMap<SysctlKey, String>>,
 }
 
+/// Trait for executing system commands
+trait CommandExecutor {
+    fn execute(&self, cmd: &str) -> std::io::Result<(std::process::ExitStatus, String)>;
+}
+
+/// Real command executor that runs actual system commands
+struct SystemCommandExecutor;
+
+impl CommandExecutor for SystemCommandExecutor {
+    fn execute(&self, cmd: &str) -> std::io::Result<(std::process::ExitStatus, String)> {
+        let output = process::Command::new(cmd).output()?;
+        let stderr = String::from_utf8_lossy(&output.stderr).into_owned();
+        Ok((output.status, stderr))
+    }
+}
+
 /// Main entry point.
 fn run() -> Result<()> {
     let args = parse_args(env::args());
@@ -56,7 +75,7 @@ fn run() -> Result<()> {
             let kernel = get_kernel_settings(args.config_path)?;
             if let Some(sysctls) = kernel.sysctl {
                 debug!("Applying sysctls: {:#?}", sysctls);
-                set_sysctls(sysctls);
+                set_sysctls(sysctls)?;
             }
         }
         "lockdown" => {
@@ -93,34 +112,75 @@ where
     toml::from_str(config_str.as_str()).context(error::DeserializationSnafu)
 }
 
-fn sysctl_path<S>(name: S) -> PathBuf
-where
-    S: AsRef<str>,
-{
-    let name = name.as_ref();
-    let mut path = PathBuf::from(SYSCTL_PATH_PREFIX);
-    path.extend(name.replace('.', "/").split('/'));
-    trace!("Path for {}: {}", name, path.display());
-    path
-}
-
-/// Applies the requested sysctls to the system.  The keys are used to generate the appropriate
-/// path, and the value its contents.
-fn set_sysctls<K>(sysctls: HashMap<K, String>)
+/// Generate sysctl config file content from key-value pairs
+fn generate_sysctl_config<K>(sysctls: &HashMap<K, String>) -> String
 where
     K: AsRef<str>,
 {
+    let mut config_content = String::new();
     for (key, value) in sysctls {
         let key = key.as_ref();
-        let path = sysctl_path(key);
-        if let Err(e) = fs::write(path, value) {
-            // We don't fail because sysctl keys can vary between kernel versions and depend on
-            // loaded modules.  It wouldn't be possible to deploy settings to a mixed-kernel fleet
-            // if newer sysctl values failed on your older kernels, for example, and we believe
-            // it's too cumbersome to have to specify in settings which keys are allowed to fail.
-            error!("Failed to write sysctl value '{}': {}", key, e);
-        }
+        // Prepend a dash to keys that don't already start with one to ensure systemd-sysctl
+        // doesn't fail on applying any sysctls.
+        //
+        // We don't fail because sysctl keys can vary between kernel versions and depend on
+        // loaded modules.  It wouldn't be possible to deploy settings to a mixed-kernel fleet
+        // if newer sysctl values failed on your older kernels, for example, and we believe
+        // it's too cumbersome to have to specify in settings which keys are allowed to fail.
+        let prefix = if !key.starts_with('-') { "-" } else { "" };
+        config_content.push_str(&format!("{}{} = {}\n", prefix, key, value.trim()));
     }
+    config_content
+}
+
+/// Write sysctl config to a file, using a temporary file and atomic rename
+fn persist_sysctl_config(
+    config_content: &str,
+    config_dir: &str,
+    config_filename: &str,
+) -> Result<PathBuf> {
+    // Create a temporary file in the sysctl config directory
+    let tempfile = NamedTempFile::new_in(config_dir).context(error::CreateTempFileSnafu {
+        path: PathBuf::from(config_dir),
+    })?;
+
+    // Write the config to the temporary file
+    fs::write(tempfile.path(), config_content).context(error::WriteTempFileSnafu)?;
+
+    // Construct the final path and atomically move the temporary file to it
+    let config_path = Path::new(config_dir).join(config_filename);
+    tempfile
+        .persist(&config_path)
+        .context(error::PersistTempFileSnafu {
+            path: config_path.clone(),
+        })?;
+
+    Ok(config_path)
+}
+
+/// Apply sysctl settings using the given systemd-sysctl binary
+fn apply_sysctl_config(systemd_sysctl_bin: &str, executor: &dyn CommandExecutor) -> Result<()> {
+    let (status, output) = executor
+        .execute(systemd_sysctl_bin)
+        .context(error::RunSystemdSysctlSnafu)?;
+
+    if !status.success() {
+        error::SystemdSysctlFailedSnafu { status, output }.fail()?;
+    }
+
+    debug!("Successfully applied sysctl settings");
+    Ok(())
+}
+
+/// Applies the requested sysctls to the system by writing them to a config file and using systemd-sysctl.
+/// The keys are used to generate the appropriate path, and the value its contents.
+fn set_sysctls<K>(sysctls: HashMap<K, String>) -> Result<()>
+where
+    K: AsRef<str>,
+{
+    let config_content = generate_sysctl_config(&sysctls);
+    persist_sysctl_config(&config_content, SYSCTL_CONFIG_DIR, SYSCTL_CONFIG_FILENAME)?;
+    apply_sysctl_config(SYSTEMD_SYSCTL_BIN, &SystemCommandExecutor)
 }
 
 /// Generate the hugepages setting for defaults.
@@ -301,6 +361,7 @@ fn main() {
 mod error {
     use snafu::Snafu;
     use std::io;
+    use std::path::PathBuf;
 
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub(super)))]
@@ -334,22 +395,138 @@ mod error {
 
         #[snafu(display("Logger setup error: {}", source))]
         Logger { source: log::SetLoggerError },
+
+        #[snafu(display("Failed to create temporary file in {}: {}", path.display(), source))]
+        CreateTempFile { path: PathBuf, source: io::Error },
+
+        #[snafu(display("Failed to write sysctl config to temporary file: {}", source))]
+        WriteTempFile { source: io::Error },
+
+        #[snafu(display("Failed to move temporary file to {}: {}", path.display(), source))]
+        PersistTempFile {
+            path: PathBuf,
+            source: tempfile::PersistError,
+        },
+
+        #[snafu(display("Failed to run systemd-sysctl: {}", source))]
+        RunSystemdSysctl { source: io::Error },
+
+        #[snafu(display(
+            "systemd-sysctl failed with exit code: {} and output: {}",
+            status,
+            output
+        ))]
+        SystemdSysctlFailed {
+            status: std::process::ExitStatus,
+            output: String,
+        },
     }
 }
 type Result<T> = std::result::Result<T, error::Error>;
 
 #[cfg(test)]
 mod test {
+    use super::*;
+    use std::os::unix::process::ExitStatusExt;
+    use std::process::ExitStatus;
+    use tempfile::TempDir;
     use test_case::test_case;
 
-    use super::*;
+    /// Mock command executor for testing
+    struct MockCommandExecutor {
+        success: bool,
+        output: String,
+    }
+
+    impl MockCommandExecutor {
+        fn new(success: bool) -> Self {
+            Self {
+                success,
+                output: String::from("mock output"),
+            }
+        }
+    }
+
+    impl CommandExecutor for MockCommandExecutor {
+        fn execute(&self, _cmd: &str) -> std::io::Result<(std::process::ExitStatus, String)> {
+            Ok((
+                ExitStatus::from_raw(if self.success { 0 } else { 1 }),
+                self.output.clone(),
+            ))
+        }
+    }
+
+    // Helper to create a temporary directory for testing
+    fn setup_test_dir() -> TempDir {
+        TempDir::new().expect("Failed to create temp directory")
+    }
 
     #[test]
-    fn no_traversal() {
+    fn test_generate_sysctl_config() {
+        let mut sysctls = HashMap::new();
+        sysctls.insert("net.ipv4.ip_forward", "1".to_string());
+        sysctls.insert("vm.swappiness", "60".to_string());
+        sysctls.insert("kernel.pid_max", "4194304 ".to_string()); // Note the trailing space
+
+        let config = generate_sysctl_config(&sysctls);
+
+        // Split into lines and sort for consistent comparison
+        let mut lines: Vec<&str> = config.lines().collect();
+        lines.sort();
+
         assert_eq!(
-            sysctl_path("../../root/file").to_string_lossy(),
-            format!("{}/root/file", SYSCTL_PATH_PREFIX)
+            lines,
+            vec![
+                "-kernel.pid_max = 4194304",
+                "-net.ipv4.ip_forward = 1",
+                "-vm.swappiness = 60"
+            ]
         );
+    }
+
+    #[test]
+    fn test_generate_sysctl_config_empty() {
+        let sysctls: HashMap<String, String> = HashMap::new();
+        let config = generate_sysctl_config(&sysctls);
+        assert_eq!(config, "");
+    }
+
+    #[test]
+    fn test_persist_sysctl_config() {
+        let temp_dir = setup_test_dir();
+        let config = "net.ipv4.ip_forward = 1\nvm.swappiness = 60\n";
+        let filename = "test-sysctl.conf";
+
+        let config_path =
+            persist_sysctl_config(config, temp_dir.path().to_str().unwrap(), filename).unwrap();
+
+        // Verify the config file was written correctly
+        let written_config = fs::read_to_string(&config_path).unwrap();
+        assert_eq!(written_config, config);
+    }
+
+    #[test]
+    fn test_persist_sysctl_config_invalid_dir() {
+        let config = "net.ipv4.ip_forward = 1\n";
+        let result = persist_sysctl_config(config, "/nonexistent", "test.conf");
+        assert!(matches!(result, Err(error::Error::CreateTempFile { .. })));
+    }
+
+    #[test]
+    fn test_apply_sysctl_config_success() {
+        let executor = MockCommandExecutor::new(true);
+        let result = apply_sysctl_config(SYSTEMD_SYSCTL_BIN, &executor);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_apply_sysctl_config_failure() {
+        let executor = MockCommandExecutor::new(false);
+        let result = apply_sysctl_config(SYSTEMD_SYSCTL_BIN, &executor);
+        assert!(matches!(
+            result,
+            Err(error::Error::SystemdSysctlFailed { .. })
+        ));
     }
 
     #[test]
@@ -392,5 +569,37 @@ mod test {
     fn test_check_for_existing_hugepages(existing_value: String, expected_hugepages: String) {
         let actual_hugepages = check_for_existing_hugepages(existing_value);
         assert_eq!(actual_hugepages, expected_hugepages);
+    }
+
+    #[test]
+    fn test_generate_sysctl_config_dash_prefix() {
+        let mut sysctls = HashMap::new();
+        // Test with a mix of keys - some with dash prefix, some without
+        sysctls.insert("net.ipv4.ip_forward", "1".to_string());
+        sysctls.insert("-vm.swappiness", "60".to_string());
+        sysctls.insert("-kernel.pid_max", "4194304".to_string());
+
+        let config = generate_sysctl_config(&sysctls);
+
+        // Validate each line is a valid sysctl line
+        for line in config.lines() {
+            // Split the line into key and value
+            let parts: Vec<&str> = line.splitn(2, " = ").collect();
+            assert_eq!(
+                parts.len(),
+                2,
+                "Line should contain key-value pair: {}",
+                line
+            );
+
+            // Verify key portion starts with dash and contains no additional dashes
+            let key = parts[0];
+            assert!(key.starts_with('-'), "Key should start with dash: {}", key);
+            assert!(
+                !key[1..].contains('-'),
+                "Key should not contain additional dashes after prefix: {}",
+                key
+            );
+        }
     }
 }
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -1804,6 +1804,7 @@ dependencies = [
  "serde_json",
  "simplelog",
  "snafu",
+ "tempfile",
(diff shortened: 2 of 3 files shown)


aetimmes:
I don't speak Rust natively so if there's anything in here that's not idiomatic, feel free to let me know.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,104 @@ where
...
+    // Write new config
+    let mut config = String::new();
     for (key, value) in sysctls {
-        let key = key.as_ref();
-        let path = sysctl_path(key);
-        if let Err(e) = fs::write(path, value) {
-            // We don't fail because sysctl keys can vary between kernel versions and depend on
-            // loaded modules.  It wouldn't be possible to deploy settings to a mixed-kernel fleet
-            // if newer sysctl values failed on your older kernels, for example, and we believe
-            // it's too cumbersome to have to specify in settings which keys are allowed to fail.
```
aetimmes, line 120: The current implementation of this PR doesn't maintain the functionality mentioned in this comment, but we could pretty easily achieve this by prepending `-` to all the sysctls passed in by default (which would then remove the need for the backup-modify-restore shuffle, which might make things simpler overall), or by documenting the ability of users to do so. Open to feedback/direction here.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,104 @@ where
...
-            error!("Failed to write sysctl value '{}': {}", key, e);
+        config.push_str(&format!("{}={}\n", key.as_ref(), value));
+    }
+    
+    fs::write(SYSCTL_CONFIG, config).context(error::WriteSysctlConfigSnafu)?;
```
aetimmes: I need to do some double-checking around:

- Does this actually append to the file as written, or does it clobber it?
- If we append and we write duplicate sysctls to the same config file, how does systemd-sysctl handle that? (Does the most recent sysctl win?)

I haven't yet been able to get a local build working (turns out my hard drive is smaller than I thought!) but I'll investigate these shortly.

aetimmes: Answers:

- This clobbers the file (must fix)
- systemd-sysctl will use the last value in a file for the same sysctl. For example, a config file with the contents:

```
dev.raid.speed_limit_max = 199999 
dev.raid.speed_limit_max = 199998 
garbage                           
dev.raid.speed_limit_max = 199997
```

and running `systemctl restart systemd-sysctl` will set `dev.raid.speed_limit_max` to `199997` and report an error when attempting to parse the "garbage" line - the return code of the systemd unit will be 1 but all of the correctly specified sysctl k/v pairs will have been applied.

systemd-sysctl also will deduplicate multiple lines and only write when there are changes to be made, so in theory endlessly appending to a file _should_ be a reasonable way to do this.

aetimmes: Output from running systemd-sysctl with the above file in support of the above conclusions:

```
bash-5.1# SYSTEMD_LOG_LEVEL=debug /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/systemd-sysctl
Parsing /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/sysctl.d/50-default.conf
Parsing /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/sysctl.d/50-pid-max.conf
Parsing /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/sysctl.d/80-release.conf
Parsing /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/sysctl.d/90-kubelet.conf
Parsing /etc/sysctl.d/90-primary_interface.conf
Parsing /etc/sysctl.d/99-zzz-override_cilium.conf
Overwriting earlier assignment of net/ipv4/conf/all/rp_filter at '/etc/sysctl.d/99-zzz-override_cilium.conf:7'.
Parsing /etc/sysctl.d/99-zzz-zzz-testing.conf
Overwriting earlier assignment of dev/raid/speed_limit_max at '/etc/sysctl.d/99-zzz-zzz-testing.conf:2'.
/etc/sysctl.d/99-zzz-zzz-testing.conf:3: Line is not an assignment, ignoring: garbage
Overwriting earlier assignment of dev/raid/speed_limit_max at '/etc/sysctl.d/99-zzz-zzz-testing.conf:4'.
...
Setting '/proc/sys/dev/raid/speed_limit_max' to '199997'
No change in value '199997', suppressing write
```

ytsssun:
Sorry for the delay, @aetimmes. And thanks so much for opening the PR for this. I will take a look at this PR and get back to you soon. I will also help verify the behavior.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+    // Ensure value doesn't contain newlines which could break the config file format
+    if value.contains('\n') {
+        return error::InvalidSysctlValueSnafu {
+            key: key.to_string(),
+            value: value.to_string()
+        }.fail();
+    }
+
+    Ok(())
+}
```
ytsssun: We perform validation when parsing the key from userdata already. See https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/0cf857357bc799bf036b5f5bf865d3c4af4e9256/bottlerocket-settings-models/modeled-types/src/shared.rs#L703-L733.

We may need to account for a more strict validation there though.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+
+/// Applies the requested sysctls to the system using systemd-sysctl.
+fn set_sysctls<K>(sysctls: HashMap<K, String>) -> Result<()>
 where
     K: AsRef<str>,
 {
+    // TODO: make the priority configurable?
+    const SYSCTL_CONFIG: &str = "/etc/sysctl.d/95-corndog.conf";
```
ytsssun: Let's move this constant up (somewhere in L26-L28) so that we don't recreate the constant every time we invoke this function.

Also maybe add `_PATH` as the suffix suggesting it is the path of the config not the actual content of the config?

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+
+    // Validate all entries first
+    for (key, value) in &sysctls {
+        validate_sysctl_entry(key, value)?;
+    }
```
ytsssun: Again validation already happened when parsing the keys.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+
+    // Create config string
+    let mut config = String::new();
     for (key, value) in sysctls {
```
ytsssun: We had conversation that iterating through the map entry is not ideal since it does not guarantee an ordering here. 

One way to address that is to sort the sysctl entries first by the keys and then iterate based on the sorted keys. You can do something like

```
    let mut sysctl_vec: Vec<_> = sysctls
        .iter()
        .map(|(k,v)| (k.as_ref(), v))
        .collect();
    sysctl_vec.sort();
    for (key, value) in sysctl_vec {
         ...
    }
```

On another note, we may also want to preserve the order the entries are specified in the userdata. But that would require a different struct for `sysctls` that can preserve the order. For example [BTreeMap](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html), but it has performance trade-offs since the insertions, lookups, and deletions would not be as fast.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+
+    // Append to file, creating it if it doesn't exist
+    let mut file = fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(SYSCTL_CONFIG)
+        .context(error::OpenSysctlConfigSnafu)?;
+
+    file.write_all(config.as_bytes())
+        .context(error::WriteSysctlConfigSnafu)?;
```
ytsssun:
```
fs::write("path/to/file.txt", config).context(error::WriteSysctlConfigSnafu)?;
```

This would try to create the file if it does not exist.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+
+    // Restart systemd-sysctl to apply changes
+    let status = Command::new("systemctl")
+        .args(["restart", "systemd-sysctl"])
+        .status()
+        .context(error::SystemctlSnafu)?;
```
ytsssun: You don't really need to restart the whole sysctld service.
```
/usr/lib/systemd/systemd-sysctl /etc/sysctl.d/99-corndog.conf
```
Would apply this specific file to sysctld.

And could you wrap the binary path and args into constants? Something like below and put it along with the definition of the `SYSCTL_CONFIG_PATH`.
```
const SYSTEMD_SYSCTL_BIN: &str = "/usr/lib/systemd/systemd-sysctl";
```

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+
+    if !status.success() {
+        return error::SystemctlFailedSnafu.fail();
+    }
```
ytsssun: When hitting this case, it would be also good to print the errors from the output. So I would get the output of the command and check the status and the stderr.

```
    let output = Command::new(SYSTEMD_SYSCTL_BIN)
        .arg(SYSCTL_CONFIG_PATH)
        .output()
        .context(ExecutionSnafu{command: format!("{} {}", SYSTEMD_SYSCTL_BIN, SYSCTL_CONFIG_PATH)})?;
    if !output.status.success() {
        warn!(
            "Failed to apply sysctl config: {}",
            String::from_utf8_lossy(&output.stderr) # account for invalid utf8 encoding
        );
    }
```

ytsssun:
Left some comments, the change looks good overall. I will run some test with the change to verify the behavior. I will get back to you on that.

sources/api/corndog/src/main.rs
aetimmes: I'll need to confirm this, but I believe only applying the specific file may not necessarily respect the ordering of the rest of the config files, so `sysctl restart systemd-sysctl` is correct IMO.

For example, the Cilium sysctl.conf file has a prefix of `99-zzz`, so applying the `95` priority corndog file by itself may clobber any entries that are also set in the Cilium conf file.

If we kept `sysctl restart systemd-ctl`, would we still want these wrapped into constants?

sources/api/corndog/src/main.rs
aetimmes: We probably need to adjust in a couple directions - systemd-sysctl also allows for wildcards in the keys, which it looks like the current format doesn't allow for. I can open a separate PR for that.

sources/api/corndog/src/main.rs
aetimmes: Good point - we definitely want to maintain ordering becuase of the interaction between wildcards and overrides. I'm not _overly_ worried about performance here since the upper bound of the size of this structure is probably on the order of thousands.

aetimmes: So it looks like using a BTreeMap would complicate the serde bits a fair amount since BTreeMap doesn't implement several of the interfaces needed by serde - how do we feel about a `Vec` here instead?

sources/api/corndog/src/main.rs
bcressey: What is the purpose of the additional validation here? Is something being allowed in as a setting that would be rejected by `systemd-sysctl`?

The settings API is intended to be the first and best line of defense - the only time checks should be needed at apply-time is if there's invalid input that the API isn't currently capable of detecting, like two mutually exclusive or dependent settings.

If we do need some extra validation for compatibility, then it'd be better to warn and remove the setting from the output, so that it's not a hard error. But even that is a potential compatibility break on upgrade that's best avoided, if there were sysctls that would have been applied before but that can't be applied now because of more restrictive rules.

sources/api/corndog/src/main.rs
```diff
@@ -104,23 +109,83 @@ where
...
+
+    Ok(())
+}
```
bcressey: Restarting a service can lead to unexpected behavior if units depend on that service, or if for whatever reason it hasn't started yet or failed to start successfully. This is also using the blocking form which can lead to trouble in early boot.

It'd be better to avoid all of those issues and just invoke `systemd-sysctl` directly. I read your other comment and understand that we can't pass just the one config file, which makes sense, but the unit just runs `/usr/lib/systemd/systemd-sysctl` with no arguments which we can do here also.

sources/api/corndog/src/main.rs
```diff
@@ -18,6 +18,11 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::string::String;
 use std::{env, process};
+use std::process::Command;
+use tempfile::NamedTempFile;
```
bcressey, line 21: This doesn't appear to be used, but it would be good to use it and follow the usual pattern:
1. create a tempfile in the same directory as the target file
2. write all the content to it
3. persist the tempfile to the final target once it's ready

(We want to avoid changing the original file if some other process might have it open, and we also want to avoid repeatedly appending to the same file on subsequent runs, which could lead to unexpected behavior or resource exhaustion.)

sources/api/corndog/src/main.rs
aetimmes: > What is the purpose of the additional validation here

No purpose - I was unaware of the bottlerocket-settings-sdk repo when I took the first pass at this PR; I agree that the validation in this MR should be removed. 

I would still like to update the validation in the bottlerocket-settings-sdk repo to allow for wildcards where possible since they're supported by systemd-sysctl in general, but given that that PR would be allowing for extra functionality, I don't consider it a blocker for this change.

sources/api/corndog/src/main.rs
aetimmes: That makes sense to me; I'll adjust accordingly.

sources/api/corndog/src/main.rs
aetimmes: > want to avoid repeatedly appending to the same file on subsequent runs

To be explicitly clear - if I make a new api client request to set sysctl userdata on a host that already has an existing `/etc/sysctl.d/95-corndog.conf` file, which of these is the desired behavior?

- append the new sysctls to the existing list of sysctls (allowing for duplicates; gets expensive if many calls are made but systemd-sysctl handles deduping for us)
- upsert the new sysctls into the existing list of sysctls (if a previously detected entry exists, remove the existing key and append the new k/v - more expensive but no duplicates)
- replace the old sysctls with the new ones (seems like a breaking change)

In the case of the first two options, is the suggestion above to make a copy of the existing sysctl conf file as the tempfile (as opposed to operating on the original)?

(Apologies for being pedantic; I just want to make sure I understand the requirements here)

aetimmes:
Thanks for the reviews! The feedback all makes sense (modulo the couple of open questions I have) - I'm weirdly struggling to get my local Rust install to play nice with the pre-existing tests so my follow-up will be slightly delayed while I debug my machine.

sources/api/corndog/src/main.rs
ytsssun: Taking a step back. I think this request is not necessarily blocking the PR. It is ok if you want to leave this part untouched. It can be tracked separately.

ytsssun:
@aetimmes Let me know if you need help setting up the environment. Or I can even help with testing the change by building a test AMI. I can either do that with your updated PR, or you can point me to your fork, whichever way works for you.

aetimmes:
Okay - turns out my "environment issues" were just me being bad at Rust and not understanding mocking.

I just force-pushed a refactor that:
- uses temp files instead of appending
- decomposes the top-level function into config gen/config persistence/command execution
- does some wrapping/dependency injection to allow for running tests without `#[cfg(not(test))]` declarations

This is ready for re-review; with the following notes:
- In lieu of an answer to [this](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/333#discussion_r1915642184) question above, I just assumed "we want to clobber the old config file"
- I punted on the ordered deserialization for now but that probably does block the inclusion of wildcards in validation for now (not the end of the world for me personally since we're not currently using them in userdata 😎)
- Do we want to use systemd-sysctl for hugepages too? (Perhaps using a separate conf file?)

sources/api/corndog/src/main.rs
bcressey: > To be explicitly clear - if I make a new api client request to set sysctl userdata on a host that already has an existing /etc/sysctl.d/95-corndog.conf file, which of these is the desired behavior?

The expected behavior is the third option:
> replace the old sysctls with the new ones (seems like a breaking change)

`/etc` is considered to be entirely under the host's control, so the only copy of `/etc/sysctl.d/95-corndog.conf` that should exist is the one generated by `corndog`. Hence there's no need to handle the case of unexpected modifications. (Even privileged containers cannot write to `/etc` without opting into the `super_t` SELinux label for break-glass access.)

I'm curious about what you mean by a breaking change, though - possibly you've spotted an edge case that I haven't, if it's something other than that we might be clobbering a file in `/etc` that we didn't previously "own".

aetimmes: (Caveat: I'm making the assumption here that `/etc/sysctl.d/95-corndog.conf` is the source of truth and that there's no intermediate datastore where these sysctls are accumulated.)

The scenario I'm envisioning is less clobbering someone else's file and more about unintendedly reverting previous changes. For example:

- default kernel sysctl config files are installed in the image
- userdata is provided on node provisioning
  - corndog turns this into `/etc/sysctl.d/95-corndog.conf` and runs `systemd-sysctl`, which runs through all the conf files (kernel+userdata) in priority order
- another apiclient sysctl request is made while the system is live
  - this overwrites the original userdata-generated `/etc/sysctl.d/95-corndog.conf`
  - `systemd-sysctl` runs through all the conf files again (kernel+corndog)

If a specific sysctl is set by both the default kernel sysctl conf files AND the original userdata request, my (potentially incorrect) understanding is that if we replace `/etc/sysctl.d/95-corndog.conf` on every invocation, any subsequent apiclient request will have the effect of unsetting any previous corndog-set sysctls that are also overridden by the kernel-default config files (with priority <95) - whereas with the previous behavior, since we were sending bits into procfs directly, apiclient sysctl commands didn't have any side effects on any other sysctls.

On the other hand, having apiclient append to the existing `95-corndog.conf` file would behave relatively in line with the previous behavior (with the caveat that it could no longer override sysctls set by conf files with higher priority, which is part of the purpose of this change).

sources/api/corndog/src/main.rs
bcressey: > we could pretty easily achieve this by prepending `-` to all the sysctls passed in by default 

I missed this note during my first review. Please go ahead and make this change.

In addition to the concern called out in the comment - that the same set of sysctls might not be supported across a fleet that uses the same user data but different kernels at the node level - another risk is that the sysctls might be valid at the time they are added by an admin running `apiclient`, but not be valid when the node is rebooted, if they depend on certain kernel modules to be loaded.

It's also possible for the sysctl to be completely invalid, but for the admin to miss the error before rebooting the node, which would again lead to boot failure.

sources/api/corndog/src/main.rs
bcressey: > any subsequent apiclient request will have the effect of unsetting any previous corndog-set sysctls that are also overridden by the kernel-default config files (with priority <95)

The current API doesn't allow you to delete existing settings, only to replace them with some other value. So if a sysctl is provided by userdata, it will continue to be in `95-corndog.conf` with either its original value or with the new value from `apiclient`. The config file that corndog reads from is also re-rendered in its entirety whenever a sysctl setting is added or updated - it won't just be the sysctls from the last committed transaction.

Those two properties should make it so `95-corndog.conf` always has the same or more sysctl keys, with the latest values from the API. If it ever overrides something because of its priority order, then it will keep doing so.

ytsssun:
Hi @aetimmes , just wanted to give you an update. I was able to run a test with your code to verify the behavior. And for initial behavioral testing, it works as expected! I will go through the code tomorrow again to see if there is any outstanding comments that we need to address.

My verification steps:
1. Built the test AMI and launch a BR node with the AMI.
2. Set the sysctl setting via apiclient.
```
bash-5.1# apiclient set --json '{"kernel": {"sysctl": {"user.max_user_namespaces": "16384"}}}'
bash-5.1# apiclient get settings.kernel
{
  "settings": {
    "kernel": {
      "lockdown": "integrity",
      "sysctl": {
        "user.max_user_namespaces": "16384",
        "vm/nr_hugepages": "0"
      }
    }
  }
}
```
3. Verify that the sysctl value is applied and that the sysctld config is persisted.
```
bash-5.1# cat /proc/sys/user/max_user_namespaces
16384

bash-5.1# cat /etc/sysctl.d/95-corndog.conf
vm/nr_hugepages = 0
user.max_user_namespaces = 16384
```

4. Restart the sysctld service
```
systemctl restart systemd-sysctl.service
```

5. Verify that the sysctl value remains the same and the sysctld config file and API settings are all consistent:
```
bash-5.1# cat /proc/sys/user/max_user_namespaces
16384
bash-5.1# apiclient get settings.kernel.sysctl
{
  "settings": {
    "kernel": {
      "sysctl": {
        "user.max_user_namespaces": "16384",
        "vm/nr_hugepages": "0"
      }
    }
  }
}
bash-5.1# cat /etc/sysctl.d/95-corndog.conf
vm/nr_hugepages = 0
user.max_user_namespaces = 16384
```

sources/api/corndog/src/main.rs
ytsssun: Hi @aetimmes, could you address this by prepending the - to the entries to ensure we achieve same behavior after migrating to this new approach?

If you don't mind, I can help add that change in to speed things up. Code looks good to me otherwise.

aetimmes:
Ahh, I'm sorry - I've had that commit locally and forgot to push it. I'll
do so shortly!

aetimmes:
Okay - just rebased against latest develop and pushed changes to prepend the dashes; this is ready for final review.

sources/api/corndog/src/main.rs
```diff
@@ -93,34 +112,72 @@ where
...
-    let mut path = PathBuf::from(SYSCTL_PATH_PREFIX);
-    path.extend(name.replace('.', "/").split('/'));
-    trace!("Path for {}: {}", name, path.display());
-    path
+    let mut config_content = String::new();
+    for (key, value) in sysctls {
+        let key = key.as_ref();
+        // We prepend a dash to keys that don't already start with one.
+        // This ensures that systemd-sysctl doesn't fail on applying any sysctls set by
+        // userdata, which would make managing a homogenous set of nodes much more
```
ytsssun: nit: can we be more general here and stick to the original comments? We can apply sysctl settings not just from userdata. Any `apiclient set` would trigger the `set_sysctls` function.

sources/api/corndog/src/main.rs
```diff
@@ -393,4 +567,31 @@ mod test {
         let actual_hugepages = check_for_existing_hugepages(existing_value);
         assert_eq!(actual_hugepages, expected_hugepages);
     }
+
+    #[test]
+    fn test_generate_sysctl_config_dash_prefix() {
```
ytsssun, line 575: This test failed.

sources/api/corndog/src/main.rs
```diff
@@ -334,24 +392,140 @@ mod error {
...
+    }
+
+    // Helper to create a temporary directory for testing
+    fn setup_test_dir() -> TempDir {
+        TempDir::new().expect("Failed to create temp directory")
+    }
 
     #[test]
-    fn no_traversal() {
+    fn test_generate_sysctl_config() {
```
ytsssun, line 465: This test failed.

ytsssun (commented):
You have a few unit tests that needs fixing
```
test test::brackets ... ok
test test::test_apply_sysctl_config_failure ... ok
test test::test_check_for_existing_hugepages::_1_to_string_0_to_string_expects ... ok
test test::test_compute_hugepages_for_efa ... ok
test test::test_generate_sysctl_config ... FAILED
test test::no_brackets ... ok
test test::test_generate_sysctl_config_dash_prefix ... FAILED
test test::test_apply_sysctl_config_success ... ok
test test::test_check_for_existing_hugepages::_100_to_string_100_to_string_expects ... ok
test test::test_check_for_existing_hugepages::_0_to_string_0_to_string_expects ... ok
test test::test_generate_sysctl_config_empty ... ok
test test::test_check_for_existing_hugepages::_to_string_0_to_string_expects ... ok
test test::test_check_for_existing_hugepages::_abc_to_string_0_to_string_expects ... ok
test test::test_persist_sysctl_config_invalid_dir ... ok
test test::test_persist_sysctl_config ... ok

failures:

---- test::test_generate_sysctl_config stdout ----
thread 'test::test_generate_sysctl_config' panicked at api/corndog/src/main.rs:474:9:
assertion `left == right` failed
  left: ["-kernel.pid_max = 4194304", "-net.ipv4.ip_forward = 1", "-vm.swappiness = 60"]
 right: ["kernel.pid_max = 4194304", "net.ipv4.ip_forward = 1", "vm.swappiness = 60"]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::test_generate_sysctl_config_dash_prefix stdout ----
thread 'test::test_generate_sysctl_config_dash_prefix' panicked at api/corndog/src/main.rs:585:13:
Line should not start with double dash: --double.dash.key = 123


failures:
    test::test_generate_sysctl_config
    test::test_generate_sysctl_config_dash_prefix

test result: FAILED. 13 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass `--bin corndog`
```

Can you make sure you run `cargo test` under the /corndog directory to validate?

Another thing is, can you run `cargo fmt` to format the code? 

I have left some minor feedback on the code comment.

I also have verified that the code works
```
bash-5.1# cat /etc/sysctl.d/95-corndog.conf
-vm/nr_hugepages = 0
-user.max_user_namespaces = 16384
```

I should be able to approve the PR once you address the feedbacks!

sources/api/corndog/src/main.rs
aetimmes: Fixed in https://github.com/bottlerocket-os/bottlerocket-core-kit/commit/28da3d0d0c28a0aa74cc433e8b6321bfd2808e3b

sources/api/corndog/src/main.rs
aetimmes: Fixed in https://github.com/bottlerocket-os/bottlerocket-core-kit/commit/28da3d0d0c28a0aa74cc433e8b6321bfd2808e3b

sources/api/corndog/src/main.rs
aetimmes: Fixed in https://github.com/bottlerocket-os/bottlerocket-core-kit/commit/28da3d0d0c28a0aa74cc433e8b6321bfd2808e3b

aetimmes:
Okay - comment was reverted, tests fixed, and code formatted. _Now_ it should be good to go.
 
```
$ cargo test
   Compiling corndog v0.1.0 (/Users/andrewtimmes/code/src/github/bottlerocket-core-kit-personal/sources/api/corndog)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1.43s
     Running unittests src/main.rs (/Users/andrewtimmes/code/src/github/bottlerocket-core-kit-personal/sources/target/debug/deps/corndog-23777922fba7b07d)

running 15 tests
test test::no_brackets ... ok
test test::test_apply_sysctl_config_failure ... ok
test test::brackets ... ok
test test::test_apply_sysctl_config_success ... ok
test test::test_check_for_existing_hugepages::_0_to_string_0_to_string_expects ... ok
test test::test_check_for_existing_hugepages::_100_to_string_100_to_string_expects ... ok
test test::test_check_for_existing_hugepages::_1_to_string_0_to_string_expects ... ok
test test::test_check_for_existing_hugepages::_abc_to_string_0_to_string_expects ... ok
test test::test_check_for_existing_hugepages::_to_string_0_to_string_expects ... ok
test test::test_compute_hugepages_for_efa ... ok
test test::test_generate_sysctl_config ... ok
test test::test_generate_sysctl_config_dash_prefix ... ok
test test::test_generate_sysctl_config_empty ... ok
test test::test_persist_sysctl_config_invalid_dir ... ok
test test::test_persist_sysctl_config ... ok

test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

```

bcressey (approved):
Nice, thank you!

ytsssun:
I gave it another round of testing just to make sure. Approving! Thanks a lot on the contribution!

ytsssun (approved):

aetimmes:
Thanks for the reviews and for bearing with me while I stumbled my way through Rust!

PR #294, opened 2024-12-04
Host container migrations

vyaghras:
**Issue number:**

Closes #

**Description of changes:**
In case of host container upgrade in BoB repo, we need to write migrations. By this PR we are incorperating the concept of weak and strong setting, where a weak setting will be deleted on upgrade and downgrade and default setting will be populated from the settings-defaults. 

```
migrator: remove all the weak setting and settings-generator
apiserver: add version 2 for /tx and /metadata/settings-generator 
apiclient: update README to document version 2 of /tx API 
datastore: support committing metadata from transactions 
models: add strength enum and Settings generator struct 
constants: Use version 2 API to get setting-generators 
openapi: Add version 2 for /tx and /metadata/settings-generator 
settings-committer: change transaction API endpoint to /v2/tx 
sundog: Parse settings generators as a table 
storewolf: enable parsing setting-generator metadata as table
```

**Testing done:**
Refer https://gist.github.com/vyaghras/cc7391ade3b276b223d1814a8770eea7

**Terms of contribution:**
By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #335, opened 2025-01-02
Conditionalize source and mode in Bootstrap container template

vyaghras:
**Issue number:**

Closes #

**Description of changes:**
Add a option in Bootstrap-containers binary to render the source for the Bootstrap container when unavailable. We will use the provided source if it is present.

**Testing done:**
- Confirm that template works.

**Terms of contribution:**

### DIFF
--- a/packages/os/bootstrap-containers-toml
+++ b/packages/os/bootstrap-containers-toml
@@ -5,8 +5,12 @@ std = { version = "v1", helpers = ["if_not_null"] }
 {{#if_not_null settings.bootstrap-containers}}
 {{#each settings.bootstrap-containers}}
 ["{{@key}}"]
+{{#if_not_null this.source}}
 source = "{{{this.source}}}"
+{{/if_not_null}}
+{{#if_not_null this.mode}}
 mode = "{{{this.mode}}}"
+{{/if_not_null}}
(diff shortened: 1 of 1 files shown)


vyaghras:
This change is blocked by https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/294 .

packages/os/bootstrap-containers-toml
```diff
@@ -5,7 +5,9 @@ std = { version = "v1", helpers = ["if_not_null"] }
 {{#if_not_null settings.bootstrap-containers}}
 {{#each settings.bootstrap-containers}}
 ["{{@key}}"]
+{{#if_not_null this.source}}
 source = "{{{this.source}}}"
+{{/if_not_null}}
 mode = "{{{this.mode}}}"
```
bcressey, line 12: PR title should be updated. If we need this change, should we also conditionalize the rendering of `mode`? It's strange to assume that's the only one that will always be defined.

bcressey (approved):
For the commit message:
> We need to make source and mode optional.

Can you say more about why? What's changing or what was wrong about the original assumptions that these fields would be set?

cbgbt (approved):

PR #350, opened 2025-01-15
Fix errors with apiserver and bootstrap-commands

vigh-m:
[Relevant Comment](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/320#issuecomment-2573821166)

**Description of changes:**
- Added a step to the `bind` command in the apicient/server to normalize the input by trimming the trailing "/"
- Explicitly check the command output in `bootstrap-commands` and throw an error on failure. Also adds a warning to the journal log on command failure

**Testing done:**
- Tested launching a container with the following userdata
  ```
  commands = [
    ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/mnt/", "/mnt/testDir", "/var/lib/containerd", "/var/lib/kubelet", "/var/log/. pods"]
  ]
  essential = true
  mode = "always"
  ```
  It now fails to boot when trying to mount `/mnt/`

- Tested with `essential = false`and the boot completes but logs the error as expected

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/ephemeral_storage.rs
+++ b/sources/api/apiserver/src/server/ephemeral_storage.rs
@@ -118,6 +118,12 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
         _ => format!("{}{}", RAID_DEVICE_DIR, RAID_DEVICE_NAME),
     };
 
+    // Normalize input by trimming trailing "/"
+    let dirs: Vec<String> = dirs
+        .into_iter()
+        .map(|dir| dir.trim_end_matches("/").to_string())
+        .collect();
+
     let mount_point = format!("/mnt/{}", EPHEMERAL_MNT);
     let mount_point = Path::new(&mount_point);
     let allowed_dirs = allowed_bind_dirs(variant);
--- a/sources/bootstrap-commands/src/main.rs
+++ b/sources/bootstrap-commands/src/main.rs
@@ -36,7 +36,7 @@ prior to `preconfigured.target` while `host-containers@.service` which is a requ
 running "exec" commands are launched after preconfigured.target.
 */
 
-use log::info;
+use log::{info, warn};
 use serde::Deserialize;
 use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
 use snafu::{ensure, OptionExt, ResultExt};
@@ -100,10 +100,15 @@ where
 {
     let mut command = Command::new(bin_path);
 
-    command
+    let output = command
         .args(args)
-        .status()
+        .output()
         .context(error::ExecutionFailureSnafu { command })?;
+
+    ensure!(
+        output.status.success(),
+        error::CommandFailureSnafu { bin_path, output }
+    );
 
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -118,6 +118,9 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
         _ => format!("{}{}", RAID_DEVICE_DIR, RAID_DEVICE_NAME),
     };
 
+    // Normalize input by trimming trailing "/"
+    dirs.iter_mut().for_each(|dir| *dir = dir.trim_end_matches("/").to_string());
```
ytsssun: Nit: I think we are fine for now since the input `dirs` are only going to be used within this function. But this introduced a side effect of the input - the dirs will be modified from the original userdata. If we want to keep the input "dirs" immutable, we can introduce a local var (not a huge overhead given the size of the command args is going to me pretty small) or perform the trim in the loop.

ytsssun:
Left a minor comment, the change looks good!

sources/api/apiserver/src/server/ephemeral_storage.rs
yeazelm: I think this is fine to make it mutable but we should update the docs to call out that we recommend you not include a training`/` and we'll just strip it anyway, so folks aren't surprised by the mutation.

yeazelm (approved):

ytsssun (approved):

KCSesh (approved):

yeazelm (approved):

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -118,10 +118,16 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
         _ => format!("{}{}", RAID_DEVICE_DIR, RAID_DEVICE_NAME),
     };
 
+    // Normalize input by trimming trailing "/"
+    let normalized_dirs: Vec<String> = dirs
```
bcressey: This is a case where I'd just create a new local with the same name to shadow the previous binding and ensure that it's not subsequently used by mistake:
```suggestion
    let dirs: Vec<String> = dirs
```

Otherwise, we have to keep the context in mind that "dirs is unsafe" and "normalized_dirs is safe" for subsequent uses in this function.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -118,10 +118,16 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
...
+        .into_iter()
+        .map(|dir| dir.trim_end_matches("/").to_string())
```
bcressey, line 124: Nice, `trim_end_matches` vs `strip_suffix` is indeed the better choice here.

bcressey (approved):

vigh-m:
⬆️ changed `normalized_dirs` to instead shadow `dirs` as recommended.

ytsssun (approved):

PR #362, opened 2025-01-24
updata: fix fetching max version in remove-update

### LINKED ISSUE #366: `updata` reporting wrong max version in remove-update command
**Platform I'm building on:**
x86_64

`updata` to report correct max version available in the tuf repo. (In my case it is supposed to be `1.32.0`)

**What actually happened:**
`updata` reported wrong value. (In my case it showed `1.18.0`)

1. Download the tuf manifest using `tuftool`
```
tuftool download \
    --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.29/x86_64/" \
    --targets-url "https://updates.bottlerocket.aws/targets" \
    --root ./default.root.json \
    --target-name manifest.json \
    ./aws-k8s-1.29/x86_64/tuf_in
```

2. Call `updata -- remove-update`
```
updata -- remove-update ./aws-k8s-1.29/x86_64/tuf_in/manifest.json \
  --arch x86_64 \
  --version 1.30.0 \
  --variant aws-k8s-1.29
20:08:47 [INFO] Update x86_64-aws-k8s-1.29-1.30.0 removed. Current maximum version: 1.18.0     <== Wrong version 
``` 

Sample manifest.json

```
{
  "updates": [
    {
      "variant": "aws-k8s-1.29",
      "arch": "x86_64",
      "version": "1.18.0",
      "max_version": "1.32.0",
      ...
    },
    ....
    {
      "variant": "aws-k8s-1.29",
      "arch": "x86_64",
      "version": "1.30.0",
      "max_version": "1.32.0",
      ...
    },
    {
      "variant": "aws-k8s-1.29",
      "arch": "x86_64",
      "version": "1.31.0",
      "max_version": "1.32.0",
      ...
    },
    {
      "variant": "aws-k8s-1.29",
      "arch": "x86_64",
      "version": "1.32.0",
      "max_version": "1.32.0",
      ...
    }
  ],
  "migrations": {
     ...
  }
```

The `max_version` in the updates field is showing correct data. But the code is referencing the `version` field of the [first item](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/updater/updog/src/bin/updata.rs#L121-L125) of the updates, which is why the max version is not correctly captured. The fix could be sort the manifest and get the latest version from the manifest version, or we can simply use the `max_version` field.

ytsssun:
**Issue number:**

Closes #366

**Description of changes:**
- When running the remove-update command it would report the max version available in the manifest. But it is checking the wrong order so that the latest version is not correctly reported. This is a simple fix for that.

**Testing done:**
Before the fix:
```
cargo run --bin updata -- remove-update "${MANIFEST_PATH}/aws-k8s-1.29/x86_64/tuf_in/manifest.json" \
    --arch x86_64\
    --version 1.30.0 \
    --variant aws-k8s-1.29

22:56:06 [INFO] Update x86_64-aws-k8s-1.29-1.30.0 removed. Current maximum version: 1.18.0
```

After:
```
cargo run --bin updata -- remove-update "${MANIFEST_PATH}/aws-k8s-1.29/x86_64/tuf_in/manifest.json" \
    --arch x86_64\
    --version 1.30.0 \
    --variant aws-k8s-1.29

22:56:06 [INFO] Update x86_64-aws-k8s-1.29-1.30.0 removed. Current maximum version: 1.31.0
```

**Terms of contribution:**

### DIFF
--- a/sources/updater/updog/src/bin/updata.rs
+++ b/sources/updater/updog/src/bin/updata.rs
@@ -121,7 +121,7 @@ impl RemoveUpdateArgs {
         if let Some(current) = manifest.updates.first() {
             info!(
                 "Update {}-{}-{} removed. Current maximum version: {}",
-                self.arch, self.variant, self.image_version, current.version
+                self.arch, self.variant, self.image_version, current.max_version
             );
         } else {
             info!(


sources/updater/updog/src/bin/updata.rs
```diff
@@ -118,7 +118,7 @@ impl RemoveUpdateArgs {
         });
         // Note: We don't revert the maximum version on removal
         update_metadata::write_file(&self.file, &manifest)?;
-        if let Some(current) = manifest.updates.first() {
+        if let Some(current) = manifest.updates.last() {
```
yeazelm: Is this just not sorted? I would expect to need to sort it to always get the max version?

ytsssun: It is sorted in a reversed order. I will adjust the code to enforce a sort here so that we get deterministic result.

bcressey:
> Closes # N/A

Can you open an issue with more details about the bug you're fixing?

"max version" is (incorrectly) associated with each update's metadata instead of being set once at the top level of `manifest.json`. This is a longstanding bug that would need a repo pivot to fix.

In the meantime, tools (`pubsys`) should all be updating the max version for every entry in the manifest as part of the  process of adding a new update entry. If that's not happening, it's potentially a bug in the tool.

Just from a practical perspective: fixing `updata` is not going to fix the behavior on any deployed systems, which will continue to find the wrong version if the constraint is violated. Better not to violate the constraint than to give up on those systems.

ytsssun:
> > Closes # N/A
> 
> Can you open an issue with more details about the bug you're fixing?
> 
> "max version" is (incorrectly) associated with each update's metadata instead of being set once at the top level of `manifest.json`. This is a longstanding bug that would need a repo pivot to fix.

Just created the issue - https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/366. Right now `updata` is not even referencing the `max_version` field in manifest.json. I can confirm that all entires in the `updates` in manifest.json does have the correct `max_version` set. 

Now I think the simple fix would probably be to just use `max_version` instead of `version` to get the max version. Or we can sort the `updates` array by `version` and get the max version with the sorted array.

bcressey (approved):

yeazelm (approved):

PR #363, opened 2025-01-25
feat: tag sd* drives as ephemeral also

### LINKED ISSUE #4370: Add support for "scsi disks" to be tagged "ephemeral" in udev
Currently here:
https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/edec0b247f770dd13912fd11293e02bf5981891a/packages/os/ephemeral-storage.rules#L10

and in a couple of other rules, there is a whitelist of device prefixes that does not include `sd*` devices. I was hoping to rely on the `BOTTLEROCKET_DEVICE_TYPE` to trigger automatic encryption using udev rules, but since `sd*` is not included by default, I have now added my own rule that tags them also using ghostdog.

Would you want to upstream this as part of the core ruleset?

mikn:
**Issue number:**

Closes [#4370](https://github.com/bottlerocket-os/bottlerocket/issues/4370)

**Description of changes:**
Minor change, adding the `sd*` pattern to the allowlist of disks to be tagged by the `ghostdog` utility. 

**Testing done:**
Mirrored these two rules in one custom rule, packaged it in our custom image, installed it on a system with `sd*` disks and saw them tagged.

**Terms of contribution:**

### DIFF
--- a/packages/os/ephemeral-storage.rules
+++ b/packages/os/ephemeral-storage.rules
@@ -7,7 +7,7 @@ ENV{DEVTYPE}!="disk", GOTO="ephemeral_storage_end"
 # Known sources of ephemeral disks:
 # - EC2 instance types with instance storage volumes
 # - KVM VMs with additional virtio disks
-KERNEL!="nvme*|vd*", GOTO="ephemeral_storage_end"
+KERNEL!="nvme*|vd*|sd*", GOTO="ephemeral_storage_end"
 
 # We can't be sure that devices with similar kernel names are all ephemeral.
 # Our udev helper checks for known system partitions to classify the device.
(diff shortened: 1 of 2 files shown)


packages/os/ephemeral-storage.rules
```diff
@@ -7,7 +7,7 @@ ENV{DEVTYPE}!="disk", GOTO="ephemeral_storage_end"
 # Known sources of ephemeral disks:
 # - EC2 instance types with instance storage volumes
 # - KVM VMs with additional virtio disks
-KERNEL!="nvme*|vd*", GOTO="ephemeral_storage_end"
+KERNEL!="nvme*|vd*|sd*", GOTO="ephemeral_storage_end"
```
bcressey, line 10: To create symlinks under `/dev/disk/ephemeral` for the non-system disks, you'd also need a rule like this after the `nvme*` and `vd*` ones below:
```
# Create symlinks for any matching SCSI devices.
KERNEL=="sd*", \
  SYMLINK+="disk/ephemeral/virtio-$env{ID_SERIAL}"
```

PR #365, opened 2025-01-30
package updates: aws-iam-authenticator to 0.6.29 and aws-signing-helper to 1.4.0

Sparksssj:
**Description of changes:**
Updated aws-iam-authenticator to 0.6.29, and aws-signing helper to 1.4.0

**Testing done:**

Quick tests and smoke tests passed.

```
aarch64-aws-k8s-130-quick                    Test               passed           
 x86-64-aws-k8s-130-quick                    Test               passed    
 x86-64-aws-k8s-130-nvidia-quick             Test               passed     
 aarch64-aws-k8s-130-nvidia-quick            Test               passed    
 
NAME                STATUS     COMPLETIONS   DURATION   AGE
nvidia-smoke-test   Complete   1/1           89s        7m49s
```

**Terms of contribution:**

### DIFF
--- a/packages/aws-iam-authenticator/aws-iam-authenticator.spec
+++ b/packages/aws-iam-authenticator/aws-iam-authenticator.spec
@@ -2,7 +2,7 @@
 %global gorepo aws-iam-authenticator
 %global goimport %{goproject}/%{gorepo}
 
-%global gover 0.6.28
+%global gover 0.6.29
 %global rpmver %{gover}
 
 %global _dwz_low_mem_die_limit 0
@@ -49,7 +49,7 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
 
 %build
 %set_cross_go_flags
-export GO_MAJOR="1.22"
+export GO_MAJOR="1.23"
 go build -ldflags="${GOLDFLAGS}" -o aws-iam-authenticator ./cmd/aws-iam-authenticator
 gofips build -ldflags="${GOLDFLAGS}" -o fips/aws-iam-authenticator ./cmd/aws-iam-authenticator
 
--- a/packages/aws-iam-authenticator/Cargo.toml
+++ b/packages/aws-iam-authenticator/Cargo.toml
@@ -12,8 +12,8 @@ path = "../packages.rs"
 releases-url = "https://github.com/kubernetes-sigs/aws-iam-authenticator/releases"
 
 [[package.metadata.build-package.external-files]]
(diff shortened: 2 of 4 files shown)


koooosh:
Changes LGTM, but a nit on the commit messages, where we prefer:
- packages: update aws-iam-authenticator to 0.6.29
- packages: update aws-signing-helper to 1.4.0

Also, let's update the PR title to mention we are updating aws-iam-authenticator and aws-signing-helper specifically for clarification

koooosh (changes requested):
Changes LGTM, but a nit on the commit messages, where we prefer:
- packages: update aws-iam-authenticator to 0.6.29
- packages: update aws-signing-helper to 1.4.0

Also, let's update the PR title to mention we are updating aws-iam-authenticator and aws-signing-helper specifically for clarification

koooosh (approved):

packages/aws-iam-authenticator/aws-iam-authenticator.spec
```diff
@@ -49,7 +49,7 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
 
 %build
 %set_cross_go_flags
-export GO_MAJOR="1.22"
+export GO_MAJOR="1.23"
```
bcressey, line 52: Is this change strictly required to build? You are overriding the SDK default and it would be better to leave it alone if it's not required.

bcressey: In other words - I'd rather delete this line than change it.

Sparksssj: Sorry for the delay - I tired to build without changing the GO_MAJOR or delete this line but it won't build. I think the reason is they updated go version in v0.6.29.
https://github.com/kubernetes-sigs/aws-iam-authenticator/releases/tag/v0.6.29#:~:text=3b3bf6f%20chore%3A%20Update%20to%20go%201.23.4

KCSesh: Seems a similar issue in the past occured, the SDK was on 1.21:
https://github.com/bottlerocket-os/bottlerocket-sdk/commit/acf1d651134bee749cf98acf7e965d4aba6649d2

And we moved this package to GO MAJOR 1.22:
https://github.com/bottlerocket-os/bottlerocket-core-kit/commit/0ee5cbaa1421b39106db9b0852dc794478266e71

Which matches the go version on v0.6.21:
https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/v0.6.21/.go-version

I also confirmed the `toolchain` error on check-out. 
LGTM

KCSesh (approved):

PR #370, opened 2025-02-06
packages: update libglib to v2.83.3

koooosh:
**Issue number:**

N/A

**Description of changes:**

Updating the `libglib` package from v2.83.0 to v2.83.3. Since the Bottlerocket SDK still uses Fedora 39, we need to maintain the `0001-require-older-meson.patch`.

The patch needed to be updated to also revert upstream commit `0b776bc20cfdbb529a97a7f8f8a4d88ddbef9329` which was added in v2.83.1 and was causing the build to break.

**Testing done:**

Built and booted a `vmware-dev` variant. VM successfully responded and shutdown with a single power off command (`open-vm-tools` test):

```
$ govc vm.power -off bottlerocket-vmware-dev-x86_64-1.32.0-0221a7fa-dirty
Powering off VirtualMachine:vm-19260... OK
```

**Terms of contribution:**


koooosh:
Closing out this PR since we prefer to wait for the v2.84.X libglib release as the even-numbered releases are stable

PR #381, opened 2025-02-20
fix incorrect hyphens used in BRSA for nvidia-container-toolkit CVE-2025-23359

ginglis13:
**Description of changes:**

Text copied directly from https://nvidia.custhelp.com/app/answers/detail/a_id/5616 for `CVE-2025-23359` uses a special "non-breaking hyphen" character for hyphens, resulted in encodings `CVE&#x2011;2025&#x2011;23359` instead of using a normal "-"

**Terms of contribution:**

### DIFF
--- a/advisories/5.4.2/BRSA-i3wlupztu5jm.toml
+++ b/advisories/5.4.2/BRSA-i3wlupztu5jm.toml
@@ -1,7 +1,7 @@
 [advisory]
 id = "BRSA-i3wlupztu5jm"
-title = "nvidia-container-toolkit CVE‑2025‑23359"
-cve = "CVE‑2025-23359"
+title = "nvidia-container-toolkit CVE-2025-23359"
+cve = "CVE-2025-23359"
 severity = "high"
 description = "A flaw was found in the NVIDIA Container Toolkit with its default configuration where a specially crafted container could gain access to the host filesystem"
 


rpkelly (approved):

advisories/5.4.2/BRSA-i3wlupztu5jm.toml
```diff
@@ -1,7 +1,7 @@
 [advisory]
 id = "BRSA-i3wlupztu5jm"
-title = "nvidia-container-toolkit CVE‑2025‑23359"
-cve = "CVE‑2025-23359"
+title = "nvidia-container-toolkit CVE-2025-23359"
+cve = "CVE-2025-23359"
```
bcressey, line 4: Can you open an issue somewhere - Twoliter? - to add a lint for this?

Stuff like this isn't reasonable to expect humans to catch in review, so this will keep happening unless we have automated checks, or else more tolerant downstream tools.

bcressey: Something like:
```
#!/bin/bash

if find advisories -name '*.toml' -type f >/dev/null 2>&1 ; then
  export LC_ALL=C
  if grep --include '*.toml' -R -l -P ['\x80'-'\xFF'] advisories ; then
    echo "error: found non-ASCII characters in advisories" >&2
    exit 1
  fi
fi

❯ make twoliter check-advisories
advisories/1.0.4/BRSA-6jzsr8lqncrq.toml
advisories/1.0.4/BRSA-tdebxkarxymw.toml
advisories/1.0.4/BRSA-4mz3mlhteng6.toml
advisories/1.0.4/BRSA-jd3wf9wpwu5l.toml
advisories/1.0.4/BRSA-bl2apnymyndl.toml
error: found non-ASCII characters in advisories
```

ginglis13: Opened https://github.com/bottlerocket-os/twoliter/issues/466 to get some lints into Twoliter

ytsssun (approved):

advisories/5.4.2/BRSA-i3wlupztu5jm.toml
ginglis13: Opened PR https://github.com/bottlerocket-os/twoliter/pull/467 to add a lint task for twoliter. Also merged https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/55 to fix this same issue in the kernel-kit.

PR #395, opened 2025-03-08
Feature: `--ebs-volumes` option for ephemeral-storage init 

### LINKED ISSUE #389: Feature: Allow EBS disks in ephemeral storage init
**What I'd like:** I'd like to use `apiclient ephemeral-storage init --disks /dev/xvdba /dev/xvdbb ...` to mount multiple EBS disks in RAID0 for use as ephemeral storage. This is currently not possible as the allowed disk paths are limited to those covered by the ephemeral disk autodiscovery, which intentionally excludes EBS disks as they have other uses as well.

This feature could be implemented by allowing explicitly selected EBS disks in [ephemeral_storage::initialize()](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/6db0fbef70d71b7f1bf8897f81f573b5efa01f3e/sources/api/apiserver/src/server/ephemeral_storage.rs#L35), possibly limiting the allowed paths according to the [EBS device naming patterns](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html).

**Any alternatives you've considered:** I currently use an init script to do this on AL2 (symlink EBS disks to appear as instance volumes). The same workaround on Bottlerocket would require the use of a bootstrap container, which seems a bit much for such a simple task.

jesseanttila-cai:
**Issue number:**

Closes #389

**Description of changes:**

Adds two new options to `ephemeral-storage init`:

* `--ebs-volumes` specifies EBS volumes to be included in the ephemeral storage array (e.g. `--ebs-volumes /dev/disk/ephemeral-ebs/xvdda /dev/disk/ephemeral-ebs/xvddb`

* `--prefer` specifies combinations of storage types to be included in the ephemeral storage array in order of preference. Currently supported storage types are `ephemeral-disk` and `ebs-volume`. Storage types can be combined with the `+` operator.  For example, `--prefer ephemeral-disk ebs-volume` uses ephemeral disks if available and ebs volumes otherwise. To use a mix of both types, specify `--prefer ephemeral-disk+ebs-volume`.

**Testing done:**

Manual testing of a few different configurations on EKS (see [this comment](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/395#issuecomment-3224069765))

**Terms of contribution:**

### DIFF
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -150,6 +150,7 @@ enum EphemeralStorageSubcommand {
     Init(EphemeralStorageInitArgs),
     Bind(EphemeralStorageBindArgs),
     ListDisks(EphemeralStorageFormatArgs),
+    ListEbsVolumes(EphemeralStorageFormatArgs),
     ListDirs(EphemeralStorageFormatArgs),
 }
 
@@ -167,7 +168,7 @@ struct EphemeralStorageInitArgs {
 struct EphemeralStorageBindArgs {
     targets: Vec<String>,
 }
-/// Stores user-supplied arguments for the 'ephemeral-storage list-disks/list-dirs' subcommand.
+/// Stores user-supplied arguments for the 'ephemeral-storage list-disks/list-ebs-volumes/list-dirs' subcommand.
 #[derive(Debug)]
 struct EphemeralStorageFormatArgs {
     format: Option<String>,
@@ -203,6 +204,8 @@ fn usage() -> ! {
             ephemeral-storage bind     Bind directories to previously initialized ephemeral storage.
             ephemeral-storage list-disks
                                        List the discovered ephemeral disks that can be initialized.
+            ephemeral-storage list-ebs-volumes
+                                       List the discovered ephemeral ebs volumes that can be initialized.
             ephemeral-storage list-dirs
                                        List the directories that can be bound to ephemeral storage.
 
@@ -274,7 +277,7 @@ fn usage() -> ! {
             --ebs-volumes VOLUME [VOLUME ..]
                                        EBS volumes in the `xvdda`-`xvddx` range to configure for storage.
             --prefer PREFERENCE [PREFERENCE ..]
-                                       Epheremeral storage type preference, in descending order. Allowed
+                                       Ephemeral storage type preference, in descending order. Allowed
                                        values: `ephemeral-disk`, `ebs-volume` or a combination of these
                                        joined by `+`. Defaults to `ephemeral-disk` only. This option is
                                        ignored if `--disks` or `--ebs-volumes` is set.
@@ -286,6 +289,9 @@ fn usage() -> ! {
 
         ephemeral-storage list-disks options:
             -f, --format               Format of the disk listing (text or json). Default format is text.
+
+        ephemeral-storage list-ebs-volumes options:
+            -f, --format               Format of the volume listing (text or json). Default format is text.   
 
         ephemeral-storage list-dirs options:
             -f, --format               Format of the directory listing (text or json). Default format is text.
@@ -757,7 +763,7 @@ fn parse_ephemeral_storage_args(args: Vec<String>) -> Subcommand {
     for arg in args.into_iter() {
         match arg.as_ref() {
             // Subcommands
-            "init" | "bind" | "list-disks" | "list-dirs"
+            "init" | "bind" | "list-disks" | "list-ebs-volumes" | "list-dirs"
                 if subcommand.is_none() && !arg.starts_with('-') =>
             {
                 subcommand = Some(arg)
@@ -772,6 +778,9 @@ fn parse_ephemeral_storage_args(args: Vec<String>) -> Subcommand {
         Some("init") => parse_ephemeral_storage_init_args(subcommand_args),
         Some("bind") => parse_ephemeral_storage_bind_args(subcommand_args),
         Some("list-disks") => EphemeralStorageSubcommand::ListDisks(
+            parse_ephemeral_storage_list_format_args(subcommand_args),
+        ),
+        Some("list-ebs-volumes") => EphemeralStorageSubcommand::ListEbsVolumes(
             parse_ephemeral_storage_list_format_args(subcommand_args),
         ),
         Some("list-dirs") => EphemeralStorageSubcommand::ListDirs(
@@ -880,7 +889,7 @@ fn parse_ephemeral_storage_bind_args(args: Vec<String>) -> EphemeralStorageSubco
     EphemeralStorageSubcommand::Bind(EphemeralStorageBindArgs { targets })
 }
 
-/// Parses arguments for the 'list-disks' and 'list-dirs' ephemeral-storage subcommand.
+/// Parses arguments for the 'list-disks', 'list-ebs-volumes', and 'list-dirs' ephemeral-storage subcommands.
 fn parse_ephemeral_storage_list_format_args(args: Vec<String>) -> EphemeralStorageFormatArgs {
     let mut format = None;
 
@@ -1139,6 +1148,14 @@ async fn run() -> Result<()> {
             }
             EphemeralStorageSubcommand::ListDisks(bind_args) => {
                 let body = ephemeral_storage::list_disks(&args.socket_path, bind_args.format)
+                    .await
+                    .context(error::EphemeralStorageSnafu)?;
+                if !body.is_empty() {
+                    print!("{body}");
+                }
+            }
+            EphemeralStorageSubcommand::ListEbsVolumes(bind_args) => {
+                let body = ephemeral_storage::list_ebs_volumes(&args.socket_path, bind_args.format)
                     .await
                     .context(error::EphemeralStorageSnafu)?;
                 if !body.is_empty() {
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -8,7 +8,7 @@
 
 use apiclient::{apply, ephemeral_storage, exec, get, reboot, report, set, update, SettingsInput};
 use log::{info, log_enabled, trace, warn};
-use model::ephemeral_storage::Filesystem;
+use model::ephemeral_storage::{Filesystem, Preference};
 use serde::{Deserialize, Serialize};
 use simplelog::{
     ColorChoice, ConfigBuilder as LogConfigBuilder, LevelFilter, TermLogger, TerminalMode,
@@ -157,6 +157,8 @@ enum EphemeralStorageSubcommand {
 #[derive(Debug)]
 struct EphemeralStorageInitArgs {
     disks: Option<Vec<String>>,
+    ebs_volumes: Option<Vec<String>>,
+    prefer: Option<Vec<Preference>>,
     filesystem: Option<Filesystem>,
 }
 
@@ -269,7 +271,13 @@ fn usage() -> ! {
                                        operation does nothing.
             --disks DISK [DISK ...]    Local disks to configure for storage. Default is all ephemeral
                                        disks.
-
+            --ebs-volumes VOLUME [VOLUME ..]
+                                       EBS volumes in the `xvdda`-`xvddx` range to configure for storage.
+            --prefer PREFERENCE [PREFERENCE ..]
+                                       Epheremeral storage type preference, in descending order. Allowed
+                                       values: `ephemeral-disk`, `ebs-volume` or a combination of these
+                                       joined by `+`. Defaults to `ephemeral-disk` only. This option is
+                                       ignored if `--disks` or `--ebs-volumes` is set.
         ephemeral-storage bind options:
             --dirs DIR [DIR ...]       Directories to bind to configured ephemeral storage
                                        (e.g. /var/lib/containerd). If not specified, uses platform (k8s vs. ECS)
@@ -777,6 +785,8 @@ fn parse_ephemeral_storage_args(args: Vec<String>) -> Subcommand {
 /// Parses arguments for the 'init' ephemeral-storage subcommand.
 fn parse_ephemeral_storage_init_args(args: Vec<String>) -> EphemeralStorageSubcommand {
     let mut disks: Option<Vec<String>> = None;
+    let mut ebs_volumes: Option<Vec<String>> = None;
+    let mut prefer: Option<Vec<Preference>> = None;
     let mut filesystem = None;
     let mut iter = args.into_iter().peekable();
     while let Some(arg) = iter.next() {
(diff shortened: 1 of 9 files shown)


sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -32,35 +32,53 @@ pub struct BindDirs {
...
+                    return Ok(());
+                }
+                // no disks specified, so use the default
+                known_disks
             }
-            // no disks specified, so use the default
-            known_disks
+        };
+        for ebs_volume in ebs_volumes.iter().flatten() {
+            let link_path = format!("/dev/disk/by-ebs-id/{ebs_volume}");
```
bcressey: Let's add an `ephemeral_ebs_volumes` function, like `ephemeral_devices`, that returns available EBS volumes.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -32,35 +32,53 @@ pub struct BindDirs {
...
-                info!("no ephemeral disks found, skipping ephemeral storage initialization");
-                return Ok(());
+            None => {
+                // if there are no disks specified, and none are available we treat the init as a
+                // no-op to allow "ephemeral-storage init"/"ephemeral-storage bind" to work on instances
+                // with and without ephemeral storage
+                if known_disks.is_empty() && ebs_volumes.is_none() {
+                    info!("no ephemeral disks found, skipping ephemeral storage initialization");
+                    return Ok(());
+                }
```
bcressey: There are a couple of design questions here that are worth considering, in the context of the patch that I shared that designates the `xvdda` - `xvddx` range for "ephemeral" EBS volumes.

Should we allow both EBS and instance store volumes in the same RAID-0 array? I would default to "no", and reject the API request if we're given both. We can change our minds later, but it seems weird to mix devices like this in the same array.

If both lists are empty, and both instance store disks and ephemeral EBS volumes are present, and we can't mix the two in the same array - then which devices do we include?

Some options:
1. Add another API flag, like `--prefer-ebs-volumes`, which controls the behavior when presented with two empty lists.
    1. If it's true:
        - If ephemeral EBS volumes are present: we'll use them in the array.
        - If no ephemeral EBS volumes are present: we'll use ephemeral disks, if any.
    2. If it's false:
         - If ephemeral disks are present: we'll use them in the array.
         - If no ephemeral disks are present: we'll use ephemeral EBS volumes, if any.
2. Punt on the problem and don't support implicit array creation for EBS volumes.

`--prefer-ebs-volumes` could also solve the first issue - if we're given two lists and told which one to prefer, we can ignore the other one rather than return an error.

(For the actual implemention, I'd favor using an enum and having `--prefer ephemeral-disks` and `--prefer ebs-volumes` rather than a boolean. That will be easier to extend if we end up with another storage category. You can put it in `models/src/ephemeral_storage.rs` so that the client and server can share the same type definition.)

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -92,7 +92,13 @@ pub fn initialize(
     info!("initializing ephemeral storage disks={:?}", disks);
     // with a single disk, there is no need to create the array
     let device_name = match disks.len() {
-        1 => disks.first().expect("non-empty").clone(),
+        1 => {
+            let disk_name = disks.first().expect("non-empty").clone();
+            let link_name = format!("{RAID_DEVICE_DIR}{RAID_DEVICE_NAME}");
+            std::os::unix::fs::symlink(disk_name, &link_name)
+                .context(error::DiskSymlinkFailureSnafu)?;
+            link_name
```
bcressey: Here you're fixing a pre-existing bug - we effectively lost track of which disk we used during the "init" operation. That didn't matter in the case where we use all ephemeral disks, but could have mattered on an instance that had three (`nvme1`, `nvme2`, `nvme3`) where the "init" request specified only one (`nvme2`) so we didn't end up creating an array.

The symlink approach is a nice, cheap way to preserve the state we need, but it can cause trouble to create links in "owned" directories under `/dev`, like `/dev/md` (which `mdadm` "owns").

Instead I'd suggest using [pathdiff::diff_paths](https://docs.rs/pathdiff/latest/pathdiff/fn.diff_paths.html) and creating a relative symlink from `/dev/disk/ephemeral-storage` to either the single disk or EBS volume, or to the RAID-0 array.

bcressey (commented):
Here's a [patch](https://gist.github.com/bcressey/3db84ff444d073a11f3c5356bf4d8cb4) that populates `/dev/disk/ephemeral-ebs` with symlinks for non-system EBS volumes in the `xvdda` - `xvddx` range.

In combination with my other suggestions, this would let you provide user-data like this to automatically configure up to 24 EBS volumes in a RAID-0 array, if present, and fall back to instance store otherwise:
```
commands = [
    ["apiclient", "ephemeral-storage", "init", "--prefer", "ebs-volumes"],
    ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/var/lib/containerd", "/var/lib/kubelet", "/var/log/pods"]
]
essential = true
mode = "always"
```

jesseanttila-cai:
Thank you kindly for the comments and suggestions. I had some doubts about the handling of system ebs volumes and some edge-cases involving mixed instance storage/ebs volume arrays, which are solved by the `/dev/disk/ephemeral-ebs` links and the `--prefer` flag suggestion.
Unfortunately I won't be able to (read: shouldn't) implement and test any of these changes until I return from holiday on the 25th, but now that I have the AMI build set-up it shouldn't take four months this time.

sources/api/apiserver/src/server/ephemeral_storage.rs
bcressey: With the enum approach, we could also just have all four options as variants:
```
--prefer [ disks-only
         | ebs-volumes-only
         | disks-to-ebs-volumes
         | ebs-volumes-to-disks ]
```

Then the default can be `disks-only` for backwards compatibility.

sources/api/apiclient/src/main.rs
```diff
@@ -271,7 +272,12 @@ fn usage() -> ! {
             --disks DISK [DISK ...]    Local disks to configure for storage. Default is all ephemeral
                                        disks.
             --ebs-volumes VOLUME [VOLUME ..]
-                                       Additional EBS volumes to configure for storage.
+                                       EBS volumes in the `xvdda`-`xvddx` range to configure for storage.
+            --prefer PREFERENCE [PREFERENCE ..]
+                                       Epheremeral storage type preference, in descending order. Allowed
+                                       values: `ephemeral-disk`, `ebs-volume` or a combination of these
+                                       joined by `+`. Defaults to `ephemeral-disk` only. This option is
+                                       ignored if `--disks` or `--ebs-volumes` is set.
```
jesseanttila-cai, line 283: Instead of the enum solution I went with an ordered list of "preferences", which are sets of allowed storage types. This should be easier to expand in case more storage types are added in the future.

There are effectively five distinct preference options available:
* `--prefer ephemeral-disk`: use only ephemeral instance storage (default)
* `--prefer ebs-volume`: use only ebs volumes
* `--prefer ephemeral-disk ebs-volume`: use ephemeral instance storage if available, otherwise ebs volumes
* `--prefer ebs-volume ephemeral-disk`: same as above except in the opposite order
* `--prefer ephemeral-disk+ebs-volume`: use all available storage of either type

The `--disks` and `--ebs-volumes` options override `--prefer`, which allows for a specific custom storage setup to be defined. This is also backwards compatible with the previous behaviour of the `--disks` option.

sources/models/src/ephemeral_storage.rs
```diff
@@ -18,12 +18,58 @@ impl Display for Filesystem {
...
+        let mut preference = Self {
+            ephemeral_disk: false,
+            ebs_volume: false,
+        };
+        for i in value.split("+") {
+            match i {
+                "ephemeral-disk" => preference.ephemeral_disk = true,
+                "ebs-volume" => preference.ebs_volume = true,
+                x => return Err(x),
+            }
```
jesseanttila-cai, line 61: Should `""` be allowed here? It would allow defining an empty set (not very useful) and superfluous `+`s, e.g. `ephemeral-disk+`. The main benefit would probably be enabling round-tripping through `Display` and `TryFrom<&str>` for the empty set case.

jesseanttila-cai:
Results from manual testing with the latest commit:

| instance config | A | B | C | D | E |
|------------------|---|---|---|---|---|
|m5.large|20G|20G|20G|20G|X|
|m5.large+4x8GB|20G|32G|32G|32G|16G|
|m5.large+16x4GB|20G|64G|64G|64G|8G|
|m5d.large|70G|70G|70G|70G|X|
|m5d.large+4x8GB|70G|32G|70G|102G|16G|
|m5d.large+16x4GB|70G|64G|70G|134G|8G|

where
* **A**: default (equivalent to `--prefer ephemeral-disk`)
* **B**: `--prefer ebs-volume ephemeral-disk`
* **C**: `--prefer ephemeral-disk ebs-volume`
* **D**: `--prefer ephemeral-disk+ebs-volume`
* **E**: `--ebs-volumes /dev/disk/ephemeral-ebs/xvddb /dev/disk/ephemeral-ebs/xvddd`

All test cases reported reasonable values for ephemeral storage capacity, except for two cases where the bootstrap command failed due to missing ebs volumes. A capacity of 20G corresponds to the default EBS volume added by Karpenter, which is used when `ephemeral-storage init` does not find any storage devices of an allowed type.

bcressey (approved):
Apologies for the delayed review! LGTM.

sources/models/src/ephemeral_storage.rs
bcressey: The round-trip property is nice, though I don't feel strongly. Up to you!

sources/api/apiclient/src/main.rs
bcressey: Makes sense; I agree the enum variants would get unwieldy if more permutations of storage were added, while this will be easy to extend.

bcressey:
Let me know if you're ready to mark this as "ready for review" and I can kick off CI tests. The only change I'd suggest is squashing the commits together and splitting them back out into a pair of apiclient and apiserver commits.

jesseanttila-cai:
I've refactored the PR to two commits, the apiserver commit also includes the changes to `models` & `os.spec`. There was also a very minor conflict with commit 3fd2ad12388e3c38bd3997883e6ec5d14ce6234a which has now been resolved.

bcressey (approved):

sources/api/apiclient/src/main.rs
```diff
@@ -269,7 +271,13 @@ fn usage() -> ! {
                                        operation does nothing.
             --disks DISK [DISK ...]    Local disks to configure for storage. Default is all ephemeral
                                        disks.
-
+            --ebs-volumes VOLUME [VOLUME ..]
+                                       EBS volumes in the `xvdda`-`xvddx` range to configure for storage.
+            --prefer PREFERENCE [PREFERENCE ..]
+                                       Epheremeral storage type preference, in descending order. Allowed
```
ytsssun: Nit: typo
```suggestion
                                       Ephemeral storage type preference, in descending order. Allowed
```

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -298,6 +338,29 @@ pub fn ephemeral_devices() -> Result<Vec<String>> {
     Ok(filenames)
 }
 
+/// ephemeral_ebs_volumes returns the full path name to the ebs volumes in /dev/disk/ephemeral-ebs
+pub fn ephemeral_ebs_volumes() -> Result<Vec<String>> {
+    const EPHEMERAL_EBS_PATH: &str = "/dev/disk/ephemeral-ebs";
```
ginglis13: nit: I'd prefer this defined at the top of this file as `static` below `EPHEMERAL_STORAGE_LINK`

sources/api/apiclient/src/main.rs
```diff
@@ -269,7 +271,13 @@ fn usage() -> ! {
                                        operation does nothing.
             --disks DISK [DISK ...]    Local disks to configure for storage. Default is all ephemeral
                                        disks.
-
+            --ebs-volumes VOLUME [VOLUME ..]
```
ginglis13, line 277: Won't block on it but would be nice to update the README for `apiclient` with these changes

ytsssun (approved):

ginglis13 (approved):

ginglis13 (approved):

ginglis13:
I performed some more extensive manual testing today, everything LGTM with 2 non-blocking observations:

* missing `list-volumes` sub-command. We have an equivalent `list-disks`, this would be nice to have for parity
* once a preference is set via `--prefer`, it cannot be updated

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -95,22 +139,21 @@ pub fn initialize(fs: Option<Filesystem>, disks: Option<Vec<String>>) -> Result<
         info!("{device_name:?} is already formatted as {fs}, skipping format");
     }
 
+    // Create link to formatted device for use in `bind`
+    std::os::unix::fs::symlink(&device_name, EPHEMERAL_STORAGE_LINK)
+        .context(error::DiskSymlinkFailureSnafu {})?;
```
bcressey, line 153: `init` now fails after a second invocation:
```
> apiclient ephemeral-storage init --disks /dev/disk/ephemeral/virtio-eph01234
Failed to initialize ephemeral storage:
  Failed POST request to '/actions/ephemeral-storage/init':
    Status 500 when POSTing /actions/ephemeral-storage/init:
      Unable to initialize ephemeral storage:
        Failed to create disk symlink File exists (os error 17)
```

Not sure how reliable a second `init` was in the past but the intent is for it to be idempotent.

jesseanttila-cai:
I've added a `list-ebs-volumes` subcommand to match `list-disks`, and corrected the documentation of `list-dirs` to match its actual behavior. I haven't had a chance to test these changes yet, but hopefully will by the end of the week.

bcressey:
Testing LGTM:
```
# Attached two EBS volumes which can be listed.
bash-5.1# apiclient ephemeral-storage list-ebs-volumes
/dev/disk/ephemeral-ebs/xvddb
/dev/disk/ephemeral-ebs/xvdda

# Volumes were added to the array.
bash-5.1# apiclient ephemeral-storage init --prefer ebs-volume

bash-5.1# cat /proc/mdstat
Personalities : [raid0]
md127 : active raid0 nvme3n1[1] nvme4n1[0]
      209582080 blocks super 1.2 256k chunks

bash-5.1# nvme amzn id-ctrl /dev/nvme4n1|grep dev
bdev      : /dev/xvddb

bash-5.1# nvme amzn id-ctrl /dev/nvme3n1|grep dev
bdev      : /dev/xvdda

# init operation is now idempotent
bash-5.1# apiclient ephemeral-storage init
```

bcressey (approved):

ytsssun (approved):

ytsssun (approved):

ginglis13 (approved):

bcressey:
Nice! Thanks for contributing this!

PR #401, opened 2025-03-11
build: fix upstream-source-fallback variable

vigh-m:
**Description of changes:**
Fix the `UPSTREAM_SOURCE_FALLBACK` variable. It now throws the right error when building a package that is not available in the the cache.

**Testing done:**

- `make build` resulted in build failures
- `UPSTREAM_SOURCE_FALLBACK=true make build` resulted in successful builds

**Terms of contribution:**


koooosh (approved):
Pulled in this change in a fresh repo and verified testing results.

bcressey (approved):

PR #435, opened 2025-03-20
sources: remove bottlerocket-variant crate

### LINKED ISSUE #396: Remove ecs-settings-applier folder
ECS Settings applier was deprecated in https://github.com/bottlerocket-os/bottlerocket/pull/3834, but its directoy was re-added to the core-kit in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/70. Currently, it only containers a `Cargo.toml` file so it is safe to remove it.

bcressey:
**Issue number:**
Fixes #396

**Description of changes:**
`ecs-settings-applier` was the last user of `bottlerocket-variant`, so let's clean it up.

**Testing done:**
Built some kits.

**Terms of contribution:**


arnaldo2792 (approved):

arnaldo2792:
If we remove `bottlerocket-variant` does that mean we won't be able to write "variant agnostic" [migrations]:

[migrations]: 
https://github.com/bottlerocket-os/bottlerocket/blob/91855a9712060069c017cd099fdbbaa4bc837c83/sources/settings-migrations/archived/v1.20.0/container-runtime-nvidia/src/main.rs#L6

bcressey:
> If we remove `bottlerocket-variant` does that mean we won't be able to write "variant agnostic" migrations

That's already the case, since Twoliter no longer supports conditional compilation based on the selected variant.

koooosh (approved):

ytsssun (approved):

PR #443, opened 2025-03-23
optimize gzip decompression for containerd 

### LINKED ISSUE #392: adopt zlib-ng as a replacement for zlib
**What I'd like:**
I'd like to switch over to zlib-ng as the source for the `libz` package. zlib-ng contains numerous optimizations for modern processors that make decompression faster and more CPU efficient.

[Fedora 41 moved to zlib-ng by default](https://fedoraproject.org/wiki/Changes/ZlibNGTransition), and [Amazon Linux 2023 is considering it](https://github.com/amazonlinux/amazon-linux-2023/issues/872).

zlib-ng can be built in a compatibility mode for use as a drop-in replacement.

Stay with the existing `zlib` upstream, which leaves some significant performance wins on the table.

There are various other zlib forks, like Cloudflare's and Chromium's. zlib-ng rolls up some (most?) of the optimizations found in these forks.

bcressey:
**Issue number:**
Fixes #392

**Description of changes:**
Switch to zlib-ng as the upstream for zlib, which improves the performance of `pigz`, and add `igzip` as an optimized decompressor for x86_64.

Backport support for `igzip` to containerd 1.7, and adjust the dependencies so that `pigz` is preferred for aarch64 and `igzip` is preferred for x86_64.

Add a patch to `igzip` to adjust the pipe size for stdin and stdout to 1 MiB, which greatly improves performance if either one is a pipe.

**Testing done:**
Pulled a 23 GiB image via `ctr`. 
```
ctr image pull nvcr.io/nvidia/pytorch:25.02-py3
```

For x86_64, I used an `m6id.16xlarge` instance type. For aarch64, I used an `m7gd.16xlarge` instance type. Both had the instance store disks configured as the backing store for container images:
```
[settings.bootstrap-commands.ephemeral-storage]
commands = [
  ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/var/lib/containerd", "/var/lib/kubelet", "/var/log/pods"]]
essential = true
mode = "always"
```

Before, using the official `aws-k8s-1.32` v1.34.0 AMIs:
```
# x86_64 with pigz+zlib
unpacking linux/amd64 sha256:50e2ea138815ed088a60e6d642c9d0298b98a42b90666f7f19924640ccc519eb...
done: 2m20.592503638s

# aarch64 with pigz+zlib
unpacking linux/arm64/v8 sha256:50e2ea138815ed088a60e6d642c9d0298b98a42b90666f7f19924640ccc519eb...
done: 1m40.090797734s
```

After, using `aws-k8s-1.32` AMIs built with these changes:
```
# x86_64 with igzip
unpacking linux/amd64 sha256:50e2ea138815ed088a60e6d642c9d0298b98a42b90666f7f19924640ccc519eb...
done: 1m5.287939797s

# aarch64 with pigz+zlib-ng
unpacking linux/arm64/v8 sha256:50e2ea138815ed088a60e6d642c9d0298b98a42b90666f7f19924640ccc519eb...
done: 1m22.394084064s
```

For x86_64, image unpack time decreased by 53%. For aarch64, image unpack time decreased by 18%.

**Terms of contribution:**

### DIFF
--- a/packages/containerd/1001-Use-Intel-ISA-L-s-igzip-if-available.patch
+++ b/packages/containerd/1001-Use-Intel-ISA-L-s-igzip-if-available.patch
@@ -1,0 +1,183 @@
+From 796ca04aa424a7770f79079917b7fcd18c993a57 Mon Sep 17 00:00:00 2001
+From: Kazuyoshi Kato <kaz@fly.io>
+Date: Thu, 5 Oct 2023 22:11:14 -0700
+Subject: [PATCH 1/2] Use Intel ISA-L's igzip if available
+
+Intel ISA-L is Intel's open source (BSD) library that outperforms both
+gzip and pigz. This commit checks and uses igzip if available.
+
+Signed-off-by: Kazuyoshi Kato <kaz@fly.io>
+(cherry picked from commit 34378ec9b4579c9f647dd6724fcb04b3390bc3dd)
+---
+ archive/compression/benchmark_test.go   | 28 +++++++++++++------
+ archive/compression/compression.go      | 36 ++++++++++++++-----------
+ archive/compression/compression_test.go | 10 +++----
+ 3 files changed, 46 insertions(+), 28 deletions(-)
+
+diff --git a/archive/compression/benchmark_test.go b/archive/compression/benchmark_test.go
+index f8340802a..845f5f9e8 100644
+--- a/archive/compression/benchmark_test.go
++++ b/archive/compression/benchmark_test.go
+@@ -20,6 +20,7 @@ import (
+ 	"fmt"
+ 	"io"
+ 	"net/http"
++	"os/exec"
+ 	"testing"
+ 
+ 	"github.com/stretchr/testify/require"
+@@ -49,19 +50,30 @@ func BenchmarkDecompression(b *testing.B) {
+ 		zstd := testCompress(b, data, Zstd)
+ 
+ 		b.Run(fmt.Sprintf("size=%dMiB", sizeInMiB), func(b *testing.B) {
+-			b.Run("gzip", func(b *testing.B) {
+-				testDecompress(b, gz)
+-			})
++			original := gzipPath
++			defer func() {
++				gzipPath = original
++			}()
++
+ 			b.Run("zstd", func(b *testing.B) {
+ 				testDecompress(b, zstd)
+ 			})
+-			if unpigzPath != "" {
+-				original := unpigzPath
+-				unpigzPath = ""
+-				b.Run("gzipPureGo", func(b *testing.B) {
++
++			gzipPath = ""
++			b.Run("gzipPureGo", func(b *testing.B) {
++				testDecompress(b, gz)
++			})
++			gzipPath, err = exec.LookPath("igzip")
++			if err == nil {
++				b.Run("igzip", func(b *testing.B) {
++					testDecompress(b, gz)
++				})
++			}
++			gzipPath, err = exec.LookPath("unpigz")
++			if err == nil {
++				b.Run("unpigz", func(b *testing.B) {
+ 					testDecompress(b, gz)
+ 				})
+-				unpigzPath = original
+ 			}
+ 		})
+ 	}
+diff --git a/archive/compression/compression.go b/archive/compression/compression.go
+index 23ddfab1a..5ad13eccf 100644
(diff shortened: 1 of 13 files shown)


packages/libisal/libisal.spec
```diff
@@ -0,0 +1,60 @@
...
+%{_cross_libdir}/*.so
+%{_cross_includedir}/isa-l.h
+%dir %{_cross_includedir}/isa-l
+%{_cross_includedir}/isa-l/*.h
+%{_cross_pkgconfigdir}/*.pc
+
+%files -n %{_cross_os}igzip
+%{_cross_bindir}/igzip
+
+%changelog
```
arnaldo2792, line 64: nit: Elsewhere in a different PR you removed `%changelog`

packages/libisal/0001-igzip-increase-stdin-and-stdout-pipe-sizes.patch
```diff
@@ -0,0 +1,49 @@
+From 27c24697b497495d5af5a8d8040ebbe6a2fdc23c Mon Sep 17 00:00:00 2001
```
arnaldo2792, line 1: Do you think this is a patch we could upstream? Or is this one of those patches that will be unique to Bottlerocket?

packages/containerd/containerd.spec
```diff
@@ -62,6 +66,30 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
...
+%description pigz
+%{summary}.
+
+%package igzip
+Summary: Prefer igzip for gzip decompression
+Requires: %{_cross_os}igzip
+Requires: %{name}
+Conflicts: %{name}-pigz
+%if "%{_cross_arch}" == "x86_64"
+Provides: %{_cross_os}containerd(optimized-gunzip) = 2:
```
arnaldo2792, line 85: Will this package be selected for installation due to its higher `epoch` number compared to the `%{name}-pigz` package?

arnaldo2792 (approved):

packages/containerd/containerd.spec
bcressey: Yes, it'll either have a higher or lower precedence depending on the architecture.

packages/libisal/0001-igzip-increase-stdin-and-stdout-pipe-sizes.patch
bcressey: > Do you think this is a patch we could upstream? Or is this one of those patches that will be unique to Bottlerocket?

I'm more confident in Bottlerocket usage, since we only invoke `igzip` (and `unpigz`) with stdin and stdout set to a pipe. Not sure it'd be acceptable in this form upstream.

packages/containerd/containerd.spec
ytsssun: TIL, this is a clever approach to implement the precedence using epoch number. Is this a preferred approach for cases like this. Would it be worth it to document this pattern if it is used elsewhere in the codebase? (Not blocking)

packages/containerd/1001-Use-Intel-ISA-L-s-igzip-if-available.patch
ytsssun, line 1: This might be a question for the upstream containerd folks, but do you know if they have plan to backport these to the 1.7 versions - in which case we may need to drop them upon upgrading to the future 1.7 versions.

ytsssun (approved):

packages/containerd/1001-Use-Intel-ISA-L-s-igzip-if-available.patch
bcressey: > This might be a question for the upstream containerd folks, but do you know if they have plan to backport these to the 1.7 versions

I wouldn't expect them to backport new features like this.

packages/containerd/containerd.spec
bcressey: > Is this a preferred approach for cases like this?

We used it a lot during the early years of Amazon Linux to control package priorities for different versions of Java and Python. The combination of the synthetic "Provides" capability (which shouldn't match any real package name) and an Epoch (for simple integer precedence) is fairly robust.

Not sure where to document it, except perhaps to use it in other places so it's less surprising.

PR #438, opened 2025-03-20
systemd: refactor to list packaged files

bcressey:
**Issue number:**
Related: #327

**Description of changes:**
The current wildcard-based file includes make it harder to review the impact of changing the build time configuration to enable features.

For optional functionality, it also makes it more tedious to place files into the correct subpackage, since each file has to be excluded from the main package and then also listed in the subpackage. It's easy to miss a related file and then it ends up packaged twice or not moved to the subpackage.

Remove all wildcard includes and replace with lists of files. While we're at it, remove a handful of units that are unlikely to ever be useful for Bottlerocket.

**Testing done:**
Here's a cleaned-up diff of the RPM contents before and after.

```
diff --git a/before/bottlerocket-systemd.aarch64.txt b/after/bottlerocket-systemd.aarch64.txt
index 7455299..fa7f9b9 100644
--- a/before/bottlerocket-systemd.aarch64.txt
+++ b/after/bottlerocket-systemd.aarch64.txt
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/modprobe.d/README
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/sysctl.d/README
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.be.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.be@latin.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.bg.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.da.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.de.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.fr.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.hr.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.hu.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.it.catalog
-/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/catalog/systemd.ko.catalog
[...]

PR #444, opened 2025-03-23
add dependencies for cryptsetup

bcressey:
**Issue number:**
Related: #327

**Description of changes:**
Add dependencies for cryptsetup. The integration with systemd is postponed until #438 is merged.

I'd intended to include only the cryptsetup libraries, but it was difficult to test the functionality without the actual binaries (`cryptsetup`, `integritysetup`, `veritysetup`). Likewise for libtss2 - the various `tpm2_*` tools are useful to prove the libraries still work.

These changes are limited to adding new packages so that the pull request will be safe to merge without needing to scrutinize any changes on the systemd side.

**Testing done:**
`cryptsetup` and `systemd-cryptenroll`  work as expected:
```
# cryptsetup luksFormat --pbkdf argon2id volume 

WARNING!
========
This will overwrite data on volume irrevocably.

Are you sure? (Type 'yes' in capital letters): YES
Enter passphrase for volume: 
Verify passphrase: 

# systemd-cryptenroll --tpm2-device=auto volume

🔐 Please enter current passphrase for disk volume: ****                    
New TPM2 token enrolled as key slot 1.

# systemd-cryptenroll volume 
SLOT TYPE    
   0 password
   1 tpm2

# /usr/lib/systemd/systemd-cryptsetup attach crypt volume none tpm2-device=auto

# ls -latr /dev/mapper/crypt 
brw-rw----. 1 root root 252, 1 Mar 23 21:03 /dev/mapper/crypt
```

Various `tpm2` tools also work:
```
bash-5.1# tpm2_pcrread sha256:7+sha512:7
  sha256:
    7 : 0xD5A213BBD7D72E5EEADBE8C4DC13078B6D63A5B372FD33CA9BF17C1BC8AE89DB
  sha512:
    7 : 0xF15855D78F2984690339540A809EAE43DF9F5642611340175AFCD226991F35CF1F08A119061A2B7B01D3BCE6628D9CD99BB0FDF5419C5D0030F4C250D1F0926F
```

**Terms of contribution:**


packages/tpm2-tools/tpm2-tools.spec
```diff
@@ -0,0 +1,152 @@
...
+URL: https://github.com/tpm2-software/tpm2-tools
+Source0: %{url}/releases/download/%{version}/tpm2-tools-%{version}.tar.gz
+Source1: %{url}/releases/download/%{version}/tpm2-tools-%{version}.tar.gz.asc
+Source2: gpgkey-D533275B0123D0A679F51FF48F4F9A45D7FFEE74.asc
+
+# aws-lc doesn't have SM2 or SM3
+Patch0001: 0001-tpm2-tools-disable-SM2-and-SM3-checks.patch
+
+# libcurl isn't available
+Patch0002: 0002-tpm2-tools-disable-tpm2_getekcertificate.patch
```
arnaldo2792, line 15: nit: the commit message in the patch could explain that `tpm2_getekcertificate` isn't needed due to the lack of `libcurl`.

packages/libpopt/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "http://ftp.rpm.org/popt/releases/popt-1.x"
+
+[[package.metadata.build-package.external-files]]
+url = "http://ftp.rpm.org/popt/releases/popt-1.x/popt-1.19.tar.gz"
+sha512 = "5d1b6a15337e4cd5991817c1957f97fc4ed98659870017c08f26f754e34add31d639d55ee77ca31f29bb631c0b53368c1893bd96cf76422d257f7997a11f6466"
+force-upstream = true
```
arnaldo2792: Is this required? Is there a problem with us distributing these sources?

packages/cryptsetup/cryptsetup.spec
```diff
@@ -0,0 +1,89 @@
...
+%{summary}.
+
+%prep
+%{gpgverify} --data=<(xzcat %{S:0}) --signature=%{S:1} --keyring=%{S:2}
+%autosetup -n cryptsetup-%{version} -p1
+
+%build
+%cross_configure \
+  --disable-asciidoc \
+  --disable-fips \
```
arnaldo2792: Does this mean that there will be a separate copy of `cryptsetup` that enables FIPS?

arnaldo2792 (approved):

packages/libpopt/Cargo.toml
bcressey: > Is this required? Is there a problem with us distributing these sources?

Oops, I missed this - I'll remove it in the next revision.

packages/cryptsetup/cryptsetup.spec
bcressey: I'm not sure the best way to handle this.

For AWS-LC, I'm not confident that builds with and without FIPS enabled at compile time will always be ABI-compatible.

For cryptsetup, I could build the binaries with and without FIPS, but I'm reluctant to do that for the libraries, and without that the FIPS checks won't necessarily be enforced.

bcressey:
⬆️ force push to rebase (no changes to these commits)

bcressey:
⬆️ force push addresses feedback from @arnaldo2792

On `aws-k8s-1.32-fips` with the `cryptsetup` package added, the test case now prints:
```
# cryptsetup luksFormat --pbkdf argon2id volume 

WARNING!
========
This will overwrite data on volume irrevocably.

Are you sure? (Type 'yes' in capital letters): YES
Enter passphrase for volume: 
Verify passphrase: 
Only PBKDF2 is supported in FIPS mode.
Failed to set pbkdf parameters.
```

On a non-FIPS variant, argon2 can still be used.

packages/cryptsetup/cryptsetup.spec
arnaldo2792: Oh FIPS :melting_face:

packages/libcrypto/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/aws/aws-lc/commits/fips-2024-09-27/"
```
arnaldo2792, line 12: I wonder if we will be able to vend the one version of AWS-LC that all Bottlerocket components consume, with this now we have 3 (aws-lc-rs, libcrypto in the SDK for `gofips`). Doesn't harm but it might be confusing.

packages/libtss2/tss-sysusers.conf
```diff
@@ -0,0 +1,2 @@
+# User for TPM devices
+u tss 59
```
arnaldo2792, line 2: Was this user selected to match what other Linux distros use?

arnaldo2792 (approved):

packages/libtss2/tss-sysusers.conf
bcressey: > Was this user selected to match what other Linux distros use?

Yes, I picked the same one as Fedora.

packages/tpm2-tools/0001-tpm2-tools-disable-SM2-and-SM3-checks.patch
```diff
@@ -0,0 +1,51 @@
+From 7c0afbd0f871c1973876dc1c20648c82d01ad84d Mon Sep 17 00:00:00 2001
+From: Ben Cressey <bcressey@amazon.com>
+Date: Sat, 22 Mar 2025 16:57:25 +0000
+Subject: [PATCH 1/2] tpm2-tools: disable SM2 and SM3 checks
```
KCSesh:
```suggestion
Subject: [PATCH] tpm2-tools: disable SM2 and SM3 checks
```

packages/tpm2-tools/0002-tpm2-tools-disable-tpm2_getekcertificate.patch
```diff
@@ -0,0 +1,54 @@
+From ea01f54f0a108e35149f9b17e622dd20cc549e92 Mon Sep 17 00:00:00 2001
+From: Ben Cressey <bcressey@amazon.com>
+Date: Sat, 22 Mar 2025 16:59:46 +0000
+Subject: [PATCH 2/2] tpm2-tools: disable tpm2_getekcertificate
```
KCSesh:
```suggestion
Subject: [PATCH] tpm2-tools: disable tpm2_getekcertificate
```

packages/cryptsetup/0001-pbkdf-check-whether-FIPS-is-enabled-at-runtime.patch
```diff
@@ -0,0 +1,55 @@
...
+ 	uint32_t old_flags, memory_kb;
+ 	int r;
++	int fips_enabled = 0;
++	FILE *f;
++
++	f = fopen("/proc/sys/crypto/fips_enabled", "r");
++	if (f) {
++		r = fscanf(f, "%1d", &fips_enabled);
++		if (r != 1)
++			fips_enabled = 0;
```
ytsssun, line 44: Is it expected to default to non-fips mode if file open failed. Also would it make sense to log the file open behavior instead of silently fallback to non-fips mode?

ytsssun (approved):

packages/cryptsetup/0001-pbkdf-check-whether-FIPS-is-enabled-at-runtime.patch
bcressey: > Is it expected to default to non-fips mode if file open failed. Also would it make sense to log the file open behavior instead of silently fallback to non-fips mode?

If a kernel doesn't enable FIPS at all, then that file might not exist (and the kernel is not in FIPS mode).

I don't really want to add more logging here, since it'll change the output of what should be a standard program in ways that are unique to Bottlerocket.

bcressey:
⬆️ force push fixes patch numbering per @KCSesh

KCSesh (approved):

PR #448, opened 2025-03-27
nvidia-container-toolkit: add a comment in spec file to clarify why do we need non-template config files

### LINKED ISSUE #450: Remove unnecessary files in nvidia-container-toolkit
In https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/48/, we moved from a static file copied at runtime, to a dynamic configuration generated from a template, but forgot to clean up the older configuration file.

These two files aren't needed since they were superseded by the dynamic generation at runtime:

- https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/a4dca72f0450e63ce99cbbf0705c77520736bcc0/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s.toml
- https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/a4dca72f0450e63ce99cbbf0705c77520736bcc0/packages/nvidia-container-toolkit/nvidia-container-toolkit-tmpfiles-k8s.conf

sky1122:
**Issue number:**
Closes [#450](https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/450)

**Description of changes:**
- add a comment in nvidia-container-runtim.spce for mom-template config files

```
bash-5.1# cat /etc/nvidia-container-runtime/config.toml
### generated from the template file ###
accept-nvidia-visible-devices-as-volume-mounts = true
accept-nvidia-visible-devices-envvar-when-unprivileged = false

[nvidia-container-cli]
root = "/"
environment = []
ldconfig = "@/sbin/ldconfig"
```

**Testing done:**
build a AMI and testing if the nvidia gpu is still working
- checking the nvidia and OS version
```
[root@admin]# apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "833cf9ad-dirty",
    "pretty_name": "Bottlerocket OS 1.35.0 (aws-k8s-1.27-nvidia)",
    "variant_id": "aws-k8s-1.27-nvidia",
    "version_id": "1.35.0"
  }
}
```
- running a pod and running a nvidia workload test
```
╰$ kubectl get pod
NAME        READY   STATUS    RESTARTS   AGE
move-test   1/1     Running   0          2m23s
╭─ in ~/k8s
╰$ kubectl logs move-test
╭─ in ~/k8s
╰$ cat pod.yaml
apiVersion: v1
kind: Pod
metadata:
  name: move-test
spec:
  restartPolicy: OnFailure
  containers:
    - name: test
      image: nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda10.2
      command: ["sleep", "infinity"]
      resources:
        limits:
          nvidia.com/gpu: 1

╭─ in ~/k8s
╰$ kubectl exec -it move-test -- /bin/bash
root@move-test:/# find / -name vectorAdd
/tmp/vectorAdd
root@move-test:/# /tmp/vectorAdd
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
root@move-test:/#
```

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
@@ -14,6 +14,8 @@ License: Apache-2.0
 URL: https://%{goimport}
 
 Source0: https://%{goimport}/archive/v%{gover}/nvidia-container-toolkit-%{gover}.tar.gz
+# non-templated version of the config files for k8s are provided for downstream
+# builders that don't use the NVIDIA Container Runtime settings
 Source1: nvidia-container-toolkit-config-k8s.toml
 Source2: nvidia-container-toolkit-config-ecs.toml
 Source3: nvidia-oci-hooks-json


yeazelm (approved):

ytsssun (approved):

arnaldo2792 (approved):

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
```diff
@@ -14,13 +14,11 @@ License: Apache-2.0
 URL: https://%{goimport}
 
 Source0: https://%{goimport}/archive/v%{gover}/nvidia-container-toolkit-%{gover}.tar.gz
-Source1: nvidia-container-toolkit-config-k8s.toml
-Source2: nvidia-container-toolkit-config-ecs.toml
-Source3: nvidia-oci-hooks-json
-Source4: nvidia-gpu-devices.rules
-Source5: nvidia-container-toolkit-tmpfiles-ecs.conf
-Source6: nvidia-container-toolkit-tmpfiles-k8s.conf
-Source7: nvidia-container-toolkit-config-k8s
```
bcressey: The point of the non-templated version of the config file is to ensure that some valid configuration is present even if the downstream variant isn't configured with the corresponding settings. I don't think we want to drop it.

bcressey (changes requested):
Not sure these files are truly unnecessary.

sky1122:
> Not sure these files are truly unnecessary.

Left a comment on the spec file to explain why we need to keep the non-template version of the config files so that in the future team will know why we keep this.

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
```diff
@@ -13,6 +13,8 @@ Summary: Tool to build and run GPU accelerated containers
 License: Apache-2.0
 URL: https://%{goimport}
 
+# non-templated version of the config files for k8s are to ensure the downstream
+# variant have ability to config with their own correspending settings.
```
arnaldo2792: Also, you can move the comment on top of `Source:1`

```suggestion
# non-templated version of the config files for k8s are provided for downstream
# builders that don't use the NVIDIA Container Runtime settings
```

arnaldo2792 (approved):

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
sky1122: Changed

bcressey (approved):

ytsssun (approved):

PR #451, opened 2025-03-29
fix aws-signing-helper and IAM Roles Anywhere

### LINKED ISSUE #4390: Add `aws-iam-authenticator` package to vmware variants
We would like to attach Bottlerocket OVAs as [hybrid nodes](https://aws.amazon.com/eks/hybrid-nodes/) to an EKS cluster but currently this is not possible as `aws-iam-authenticator` binary is not available on the `vmware-k8s-*` variants out of the box. `aws-iam-authenticator` is required to authenticate the kubelet on the worker node against the kube-apiserver as a client, for which it will use the AWS credentials provided by SSM or IAM Roles Anywhere.

Related to #2823

### LINKED ISSUE #4456: kubelet process "sh" access for IAM Roles Anywhere ECR pulls
Hi,

We are having an issue with Bottlerocket when trying to use the Credential Provider Package (IAM Roles Anywhere) for EKS Anywhere.

The package makes use of the ecr-credential-provider for kubelet, and it looks as though the flow goes like:

kubelet process -> ecr-credential-provider -> AWS config "credential_process" -> aws-signing-helper binary

The AWS config on the hosts looks like so:

```
[default]
region = eu-west-2
credential_process = aws_signing_helper credential-process --certificate /var/lib/kubelet....
```

The issue is that it looks as though "credential_process" in an AWS config file spawns a subprocess shell via "sh" to run the command that is specified. I can't find any mention of this subprocess spawn in the AWS docs (https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-sourcing-external.html).
In this case, the sandboxed environment that the kubelet process runs from (via systemctl) does not have access to "sh", resulting in an error:

```
bash-5.1# nsenter --target $(pidof kubelet) --mount --uts --ipc --net --pid -- ls /bin/sh
ls: cannot access '/bin/sh': No such file or directory
```

**Image I'm using:**
VERSION="1.32.0 (vmware-k8s-1.32)"
PRETTY_NAME="Bottlerocket OS 1.32.0 (vmware-k8s-1.32)"
VARIANT_ID=vmware-k8s-1.32
VERSION_ID=1.32.0
BUILD_ID=cacc4ce9

I can pull images from a private ECR repo from kubelet through the use of ecr-credential-provider and aws-signing-helper (for IAM Roles Anywhere).

**What actually happened:**

In our Kubelet logs on our worker nodes, we have errors like:
```
kubelet[5703]: E0327 08:14:35.655164    6910 main.go:262] Error running credential provider plugin: operation error ECR: GetAuthorizationToken, get identity: get credentials: failed to refresh cached credentials, process provider error: error in credential_process: exec: "sh": executable file not found in $PATH
```

and private ECR repo images cannot be pulled by kubelet.

I'm wondering if there is a boot or kernel parameter that can be used to grant the kubelet process access to the "sh" binary? These seem to be the only configurable options from the EKS-A curated package: https://anywhere.eks.amazonaws.com/docs/getting-started/optional/hostosconfig/#bottlerocketconfiguration

It seems that we are encountering a similar issue to this: https://github.com/aws/aws-sdk-go-v2/issues/2879

For additional context - I am following the exact instructions on this page: https://anywhere.eks.amazonaws.com/docs/packages/credential-provider-package/iam_roles_anywhere/ 

Thanks for any help!

bcressey:
**Issue number:**
Fixes [#4390](https://github.com/bottlerocket-os/bottlerocket/issues/4390), fixes [#4456](https://github.com/bottlerocket-os/bottlerocket/issues/4456)

**Description of changes:**
Add `brush` as the preferred handler for the `sh -c '...'` invocations that the AWS SDK for Go uses to execute credential processes.

Add a dependency to `aws-iam-authenticator` and `ecr-credential-provider` on `aws-signing-helper`, since these programs need that signing helper in order to work with IAM Roles Anywhere.

Simplify `kubelet` dependencies so it only depends on packages it uses (or can use) directly, rather than a transitive dependency like the signing helper.

**Testing done:**
Verified that `vmware-k8s-1.32` nodes could join an EKS cluster with the appropriate IAM RA setup.

`brush` in action:
```
# sh -c 'aws_signing_helper --help'
A tool that utilizes certificates and their associated private keys to 
sign requests to AWS IAM Roles Anywhere's CreateSession API and retrieve temporary 
AWS security credentials. This tool exposes multiple commands to make credential 
retrieval and rotation more convenient.

# sh -c 'foo bar'
'foo' could not be found

Caused by these errors (recent errors listed first):
  1: failed to locate 'foo' in PATH
  2: cannot find binary path

# sh -c 'echo bar'
'echo' is not allowed

Caused by these errors (recent errors listed first):
  1: failed to canonicalize '/usr/libexec/brush/allowed-programs/echo'
  2: No such file or directory (os error 2)
```

Verified that the epoch + capability combo works as expected:
* `brush` is installed by default for `aws-k8s-*` and `vmware-k8s-*` variants
* `aws-dev` uses `bash` if `aws-signing-helper` is added to its package list

**Terms of contribution:**

### DIFF
--- a/sources/brush/src/main.rs
+++ b/sources/brush/src/main.rs
@@ -1,0 +1,625 @@
+/*!
+# Introduction
+
+*brush* is a sanitizing pseudo-shell. Its purpose is to parse `sh -c '<program> ...'` invocations to
+decide whether the requested program should be allowed to run. It splits its third argument based
+on POSIX shell rules. Any additional input validation is left up to the invoked program.
+
+Programs can be allowed by creating a symlink in the `allowed-programs` directory:
+```bash
+/usr/libexec/brush/allowed-programs/my-program -> ../path/to/program
+```
+
+In most cases the symlink target should be in `/usr/bin` or `/usr/sbin`, since *brush* will use
+those paths to find the requested program if it is given as a bare filename.
+
+If the requested program is allowed, *brush* will execute it directly and never return. Otherwise,
+it will print an error summary and return with exit code 127 to indicate that the command could not
+be executed.
+
+Arguments for programs can be further restricted with a configuration file in this directory:
+```bash
+/usr/share/brush/my-program.toml
+```
+
+Both required arguments and blocked arguments can be specified:
+```toml
+# Required args must appear in positional order.
+required-args = ["first", "second", "third"]
+
+# Blocked args must not appear.
+blocked-args = ["--spill-secrets"]
+```
+*/
+
+use std::collections::HashSet;
+use std::ffi::{OsStr, OsString};
+use std::fs;
+use std::io::ErrorKind;
+use std::os::unix::process::CommandExt;
+use std::path::{Path, PathBuf};
+use std::process::{Command, ExitCode};
+
+use serde::{Deserialize, Serialize};
+use shlex::Shlex;
+use snafu::{prelude::*, FromString, Report, ResultExt, Whatever};
+use which::CanonicalPath;
+
+type Result<T> = std::result::Result<T, Whatever>;
+
+// These should be read-only directories to prevent trivially altering either the set of allowed
+// programs, or the arguments that are required or blocked.
+static ALLOWED_PROGRAMS_DIR: &str = "/usr/libexec/brush/allowed-programs";
+static CONFIG_DIR: &str = "/usr/share/brush";
+
+// These are the paths that will be searched if the program is requested by name.
+static ALLOWED_PATHS: &str = "/usr/sbin:/usr/bin";
+
+// Ensure that the provided arguments conform to the `sh -c '...'` interface, parse the third
+// argument with shlex, and return the program name and arguments.
+fn parse_args(args: Vec<String>) -> Result<(String, Vec<String>)> {
+    let argc = args.len();
+    let mut args_iter = args.into_iter();
+    ensure_whatever!(argc == 3, "expected 3 args, found {argc}");
+
+    let arg0 = args_iter.next().unwrap();
+    ensure_whatever!(arg0 == "sh", "expected first arg 'sh', not '{arg0}'");
+
+    let arg1 = args_iter.next().unwrap();
+    ensure_whatever!(arg1 == "-c", "expected second arg '-c', not '{arg1}'");
+
+    let arg2 = args_iter.next().unwrap();
+    let mut split_input = Shlex::new(&arg2);
+    let split_arg0 = split_input
+        .next()
+        .with_whatever_context(|| "failed to identify requested program from shell input")?;
+    let split_argv = split_input.collect();
+
+    Ok((split_arg0, split_argv))
+}
+
+// Find the canonical path to the provided program, which can be relative or absolute, resolving
+// any symlinks or relative path components.
+fn resolve_program(program: impl AsRef<Path>, paths: impl AsRef<str>) -> Result<CanonicalPath> {
+    let program = program.as_ref();
+    let paths = paths.as_ref();
+    let cwd = std::env::current_dir().unwrap_or_default();
+    which::CanonicalPath::new_in(program, Some(paths), cwd)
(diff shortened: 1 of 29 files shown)


sources/brush/src/main.rs
```diff
@@ -0,0 +1,429 @@
+/*!
+# Introduction
+
+*brush* is a sanitizing pseudo-shell. Its purpose is to parse `sh -c '<program> ...'` invocations to
+decide whether the requested program should be allowed to run. It splits its third argument based
+on POSIX shell rules. Any additional input validation is left up to the invoked program.
+
+Programs can be allowed by creating a symlink in `allowed-programs` directory:
+```
```
arnaldo2792: nit: can you force `bash` or `text` here? Otherwise the generated README uses `rust`.

arnaldo2792 (approved):

sources/brush/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+        resolved_program_path.to_string_lossy(),
+    );
+
+    Ok((verified_program_path, requested_program_name))
+}
+
+// Wrapper function that either executes the requested program, if allowed, or returns an error.
+fn run(args: Vec<String>, allowed_programs_dir: &str, allowed_paths: &str) -> Result<()> {
+    let (program, args) = parse_args(args)?;
+    let (program, arg0) = verify_program(allowed_programs_dir, allowed_paths, &program)?;
```
yeazelm, line 234: nit: although more efficient to reuse the same variable, I think I'd feel a bit better about keeping the unverified `program` in a separate variable to avoid future refactoring that loses the safety of only executing the verified program. 
```suggestion
    let (verified_program, arg0) = verify_program(allowed_programs_dir, allowed_paths, &program)?;
```

sources/brush/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+    fn test_parse_args() {
+        let good = args_from(&["sh", "-c", "qwerty --uiop asdf"]);
+        let expected = cmd_from(&["qwerty", "--uiop", "asdf"]);
+        assert_eq!(parse_args(good).unwrap(), expected);
+
+        let bad = args_from(&["sh", "zxcvb"]);
+        assert!(parse_args(bad).is_err());
+
+        let bad = args_from(&["sh", "-c"]);
+        assert!(parse_args(bad).is_err());
```
yeazelm, line 521: nit: I'd like to catch the other case I could see a user doing where they pass all of the arguments as separate args and we want to explicitly signal this isn't a bug but one should pass the entire list as a single string for the program execution.
```suggestion
        assert!(parse_args(bad).is_err());
        
        let bad = args_from(&["sh", "-c", "querty", "--uiop", "asdf"]);
        assert!(parse_args(bad).is_err());
```

yeazelm (approved):
Just a few nits but looks great!

sources/brush/src/main.rs
bcressey: > nit: although more efficient to reuse the same variable

Rust uses SSA so there is no efficiency gain or loss either way

> I'd feel a bit better about keeping the unverified program in a separate variable to avoid future refactoring that loses the safety of only executing the verified program.

I didn't want the "old" binding for `program` here to stay alive, but having `verify_program` take an owned value for `program` is annoying for unit tests.

sources/brush/src/main.rs
bcressey: Added this test case.

bcressey:
⬆️ force push adds a patch to disable `--debug` for `aws_signing_helper`, which can cause credentials to end up logged to the journal.

bcressey:
⬆️ force push drops the `--debug` patch for `aws_signing_helper`, and adds argument validation to `brush` instead.

bcressey:
⬆️ force push to set the edition back to 2021, pending a tree-wide update to 2024

arnaldo2792 (approved):

yeazelm (approved):

PR #456, opened 2025-04-02
migrator: clear configuration-files and services

cbgbt:
**Description of changes:**
In `bottlerocket-core-kit` [v6.0.0](https://github.com/bottlerocket-os/bottlerocket-core-kit/releases/tag/v6.0.0) we began to clear weak settings and all metadata from Bottlerocket's datastore whenever the migrator performed an upgrade or downgrade.

This change expands this behavior to "transient data" which also now includes:
* `services` -- configurations for processes to restart when certain Bottlerocket settings change
* `configuration-files` -- Configuration files to re-render from templates when certain Bottlerocket settings change.

The change also includes a slight refactor of migrator to make it more clear that the process will exit when no version change has occurred.

**Testing done:**
* [x] Unit testing
* [x] Ensure configuration-files and services are cleared & restored on upgrade
* [x] Ensure configuration-files and services are cleared & restored on downgrade

**Terms of contribution:**

### DIFF
--- a/sources/api/migration/migrator/src/main.rs
+++ b/sources/api/migration/migrator/src/main.rs
@@ -75,11 +75,16 @@ async fn main() {
     }
 }
 
-async fn get_current_version<P>(datastore_dir: P) -> Result<Version>
+async fn get_current_version<P>(datastore_path: P) -> Result<Version>
 where
     P: AsRef<Path>,
 {
-    let datastore_dir = datastore_dir.as_ref();
+    let datastore_path = datastore_path.as_ref();
+    let datastore_dir = datastore_path
+        .parent()
+        .context(error::DataStoreLinkToRootSnafu {
+            path: datastore_path,
+        })?;
 
     // Find the current patch version link, which contains our full version number
     let current = datastore_dir.join("current");
@@ -116,27 +121,9 @@ where
 }
 
 pub(crate) async fn run(args: &Args) -> Result<()> {
-    let migrated_datastore = perform_migrations(args).await?;
+    let current_version = get_current_version(&args.datastore_path).await?;
 
-    // Remove all the weak setting and all metadata
-    let datastore =
-        remove_weak_settings_and_metadata(migrated_datastore, &args.migrate_to_version).await?;
-    flip_to_new_version(&args.migrate_to_version, datastore).await?;
-
-    Ok(())
-}
-
-pub(crate) async fn perform_migrations(args: &Args) -> Result<PathBuf> {
-    // Get the directory we're working in.
-    let datastore_dir = args
-        .datastore_path
-        .parent()
-        .context(error::DataStoreLinkToRootSnafu {
-            path: &args.datastore_path,
-        })?;
-
-    let current_version = get_current_version(datastore_dir).await?;
-    let direction = Direction::from_versions(&current_version, &args.migrate_to_version)
+    let migration_version_meta = MigrationVersionMeta::new(current_version, args.migrate_to_version.clone())
         .unwrap_or_else(|| {
             info!(
                 "Requested version {} matches version of given datastore at '{}'; nothing to do. Exiting from migrator.",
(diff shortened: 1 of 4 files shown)


bcressey (approved):
Nice!

cbgbt:
^ force push a slight refactoring of the `migrator` to make it more clear in `run()` that migrator will early exit if there is nothing to do.

cbgbt:
Test results: https://gist.github.com/cbgbt/5b1c3ff337183d367694cdbcebf3ea07

sources/api/migration/migrator/src/main.rs
```diff
@@ -116,27 +121,37 @@ where
...
 
 pub(crate) async fn run(args: &Args) -> Result<()> {
-    let migrated_datastore = perform_migrations(args).await?;
+    let current_version = get_current_version(&args.datastore_path).await?;
+
+    exit_unless_version_change(&current_version, args).await?;
+
+    let migrated_datastore = perform_migrations(&current_version, args).await?;
 
     // Remove all the weak setting and all metadata
```
ytsssun: Nit: do we also need to update the comments here now that we are clearing more resources.

sources/api/migration/migrator/src/main.rs
```diff
@@ -323,6 +345,22 @@ fn remove_metadata_from_datastore(datastore: &mut DataStoreData) -> Result<()> {
     Ok(())
 }
 
+// Removes all data from the datastore where the key begins with the given prefix
+fn remove_keys_with_prefix(datastore: &mut DataStoreData, prefix: &str) -> Result<()> {
+    let keys_to_remove = datastore
+        .data
+        .keys()
+        .filter(|key| key.starts_with(prefix))
+        .cloned()
```
ytsssun: Open question - is it possible to avoid the clone here and perform the removal in the one-pass? Not sure how costly the copy would be. If performance is not a concern, I am good with the change as the readability is great.

ytsssun (approved):

sources/api/migration/migrator/src/main.rs
cbgbt: Nice catch. I think the readability here is strong enough that we can just delete the comment -- I'll do it.

sources/api/migration/migrator/src/main.rs
```diff
@@ -116,27 +121,37 @@ where
...
+    args: &Args,
+) -> Result<()> {
+    if current_version == &args.migrate_to_version {
+        info!(
+            "Requested version {} matches version of given datastore at '{}'; nothing to do. Exiting from migrator.",
+            args.migrate_to_version,
+            args.datastore_path.display()
+        );
+        process::exit(0);
+    }
```
bcressey: This same check is duplicated in `perform_migrations`, which runs right after `exit_unless_version_change` is called. Do we need both?

sources/api/migration/migrator/src/main.rs
cbgbt: In general, not without writing unsafe code because the first iterator borrows the key immutably, which in turn borrows the datastore. Since that borrow is held, we can't simultaneously perform a mutable borrow to perform the deletion.

But! The HashMap implementation actually has the requisite `unsafe` code [abstracted away in the `retain` method](https://github.com/rust-lang/hashbrown/blob/master/src/map.rs#L917). Thanks for prompting me to look!

cbgbt:
^ push to address feedback from @ytsssun

cbgbt:
^ push to address feedback from @bcressey

sources/api/migration/migrator/src/main.rs
cbgbt: Technically no, but we need to compute `Direction`, at which point we need to handle the `None` case.

I refactored it a bit to compute `Direction` at the same time that we make the decision whether or not to exit. It cost a bit in readability (I really liked the `exit_unless()` function for that reason) but I think it's still pretty clear from looking at `run()` that the migrator will flatly exit if there's no version change.

bcressey (approved):
Nice!

sources/api/migration/migrator/src/main.rs
```diff
@@ -323,17 +351,22 @@ fn remove_metadata_from_datastore(datastore: &mut DataStoreData) -> Result<()> {
     Ok(())
 }
 
+// Removes all data from the datastore where the key begins with the given prefix
+fn remove_keys_with_prefix(datastore: &mut DataStoreData, prefix: &str) -> Result<()> {
+    datastore.data.retain(|key, _| !key.starts_with(prefix));
+    Ok(())
```
ytsssun, line 357: Wow, this is neat!

ytsssun (approved):

cbgbt:
I'm holding off on merging this until I have a PR open for the related migration in bottlerocket-os/bottlerocket

KCSesh (approved):
Cool! 😄

PR #459, opened 2025-04-08
Support both CDI and Legacy NVIDIA Container Runtime modes

### LINKED ISSUE #468: Add support for CDI in Kubernetes
The NVIDIA Container Toolkit supports the Container Device Interface (CDI), which is a fresher, more standard mechanism to support third-party hardware vendors.

sky1122:
**Issue number:**

Closes #468

**Description of changes:**

- Set DriverRoot and DefaultContainterDriverRoot in nvidia-k8s-device-plugin
    - [k8s-device-plugin](https://github.com/NVIDIA/k8s-device-plugin/blob/bafe5afdc325c9185827e50ed548415b221f5c27/api/config/v1/consts.go#L55) always mount the DefaultContainterDriverRoot at [/driver-root](https://github.com/NVIDIA/k8s-device-plugin/blob/bafe5afdc325c9185827e50ed548415b221f5c27/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml#L202), but for bottlerocket the driver is in the root filesystem.
 - Select the correct `mode` for the default NVIDIA container runtime, based off the configurations set for the `device-list-strategy` in the k8s device plugin
 - Disable mode detection in the prestart hook, as it prevents the NVIDIA legacy runtime from working when the `mode` for the default NVIDIA runtime is set to `cdi`.

**Testing done:**
In combination with: 

- [settings-sdk](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/77)
- [bottlerocket-set k8s deviceliststrategy default config](https://github.com/bottlerocket-os/bottlerocket/pull/4475)
- [bottlerocket-remove unnecessary settings for OCI Hooks API ](https://github.com/bottlerocket-os/bottlerocket/pull/4474)
- https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/511

<details>

In variants: `aws-k8s-1.29`, `aws-k8s-1.31`, `aws-k8s-1.32`, `aws-k8s-1.33` for x86_64 and aarch64:

- Verified that all the containerd runtimes are functional
- Verified that the correct mode is configured depending on the `device-list-strategy` values:

<table>
  <thead>
    <tr>
      <th colspan=3>API Value</th>
      <th>Mode</th>
      <th>Status</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>envvar</td>
      <td>-</td>
      <td>-</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>envvar</td>
      <td>volume-mounts</td>
      <td>-</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>envvar</td>
      <td>volume-mounts</td>
      <td>cdi-cri</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>envvar</td>
      <td>cdi-cri</td>
      <td>-</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>ennvar</td>
      <td>cdi-cri</td>
      <td>volume-mounts</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>volume-mounts</td>
      <td>-</td>
      <td>-</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>volume-mounts</td>
      <td>envvar</td>
      <td>-</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>volume-mounts</td>
      <td>envvar</td>
      <td>cdi-cri</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>volume-mounts</td>
      <td>cdi-cri</td>
      <td>-</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>volume-mounts</td>
      <td>cdi-cri</td>
      <td>envvar</td>
      <td>legacy</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>cdi-cri</td>
      <td>-</td>
      <td>-</td>
      <td>cdi</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>cdi-cri</td>
      <td>volume-mounts</td>
      <td>-</td>
      <td>cdi</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>cdi-cri</td>
      <td>volume-mounts</td>
      <td>envvar</td>
      <td>cdi</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>cdi-cri</td>
      <td>envvar</td>
      <td>-</td>
      <td>cdi</td>
      <td>Pass</td>
    </tr>
    <tr>
      <td>cdi-cri</td>
      <td>envvar</td>
      <td>volume-mounts</td>
      <td>cdi</td>
      <td>Pass</td>
    </tr>
  </tbody>
</table>

</details>

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
@@ -31,3 +31,9 @@ mode="legacy"
 {{else}}
 mode="legacy"
 {{/if}}
+
+[nvidia-container-runtime-hook]
+# For the legacy NVIDIA runtime, skip detecting the mode used in the
+# NVIDA Container Runtime. This prevents failures in the legacy NVIDIA runtime
+# when the selected mode is 'cdi'.
+skip-mode-detection = true
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
@@ -1,14 +1,33 @@
 [required-extensions]
 nvidia-container-runtime = "v1"
-std = { version = "v1", helpers = ["default"] }
+kubelet-device-plugins = "v1"
+std = { version = "v1", helpers = ["default", "is_array"] }
 
 +++
 ### generated from the template file ###
 accept-nvidia-visible-devices-as-volume-mounts = {{default true settings.nvidia-container-runtime.visible-devices-as-volume-mounts}}
 accept-nvidia-visible-devices-envvar-when-unprivileged = {{default false settings.nvidia-container-runtime.visible-devices-envvar-when-unprivileged}}
-
 [nvidia-container-cli]
 root = "/"
 path = "/usr/bin/nvidia-container-cli"
 environment = []
 ldconfig = "@/sbin/ldconfig"
+
+[nvidia-container-runtime]
+{{#if settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+{{~#if (is_array settings.kubelet-device-plugins.nvidia.device-list-strategy) ~}}
+{{~#if (eq settings.kubelet-device-plugins.nvidia.device-list-strategy.[0] "cdi-cri") ~}}
+mode="cdi"
+{{~else~}}
+mode="legacy"
+{{~/if~}}
+{{~else~}}
+{{~#if (eq settings.kubelet-device-plugins.nvidia.device-list-strategy "cdi-cri") ~}}
+mode="cdi"
+{{~else~}}
+mode="legacy"
+{{~/if~}}
+{{/if}}
+{{else}}
+mode="legacy"
+{{/if}}
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
@@ -1,6 +1,6 @@
 [required-extensions]
 kubelet-device-plugins = "v1"
-std = { version = "v1", helpers = ["default"] }
+std = { version = "v1", helpers = ["default", "join_array", "is_array"] }
 
 +++
 version: v1
@@ -18,7 +18,15 @@ flags:
   nvidiaDriverRoot: "/"
   plugin:
     passDeviceSpecs: {{default true settings.kubelet-device-plugins.nvidia.pass-device-specs}}
-    deviceListStrategy: {{default "volume-mounts" settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+    {{#if settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+    {{#if (is_array settings.kubelet-device-plugins.nvidia.device-list-strategy)}}
+    deviceListStrategy: [{{join_array "," settings.kubelet-device-plugins.nvidia.device-list-strategy }}]
+    {{else}}
+    deviceListStrategy: {{settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+    {{/if}}
+    {{else}}
+    deviceListStrategy: "volume-mounts"
+    {{/if}}
     deviceIDStrategy: {{default "index" settings.kubelet-device-plugins.nvidia.device-id-strategy}}
     containerDriverRoot: "/"
 {{#if settings.kubelet-device-plugins.nvidia.device-sharing-strategy}}
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
@@ -15,10 +15,12 @@ flags:
   migStrategy: "none"
   {{/if}}
   failOnInitError: true
+  nvidiaDriverRoot: "/"
   plugin:
     passDeviceSpecs: {{default true settings.kubelet-device-plugins.nvidia.pass-device-specs}}
     deviceListStrategy: {{default "volume-mounts" settings.kubelet-device-plugins.nvidia.device-list-strategy}}
     deviceIDStrategy: {{default "index" settings.kubelet-device-plugins.nvidia.device-id-strategy}}
+    containerDriverRoot: "/"
 {{#if settings.kubelet-device-plugins.nvidia.device-sharing-strategy}}
 {{#if (eq settings.kubelet-device-plugins.nvidia.device-sharing-strategy "time-slicing")}}
 sharing:


sky1122:
force pushed to drop unrelated commits

sky1122:
Forced push to change the commit content and add subject for patch

sky1122:
Forced push to add sign off in one commit

sky1122:
forced push to add all the changes from #467 to this PR

packages/nvidia-k8s-device-plugin/1100-k8s-device-plugin-all-device-patch.patch
```diff
@@ -0,0 +1,36 @@
+From ae5521585da70ff06ae66c9b02613cf85b5c3c0a Mon Sep 17 00:00:00 2001
+From: Jingwei Wang <jweiw@amazon.com>
+Date: Tue, 18 Mar 2025 23:23:40 +0000
+Subject: [PATCH] k8s-device-plugin all device patch
```
arnaldo2792: Your commit message should be more descriptive about why the change is needed. That will also help us contribute the patch upstream.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
```diff
@@ -18,6 +18,7 @@ Source2: nvidia-k8s-device-plugin-conf
 Source3: nvidia-k8s-device-plugin-exec-start-conf
 Source4: nvidia-k8s-device-plugin-mig-conf
 
+Patch1100: 1100-k8s-device-plugin-all-device-patch.patch
```
arnaldo2792: We could begin numbering the patches at `0001`.

packages/nvidia-k8s-device-plugin/1100-k8s-device-plugin-all-device-patch.patch
```diff
@@ -0,0 +1,36 @@
+From ae5521585da70ff06ae66c9b02613cf85b5c3c0a Mon Sep 17 00:00:00 2001
```
arnaldo2792: Posting here for the lack of a better place to add this comment: you have a typo in your commit message

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -15,10 +15,12 @@ flags:
   migStrategy: "none"
   {{/if}}
   failOnInitError: true
+  nvidiaDriverRoot: "/"
```
arnaldo2792, line 18: These changes don't technically enable the CDI support in the device plugin. You could say:

```
nvidia-device-plugin: update driver root

The NVIDIA Device Plugin defaults to `/driver-root` because it expects to be running
as a container image where the driver root is mounted at that location. In Bottlerocket,
the driver root is the actual root filesystem.
```

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
```diff
@@ -12,3 +12,12 @@ root = "/"
 path = "/usr/bin/nvidia-container-cli"
 environment = []
 ldconfig = "@/sbin/ldconfig"
+
+[nvidia-contaienr-runtime]
```
arnaldo2792: You have a typo here

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -1,10 +1,12 @@
 [required-extensions]
 kubelet-device-plugins = "v1"
+container-runtime = "v1"
```
arnaldo2792: Why do you need the container runtime settings here? They are not used anywhere in this file.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -1,10 +1,12 @@
...
 std = { version = "v1", helpers = ["default"] }
 
 +++
 version: v1
 flags:
+
```
arnaldo2792: nit: extra comment

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -18,7 +20,15 @@ flags:
...
-    deviceListStrategy: {{default "volume-mounts" settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+    {{#if settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+    {{#if (eq settings.kubelet-device-plugins.nvidia.device-list-strategy "cdi-cri")}}
+    deviceListStrategy: "cdi-cri"
+    {{else}}
+    deviceListStrategy: {{settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+    {{/if}}
+    {{else}}
+    deviceListStrategy: "volume-mounts"
+    {{/if}}
```
arnaldo2792: You don't have to conditionally setting the `cdi-cri` value. You can just keep the line as it was before, since we control the CDI support in the NVIDIA Container Runtime configuration.

arnaldo2792 (changes requested):

sky1122:
force-pushed to fix the issue in the conversation

packages/nvidia-k8s-device-plugin/0001-Add-CDI-specs-for-the-all-device.patch
```diff
@@ -0,0 +1,45 @@
...
+
+diff --git a/internal/cdi/cdi.go b/internal/cdi/cdi.go
+index fae421b..0c5a4b4 100644
+--- a/internal/cdi/cdi.go
++++ b/internal/cdi/cdi.go
+@@ -38,6 +38,7 @@ import (
+ 
+ const (
+ 	cdiRoot = "/var/run/cdi"
++  allDeviceName = "all"
```
arnaldo2792: You should fix the indentation here, so that we could contribute this upstream if required

packages/nvidia-k8s-device-plugin/0001-Add-CDI-specs-for-the-all-device.patch
```diff
@@ -0,0 +1,45 @@
...
+Subject: [PATCH]  Add CDI specs for the "all" device
+
+The Device Plugin cannot generate CDI specifications
+when NVIDIA_VISIBLE_DEVICES is set to "all". This limitation breaks compatibility
+with containers that depend on this setting, e.g. DCGM Exporter.
+
+Fix this by mimicking what nvidia-ctk does[1], and add the "all" device
+to the CDI specifications.
+
+[1]: https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/cmd/nvidia-ctk/cdi/generate/generate.go#L300-L303
```
arnaldo2792: You should provide the perma link instead of the link to the main branch.

packages/nvidia-k8s-device-plugin/0001-Add-CDI-specs-for-the-all-device.patch
```diff
@@ -0,0 +1,45 @@
+From 1ac450cfc861200a77685dd0ba77394963217502 Mon Sep 17 00:00:00 2001
```
arnaldo2792: Adding this for the lack of a better place:

nit: you could use the description of your patch in your commit message

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
arnaldo2792: Now your commit message header is a bit larger than expected, could you please fix that?. Also some nits for my previous suggestion (:sweat: ):

```
The NVIDIA Device Plugin looks for libraries and binaries at `/driver-root`
because it expects to be running as a container where the driver root is
mounted at that location. In Bottlerocket, the driver root is the actual root
filesystem.
```

packages/containerd/containerd-config-toml_k8s_nvidia_containerd_sock
```diff
@@ -21,6 +21,7 @@ address = "/run/containerd/containerd.sock"
 [plugins."io.containerd.grpc.v1.cri"]
```
arnaldo2792: Leaving this here for the lack of a better place to post it:

In your commit message, you can mention that CDI support is required so that containerd updates the OCI spec based off the CDI specifications generated by the Device Plugin.

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
```diff
@@ -12,3 +12,12 @@ root = "/"
 path = "/usr/bin/nvidia-container-cli"
 environment = []
 ldconfig = "@/sbin/ldconfig"
+
+[nvidia-container-runtime]
```
arnaldo2792, line 16: Posting this here for the lack of a better place to post it:

You can just say

```
nvidia-container-toolkit: support both legacy and CDI stacks

Update the NVIDIA Container Toolkit configuration template to dynamically
select the mode based off the configurations used in the Kubernetes Device
Plugin.
```

sky1122:
force-pushed to address the conversation above.

sky1122:
force-pushed to add a new patch in nvidia-k8s-device-plugin

sky1122:
force-pushed to change one commit message

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
```diff
@@ -12,3 +12,12 @@ root = "/"
...
+{{#if settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+{{#if (eq settings.kubelet-device-plugins.nvidia.device-list-strategy "cdi-cri")}}
```
elezar: Setting `mode="cdi"` heres is correct, but possibly for the wrong reasons.

Note that if the mode is `cdi-cri`, the `nvidia-container-runtime` will not perform the injection since it cannot read the CRI fields set by the plugin. This means that the `nvidia-container-runtime` performs no modification to the OCI runtime specification and forwards the command to `runc`.

In the case where `cdi-cri` is used one does not strictly need the `nvidia` runtime at all. In the case of the GPU Operator we "solve" this by:
1. Not setting the `nvidia` runtime as the default in containerd.
2. Add a `RuntimeClass` for the `nvidia` runtime.
3. Ensuring that "management" containers such as the k8s-device-plugin that themselves need access to GPUs use this `runtime` class.

arnaldo2792: Hey @elezar, thanks again for the interest in this PR!

>Note that if the mode is cdi-cri, the nvidia-container-runtime will not perform the injection since it cannot read the CRI fields set by the plugin. This means that the nvidia-container-runtime performs no modification to the OCI runtime specification and forwards the command to runc.

We use `cdi-cri` to let the Kubernetes Device Plugin generate the CDI specifications, and we enabled the CDI support in `containerd` to perform the CDI updates. So `nvidia-container-runtime` effectively does nothing since all the updates were already applied by containerd.

The reason why we used `nvidia-container-runtime` at all is to allow users use `NVIDIA_VISIBLE_DEVICES` since, even though we understand that this environment variable is meant to be used solely for "management" containers, we have seen Bottlerocket users rely on this "feature" to allow over-subscription of the GPUs. As I said elsewhere, we do advice against this but we also don't want to break users that heavily rely on this type of oversubscription.

What we want to achieve with the setup that we have is a "hands-free" experience, where the users just create their pod specifications without having to worry about the `RuntimeClass` they have to use as we have it today with the legacy support. Additionally, we wanted to stop injecting the prestart hook even from management containers to fully rely on CDI.

Please let us know if you think there are problems with this approach. FWIW, we don't run the k8s device plugin as a daemonset, we run it as a service in the host but users own running other "management" containers like `dcgm-exporter`.

arnaldo2792: Just to circle back to this discussion, we decided to include the additional runtimes to be more compatible with the GPU Operator. Thanks for the feedback @elezar!

sky1122:
force pushed to change the commit format and comment to make it looks better

sky1122:
force pushed to fix the handle bar

arnaldo2792:
Last push includes:
- Add missing "header" to the k8s device plugin configuration
- Disable mode detection for prestart hook

yeazelm (approved):

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
```diff
@@ -1,14 +1,39 @@
...
 root = "/"
 path = "/usr/bin/nvidia-container-cli"
 environment = []
 ldconfig = "@/sbin/ldconfig"
+
+[nvidia-container-runtime]
+{{#if settings.kubelet-device-plugins.nvidia.device-list-strategy}}
+{{~#if (is_array settings.kubelet-device-plugins.nvidia.device-list-strategy) ~}}
+{{~#if (eq settings.kubelet-device-plugins.nvidia.device-list-strategy.[0] "cdi-cri") ~}}
+mode="cdi"
```
ytsssun, line 20: Is it expected that "cdi-cri" will always be the first item of the array? What is the behavior if we have multiple items in the array, and what if "cdi-cri" is the second item in the array? 

May not be blocking. But it may be good to implement some helper like "has/contains" - similar to what they did in the device plugin - https://github.com/NVIDIA/k8s-device-plugin/blob/6f41f70c43f8da1357f51f64cf60431acc74141f/deployments/helm/nvidia-device-plugin/templates/_helpers.tpl#L178.

Also a note here - I was going to warn the index out of bound concern, but looks like the `if` helper does check for empty list and treat it as false, so using "0" is safe here.

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
```diff
@@ -1,14 +1,39 @@
...
+{{~else~}}
+{{~#if (eq settings.kubelet-device-plugins.nvidia.device-list-strategy "cdi-cri") ~}}
+mode="cdi"
+{{~else~}}
+mode="legacy"
+{{~/if~}}
+{{/if}}
+{{else}}
+mode="legacy"
+{{/if}}
```
ytsssun, line 33: I don't love how the nested if else block here and multiple repeated `mode="legacy".

It looks like the only cases we want to set "cdi" is when the device-list-strategy is set as "cdi-cri" (in string or list form). Could we clean up the if-else logic here?

On another note, this along with the comment I have above, it might be worth consider introducing a dedicated "device-list-strategy" helper? The logic will be simplified in rust code.

Our custom helpers - https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/api/schnauzer/src/v2/import/helpers.rs#L36-L88

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
arnaldo2792: Regarding the conversation about custom helper, we had one, we were advice not to add it at least for this case:

https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/502

KCSesh (approved):

arnaldo2792 (approved):

PR #464, opened 2025-04-10
allow lookups of .local domains using unicast DNS

tzneal:
The .local suffix is reserved for Multicast DNS, but its sometimes used as a convenient suffix for internal domains. Those then fail to lookup correctly. Since Multicast DNS usage is tiny and not supported with Bottlerocket, disable this check to allow those names to resolve with unicast DNS.

**Issue number:**

Closes #

**Description of changes:**

Patches systemd to allow lookups of .local domains using unicast DNS

**Testing done:**

Before
```
root@my-shell:/# ping some.fake.local
ping: some.fake.local: Temporary failure in name resolution
```

After
```
root@my-shell:/# ping some.fake.local
PING some.fake.local (8.8.8.8) 56(84) bytes of data.
64 bytes from dns.google (8.8.8.8): icmp_seq=1 ttl=115 time=10.2 ms
64 bytes from dns.google (8.8.8.8): icmp_seq=2 ttl=115 time=8.63 ms
64 bytes from dns.google (8.8.8.8): icmp_seq=3 ttl=115 time=8.70 ms
64 bytes from dns.google (8.8.8.8): icmp_seq=4 ttl=115 time=8.65 ms
```

**Terms of contribution:**

### DIFF
--- a/packages/systemd/9015-allow-lookups-of-local-domains-using-unicast-DNS.patch
+++ b/packages/systemd/9015-allow-lookups-of-local-domains-using-unicast-DNS.patch
@@ -1,0 +1,34 @@
+From e72ca322112e62186ee4bc0b913ebeb81c97dd8a Mon Sep 17 00:00:00 2001
+From: Todd Neal <tnealt@amazon.com>
+Date: Thu, 10 Apr 2025 15:06:06 +0000
+Subject: [PATCH] allow lookups of .local domains using unicast DNS
+
+The .local suffix is reserved for Multicast DNS, but its sometimes used
+as a convenient suffix for internal domains. Those then fail to lookup
+correctly. Since Multicast DNS usage is tiny and not supported with
+Bottlerocket, disable this check to allow those names to resolve with
+unicast DNS.
+---
+ src/resolve/resolved-dns-scope.c | 6 +-----
+ 1 file changed, 1 insertion(+), 5 deletions(-)
+
+diff --git a/src/resolve/resolved-dns-scope.c b/src/resolve/resolved-dns-scope.c
+index 4085b01..862f5c3 100644
+--- a/src/resolve/resolved-dns-scope.c
++++ b/src/resolve/resolved-dns-scope.c
+@@ -705,11 +705,7 @@ DnsScopeMatch dns_scope_good_domain(
+                 }
+ 
+                 /* Exclude link-local IP ranges */
+-                if (match_link_local_reverse_lookups(domain) >= DNS_SCOPE_YES_BASE ||
+-                    /* If networks use .local in their private setups, they are supposed to also add .local
+-                     * to their search domains, which we already checked above. Otherwise, we consider .local
+-                     * specific to mDNS and won't send such queries ordinary DNS servers. */
+-                    dns_name_endswith(domain, "local") > 0)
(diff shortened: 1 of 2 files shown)


bcressey (commented):
Can you prefix the commit (not the patch) with `systemd: `? LGTM otherwise.

bcressey (approved):

ytsssun:
I did some testing on this change and was able to confirm the fix.

What I did:
1. **Prerequisite**: Created a Route53 Private host zone with a record "some.fake.local" pointing to IP address "8.8.8.8"
2. Launched nodes using both existing Bottlerocket variant in an EKS cluster and a custom AMI that includes this change, then compared behavior.
3. **Used `nslookup` from the admin container**
   
   For a Bottlerocket 1.35.0 official AMI (without this change), `nslookup` failed in the admin container:
   ```
   [root@admin]# nslookup some.fake.local
   Server:         127.0.0.53
   Address:        127.0.0.53#53

   ** server can't find some.fake.local: NXDOMAIN
   ```

   For the custom AMI with this change, `nslookup` succeeded:
   ```
   [root@admin]# nslookup some.fake.local
   Server:         127.0.0.53
   Address:        127.0.0.53#53

   Non-authoritative answer:
   Name:   some.fake.local
   Address: 8.8.8.8
   ```

4. **Ran a `dnsutils` pod on the nodes and tested `nslookup`**
   ```
   kubectl apply -f https://k8s.io/examples/admin/dns/dnsutils.yaml
   ```
   
   For a Bottlerocket 1.35.0 official AMI (without this change), `nslookup` worked as expected:
   ```
   kubectl exec -i -t dnsutils -- nslookup some.fake.local         
   Server:         10.100.0.10
   Address:        10.100.0.10#53

   Non-authoritative answer:
   Name:   some.fake.local
   Address: 8.8.8.8
   ```
   
   For the custom AMI with this change, `nslookup` also worked:
   ```
   kubectl exec -i -t dnsutils -- nslookup some.fake.local                 
   Server:         10.100.0.10
   Address:        10.100.0.10#53

   Non-authoritative answer:
   Name:   some.fake.local
   Address: 8.8.8.8
   ```

---
**Extra reading**

For some extra context, pods by default use the `ClusterFirst` DNS policy, which has a resolv.conf like this in the container (injected by kubelet):
```
search default.svc.cluster.local svc.cluster.local cluster.local us-west-2.compute.internal
nameserver 10.100.0.10
options ndots:5
``` 

It uses the CoreDNS service (on 10.100.0.10) for DNS resolution.

If we specify "Default" as the DNS policy for the pod, it would use what we provision in the kubelet configuration as resolv.conf, which in Bottlerocket is `/run/netdog/resolv.conf`:
```
search us-west-2.compute.internal
nameserver 192.168.0.2
```

For the admin containers, the /etc/resolv.conf looks like:
```
nameserver 127.0.0.53
options edns0 trust-ad
search us-west-2.compute.internal
```

For the 1.35.0 Bottlerocket AMI without this change, explicitly pointing to `192.168.0.2` when running nslookup on the admin container would work (same reason why the dns lookup on the pod would work for pods using `Default` dnsPolicy):
```
[root@admin]# nslookup some.fake.local 192.168.0.2
Server:         192.168.0.2
Address:        192.168.0.2#53

Non-authoritative answer:
Name:   some.fake.local
Address: 8.8.8.8
```

ytsssun (approved):

PR #465, opened 2025-04-10
release: restrict load-crash-kernel to x86_64

larvacea:
The kexec-tools package is unable to parse ARM EFI_ZBOOT images, so only enable the crash kernel on x86_64.

**Description of changes:**

Add a ConditionArchitecture=x86-64 to the load-crash-kernel unit file. The `kexec-tools` package is not able to decompress and parse a compressed kernel image on aarch64, so restrict crash kernel loading to x86_64 hosts which meet the other conditions specified in the unit file. At present, we do not release aarch64 AMIs with a crash kernel, so nothing changes in our AMI behavior.

**Testing done:**

Booted a sufficiently large aws-dev AMI (which does enable the crash kernel by default) on both x86_64 and aarch64. Verified that the crash kernel loads on x86_64, and the load-crash-kernel service is skipped on aarch64. Verified that the follow-on disable-kexec-load service runs on both architectures (and disables kexec loading as it should).

**Terms of contribution:**

### DIFF
--- a/packages/release/load-crash-kernel.service
+++ b/packages/release/load-crash-kernel.service
@@ -1,5 +1,8 @@
 [Unit]
 Description=Load crash kernel
+# No kexec support yet for arm64's zstd-compressed zboot image on 6.12+
+ConditionKernelVersion=|<6.12
+ConditionArchitecture=|!arm64
 ConditionKernelCommandLine=crashkernel
 ConditionPathExists=!/proc/vmcore
 RefuseManualStart=true


packages/release/load-crash-kernel.service
```diff
@@ -1,5 +1,6 @@
 [Unit]
 Description=Load crash kernel
+ConditionArchitecture=x86-64
```
bcressey: The commit message doesn't really make it clear that this is specific to the arm64 zboot image in 6.12.

In addition to fixing the commit message, you should be able to tighten up the conditions to reflect this, using triggering conditions to achieve a logical "or" effect:
```suggestion
# No kexec support yet for arm64's zstd-compressed zboot image on 6.12+
ConditionKernelVersion=|<6.12
ConditionArchitecture=|!arm64
```

larvacea:
The new commit message:
```
    release: disable load-crash-kernel on arm64 6.12
    
    Bottlerocket enables arm64 EFI_ZBOOT kernel compression in kernel 6.12.
    The kexec-tools package is unable to parse this image, so do not run the
    load-crash-kernel service on arm64 kernel 6.12 or later.
```
Tested and verified that 6.1 loads the crash kernel on either architecture, and 6.12 loads on x86_64 and does not load on aarch64 (neither triggering condition fires and systemd skips the service). Also re-verified that the disable-kexec-load service runs unconditionally and disables kexec-load everywhere, always.

bcressey (approved):

arnaldo2792 (approved):

PR #458, opened 2025-04-04
Packages: Drop shimpei oci add hook

sky1122:
**Issue number:**

Closes #453 

**Description of changes:**

Replace Shimpei and the OCI-add-hooks packages with NVIDIA Container Runtime to handle OCI hook injection.

**Testing done:**
based on the current branch, running MIG test on k8s-1.32-nvidia variant.
```
{
  "os": {
    "arch": "x86_64",
    "build_id": "e5cd08cf-dirty",
    "pretty_name": "Bottlerocket OS 1.36.0 (aws-k8s-1.32-nvidia)",
    "variant_id": "aws-k8s-1.32-nvidia",
    "version_id": "1.36.0"
  }
}
```

Below is some of the pods
```
╰$ kubectl get pods -A
NAMESPACE     NAME                             READY   STATUS              RESTARTS      AGE
default       nvidia-cuda-workload-45j4d       0/1     ContainerCreating   0             10s
default       nvidia-cuda-workload-4tsv8       0/1     ContainerCreating   0             10s
default       nvidia-cuda-workload-67q25       0/1     ContainerCreating   0             10s
```

```
╰$ kubectl logs nvidia-cuda-workload-45j4d
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done

╰$ kubectl logs nvidia-cuda-workload-4tsv8
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done

 kubectl logs nvidia-cuda-workload-6ggq4
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
```
Time Slicing and MIG cross setting looks fine
```
Allocatable:
  cpu:                    95690m
  ephemeral-storage:      15751651301
  hugepages-1Gi:          0
  hugepages-2Mi:          0
  memory:                 1167612760Ki
  nvidia.com/gpu:         0
  nvidia.com/gpu.shared:  224
```
Running the CUDA work load and also pass all the test.

- `aws-ecs-2-nvidia`
```
{
    "events": [
        {
            "timestamp": 1743733927639,
[...]

PR #471, opened 2025-04-15
Enable CDI support for ECS

### LINKED ISSUE #470: Support CDI for ECS
Docker 25 supports CDI, so it is possible to migrate off the NVIDIA Container Toolkit prestart hook and use CDI instead.

arnaldo2792:
**Issue number:**

Closes #470

**Description of changes:**

The first commit in the series improves the dependency management for the configuration files provided for each orchestrator. This allows to depend only on `nvidia-container-toolkit` on the variant definition, and get the required configuration files depending on the variant built.

The remaining commits enable CDI support for ECS variants. A new package `nvidia-container-toolkit-cdi-specs` provides an additional service to generate CDI specifications. This new service can be used by downstream builders that support CDI. This new package is automatically installed for ECS variants.

The `nvidia-container-toolkit` configuration for ECS now forces the `cdi` mode. The reason for this is twofold:

- This truly enables CDI, and drops the dependency on the `prestart` hook provided by `nvidia-container-toolkit`
- This mode allows to use the `NVIDIA_VISIBLE_DEVICES` environment variable, which is what the ECS agent uses to "pass down" the devices that must be configured in the container

**Testing done:**

Ran an ECS task using `aws-ecs-2-nvidia`, and confirmed:
- Task runs successfully
- Device is accessible through the containers
- `NVIDIA_VISIBLE_DEVICES=all` doesn't have any effect in ECS as always, since the ECS agent clobbers the environment variable
- Binaries and libraries are mounted as expected

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-container-toolkit/generate-cdi-specs.service
+++ b/packages/nvidia-container-toolkit/generate-cdi-specs.service
@@ -1,0 +1,34 @@
+[Unit]
+Description=Generate CDI specifications
+# This has to be executed after the kernel modules are loaded
+# otherwise the userspace component of the driver will fail to
+# query the /dev devices
+After=load-tesla-kernel-modules.service load-open-gpu-kernel-modules.service
+# Running this unit after nvidia persistenced ensures that
+# the /dev devices are created and the hardware set to
+# persistence mode.
+Requires=nvidia-persistenced.service
+After=nvidia-persistenced.service
+# Block manual interactions with this service. It doesn't
+# make sense to regenerate CDI specifications as features
+# that might change the GPU (e.g. MIG) are not supported in
+# ECS
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+# Explanation of the options:
+# --format json: to be consistent across Bottlerocket's variants
+# --mode nvml: the default mode ("auto") resolves to this already, make it explicit
+# --device-name-strategy uuid: the ECS agent only supports device UUIDs
+# --output /etc/cdi/nvidia.json: store the CDI specifications at this location
+ExecStart=/usr/bin/nvidia-ctk cdi generate --format json \
+  --mode nvml \
+  --device-name-strategy uuid \
+  --output /etc/cdi/nvidia.json
+RemainAfterExit=true
+StandardError=journal+console
+
+[Install]
+RequiredBy=preconfigured.target
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-ecs.toml
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-ecs.toml
@@ -3,3 +3,6 @@ root = "/"
 path = "/usr/bin/nvidia-container-cli"
 environment = []
 ldconfig = "@/sbin/ldconfig"
+
+[nvidia-container-runtime]
+mode = "cdi"
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
@@ -22,6 +22,7 @@ Source3: nvidia-gpu-devices.rules
 Source4: nvidia-container-toolkit-tmpfiles-ecs.conf
 Source5: nvidia-container-toolkit-tmpfiles-k8s.conf
 Source6: nvidia-container-toolkit-config-k8s
+Source7: generate-cdi-specs.service
 
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}libnvidia-container
@@ -34,6 +35,7 @@ Requires: (%{name}-k8s if %{_cross_os}variant-family(aws-k8s))
 %package ecs
 Summary: Files specific for the ECS variants
 Requires: %{name}
+Requires: %{name}-cdi-specs
 Conflicts: %{name}-k8s
 
 %description ecs
@@ -45,6 +47,13 @@ Requires: %{name}
 Conflicts: %{name}-ecs
 
 %description k8s
+%{summary}.
+
+%package cdi-specs
+Summary: Tools to generate CDI specifications
+Requires: %{name}
+
+%description cdi-specs
 %{summary}.
 
 %prep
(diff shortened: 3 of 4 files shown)


packages/nvidia-container-toolkit/generate-cdi-specs.service
```diff
@@ -0,0 +1,34 @@
+[Unit]
+Description=Generate CDI specifications
+# This has to be executed after the kernel modules are loaded
+# otherwise the userspace component of the driver will fail to
+# query the /dev devices
+After=load-tesla-kernel-modules.service load-open-gpu-kernel-modules.service
```
koooosh, line 6: Why do we not include a `Requires=` for these units?

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
```diff
@@ -22,6 +22,7 @@ Source3: nvidia-gpu-devices.rules
 Source4: nvidia-container-toolkit-tmpfiles-ecs.conf
 Source5: nvidia-container-toolkit-tmpfiles-k8s.conf
 Source6: nvidia-container-toolkit-config-k8s
+Source7: generate-cdi-specs.service
```
koooosh, line 25: nit: Should we include a comment above this new source on what it's for? Perhaps we can include that currently it's just added to ECS for clarity

koooosh (approved):
Besides my question and nit, LGTM -- nice!

sky1122 (approved):

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
arnaldo2792: Its not just for ECS, its for any downstreams that want to generate CDI specifications. I think the description of the package covers what you are suggesting, along with the name of the service. The description of the package where this is included states that should be used to generate CDI specifications.

packages/nvidia-container-toolkit/generate-cdi-specs.service
arnaldo2792: This is a great catch, I can fix it. Also, we should fix the other services that have the same problem (I copied this from `ecs-gpu-init.service` but, `nvidia-persistenced` has the same problem).

arnaldo2792: I'll actually pause on this change. `Requires=load-tesla-kernel-modules.service load-open-gpu-kernel-modules.service` will introduce a bit of a weird state. Not both services will be started all the time, so by making them required I worry that systemd will issue Jobs to start them on every target transition, since technically only one of them will be started. If loading the drivers failed, this unit will fail and the boot will halt so at least that will be consistent. I can create an issue to track this problem.

ytsssun: Not blocking: Can we introduce a custom target here like `gpu-kernel-load.target` and enforce the ordering using the target instead of these services given that they will never be loaded at the same time?

ytsssun (approved):

KCSesh (approved):
LGTM - nice!

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-ecs.toml
```diff
@@ -3,3 +3,6 @@ root = "/"
 path = "/usr/bin/nvidia-container-cli"
 environment = []
 ldconfig = "@/sbin/ldconfig"
+
+[nvidia-container-runtime]
+mode = "cdi"
```
ytsssun, line 8: Question - per the description 
> This truly enables CDI, and drops the dependency on the prestart hook provided by nvidia-container-toolkit

I am trying to connect this with https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/458/files#diff-bf719304fec29a1f2736a935447edf3308cd718549ba13130f2166377c1c6e1b which already dropped the pre-start hook injection from nvidia-container-toolkit. Do you mean nvidia-container-runtime here?

packages/nvidia-container-toolkit/generate-cdi-specs.service
arnaldo2792: That could be something to consider, but you would like something more generic than `gpu-kernel-load.target`. Something like `prepare-devices.target` would be more generic. And then system services can be `RequiredBy=prepare-devices.target` and declare dependencies as `Requires=prepare-devices.target`.

We should be careful about adding more targets, as I mentioned if we add the target and a service declares itself as `RequiredBy=`, any transition may result on Jobs to start services that failed their conditions to `exec`.

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-ecs.toml
arnaldo2792: The change you linked didn't drop the dependency on the prestart hook, it only dropped the dependency on shimpei and oci-add-hooks to inject the prestart hook. The change that you linked still depended on the prestart hook. With this change, at least for ECS, the prestart hook is no longer needed.

NVIDIA Container Toolkit has two modes (actually three, if you count the `auto` mode). `legacy`, which is what almost all the time is used, requires the prestart hook. `cdi` doesn't, it still requires a hook (`cdi-hook`), but not the prestart hook.

In the description I say "truly" because the `auto` mode always resolves to `legacy`. This is because NVIDIA Container Runtime always fails [this condition] because when the device plugin uses `cdi-cri` as the `device-list-strategy` (which, BTW, is required to generate the CDI specs), the devices are passed as `CDIDevices`, not as `Mounts` or through `NVIDIA_VISIBLE_DEVICES`.

Since `legacy` is used, NVIDIA Container Runtime [injects] the prestart hook, even when the container is already configured with `CDIDevices`, which voids the purpose of using CDI (don't use the prestart hook). The only way to truly use CDI, is to force the `cdi` mode in the NVIDIA Container Runtime. This will skip the injection of the prestart hook.

Actually, we wouldn't even need to use `nvidia-container-runtime` if the ECS agent implemented the new CDI Devices support in Docker. But, the ECS agent uses `NVIDIA_VISIBLE_DEVICES` to configure the GPUs and `nvidia-container-runtime` is the only thing in the CDI world that is [aware] of that environment variable.

[this condition]: https://github.com/NVIDIA/nvidia-container-toolkit/blob/6394e9e9e7f8692438f2c921c0bbc95b72d693c6/internal/config/image/cuda_image.go#L201

[injects]: https://github.com/NVIDIA/nvidia-container-toolkit/blob/6394e9e9e7f8692438f2c921c0bbc95b72d693c6/internal/modifier/stable.go#L48

[aware]: https://github.com/NVIDIA/nvidia-container-toolkit/blob/6394e9e9e7f8692438f2c921c0bbc95b72d693c6/internal/modifier/cdi.go#L68

PR #483, opened 2025-04-24
ghostdog: add grid driver to options

yeazelm:
**Issue number:** Related to https://github.com/bottlerocket-os/bottlerocket/issues/4441

**Description of changes:**
Bottlerocket needs to use GRID drivers for some EC2 instances. This adds the logic to provide GRID as an option based upon PCI Subdevice ID. These IDs are known ahead of time so we create a list that can be referenced upon boot to add one more additional option (`grid`) when these devices are explicitly found.  

**Testing done:**
Validated with draft changes that include GRID drivers that when these subdevice IDs match, they choose GRID:

<details>

```
         Starting User Login Management...
[  OK  ] Finished Disable kexec load syscalls.
[  OK  ] Started User Login Management.
[   14.949740] ghostdog[1943]: open-gpu is not preferred driver: grid
[   15.060421] ghostdog[1945]: tesla is not preferred driver: grid
         Starting Load open GPU kernel modules...
         Starting Load Tesla kernel modules...
[  OK  ] Finished Generate network configuration.
[  OK  ] Reached target Preparation for Network.
[   15.067668] ghostdog[1982]: open-gpu is not preferred driver: grid
         Starting Network Configuration...
[   15.070212] ghostdog[1983]: tesla is not preferred driver: grid
[  OK  ] Started Network Configuration.
[  OK  ] Reached target Network.
         Starting Wait for Network to be Configured...
[   15.617122] migrator[1947]: Data store does not exist at given path, exiting (/var/lib/bottlerocket/datastore/current)
[  OK  ] Finished Bottlerocket data store migrator.
         Starting Call signpost to mark the …r all required targets are met....
         Starting Datastore creator...
[  OK  ] Finished Call signpost to mark the …ter all required targets are met..
[  OK  ] Finished Datastore creator.
         Starting Bottlerocket API server...
[  OK  ] Finished Wait for Network to be Configured.
         Starting Write network status...
[  OK  ] Finished Write network status.
[  OK  ] Reached target Network is Online.
[  OK  ] Started Bottlerocket API server.
         Starting Bottlerocket userdata configuration system...
[  OK  ] Finished Bottlerocket userdata configuration system.
         Starting User-specified setting generators...
[  OK  ] Finished Copy GRID kernel modules.
         Starting Load GRID kernel modules...
```

</details>

When on instances that have similar PCI Device IDs but not subdevice IDs, the `open-gpu` driver is still recommended:

<details>

```
[   22.529857] ghostdog[2058]: grid is not preferred driver: open-gpu
         Starting Link Tesla kernel modules...
[  OK  ] Started NVIDIA fabric manager service.
         Starting Reboot system if needed fo…guration changes to take effect...
         Starting Send boot success...
[   22.540520] ghostdog[2067]: tesla is not preferred driver: open-gpu
         Starting Wait for the instance to be in service...
[  OK  ] Finished Wait for the instance to be in service.
         Starting Load GRID kernel modules...
         Starting Load Tesla kernel modules...
[  OK  ] Finished Isolates multi-user.target.
[   22.559612] ghostdog[2081]: grid is not preferred driver: open-gpu
[   22.566336] ghostdog[2083]: tesla is not preferred driver: open-gpu
```

</details>

**Terms of contribution:**

### DIFF
--- a/sources/ghostdog/src/main.rs
+++ b/sources/ghostdog/src/main.rs
@@ -21,7 +21,22 @@ use std::process::Command;
 const NVME_CLI_PATH: &str = "/sbin/nvme";
 const NVME_IDENTIFY_DATA_SIZE: usize = 4096;
 const NVIDIA_VENDOR_ID: &str = "10de";
+const NVIDIA_GRID_DEVICE_ID: &str = "27b8";
 const OPEN_GPU_SUPPORTED_DEVICES_PATH: &str = "/usr/share/nvidia/open-gpu-supported-devices.json";
+
+// Generate a list of Subdevice IDs that match the format of the file at OPEN_GPU_SUPPORTED_DEVICES_PATH
+// but are instead sourced here. The format in the JSON file has each ID starting with `0x` and are all upper
+// case where `pciclient` will provide just the 4 character ID with no prefix. `pciclient` output has to be
+// prepended and moved to uppercase to match these just as if they were sourced from the JSON file.
+lazy_static! {
+    static ref NVIDIA_GRID_SUBDEVICES: HashSet<&'static str> = {
+        let mut m = HashSet::new();
+        m.insert("0x1733");
+        m.insert("0x1735");
+        m.insert("0x1737");
+        m
+    };
+}
 
 #[derive(FromArgs, PartialEq, Debug)]
 /// Manage ephemeral disks.
@@ -217,7 +232,7 @@ fn read_supported_devices_file(path: PathBuf) -> Result<SupportedDevicesConfigur
     Ok(device_configuration)
 }
 
-/// Search the Open GPU Supported Devices File to determine if the Open GPU Driver should be used based upon PCI devices present
+/// Search the Open GPU Supported Devices File to determine which driver should be used based upon PCI devices present
 fn find_preferred_driver() -> Result<String> {
     let open_gpu_devices = read_supported_devices_file(OPEN_GPU_SUPPORTED_DEVICES_PATH.into())?;
     let list_input = pciclient::ListDevicesParam::builder()
@@ -239,6 +254,27 @@ fn find_preferred_driver() -> Result<String> {
             .map(|x| &x.device_id)
             .collect::<HashSet<_>>(),
     };
+
+    // If the PCI device ID is one that could potentially use GRID, collect the Subdevice IDs
+    let mut subdevice_ids = present_devices
+        .iter()
+        .filter(|x| x.device().starts_with(NVIDIA_GRID_DEVICE_ID))
+        .map(|x| {
+            format!(
+                "0x{}",
+                x.subsystem_device()
+                    .as_ref()
+                    .unwrap_or(&"".to_string())
+                    .to_uppercase()
+            )
+            .clone()
+        })
+        .collect::<HashSet<_>>()
+        .into_iter();
+    // Return early with grid if a match is made for these subdevices
+    if subdevice_ids.any(|subdevice| NVIDIA_GRID_SUBDEVICES.contains(subdevice.as_str())) {
+        return Ok("grid".to_string());
+    }
 
     if unique_ids.any(|input_device| open_gpu_device_set.contains(&input_device)) {
         Ok("open-gpu".to_string())


sources/ghostdog/src/main.rs
```diff
@@ -217,7 +228,7 @@ fn read_supported_devices_file(path: PathBuf) -> Result<SupportedDevicesConfigur
     Ok(device_configuration)
 }
 
-/// Search the Open GPU Supported Devices File to determine if the Open GPU Driver should be used based upon PCI devices present
+/// Search the Open GPU Supported Devices File to determine if which driver should be used based upon PCI devices present
```
bcressey:
```suggestion
/// Search the Open GPU Supported Devices File to determine which driver should be used based upon PCI devices present
```

sources/ghostdog/src/main.rs
```diff
@@ -21,8 +21,19 @@ use std::process::Command;
 const NVME_CLI_PATH: &str = "/sbin/nvme";
 const NVME_IDENTIFY_DATA_SIZE: usize = 4096;
 const NVIDIA_VENDOR_ID: &str = "10de";
+const NVIDIA_GRID_DEVICE_ID: &str = "27B8";
```
bcressey: nit: kind of want this to be lowercase to match `NVIDIA_VENDOR_ID`

sources/ghostdog/src/main.rs
```diff
@@ -21,8 +21,19 @@ use std::process::Command;
...
 const OPEN_GPU_SUPPORTED_DEVICES_PATH: &str = "/usr/share/nvidia/open-gpu-supported-devices.json";
 
+lazy_static! {
+    static ref NVIDIA_GRID_SUBDEVICES: HashSet<&'static str> = {
+        let mut m = HashSet::new();
+        m.insert("0x1733");
+        m.insert("0x1735");
+        m.insert("0x1737");
```
bcressey, line 36: nit: it doesn't matter now, but since everything else in the file uses lowercase hex, it's probably better to treat this as a set of lowercase identifiers rather than uppercase

bcressey (approved):

sources/ghostdog/src/main.rs
yeazelm: I've made this a bit of a mess and not documented my thoughts. These values currently don't show up in the `supported-gpus.json` file but I wanted them to feel like they do when working with them later on. The `supported-gpus.json` file provide IDs like this:
```
{
    "devid": "0x27B8",
    "subdevid": "0x16CA",
    "subvendorid": "0x10DE",
    "name": "NVIDIA L4",
    "features": [
        "dpycbcr420",
        "dpgsynccompatible",
        "hdmi4k60rgb444",
        "hdmigsynccompatible",
        "runtimepm",
        "vdpaufeaturesetK",
        "kernelopen",
        "gsp_proprietary_supported"
    ]
},
``` 
Where each ID is prefixed with `0x` and is uppercase. The intention here is that sources of known IDs (`supported-gpus.json` or this additional list of identifiers) are expected to all have `0x` at the beginning and be all uppercase, but the `pciclient` or other dynamic sources may or may not provide values in uppercase or with `0x` at the beginning, so the proper experience is to normalize the dynamic input into this format for matching. This allows the implementation to only do `format!()` on one set (the present PCI IDs) rather than both sets which is expected to be much larger. So in this case, if we statically create these as lowercase, we need to either do the transformation at runtime or be aware these are "special" and don't follow the convention like Subdevice IDs that *are* sourced in `supported-gpus.json`. 

I do want to point out this is all with an eye towards future updates because, as you point out, we could just use the 4 character lower case to match this list as it stands today but that logic would be very different from the matching we do from the file so I wanted to keep some consistency.

KCSesh: I see that makes sense. 

It definitely feels like this could be information we fetch from `supported-gpus.json` like we do here: https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/develop/packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec#L165

Maybe we can re-evaluate on the next addition.

KCSesh (approved):

yeazelm:
^ Updated to lowercase for constant and updated comments for the static set.

bcressey (approved):

PR #484, opened 2025-04-25
Drop kubernetes-1.25

piyush-jena:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket/issues/4496

**Description of changes:**

Amazon EKS will end extended support of [Kubernetes 1.25 on May 1, 2025](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-release-calendar). Support for the `aws-k8s-1.25*` variants of Bottlerocket will end on the same date. As a result, the core kit will no longer maintain the kubernetes-1.25
package.

**Testing done:**

`make ARCH=aarch64 && make ARCH=x86_64`

**Terms of contribution:**


koooosh (commented):
nit: I believe we should remove the `ecr-credential-provider-1.25` package in separate commits from `kubernetes-1.25`

KCSesh (approved):

yeazelm:
I think the split in commits was a bit too much? I don't think we want to separate out the Cargo.toml change from rest of the removal.

piyush-jena:
Refreshed the PR to account for the last comment.

koooosh (approved):

arnaldo2792 (approved):

yeazelm (approved):

PR #487, opened 2025-04-30
Support Kubelet Config containerLogMaxWorkers and containerLogMonitorInterval

ytsssun:
**Issue number:**

Relevant Issue:
bottlerocket-os/bottlerocket#4470

**Description of changes:**
This change is paired with https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/80

Support kubernetes settings:
- `kubernetes.container-log-max-workers` 
- `kubernetes.container-log-monitor-interval` 

Which will be used to render the KubeletConfiguration:
- `containerLogMaxWorkers`
- `containerLogMonitorInterval`

Note that these [two fields are introduced](https://github.com/kubernetes/kubernetes/pull/114301) in k8s 1.30. So we will only render the kubelet config for **1.30+** k8s variants. For k8s variants older than 1.30, the settings will be silently skipped and not rendered to the kubelet config.

**Testing done:**
**Test 1.** Built custom k8s 1.30 variant and set the userdata in my nodegroup spec and deployed via `eksctl`
```
nodeGroups:
  - name: my-ng-bottlerocketv1
    instanceType: m5.large
    ...
    amiFamily: Bottlerocket
    ...
    bottlerocket:
      settings:
        kubernetes:
          container-log-max-workers: 3
          container-log-monitor-interval: 20s
```
The settings were persisted
```
bash-5.1# apiclient get settings.kubernetes
{
  "settings": {
    "kubernetes": {
      ...
      "container-log-max-workers": 3,
      "container-log-monitor-interval": "20s",
      ...
    }
  }
}

```
Kubelet config is properly rendered
```
bash-5.1# cat /etc/kubernetes/kubelet/config | grep containerLog 
containerLogMaxWorkers: 3
containerLogMonitorInterval: 20s
```
Node joined the cluster
```
» k get nodes                                           
NAME                                            STATUS   ROLES    AGE     VERSION
ip-192-168-128-126.us-west-2.compute.internal   Ready    <none>   6m11s   v1.30.10-eks-1a9dacd
```

**Test 2:** Built custom k8s 1.29 ami and set same userdata as Test 1 and verify that the settings will be skipped and not rendered to the kubelet config.
Settings were persisted:
```
bash-5.1# apiclient get settings.kubernetes | grep container
      "container-log-max-workers": 3,
      "container-log-monitor-interval": "20s",
``` 
But we will not apply it to the kubelet config given that the fields are only supported in 1.30+ k8s versions. Below command shows that the relevant fields are not rendered.
```
bash-5.1# cat /etc/kubernetes/kubelet/config | grep container
bash-5.1#
```
Node joined the cluster.

**Test 3:** verify that the change is compatible with older settings sdk
Built a custom k8s 1.30 ami that consumes this core-kit change and use the old settings sdk. Launch the instance without the new settings.

- The node joined the cluster.
- The kubelet config was properly rendered without the new fields.

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.30/kubelet-config
+++ b/packages/kubernetes-1.30/kubelet-config
@@ -161,6 +161,12 @@ containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
 {{#if settings.kubernetes.container-log-max-files includeZero=true}}
 containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
 {{/if}}
+{{#if settings.kubernetes.container-log-max-workers includeZero=true}}
+containerLogMaxWorkers: {{settings.kubernetes.container-log-max-workers}}
+{{/if}}
+{{#if settings.kubernetes.container-log-monitor-interval}}
+containerLogMonitorInterval: {{settings.kubernetes.container-log-monitor-interval}}
+{{/if}}
 {{#if settings.kubernetes.shutdown-grace-period}}
 shutdownGracePeriod: {{settings.kubernetes.shutdown-grace-period}}
 {{/if}}
--- a/packages/kubernetes-1.33/kubelet-config
+++ b/packages/kubernetes-1.33/kubelet-config
@@ -161,6 +161,12 @@ containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
 {{#if settings.kubernetes.container-log-max-files includeZero=true}}
 containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
 {{/if}}
+{{#if settings.kubernetes.container-log-max-workers includeZero=true}}
+containerLogMaxWorkers: {{settings.kubernetes.container-log-max-workers}}
+{{/if}}
+{{#if settings.kubernetes.container-log-monitor-interval}}
+containerLogMonitorInterval: {{settings.kubernetes.container-log-monitor-interval}}
+{{/if}}
 {{#if settings.kubernetes.shutdown-grace-period}}
 shutdownGracePeriod: {{settings.kubernetes.shutdown-grace-period}}
 {{/if}}
--- a/packages/kubernetes-1.31/kubelet-config
+++ b/packages/kubernetes-1.31/kubelet-config
@@ -161,6 +161,12 @@ containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
 {{#if settings.kubernetes.container-log-max-files includeZero=true}}
 containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
 {{/if}}
+{{#if settings.kubernetes.container-log-max-workers includeZero=true}}
+containerLogMaxWorkers: {{settings.kubernetes.container-log-max-workers}}
+{{/if}}
+{{#if settings.kubernetes.container-log-monitor-interval}}
+containerLogMonitorInterval: {{settings.kubernetes.container-log-monitor-interval}}
+{{/if}}
 {{#if settings.kubernetes.shutdown-grace-period}}
 shutdownGracePeriod: {{settings.kubernetes.shutdown-grace-period}}
 {{/if}}
--- a/packages/kubernetes-1.32/kubelet-config
+++ b/packages/kubernetes-1.32/kubelet-config
@@ -161,6 +161,12 @@ containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
 {{#if settings.kubernetes.container-log-max-files includeZero=true}}
 containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
 {{/if}}
+{{#if settings.kubernetes.container-log-max-workers includeZero=true}}
+containerLogMaxWorkers: {{settings.kubernetes.container-log-max-workers}}
+{{/if}}
+{{#if settings.kubernetes.container-log-monitor-interval}}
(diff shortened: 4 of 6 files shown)


packages/kubernetes-1.30/kubelet-config
```diff
@@ -161,6 +161,12 @@ containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
 {{#if settings.kubernetes.container-log-max-files includeZero=true}}
 containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
 {{/if}}
+{{#if settings.kubernetes.container-log-max-workers includeZero=true}}
+containerLogMaxWorkers: {{settings.kubernetes.container-log-max-workers}}
+{{/if}}
+{{#if settings.kubernetes.container-log-monitor-internal}}
```
koooosh: Same comment as the other PR, applies to each config + description:

Should be: `container_log_monitor_interval`, right?

koooosh (changes requested):

packages/kubernetes-1.30/kubelet-config
bcressey: > Should be: container_log_monitor_interval, right?

No, definitely not!

packages/kubernetes-1.30/kubelet-config
```diff
@@ -161,6 +161,12 @@ containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
 {{#if settings.kubernetes.container-log-max-files includeZero=true}}
 containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
 {{/if}}
+{{#if settings.kubernetes.container-log-max-workers includeZero=true}}
+containerLogMaxWorkers: {{settings.kubernetes.container-log-max-workers}}
+{{/if}}
```
bcressey, line 166: Are you certain you don't need guard these via `if_not_null` blocks? Have you tested this on a variant build that isn't using your new settings SDK with the newer `KubernetesSettings` shape?

packages/kubernetes-1.33/kubelet-config
```diff
@@ -161,6 +161,12 @@ containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
 {{#if settings.kubernetes.container-log-max-files includeZero=true}}
 containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
 {{/if}}
+{{#if settings.kubernetes.container-log-max-workers includeZero=true}}
+containerLogMaxWorkers: {{settings.kubernetes.container-log-max-workers}}
+{{/if}}
+{{#if settings.kubernetes.container-log-monitor-interval}}
+containerLogMonitorInterval: {{settings.kubernetes.container-log-monitor-interval}}
+{{/if}}
```
cbgbt, line 169: I think we have to be more defensive with our conditionals on new settings values.

I'm *pretty sure* that if someone takes this `core-kit` update *without* the settings SDK change in their variant definition, then this template will fail to render because `settings.kubernetes.container-log-max-workers` won't be null -- it just won't be present at all, which will result in a template render failure.

packages/kubernetes-1.30/kubelet-config
ytsssun: @bcressey I think @koooosh meant to say the typo `container-log-monitor-internal` -> `container-log-monitor-interval`. I just pushed a change to fix the typo.

I have been using kebab case in this core-kit change.

koooosh: Ah yeah, I meant to correct `internal` to `interval` and copied my comment from the other PR. Sorry for the confusion

packages/kubernetes-1.30/kubelet-config
ytsssun: TIA, I haven't test such case. I will make this change you suggested and test that use case!

packages/kubernetes-1.33/kubelet-config
ytsssun: Thanks for the detailed insights. I will account for this change and test this particular case also suggested by @bcressey

packages/kubernetes-1.30/kubelet-config
ytsssun: I have tested the case, and "if" indeed works! See the Test 3 in the description for detail.

packages/kubernetes-1.33/kubelet-config
ytsssun: I have tested the case, and "if" indeed works! See the Test 3 in the description for detail.

koooosh (approved):
LGTM, can we drop the Skip CI commit?

ytsssun:
> LGTM, can we drop the Skip CI commit?

Yes, I will drop that when I make the last change to point to the released settings sdk upstream.

koooosh (approved):

arnaldo2792:
LGTM, provided CI passes :smile:

arnaldo2792 (approved):

packages/kubernetes-1.33/kubelet-config
cbgbt: Thanks! After discussing this offline with @ytsssun, we realized that `if` itself is not problematic, and when we had been bitten in the past it was because of boolean operators like `eq` and `ne` -- those are `strict` in the sense that they error on undefined values rather than treating them as `null`.

cbgbt (approved):

KCSesh (approved):

PR #489, opened 2025-05-02
Support for singleProcessOOMKill in Kubernetes 1.32 and 1.33

arnaldo2792:
**Description of changes:**

This adds support for the singleProcessOOMKill [[1](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.32.md)] flag used by the kubelet starting with version 1.32.

**Testing done:**

In combination with https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/81 and a local change to the Bottlerocket repo I confirmed that I can set the setting:

```bash
[root@admin]# apiclient get settings.kubernetes.single
{
  "settings": {
    "kubernetes": {
      "single-process-oom-kill": true
    }
  }
}
```

Confirmed that the setting was rendered:

```
staticPodPath: "/etc/kubernetes/static-pods/"
failSwapOn: false
singleProcessOOMKill: true
```

Confirmed that the cgroup for the pod was configured:

```
# With singleProcessOOMKill: false
bash-5.1# find /sys -name "*memory.oom*" -exec cat {} \; | grep 1
1

# With singleProcessOOMKill: true
bash-5.1# find /sys -name "*memory.oom*" -exec cat {} \; | grep 1
# Nothing is returned
```

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.32/kubelet-config
+++ b/packages/kubernetes-1.32/kubelet-config
@@ -200,3 +200,6 @@ reservedMemory:
 reservedSystemCPUs: {{settings.kubernetes.reserved-cpus}}
 {{/if}}
 failSwapOn: false
+{{#if settings.kubernetes.single-process-oom-kill}}
+singleProcessOOMKill: {{settings.kubernetes.single-process-oom-kill}}
+{{/if}}
--- a/packages/kubernetes-1.33/kubelet-config
+++ b/packages/kubernetes-1.33/kubelet-config
@@ -200,3 +200,6 @@ reservedMemory:
 reservedSystemCPUs: {{settings.kubernetes.reserved-cpus}}
 {{/if}}
 failSwapOn: false
+{{#if settings.kubernetes.single-process-oom-kill}}
+singleProcessOOMKill: {{settings.kubernetes.single-process-oom-kill}}
+{{/if}}


ytsssun:
Have you also verified that memory.oom.group is set to 0 for the cgroups that are associated with the k8s pods?

ytsssun:
A stretch testing:

It would also be helpful to verify the oom event stats, I would expect the `oom_kill` stats to be none zero and `oom_kill_group` to remain zero.

E.g.
```
bash-5.1# cat /sys/fs/cgroup/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod1e3fd44f_647f_467b_a2b5_5e9dfcf407b3.slice/cri-containerd-bdaac2585aaf776b81f0d9e17733fd41dc858b3bc90a3310184339e2d21c6aaf.scope/memory.events
low 0
high 0
max 0
oom 0
oom_kill xxx   <--- this to be none zero
oom_group_kill 0
```

You could use this setup to verify the oom killing events - https://github.com/bottlerocket-os/bottlerocket/issues/4075#issuecomment-2516196486.


I am fine with at least verifying that this setting flip the `memory.oom.group` value back to `0` for the k8s pods.

arnaldo2792:
> Have you also verified that memory.oom.group is set to 0 for the cgroups that are associated with the k8s pods?

Yes, this `find` command:

```bash
find /sys -name "*memory.oom*" -exec cat {} \; | grep 1
```

Was ran in the host. When `single-process-oom-kill` is `false`, the pod's cgroup config was `0` and the above command doesn't print anything as mentioned in the description.

```bash
# Without the setting, this returns nothing
```

I updated the description to make it more clear

arnaldo2792:
> It would also be helpful to verify the oom event stats,

This was a little tricky to test since the `cgroup` for the pod was deleted as soon as it was OOM killed, but I got a setup working. In a pod that holds 250Mi, and with a limit of `200Mi` I can see it been `OOMKilled`:

```
# Ignore the name of the pod
nginx-q9pmz   0/1     OOMKilled   1 (14s ago)   17s 
``` 

I confirmed that the parent cgroup increased `oom_kill` but not `oom_group_kill`

```
# The parent cgroup is:
# /.bottlerocket/rootfs/sys/fs/cgroup/kubepods.slice/kubepods-burstable.slice/memory.events
# This increases as the pods are killed. My pods so far died 75 times
oom_kill 75
# This only increased with singleProcessOOMKill: false
oom_group_kill 20
```

ytsssun (approved):

ytsssun (commented):
LGTM, thanks for the testing!

KCSesh:
Model version bump happened here:
https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/487

LGTM

arnaldo2792:
(forced push removes `[skip ci]`)

KCSesh (approved):

koooosh (approved):

PR #499, opened 2025-05-12
Add new command to ghostdog for Infiniband detection and configuration

yeazelm:
**Description of changes:**
This change is related to and required for https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/142

This adds an additional sub command to `ghostdog`: `write-infiniband-sw-mng-guid` which matches similar other detection subcommands. Its first role is to detect the correct hardware needed for NVLSM. If this hardware is not detected it exits early. If the hardware is detected, it finds the appropriate Port GUID needed for communication between Fabric Manager and NVLSM. Once it finds this, it writes the GUID to the specified (via an argument to the command) file in the format of `GUID_ARGS="-g 0xffffff.."` which can then be read in by systemd as an environment variables for starting NVLSM and Fabric Manager. 

**Testing done:**
See the testing in https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/142. I built an `aws-k8s-1.32-nvidia` variant with both this and the kernel kit changes and confirmed that Fabric Manager works both with and without this Infiniband hardware.

On `p4d.24xlarge`, the hardware doesn't exist and so nvlsm skips due to the non-zero exit:

<details>

```
bash-5.1# systemctl status nvidia-fabricmanager
● nvidia-fabricmanager.service - NVIDIA fabric manager service
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/nvidia-fabricmanager.service; enabled; preset: enabled)
             └─00-aws-config.conf
     Active: active (running) since Fri 2025-05-16 18:56:29 UTC; 2min 18s ago
   Main PID: 15988 (nv-fabricmanage)
      Tasks: 19 (limit: 629145)
     Memory: 26.0M
        CPU: 18.245s
     CGroup: /system.slice/nvidia-fabricmanager.service
             └─15988 /usr/bin/nv-fabricmanager -c /etc/nvidia/fabricmanager.cfg

May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] GPU with nodeId = 0, gpuId =…nfigured
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] GPU with nodeId = 0, gpuId =…nfigured
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] GPU with nodeId = 0, gpuId =…nfigured
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] GPU with nodeId = 0, gpuId =…nfigured
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] GPU with nodeId = 0, gpuId =…nfigured
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] GPU with nodeId = 0, gpuId =…nfigured
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] GPU with nodeId = 0, gpuId =…nfigured
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] Multi node HA error handling disabled
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] Successfully configured all …traffic.
May 16 18:56:40 ip-192-168-32-119.us-west-2.compute.internal nv-fabricmanager[15988]: [May 16 2025 18:56:40] [INFO] [tid 15988] FM starting NvLink Inband messages
Hint: Some lines were ellipsized, use -l to show in full.
bash-5.1# systemctl status nvlsm
○ nvlsm.service - Start NVLink Subnet Manager
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/nvlsm.service; enabled; preset: enabled)
             └─00-aws-config.conf
     Active: inactive (dead) (Result: exec-condition) since Fri 2025-05-16 18:56:29 UTC; 2min 30s ago
  Condition: start condition failed at Fri 2025-05-16 18:56:29 UTC; 2min 30s ago
    Process: 15966 ExecCondition=[ -s /etc/nvidia/nvlsm.env ] (code=exited, status=1/FAILURE)
        CPU: 1ms

May 16 18:56:29 ip-192-168-32-119.us-west-2.compute.internal systemd[1]: Starting Start NVLink Subnet Manager...
May 16 18:56:29 ip-192-168-32-119.us-west-2.compute.internal systemd[1]: nvlsm.service: Skipped due to 'exec-condition'.
May 16 18:56:29 ip-192-168-32-119.us-west-2.compute.internal systemd[1]: Condition check resulted in Start NVLink Subnet Manager being skipped.
bash-5.1# systemctl status nvlsm-online
○ nvlsm-online.service - Start NVLink Subnet Manager Online Service
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/nvlsm-online.service; enabled; preset: enabled)
             └─00-aws-config.conf
     Active: inactive (dead) (Result: exec-condition) since Fri 2025-05-16 18:56:29 UTC; 2min 34s ago
  Condition: start condition failed at Fri 2025-05-16 18:56:29 UTC; 2min 34s ago
    Process: 15972 ExecCondition=[ -s /etc/nvidia/nvlsm.env ] (code=exited, status=1/FAILURE)
        CPU: 1ms

May 16 18:56:29 ip-192-168-32-119.us-west-2.compute.internal systemd[1]: Starting Start NVLink Subnet Manager Online Service...
May 16 18:56:29 ip-192-168-32-119.us-west-2.compute.internal systemd[1]: nvlsm-online.service: Skipped due to 'exec-condition'.
May 16 18:56:29 ip-192-168-32-119.us-west-2.compute.internal systemd[1]: Condition check resulted in Start NVLink Subnet Manager Online Service being skipped.
bash-5.1# systemd-analyze critical-chain
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

multi-user.target @33.290s
└─nvidia-k8s-device-plugin.service @33.053s +223ms
  └─kubelet.service @31.682s +1.238s
    └─containerd.service @31.625s +52ms
      └─runtime.slice @31.621s
        └─-.slice @1.416s
```

</details>

**Terms of contribution:**

### DIFF
--- a/packages/rdma-core/rdma-core.spec
+++ b/packages/rdma-core/rdma-core.spec
@@ -1,7 +1,7 @@
-%global abiver rdmav34
+%global abiver 57
 
 Name: %{_cross_os}rdma-core
-Version: 56.0
+Version: 57.0
 Release: 1%{?dist}
 Summary: RDMA core userspace infrastructure, including core libraries and util programs.
 License: Linux-OpenIB AND MIT
@@ -62,17 +62,21 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
 %{_cross_factorydir}%{_cross_sysconfdir}/libibverbs.d/efa.driver
 
 # Core RDMA libraries
+%{_cross_libdir}/libibmad.so.*
+%{_cross_libdir}/libibumad.so.*
+%{_cross_libdir}/libibnetdisc.so.*
 %{_cross_libdir}/libibverbs.so.*
 %{_cross_libdir}/librdmacm.so.*
 %dir %{_cross_libdir}/libibverbs
 
 # EFA libraries
 %{_cross_libdir}/libefa.so.*
-%{_cross_libdir}/libibverbs/libefa-%{abiver}.so
+%{_cross_libdir}/libibverbs/libefa-rdmav%{abiver}.so
 
 # Verification tools
 %{_cross_bindir}/ibv_devices
 %{_cross_bindir}/ibv_devinfo
+%{_cross_sbindir}/ibstat
 
 # Exclude the bits that are not needed
 %exclude %{_cross_datadir}/perl5
@@ -80,47 +84,33 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
 %exclude %{_cross_libdir}/udev
 %exclude %{_cross_libexecdir}
 %exclude %{_cross_pkgconfigdir}
-%exclude %{_cross_sbindir}
 %exclude %{_cross_sysconfdir}
 %exclude %{_cross_unitdir}
 
 # Exclude all the unused libs
 %exclude %{_cross_libdir}/ibacm*
-%exclude %{_cross_libdir}/libbnxt*
-%exclude %{_cross_libdir}/libcxgb4*
-%exclude %{_cross_libdir}/liberdma*
-%exclude %{_cross_libdir}/libhfi1*
 %exclude %{_cross_libdir}/libhns*
-%exclude %{_cross_libdir}/libibmad*
-%exclude %{_cross_libdir}/libibnetdisc*
-%exclude %{_cross_libdir}/libibumad*
 %exclude %{_cross_libdir}/libmana*
 %exclude %{_cross_libdir}/libmlx*
-%exclude %{_cross_libdir}/libmthca*
-%exclude %{_cross_libdir}/libocrdma*
-%exclude %{_cross_libdir}/libqedr*
-%exclude %{_cross_libdir}/librxe*
-%exclude %{_cross_libdir}/libsiw*
-%exclude %{_cross_libdir}/libvmw*
 %exclude %{_cross_libdir}/rsocket
 
 # Exclude specific RDMA providers (keeping only libefa)
-%exclude %{_cross_libdir}/libibverbs/libbnxt_re-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libcxgb4-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/liberdma-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libhfi1verbs-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libhns-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libipathverbs-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libirdma-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libmana-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libmlx4-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libmlx5-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libmthca-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libocrdma-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libqedr-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/librxe-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libsiw-%{abiver}.so
-%exclude %{_cross_libdir}/libibverbs/libvmw_pvrdma-%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libbnxt_re-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libcxgb4-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/liberdma-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libhfi1verbs-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libhns-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libipathverbs-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libirdma-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libmana-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libmlx4-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libmlx5-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libmthca-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libocrdma-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libqedr-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/librxe-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libsiw-rdmav%{abiver}.so
+%exclude %{_cross_libdir}/libibverbs/libvmw_pvrdma-rdmav%{abiver}.so
 
 # Exclude udev rules
 %exclude %{_cross_udevrulesdir}
@@ -146,14 +136,52 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
 %exclude %{_cross_bindir}/ucmatose
 %exclude %{_cross_bindir}/udaddy
 %exclude %{_cross_bindir}/udpong
+%exclude %{_cross_sbindir}/check_lft_balance.pl
+%exclude %{_cross_sbindir}/dump_fts
+%exclude %{_cross_sbindir}/dump_lfts.sh
+%exclude %{_cross_sbindir}/dump_mfts.sh
+%exclude %{_cross_sbindir}/ibacm
+%exclude %{_cross_sbindir}/ibaddr
+%exclude %{_cross_sbindir}/ibcacheedit
+%exclude %{_cross_sbindir}/ibccconfig
+%exclude %{_cross_sbindir}/ibccquery
+%exclude %{_cross_sbindir}/ibfindnodesusing.pl
+%exclude %{_cross_sbindir}/ibhosts
+%exclude %{_cross_sbindir}/ibidsverify.pl
+%exclude %{_cross_sbindir}/iblinkinfo
+%exclude %{_cross_sbindir}/ibnetdiscover
+%exclude %{_cross_sbindir}/ibnodes
+%exclude %{_cross_sbindir}/ibping
+%exclude %{_cross_sbindir}/ibportstate
+%exclude %{_cross_sbindir}/ibqueryerrors
+%exclude %{_cross_sbindir}/ibroute
+%exclude %{_cross_sbindir}/ibrouters
+%exclude %{_cross_sbindir}/ibsrpdm
+%exclude %{_cross_sbindir}/ibstatus
+%exclude %{_cross_sbindir}/ibswitches
+%exclude %{_cross_sbindir}/ibsysstat
+%exclude %{_cross_sbindir}/ibtracert
+%exclude %{_cross_sbindir}/iwpmd
+%exclude %{_cross_sbindir}/perfquery
+%exclude %{_cross_sbindir}/run_srp_daemon
+%exclude %{_cross_sbindir}/saquery
+%exclude %{_cross_sbindir}/sminfo
+%exclude %{_cross_sbindir}/smpdump
+%exclude %{_cross_sbindir}/smpquery
+%exclude %{_cross_sbindir}/srp_daemon
+%exclude %{_cross_sbindir}/srp_daemon.sh
+%exclude %{_cross_sbindir}/vendstat
 
 %files devel
 %dir %{_cross_includedir}/infiniband
 %dir %{_cross_includedir}/rdma
 %{_cross_includedir}/infiniband/*
 %{_cross_includedir}/rdma/*
-%{_cross_libdir}/libefa*
-%{_cross_libdir}/libibverbs*
-%{_cross_libdir}/librdmacm*
+%{_cross_libdir}/libefa.so
+%{_cross_libdir}/libibmad.so
+%{_cross_libdir}/libibnetdisc.so
+%{_cross_libdir}/libibumad.so
+%{_cross_libdir}/libibverbs.so
+%{_cross_libdir}/librdmacm.so
 
 %changelog
--- a/packages/rdma-core/rdma-core.spec
+++ b/packages/rdma-core/rdma-core.spec
@@ -38,6 +38,7 @@ Requires: %{name}
   -DCMAKE_INSTALL_BINDIR:PATH=%{_cross_bindir} \
   -DCMAKE_INSTALL_SBINDIR:PATH=%{_cross_sbindir} \
   -DCMAKE_INSTALL_SYSCONFDIR:PATH=%{_cross_sysconfdir} \
+  -DCMAKE_INSTALL_UDEV_RULESDIR:PATH=%{_cross_udevrulesdir} \
 
 %make_build
 
@@ -78,10 +79,21 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
 %{_cross_bindir}/ibv_devinfo
 %{_cross_sbindir}/ibstat
 
+# udev rule for renaming to persistent names
+%{_cross_libdir}/udev/rdma_rename
+%{_cross_udevrulesdir}/60-rdma-persistent-naming.rules
+%{_cross_udevrulesdir}/90-rdma-umad.rules
+
+# Exclude the other udev rules we don't want
+%exclude %{_cross_udevrulesdir}/60-srp_daemon.rules
+%exclude %{_cross_udevrulesdir}/75-rdma-description.rules
+%exclude %{_cross_udevrulesdir}/90-iwpmd.rules
+%exclude %{_cross_udevrulesdir}/90-rdma-hw-modules.rules
+%exclude %{_cross_udevrulesdir}/90-rdma-ulp-modules.rules
+
 # Exclude the bits that are not needed
 %exclude %{_cross_datadir}/perl5
 %exclude %{_cross_docdir}
-%exclude %{_cross_libdir}/udev
 %exclude %{_cross_libexecdir}
 %exclude %{_cross_pkgconfigdir}
 %exclude %{_cross_sysconfdir}
@@ -111,9 +123,6 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
 %exclude %{_cross_libdir}/libibverbs/librxe-rdmav%{abiver}.so
 %exclude %{_cross_libdir}/libibverbs/libsiw-rdmav%{abiver}.so
 %exclude %{_cross_libdir}/libibverbs/libvmw_pvrdma-rdmav%{abiver}.so
-
-# Exclude udev rules
-%exclude %{_cross_udevrulesdir}
 
 # Exclude all the unused binaries
 %exclude %{_cross_bindir}/cmtime
--- a/sources/ghostdog/src/infiniband.rs
+++ b/sources/ghostdog/src/infiniband.rs
@@ -1,0 +1,317 @@
+use crate::error;
+use crate::error::Result;
+use snafu::{ensure, ResultExt};
+use std::ffi::OsString;
+use std::ops::Deref;
+use std::path::Path;
+use std::str::FromStr;
+use std::{fs, str, vec};
+
+const SYS_CLASS_DIR: &str = "/sys/class";
+const INFINIBAND_CLASS_NAME: &str = "infiniband";
+const SW_MNG_VALUE: &str = "SW_MNG";
+
+/// Generic device struct found from sysfs
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub(crate) struct InfinibandDevice {
+    pub(crate) name: OsString,
+    pub(crate) ports: Vec<InfinibandPort>,
+}
+
+/// Finds all the devices in sysfs for a given class sorted by name
+pub(crate) fn find_infiniband_devices() -> Result<Vec<InfinibandDevice>> {
+    let devices_path = Path::new(SYS_CLASS_DIR).join(INFINIBAND_CLASS_NAME);
+    let mut found_devices: Vec<InfinibandDevice> = vec![];
+    if !devices_path.exists() {
+        // Nothing to do since no devices detected
+        return Ok(found_devices);
+    }
+
+    let sys_devices_dirs = fs::read_dir(devices_path).context(error::InfinibandSysDevicesSnafu)?;
+    for device in sys_devices_dirs {
+        let device = device.context(error::InfinibandDeviceSnafu)?;
+        let device_name: OsString = device.file_name();
+        found_devices.push(InfinibandDevice {
+            name: device_name,
+            ports: vec![],
+        })
+    }
+
+    found_devices.sort();
+    Ok(found_devices)
+}
+
+impl InfinibandDevice {
+    /// Returns true if the SW_MNG_VALUE is found in the Vital Product Data file for given sysfs device
+    pub(crate) fn is_device_sw_mng(&self) -> Result<bool> {
+        let vpd_file = Path::new(SYS_CLASS_DIR)
+            .join(INFINIBAND_CLASS_NAME)
+            .join(self.name.as_os_str())
+            .join("device")
+            .join("vpd");
+        if vpd_file.exists() {
+            let vpd_file_bytes: &[u8] =
+                &std::fs::read(vpd_file.as_path()).context(error::ReadFileSnafu {
+                    path: vpd_file.as_path(),
+                })?;
+            let needle = SW_MNG_VALUE.as_bytes();
+            if vpd_file_bytes
+                .windows(needle.len())
+                .any(|window| window == needle)
+            {
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+
+    // Given a device in sysfs, iterate over the known paths where ports are defined
+    // Iterate over the ports and read in the capability mask and Port GUID for each port
+    pub(crate) fn find_ports_for_device(self) -> Result<Vec<InfinibandPort>> {
+        let mut found_ports: Vec<InfinibandPort> = vec![];
+        let ports_path = Path::new(SYS_CLASS_DIR)
+            .join(INFINIBAND_CLASS_NAME)
+            .join(self.name)
+            .join("ports");
+
+        let port_directories =
+            fs::read_dir(ports_path).context(error::InfinibandSysDevicesSnafu)?;
+        for port in port_directories {
+            let port = port.context(error::InfinibandSysDevicesSnafu)?;
+            let capability_mask_path = port.path().join("cap_mask");
+            if !capability_mask_path.exists() {
+                continue;
+            }
+            let capability_mask = std::fs::read_to_string(capability_mask_path.as_path()).context(
+                error::ReadFileSnafu {
+                    path: capability_mask_path.as_path(),
+                },
+            )?;
+            let first_guid_path = port.path().join("gids").join("0");
+            if !first_guid_path.exists() {
+                continue;
+            }
+            let first_guid = std::fs::read_to_string(first_guid_path.as_path()).context(
+                error::ReadFileSnafu {
+                    path: first_guid_path.as_path(),
+                },
+            )?;
+
+            // If a capability mask and port guid have been read, create the Infiniband Port
+            found_ports.push(InfinibandPort {
+                capability_mask: CapabilityMask::from_str(capability_mask.as_str())?,
+                port_guid: PortGuid::from_str(first_guid.as_str())?,
+            });
+        }
+        Ok(found_ports)
+    }
+}
+
+/// Struct that holds the two important values for an Infiniband Port when checking for
+/// Subnet Management capabilties.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub(crate) struct InfinibandPort {
+    /// Capability Mask is read in as a string such as 0xa751e848
+    pub(crate) capability_mask: CapabilityMask,
+    /// Primary(first enumerated) PortGuid for a Port
+    pub(crate) port_guid: PortGuid,
+}
+
+impl InfinibandPort {
+    /// is_sm_enabled checks the specific capability bit for Subnet Management means its enabled
+    // The actual bit means `isSMDisabled`, so check if its not zero
+    pub(crate) fn is_sm_enabled(&self) -> bool {
+        // SM is enabled if the eleventh bit ("SM disabled") is clear
+        (*self.capability_mask & (1 << 10)) == 0
+    }
+}
+
+/// Port GUID is a u64 that is read in as a string with : separting 8 sections but is written out in a form such as 0x
+///
+/// For example "fe80:0000:0000:0000:e09d:7303:003f:3bf8" will be stored and displayed as "0xe09d7303003f3bf8"
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub(crate) struct PortGuid(String);
+
+impl FromStr for PortGuid {
+    type Err = error::Error;
+    /// String is expected to look like "fe80:0000:0000:0000:e09d:7303:003f:3bf8"
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let potential_guid = s.trim().to_string();
+        ensure!(
+            potential_guid.starts_with("fe80"),
+            error::InvalidPortGuidStringSnafu {
+                guid: potential_guid
+            }
+        );
+        let parts: Vec<&str> = potential_guid.split(":").collect();
+        ensure!(
+            parts.len() == 8,
+            error::InvalidPortGuidStringSnafu {
+                guid: potential_guid
+            }
+        );
+        for part in parts.iter() {
+            ensure!(
+                part.len() == 4,
+                error::InvalidPortGuidStringSnafu {
+                    guid: potential_guid
+                }
+            );
+        }
+
+        Ok(PortGuid(format!(
+            "0x{}{}{}{}",
+            parts[4], parts[5], parts[6], parts[7]
+        )))
+    }
+}
+
+impl Deref for PortGuid {
+    type Target = String;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+/// Capability Mask struct to help with FromStr and bit masking
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub(crate) struct CapabilityMask(u32);
+
+impl FromStr for CapabilityMask {
+    type Err = error::Error;
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let mask = hex_string_to_u32(s.trim()).context(error::CapabilityCheckSnafu { mask: s })?;
+        Ok(CapabilityMask(mask))
+    }
+}
+
+impl Deref for CapabilityMask {
+    type Target = u32;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+// Helper function to take the string Capability mask and convert to u32
+fn hex_string_to_u32(hex: &str) -> std::result::Result<u32, std::num::ParseIntError> {
+    let hex_str = hex.strip_prefix("0x").unwrap_or(hex);
+    u32::from_str_radix(hex_str, 16)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_hex_to_u32_valid_hex_with_prefix() {
+        assert_eq!(hex_string_to_u32("0xa750e84a").unwrap(), 0xa750e84a);
+        assert_eq!(hex_string_to_u32("0xffff").unwrap(), 0xffff);
+        assert_eq!(hex_string_to_u32("abcdef12").unwrap(), 0xABCDEF12);
+        assert_eq!(hex_string_to_u32("0x0").unwrap(), 0x0);
+    }
+
+    #[test]
+    fn test_hex_to_u32_valid_hex_without_prefix() {
+        assert_eq!(hex_string_to_u32("a750e84a").unwrap(), 0xa750e84a);
+        assert_eq!(hex_string_to_u32("ffff").unwrap(), 0xffff);
+        assert_eq!(hex_string_to_u32("0").unwrap(), 0x0);
+    }
+
+    #[test]
+    fn test_hex_to_u32_invalid_hex() {
+        assert!(hex_string_to_u32("invalid").is_err());
+        assert!(hex_string_to_u32("0xgggg").is_err());
+        assert!(hex_string_to_u32("").is_err());
+    }
+
+    #[test]
+    fn test_hex_to_u32_overflow() {
+        // Value larger than u32::MAX
+        assert!(hex_string_to_u32("0xffffffffff").is_err());
+    }
+
+    #[test]
+    fn test_valid_guid_creation() {
+        let valid_guid = "fe80:0000:0000:0000:e09d:7303:003f:3bf8";
+        let result = PortGuid::from_str(valid_guid);
+        assert!(result.is_ok());
+        assert_eq!(result.as_ref().unwrap().0, "0xe09d7303003f3bf8");
+        assert_eq!(*result.unwrap(), "0xe09d7303003f3bf8");
+    }
+
+    #[test]
+    fn test_valid_guid_with_whitespace() {
+        let valid_guid = "  fe80:0000:0000:0000:e09d:7303:003f:3bf8  ";
+        let result = PortGuid::from_str(valid_guid);
+        assert!(result.is_ok());
+        assert_eq!(*result.unwrap(), "0xe09d7303003f3bf8");
+    }
+
+    #[test]
+    fn test_invalid_guid_prefix() {
+        let invalid_guids = [
+            "abcd:0000:0000:0000:e09d:7303:003f:3bf8",
+            "fe81:0000:0000:0000:e09d:7303:003f:3bf8",
+            "",
+            "fe",
+            "fe8",
+        ];
+
+        for invalid_guid in invalid_guids.iter() {
+            let result = PortGuid::from_str(invalid_guid);
+            assert!(result.is_err());
+
+            match result {
+                Err(e) => {
+                    assert!(matches!(e, error::Error::InvalidPortGuidString { .. }));
+                }
+                Ok(_) => panic!("Expected error for invalid GUID"),
+            }
+        }
+    }
+
+    #[test]
+    fn test_guid_clone() {
+        let guid = PortGuid::from_str("fe80:0000:0000:0000:e09d:7303:003f:3bf8").unwrap();
+        let cloned_guid = guid.clone();
+        assert_eq!(guid.0, cloned_guid.0);
+    }
+
+    #[test]
+    fn test_guid_debug() {
+        let guid = PortGuid::from_str("fe80:0000:0000:0000:e09d:7303:003f:3bf8").unwrap();
+        let debug_output = format!("{:?}", guid);
+        assert!(debug_output.contains("0xe09d7303003f3bf8"));
+    }
+
+    #[test]
+    fn test_is_sm_enabled() {
+        let test_cases = vec![
+            // SM enabled cases (bit 10 is 0)
+            ("0x00000000", true), // All bits 0
+            ("0xa750e84a", true), // Expected mask that passes
+            ("0xfffffbff", true), // Only bit 10 is 0
+            ("0x00000001", true), // Only first bit set
+            // SM disabled cases (bit 10 is 1)
+            ("0x00000400", false), // Only bit 10 set
+            ("0xffffffff", false), // All bits set
+            ("0x00000401", false), // Bit 10 and first bit set
+        ];
+
+        for (capability_mask, expected) in test_cases {
+            let port = InfinibandPort {
+                capability_mask: CapabilityMask::from_str(capability_mask).unwrap(),
+                port_guid: PortGuid::from_str("fe80:0000:0000:0000:e09d:7303:003f:3bf8").unwrap(),
+            };
+
+            assert_eq!(
+                port.is_sm_enabled(),
+                expected,
+                "Failed for capability_mask: {}",
+                capability_mask
+            );
+        }
+    }
+}
--- a/sources/ghostdog/src/main.rs
+++ b/sources/ghostdog/src/main.rs
@@ -6,9 +6,11 @@ It can also check if devices on the PCI bus match a particular NVIDIA driver.
 */
 
 mod error;
+mod infiniband;
 
+use crate::error::Result;
+use crate::infiniband::find_infiniband_devices;
 use argh::FromArgs;
-use error::Result;
 use gptman::GPT;
 use hex_literal::hex;
 use lazy_static::lazy_static;
@@ -16,10 +18,11 @@ use serde::Deserialize;
 use signpost::uuid_to_guid;
 use snafu::{ensure, ResultExt};
 use std::collections::HashSet;
-use std::fs;
-use std::io::{Read, Seek};
-use std::path::PathBuf;
+use std::io::{Read, Seek, Write};
+use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::{fs, str};
+use tempfile::NamedTempFile;
 
 const NVME_CLI_PATH: &str = "/sbin/nvme";
 const NVME_IDENTIFY_DATA_SIZE: usize = 4096;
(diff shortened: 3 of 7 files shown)


sources/ghostdog/src/main.rs
```diff
@@ -319,6 +368,294 @@ lazy_static! {
...
+        if let Some(port) = current_port {
+            ca.ports.push(port);
+        }
+        Ok(ca)
+    }
+}
+
+/// Finds the device name using the nvme-cli
+fn get_ibstat_for_device(device: OsString) -> Result<InfinibandDevice> {
+    let device_name = device.into_string().unwrap();
```
yeazelm: snafu really was fighting me here because the Result is actually of `Result<String, OsString>` which is complicated to work with.

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +368,294 @@ lazy_static! {
...
+// 	   Capability mask: 0xa750e84a
+// 	   Port GUID: 0xe09d7303ffffffff
+// 	   Link layer: InfiniBand
+impl FromStr for InfinibandDevice {
+    type Err = Box<dyn std::error::Error>;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let mut lines = s.lines();
+        let first_line = lines.next().ok_or("Empty input")?;
+        let mut ca = InfinibandDevice {
```
yeazelm: This was quickly written and probably should be refactored, I'm open to recommendations on the "right" way to parse a format like this.

packages/rdma-core/rdma-core.spec
```diff
@@ -62,68 +63,67 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
...
+%{_cross_libdir}/libibmad.so.*
+%{_cross_libdir}/libibumad.so.*
+%{_cross_libdir}/libibnetdisc.so.*
 %{_cross_libdir}/libibverbs.so.*
 %{_cross_libdir}/librdmacm.so.*
 %dir %{_cross_libdir}/libibverbs
 
 # EFA libraries
 %{_cross_libdir}/libefa.so.*
-%{_cross_libdir}/libibverbs/libefa-%{abiver}.so
```
cbgbt, line 71: Do we not need this `.so` anymore?

sources/ghostdog/src/main.rs
```diff
@@ -89,6 +96,11 @@ struct MatchNvidiaDriverArgs {
     driver_name: String,
 }
 
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "infiniband-present")]
+/// Detect if Infiniband devices are attached.
+struct InfinibandPresentArgs {}
```
cbgbt: nit: Not sure if `argh` likes this, but you could just do

```suggestion
struct InfinibandPresentArgs;
```

packages/rdma-core/rdma-core.spec
```diff
@@ -62,68 +63,67 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
...
+%{_cross_libdir}/libibverbs/libefa-rdmav%{abiver}.so
```
bcressey, line 75: Is there a build option that avoids this new suffixing behavior, or are we stuck with it? It makes the diff more complicated to review.

packages/rdma-core/rdma-core.spec
```diff
@@ -62,68 +63,67 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
...
 
 # Verification tools
 %{_cross_bindir}/ibv_devices
 %{_cross_bindir}/ibv_devinfo
+%{_cross_sbindir}/ibstat
+
+# udev rule for renaming to persistent names
+%{_cross_libdir}/udev/rdma_rename
+%{_cross_udevrulesdir}/60-rdma-persistent-naming.rules
+%{_cross_udevrulesdir}/90-rdma-umad.rules
```
bcressey, line 85: This is a systemd alias rule, would it be possible to use that as a dependency for the nvlsm unit?

packages/rdma-core/rdma-core.spec
```diff
@@ -62,68 +63,67 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
...
 %exclude %{_cross_libdir}/ibacm*
-%exclude %{_cross_libdir}/libbnxt*
-%exclude %{_cross_libdir}/libcxgb4*
-%exclude %{_cross_libdir}/liberdma*
-%exclude %{_cross_libdir}/libhfi1*
 %exclude %{_cross_libdir}/libhns*
-%exclude %{_cross_libdir}/libibmad*
-%exclude %{_cross_libdir}/libibnetdisc*
-%exclude %{_cross_libdir}/libibumad*
+%exclude %{_cross_libdir}/libibnetdisc.so
```
bcressey: This should be included in the devel package if we're including the library.

packages/rdma-core/rdma-core.spec
```diff
@@ -146,13 +146,50 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_datadir}/logdog.d
...
+%exclude %{_cross_sbindir}/vendstat
 
 %files devel
 %dir %{_cross_includedir}/infiniband
 %dir %{_cross_includedir}/rdma
 %{_cross_includedir}/infiniband/*
 %{_cross_includedir}/rdma/*
 %{_cross_libdir}/libefa*
+%{_cross_libdir}/libibmad*
+%{_cross_libdir}/libibumad*
```
bcressey: Is this correctly including only the `.so` symlinks and not the actual DSOs?

sources/ghostdog/src/main.rs
```diff
@@ -140,6 +152,9 @@ fn run() -> Result<()> {
             let driver_name = driver.driver_name;
             nvidia_driver_supported(&driver_name)?;
         }
+        SubCommand::InfinibandPresent(_) => {
+            is_infiniband_attached()?;
+        }
```
cbgbt: Not feedback for your code, but we have a common pattern of calling a `run()` function from `main()`, where `main()` then uses the `Display` implementation for our Error types to print an error message.

In `snafu` 0.8, a new attribute macro called `report` was added which has much better output on failure. Lately when I update packages I do the following:

* Rename `run` to `main`
* Add `#[snafu::report] to `run`
* Delete the old `main`

Mind doing that here? The error message output is vastly superior to what we're doing now.

sources/ghostdog/src/main.rs
```diff
@@ -160,6 +175,51 @@ fn is_neuron_attached() -> Result<()> {
...
 }
 
+fn is_infiniband_attached() -> Result<()> {
+    // confirm device has `SW_MNG` in the VPD file
+    let infiniband_sys_devices =
+        fs::read_dir("/sys/class/infiniband/").context(error::InfinibandSysDevicesSnafu)?;
+
+    for device in infiniband_sys_devices {
+        let device = device.context(error::InfinibandSysDevicesSnafu)?;
+        let device_name: OsString = device.file_name().clone();
```
cbgbt: nit, `file_name()` already returns an owned `OsString`:

```suggestion
        let device_name: OsString = device.file_name();
```

sources/ghostdog/src/main.rs
```diff
@@ -365,6 +743,14 @@ mod error {
...
             source: std::io::Error,
         },
+        #[snafu(display("Failed to write '{}': {}", path.display(), source))]
+        WriteFile {
+            path: std::path::PathBuf,
+            source: std::io::Error,
+        },
+        #[snafu(display("Failed to create temporary file: {}", source))]
+        CreateTemporaryFile { source: std::io::Error },
+
```
bcressey: nit: either have line breaks between each enum variant or no line breaks

sources/ghostdog/src/main.rs
```diff
@@ -345,8 +701,30 @@ mod error {
...
+            source
+        ))]
+        IbstatCommand {
+            device: String,
+            source: std::io::Error,
+        },
+        #[snafu(display("Unable to read ibstat command output: {}", source))]
+        ReadIbstatOutput { source: std::str::Utf8Error },
+        #[snafu(display("Unable to parse ibstat command output: {}", source))]
+        ParseIbstatOutput { source: Box<dyn std::error::Error> },
```
bcressey: Is this boxed to address a large enum variant lint?

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+        }
+    }
+    // If there was no reference replaced, append to the end of the file
+    if !found_nvlsm_port {
+        writeln!(nvlsm_config, "guid {}", guid.clone()).context(error::WriteFileSnafu {
+            path: FABRIC_MANAGER_CONFIG_PATH.to_string(),
+        })?;
+    }
+
+    // Wait until both writes have succeeded before copying to ensure a minimal chance of unmatched files
```
bcressey: I don't understand this comment - is something watching writes to these files via inotify? If so, isn't there a race even if it's "minimal"?

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+            ca.ports.push(port);
+        }
+        Ok(ca)
+    }
+}
+
+/// Finds the device name using the nvme-cli
+fn get_ibstat_for_device(device: OsString) -> Result<InfinibandDevice> {
+    // let device_name = device.to_string_lossy();
+    let device_name = device.into_string().unwrap();
```
bcressey: definitely need to fix this unwrap

sources/ghostdog/src/main.rs
```diff
@@ -160,6 +175,51 @@ fn is_neuron_attached() -> Result<()> {
...
+        if vpd_file.exists() {
+            let vpd_file_bytes: &[u8] =
+                &std::fs::read(vpd_file.as_path()).context(error::ReadFileSnafu {
+                    path: vpd_file.as_path(),
+                })?;
+            let vpd_file_string = String::from_utf8_lossy(vpd_file_bytes);
+            if !vpd_file_string.contains("SW_MNG") {
+                continue;
+            }
+        }
```
bcressey: My main misgiving about this is that the command is very generic - `ghostdog infiniband-present` - but it ends up being super specific to a particular configuration of an NVIDIA device.

I would factor this into two parts:
1. generic function that can iterate across Infiniband device entries and check some files for properties
2. specific subcommand for the task, such as `ghostdog infiniband-sw-mng`

sources/ghostdog/src/main.rs
```diff
@@ -160,6 +175,51 @@ fn is_neuron_attached() -> Result<()> {
...
+        let vpd_file = Path::new(&device.path()).join("device").join("vpd");
+        if vpd_file.exists() {
+            let vpd_file_bytes: &[u8] =
+                &std::fs::read(vpd_file.as_path()).context(error::ReadFileSnafu {
+                    path: vpd_file.as_path(),
+                })?;
+            let vpd_file_string = String::from_utf8_lossy(vpd_file_bytes);
+            if !vpd_file_string.contains("SW_MNG") {
+                continue;
+            }
```
cbgbt: Might be worth isolating this so you can write a unit test against a real `vpd` file.

---

I've never seen a `vpd` file. If it's small, you can likely ignore this feedback. The `utf8` conversion can be skipped here though, since we don't actually care about the encoding and our "needle" is ASCII.

```suggestion
            let vpd_file_bytes: &[u8] =
                &std::fs::read(vpd_file.as_path()).context(error::ReadFileSnafu {
                    path: vpd_file.as_path(),
                })?;
            let needle = "SW_MNG".as_bytes();
            if vpd_file_bytes.windows(needle.len())
                .any(|window| window == needle) {
                continue;
            }
```

packages/rdma-core/rdma-core.spec
ytsssun: Sorry, I updated the `abiver` annotation to be more consistent with how upstream uses it
https://github.com/linux-rdma/rdma-core/blob/master/CMakeLists.txt#L86-L87

Let me know if you have different opinion. The name scheme is essentially the same from upstream.

packages/rdma-core/rdma-core.spec
ytsssun: The library was only added for the `ibstats` binary. Nothing is depending on `libibnetdisc` for the build - in which case, do we still want this to be in `devel`?

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+    #[serde(rename = "Hardware version")]
+    hardware_version: u32,
+    #[serde(rename = "Node GUID")]
+    node_guid: String,
+    #[serde(rename = "System image GUID")]
+    system_image_guid: String,
+    ports: Vec<InfinibandPort>,
+}
+
+fn hex_string_to_u32(hex: &str) -> std::result::Result<u32, std::num::ParseIntError> {
```
cbgbt: helpful nit: You can redefine the `Result` type in this module as:

```rust
type Result<T, E = error::Error> = std::result::Result<T, E>;
```

Which would then let you express this as:

```suggestion
fn hex_string_to_u32(hex: &str) -> Result<u32, std::num::ParseIntError>
```

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+            if line.contains("guid 0x") {
+                found_nvlsm_port = true;
+                writeln!(nvlsm_config, "guid {}", guid.clone()).context(error::WriteFileSnafu {
+                    path: FABRIC_MANAGER_CONFIG_PATH.to_string(),
+                })?;
+            } else {
+                writeln!(nvlsm_config, "{}", line).context(error::WriteFileSnafu {
+                    path: FABRIC_MANAGER_CONFIG_PATH.to_string(),
+                })?;
+            }
```
bcressey: The use of `FABRIC_MANAGER_CONFIG_PATH` here seems incorrect.

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+        }
+    }
+    // If there was no reference replaced, append to the end of the file
+    if !found_fabric_manager_port {
+        writeln!(fm_config, "FM_SM_MGMT_PORT_GUID={}", guid.clone()).context(
+            error::WriteFileSnafu {
+                path: FABRIC_MANAGER_CONFIG_PATH.to_string(),
+            },
```
bcressey: This is confusing since you're not actually writing to this file, but a temporary one.

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+/// existing references, and replace the reference if found. If the reference doesn't
+/// exist, it will append to the end of the file the correct reference. This uses a
+/// tempfile to avoid reading and writing to the same file and copies the temp file
+/// over the exiting file when complete.
+fn write_guid(guid: String) -> Result<()> {
+    let mut found_fabric_manager_port = false;
+    let mut found_nvlsm_port = false;
+
+    // Search and update fabricmanager.cfg
+    let mut fm_config = NamedTempFile::new().context(error::CreateTemporaryFileSnafu)?;
```
bcressey: This should be created in the same directory as the target file, so you can rename the temp file into the target one as an atomic operation.

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+        InfinibandDevice::from_str(ibstat_str).context(error::ParseIbstatOutputSnafu)?;
+    Ok(ib_device)
+}
+
+/// write_guid will read the two files that need to be coordinated, search them for
+/// existing references, and replace the reference if found. If the reference doesn't
+/// exist, it will append to the end of the file the correct reference. This uses a
+/// tempfile to avoid reading and writing to the same file and copies the temp file
+/// over the exiting file when complete.
+fn write_guid(guid: String) -> Result<()> {
```
bcressey: This behavior is too surprising as a side effect to a CLI invocation that sounds like a boolean check.

packages/rdma-core/rdma-core.spec
bcressey: > The library was only added for the ibstats binary. Nothing is depending on libibnetdisc for the build - in which case, do we still want this to be in devel?

Devel packages should have .so symlinks for everything in the corresponding library package, unless we know it's something like a plugin that can't be linked against.

sources/ghostdog/src/main.rs
cbgbt: [`Command::args` accepts as `AsRef<OsStr>`](https://doc.rust-lang.org/std/process/struct.Command.html#method.args) (which `&OsString` implements), so I think you actually don't need to do a UTF-8 check here -- you can just pass the `device_name` directly to `args()`.

---

In case I'm wrong:

The reason this is like this is that `into_string()` takes ownership of `device`; however, it returns an `Err` if the string is not valid UTF-8.

If you want to subsequently print an error message that includes the non-UTF-8 string, you'll need access to it after the call to `into_string()`. If the function didn't give you the string back, you'd have to clone it.

I think what you'll basically want to do is something like:

```suggestion
    let device_name = device.into_string().map_err(|device_name| {
        // `display` for this mentions that the device name isn't valid UTF-8
        error::Error::DeviceNameEncoding {
            device_name
        }
    });
```

sources/ghostdog/src/main.rs
cbgbt: I think the "right" way to do this would be something like:
* a custom serde deserializer
* A parser using something like `nom`

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+    #[serde(rename = "Firmware version")]
+    firmware_version: String,
+    #[serde(rename = "Hardware version")]
+    hardware_version: u32,
+    #[serde(rename = "Node GUID")]
+    node_guid: String,
+    #[serde(rename = "System image GUID")]
+    system_image_guid: String,
+    ports: Vec<InfinibandPort>,
+}
```
bcressey: Can we reduce this struct and the parser to just the set of fields we actually need?

sources/ghostdog/src/main.rs
```diff
@@ -319,6 +379,302 @@ lazy_static! {
...
+
+fn hex_string_to_u32(hex: &str) -> std::result::Result<u32, std::num::ParseIntError> {
+    // Remove "0x" if present
```
bcressey: nit: comment is redundant with the code

sources/ghostdog/src/main.rs
yeazelm: I agree with this comment. It was the primary reason this is still in draft. I wanted to have some early feedback on some aspects of the code but I do want to break out the "check if there are devices present to proceed" and the "configure the system". I'll refactor this to be better separated between these two concerns.

sources/ghostdog/src/main.rs
yeazelm: Oh, good point. I was thinking of these as an extension of the end file I want to have but this would be quite misleading to the user if they took it literally. I'll fix this to be more clear.

sources/ghostdog/src/main.rs
yeazelm: Good catch, I did have these staged as fixes and somehow lost them when rebasing onto of develop. I'll fix in a new rev.

sources/ghostdog/src/main.rs
yeazelm: I had originally copied the Fabric Manager file before getting to the nvlsm file, which failed, so the state of the system was one file updated with the guid and one not updated. This comment doesn't do service to what I'm actually trying to accomplish which is avoid touching the original files until everything is ready. I'll update it to be more clear.

sources/ghostdog/src/main.rs
yeazelm: This is an oddity from my custom `FromStr` implementation. This comment is a good reminder to come back and clean this up. I somewhat expected to be encouraged to refactor much of that function so I was planning on addressing it then. I'll fix this in a future revision.

sources/ghostdog/src/main.rs
cbgbt: @yeazelm and I took at look at `ibstat` and saw that it's getting this information from `sysfs`. Might be easiest to also grab it from there, as the format is bound to be more stable than this text output.

sources/ghostdog/src/main.rs
yeazelm: `argh` doesn't like the change.

sources/ghostdog/src/main.rs
yeazelm: Nice! I'll do this change with the rest!

sources/ghostdog/src/main.rs
yeazelm: I was trying to find a way to get this data out of what is a very small (binary) file and this works, but a cleaner way would be to parse the `lspci` output based upon the slot. Given a device `/sys/class/infiniband/ibp115s0f0` with a `device` link of `../../../0000:73:00.0` (`readlink device`), you can use that to get just data about that device:

<details>

```
73:00.0 Infiniband controller: Mellanox Technologies MT2910 Family [ConnectX-7]
	Subsystem: Mellanox Technologies Device 0087
	Physical Slot: 107-1
	Control: I/O- Mem+ BusMaster+ SpecCycle- MemWINV- VGASnoop- ParErr- Stepping- SERR- FastB2B- DisINTx+
	Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
	Latency: 0
	Interrupt: pin A routed to IRQ 254
	NUMA node: 1
	Region 0: Memory at a8000000 (64-bit, prefetchable) [size=32M]
	Capabilities: [60] Express (v2) Endpoint, IntMsgNum 0
		DevCap:	MaxPayload 512 bytes, PhantFunc 0, Latency L0s unlimited, L1 unlimited
			ExtTag+ AttnBtn- AttnInd- PwrInd- RBE+ FLReset+ SlotPowerLimit 25W TEE-IO-
		DevCtl:	CorrErr+ NonFatalErr+ FatalErr+ UnsupReq-
			RlxdOrd+ ExtTag+ PhantFunc- AuxPwr- NoSnoop+ FLReset-
			MaxPayload 128 bytes, MaxReadReq 4096 bytes
		DevSta:	CorrErr+ NonFatalErr- FatalErr- UnsupReq+ AuxPwr- TransPend-
		LnkCap:	Port #0, Speed 16GT/s, Width x2, ASPM not supported
			ClockPM- Surprise- LLActRep- BwNot- ASPMOptComp+
		LnkCtl:	ASPM Disabled; RCB 64 bytes, LnkDisable- CommClk-
			ExtSynch- ClockPM- AutWidDis- BWInt- AutBWInt-
		LnkSta:	Speed 16GT/s, Width x2
			TrErr- Train- SlotClk+ DLActive- BWMgmt- ABWMgmt-
		DevCap2: Completion Timeout: Range ABC, TimeoutDis+ NROPrPrP- LTR-
			 10BitTagComp+ 10BitTagReq+ OBFF Not Supported, ExtFmt- EETLPPrefix-
			 EmergencyPowerReduction Not Supported, EmergencyPowerReductionInit-
			 FRS- TPHComp- ExtTPHComp-
			 AtomicOpsCap: 32bit+ 64bit+ 128bitCAS+
		DevCtl2: Completion Timeout: 260ms to 900ms, TimeoutDis-
			 AtomicOpsCtl: ReqEn+
			 IDOReq- IDOCompl- LTR- EmergencyPowerReductionReq-
			 10BitTagReq+ OBFF Disabled, EETLPPrefixBlk-
		LnkCap2: Supported Link Speeds: 2.5-16GT/s, Crosslink- Retimer+ 2Retimers+ DRS-
		LnkCtl2: Target Link Speed: 16GT/s, EnterCompliance- SpeedDis-
			 Transmit Margin: Normal Operating Range, EnterModifiedCompliance- ComplianceSOS-
			 Compliance Preset/De-emphasis: -6dB de-emphasis, 0dB preshoot
		LnkSta2: Current De-emphasis Level: -6dB, EqualizationComplete+ EqualizationPhase1+
			 EqualizationPhase2+ EqualizationPhase3+ LinkEqualizationRequest-
			 Retimer- 2Retimers- CrosslinkRes: unsupported
	Capabilities: [48] Vital Product Data
		Product Name: Nvidia ConnectX-7 mezz internal for Nvidia Umbriel system
		Read-only fields:
			[PN] Part number: 692-9X760-00SE-S00
			[EC] Engineering changes: A4
			[V2] Vendor specific: 692-9X760-00SE-S00
			[SN] Serial number: MT2451601XLH
			[V3] Vendor specific: 502d1f0053bfef118000e09d733f3bf8
			[VA] Vendor specific: MLX:MN=MLNX:CSKU=V2:UUID=V3:PCI=V0:SMDL=SW_MNG:MODL=C7010Z
			[V0] Vendor specific: PCIeGen4 x2
			[VU] Vendor specific: MT2451601XLHMLNXS0D0F0
			[RV] Reserved: checksum good, 2 byte(s) reserved
		End
	Capabilities: [9c] MSI-X: Enable+ Count=64 Masked-
		Vector table: BAR=0 offset=00002000
		PBA: BAR=0 offset=00003000
	Capabilities: [c0] Vendor Specific Information: Len=18 <?>
	Capabilities: [40] Power Management version 3
		Flags: PMEClk- DSI- D1- D2- AuxCurrent=375mA PME(D0-,D1-,D2-,D3hot-,D3cold+)
		Status: D0 NoSoftRst+ PME-Enable- DSel=0 DScale=0 PME-
	Kernel driver in use: mlx5_core
```

</details>

Where there this data is easier to find and in text:
```
[VA] Vendor specific: MLX:MN=MLNX:CSKU=V2:UUID=V3:PCI=V0:SMDL=SW_MNG:MODL=C7010Z
```
I think the reading of the output might be cleaner than this, but it *is* a tiny file, even if a little strange to search a string in a binary file (thus the strange parsing I'm doing here).

sources/ghostdog/src/main.rs
yeazelm: I did look into this more and found that you *can* find the things you need at a couple of paths within the sysfs, but they are a bit fiddly to work with, under a device such as `/sys/class/infiniband/ibp115s0f0` there is a `ports` dir with a number for each one:
```
ls /sys/class/infiniband/ibp115s0f0/ports/1/
cap_mask  counters  gid_attrs  gids  hw_counters  lid  lid_mask_count  link_layer  phys_state  pkeys  rate  sm_lid  sm_sl  state
```
Where `cap_mask` is the Capability mask: 
```
cat ports/1/cap_mask
0xa751e84a
```
And the `gids` gives you the Port GUID (in :
```
cat ports/1/gids/0
fe80:0000:0000:0000:e09d:7303:003f:3bf8
```
Which line up with the output of `ibstat`. Its significantly less code to iterate over the files in the ports to fetch these two values we need.

sources/ghostdog/src/main.rs
yeazelm: I think we might sidestep parsing output of `ibstat` entirely. I'll look at implementing just the targeted items we need from `/sys/class/infiniband` to see if this can reduce the amount of new code (and dependencies)

sources/ghostdog/src/main.rs
yeazelm: I'm not sure I got myself into a place where I was doing the extra `Into` but without it works fine.

sources/ghostdog/src/main.rs
yeazelm: I've found a possible different path which involves passing the Port GUID to each process via `-g` on their command line invocation. This can be sourced from an `EnvironmentFile` which allows a much simpler interface where `ghostdog` only needs to write to an environment file if there is a suitable device present, and if not, then systemd will do the right thing and skip the right services. This means `ghostdog` will have an argument like `ghostdog write-infiniband-sw-mng-guid` which is more accurate and targeted at writing this file.

yeazelm:
^ this addresses some of the feedback but is not complete. It switches the approach to use systemd EnvironmentFile's for these guids. I'll work on addressing the rest of the feedback now but wanted to get the diff pushed to help with some of the comments.

sources/ghostdog/src/main.rs
```diff
@@ -13,12 +13,17 @@ use serde::Deserialize;
...
+use std::ffi::OsString;
+use std::io::{Read, Seek, Write};
+use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::str::FromStr;
+use std::{fs, str};
 
 const NVME_CLI_PATH: &str = "/sbin/nvme";
+const IBSTAT_CLI_PATH: &str = "/usr/sbin/ibstat";
+const NVLSM_ENV_CONFIG_PATH: &str = "/etc/nvidia/nvlsm.env";
```
bcressey: This could be an argument, perhaps with a default value, instead of a constant.

sources/ghostdog/src/main.rs
```diff
@@ -319,18 +384,239 @@ lazy_static! {
...
+    use std::fs::OpenOptions;
+
+    let mut env_file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(NVLSM_ENV_CONFIG_PATH)
+        .context(error::OpenFileSnafu {
+            path: NVLSM_ENV_CONFIG_PATH.to_string(),
+        })?;
```
bcressey: I'd prefer to use `NamedTempFile::new_in` for this file - see usage in `corndog` and `thar-be-updates`.

sources/ghostdog/src/main.rs
cbgbt: We discussed this aloud and decided that there is a slim but imagineable chance that the binary contents of the `vpd` file could trick `to_string_lossy()` into incorrectly escaping the relevant ASCII string due to misidentifying binary contents of the file as variable content-length code points.

It's probably a good idea to adopt the "bytestring search" window approach.

ytsssun:
I verified the rdma-core functionality with below aspects:
- Verified its compatibility with EFA Device Plugin. 
- Verified `ibstat` is working.
- Verified `ibv_devinfo` and `ibv_devices` works.

packages/rdma-core/rdma-core.spec
yeazelm: I looked at this built ultimately didn't find enough here to help.

sources/ghostdog/src/main.rs
yeazelm: I switched to the writing the environment file. `ghostdog` takes this file as an argument now but the last comment remains the same. I'll update the description to reflect the new approach.

sources/ghostdog/src/main.rs
yeazelm: This new approach works great! Thanks for the tip!

yeazelm:
^ Fixes clippy finding by not implementing `ToString` at all. This is a special visualization of the string `"fe80:0000:0000:0000:e09d:7303:003f:3bf8"` should be `"0xe09d7303003f3bf8"` so this is a different function entirely. Implementing `Display` would result in the underlying string if the user isn't careful.

sources/ghostdog/src/main.rs
```diff
@@ -319,66 +368,29 @@ lazy_static! {
...
+    // Construct the final path and atomically move the temporary file to it
+    tempfile
+        .persist(&config_path)
+        .context(error::PersistTempFileSnafu {
+            path: config_path.clone(),
+        })?;
 
-/// Potential errors during `ghostdog` execution.
-mod error {
-    use snafu::Snafu;
```
bcressey, line 334: Can you split the move of existing error variants into its own commit?

sources/ghostdog/src/main.rs
```diff
@@ -319,66 +368,29 @@ lazy_static! {
...
-    if let Err(e) = run() {
-        eprintln!("{}", e);
-        std::process::exit(1);
-    }
-}
+/// write_guid will write the provided guid in the environment variable format needed for
+/// starting services such as NVIDIA Fabric Manager or NVLSM to the requested path
+fn write_guid(config_path: PathBuf, guid: &str) -> Result<()> {
+    // Create a temporary file in the desired config directory
+    let env_directory = config_path.parent().unwrap_or(Path::new("/"));
```
bcressey: This should return an error instead of defaulting to `/` as the parent directory.

sources/ghostdog/src/main.rs
```diff
@@ -319,66 +368,29 @@ lazy_static! {
...
+    let mut tempfile =
+        NamedTempFile::new_in(env_directory).context(error::CreateTempFileSnafu {
+            path: PathBuf::from(env_directory),
+        })?;
+
+    // Write the config to the temporary file
+    writeln!(tempfile, "GUID_ARG=\"-g {}\"", guid).context(error::WriteTempFileSnafu)?;
```
bcressey: This function would be pretty useful as an alternative to `std::fs::write` without this. Let's prepend `GUID_ARG=` to the input parameter and rename the function accordingly.

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    found_devices.sort_by_key(|device| device.name.clone());
+    Ok(found_devices)
+}
+
+/// Returns true if the SW_MNG_VALUE is found in the Vital Product Data file for given sysfs device
+pub(crate) fn is_device_sw_mng(device: OsString) -> Result<bool> {
+    let vpd_file = Path::new(SYS_CLASS_DIR)
+        .join(SW_MNG_CLASS)
```
bcressey: `is_device_sw_mng` could be an associated function with `SysDevice` (or `InfinibandDevice`), along with `find_ports_for_device`, and then the caller wouldn't have to pass a copy of the device name.

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+const SW_MNG_VALUE: &str = "SW_MNG";
+
+/// Generic device struct found from sysfs
+#[derive(Clone, Debug)]
+pub(crate) struct SysDevice {
+    pub(crate) name: OsString,
+}
+
+/// Finds all the devices in sysfs for a given class sorted by name
+pub(crate) fn find_devices_by_class(class: String) -> Result<Vec<SysDevice>> {
```
bcressey: I would make this `find_infiniband_devices`, and have it return `Result<Vec<InfinibandDevice>>`, so we can punt on the problem of adding stronger typing than `String` to the class selector.

Later, if needed, we can make it more generic later by making class an enum, with variants like `Infiniband`, `Block` etc - but that will make the other functions and the return type more complex and it doesn't seem worth it for now.

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+use std::path::Path;
+use std::str::FromStr;
+use std::{fs, str};
+
+const SYS_CLASS_DIR: &str = "/sys/class";
+const SW_MNG_CLASS: &str = "infiniband";
+const SW_MNG_VALUE: &str = "SW_MNG";
+
+/// Generic device struct found from sysfs
+#[derive(Clone, Debug)]
```
bcressey:
```suggestion
#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
```

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    let sys_devices_dirs = fs::read_dir(devices_path).context(error::InfinibandSysDevicesSnafu)?;
+    for device in sys_devices_dirs {
+        let device = device.context(error::InfinibandDeviceSnafu)?;
+        let device_name: OsString = device.file_name();
+        found_devices.push(SysDevice { name: device_name })
+    }
+
+    found_devices.sort_by_key(|device| device.name.clone());
```
bcressey: With the derived impls:
```suggestion
    found_devices.sort();
```

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+    Ok(false)
+}
+
+/// Struct that holds the two important values for an Infiniband Port when checking for
+/// Subnet Management capabilties.
+#[derive(Clone, Debug)]
+pub(crate) struct InfinibandPort {
+    /// Capability Mask is read in as a string such as 0xa751e848
+    pub(crate) capability_mask: CapabilityMask,
+    /// Port GUID is a u64 that is read in as a : formated string but written out in a form such as 0x
```
bcressey: What is a `: formated string`?

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub(crate) port_guid: Guid,
```
bcressey: "guid" is pretty generic, I would disambiguate for clarity:
```suggestion
    pub(crate) port_guid: PortGuid,
```

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// is_sm_enabled checks the specific capability bit for Subnet Management means its enabled
+    // The actual bit means `isSMDisabled`, so check if its not zero
+    pub(crate) fn is_sm_enabled(&self) -> bool {
+        let bit_position: u32 = 10;
+        let mask: u32 = 1 << bit_position;
+        if !(&self.capability_mask & mask) != 0 {
+            return true;
+        }
+        false
+    }
```
bcressey: Negating the value after masking means that it will always be non-zero, because the zero bits in the mask get flipped to ones in the final value.

```suggestion
    pub(crate) fn is_sm_enabled(&self) -> bool {
        // SM is enabled if the eleventh bit ("SM disabled") is clear
        (&self.capability_mask & (1 << 10)) == 0
    }
```

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+        if parts.len() == 8 {
+            return format!("0x{}{}{}{}", parts[4], parts[5], parts[6], parts[7]);
+        }
+        self.0.to_string()
+    }
+}
+
+/// Capability Mask struct to help with FromStr and bit masking
+#[derive(Clone, Debug)]
+pub(crate) struct CapabilityMask(u32);
```
bcressey: You could also `impl Deref` for `CapabilityMask` and delegate everything to its inner `u32`.

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+    type Err = error::Error;
+    /// String is expected to look like "fe80:0000:0000:0000:e09d:7303:003f:3bf8"
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let potential_guid = s.trim().to_string();
+        ensure!(
+            potential_guid.starts_with("fe80"),
+            error::InvalidGuidStringSnafu {
+                guid: potential_guid
+            }
+        );
```
bcressey: What is the source of this `fe80` check?

It'd be cleaner to do this via regex; right now we just quietly return the unmodified value in `to_config_string` but it might not be in the correct format if we don't have all 8 parts.

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        let guid = Guid(s.trim().to_string());
+        Ok(guid)
```
bcressey: IMO we should store the GUID in its final string form, and perhaps impl Deref so that the inner string can be used directly.

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let port_directories = fs::read_dir(ports_path).context(error::InfinibandSysDevicesSnafu)?;
+    for port in port_directories {
+        let mut capability_mask: String = "".to_string();
+        let mut first_guid: String = "".to_string();
+
+        let port = port.context(error::InfinibandSysDevicesSnafu)?;
+        let capability_mask_path = port.path().join("cap_mask");
+        if capability_mask_path.exists() {
+            capability_mask = std::fs::read_to_string(capability_mask_path.as_path()).context(
```
bcressey: This will let you declare the `capability_mask` and `first_guid` bindings at the time you assign to them, and avoid the empty string checks at the bottom of the loop:
```
        if !capability_mask_path.exists() {
            continue;
        }
        ...
        if !first_guid_path.exists() {
            continue;
        }
```

sources/ghostdog/src/infiniband.rs
yeazelm: The GUID seems to act as an IPv6 address (this is where the `fe80` comes from). I see some references where they list the "full" GUID as something like `fe80::aaaa:bbbb:cccc:dddd` but ultimately the only uses we have right now are the final form. I think we can store the `0x` prefixed string and use Deref since that is what we really need. If a new use case comes up for using the other form, we can shift this around but it seems unlikely based upon what I've been finding with all references to this GUID being in the `0x` prefixed string.

sources/ghostdog/src/infiniband.rs
yeazelm: This is better! I'll do that!

sources/ghostdog/src/main.rs
yeazelm: Sounds good. I'll do the refactor.

sources/ghostdog/src/main.rs
yeazelm: Good catch. I've made it error when a suitable path isn't provided.

sources/ghostdog/src/infiniband.rs
yeazelm: `fe80` is the prefix of the GUIDs for infiniband devices. I agree that we should confirm we read 8 sections and the last 4 contain values other than all zeros. I'll add the additional checks into this function and store as the final string rather than the incoming string.

sources/ghostdog/src/infiniband.rs
yeazelm: This is a nice improvement!

sources/ghostdog/src/infiniband.rs
yeazelm: Thanks for catching this, I had ment to go back and confirm if this was going to work as expected.

packages/rdma-core/rdma-core.spec
ytsssun: This still needs to be updated so that we only include the .so files in the devel package @yeazelm.

sources/ghostdog/src/main.rs
```diff
@@ -160,6 +179,40 @@ fn is_neuron_attached() -> Result<()> {
...
+    for device in devices {
+        if device.is_device_sw_mng()? {
+            let ports = device.find_ports_for_device()?;
+            for port in ports {
+                if port.is_sm_enabled() {
+                    // NVIDIA Fabric Manager or NVLSM use -g to configure a Port GUID, use this as input
+                    // to those services
+                    write_config_string(
+                        env_file.clone(),
+                        format!("GUID_ARG=\"-g {}\"", *port.port_guid).as_str(),
```
bcressey, line 205: nit: this subcommand could take `env_var` and `env_val_prefix` arguments to avoid hard-coding this logic

sources/ghostdog/src/main.rs
```diff
@@ -319,73 +372,43 @@ lazy_static! {
...
+            path: config_path.clone(),
+        })?;
+
+    let mut tempfile =
+        NamedTempFile::new_in(env_directory).context(error::CreateTempFileSnafu {
+            path: PathBuf::from(env_directory),
+        })?;
+
+    // Write the config to the temporary file
+    writeln!(tempfile, "{}", content_string).context(error::WriteTempFileSnafu)?;
```
bcressey: nit: avoid the `writeln!` if the content string is empty so the file is zero bytes

sources/ghostdog/src/main.rs
```diff
@@ -476,4 +499,77 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_write_config_string_overwrite_existing() -> Result<()> {
+        // Create a temporary directory for testing
+        let temp_dir = TempDir::new().unwrap();
+        let config_path = temp_dir.path().join("config.txt");
+
+        // Write initial content
+        let initial_content = "initial content\n";
```
bcressey: nit: the newline here is distracting since newline termination is a special property of this implementation:
```suggestion
        // Write initial content
        let initial_content = "initial content";
```

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,283 @@
...
+        let potential_guid = s.trim().to_string();
+        ensure!(
+            potential_guid.starts_with("fe80"),
+            error::InvalidPortGuidStringSnafu {
+                guid: potential_guid
+            }
+        );
+        let parts: Vec<&str> = potential_guid.split(":").collect();
+        ensure!(
+            parts.len() == 8,
```
bcressey, line 148: nit: let's also assert that each part has a length of 4

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,283 @@
...
+        assert_eq!(hex_string_to_u32("abcdef12").unwrap(), 0xABCDEF12);
+        assert_eq!(hex_string_to_u32("0x0").unwrap(), 0x0);
+    }
+
+    #[test]
+    fn test_hex_to_u32_valid_hex_without_prefix() {
+        assert_eq!(hex_string_to_u32("a750e84a").unwrap(), 0xa750e84a);
+        assert_eq!(hex_string_to_u32("FFFF").unwrap(), 0xFFFF);
+        assert_eq!(hex_string_to_u32("0").unwrap(), 0x0);
+    }
```
bcressey: nit: it is a bit distracting to use uppercase for the u32 hex constants

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,283 @@
...
+        let result = PortGuid::from_str(valid_guid);
+        assert!(result.is_ok());
+        assert_eq!(*result.unwrap(), "0xe09d7303003f3bf8");
+    }
+
+    #[test]
+    fn test_invalid_guid_prefix() {
+        let invalid_guids = [
+            "abcd:0000:0000:0000:e09d:7303:003f:3bf8",
+            "FE80:0000:0000:0000:e09d:7303:003f:3bf8", // Testing case sensitivity
```
bcressey: Why is case sensitivity desirable?

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,283 @@
...
+            "",
+            "fe",
+            "fe8",
+        ];
+
+        for invalid_guid in invalid_guids.iter() {
+            let result = PortGuid::from_str(invalid_guid);
+            assert!(result.is_err());
+
+            // If you want to test the specific error type
```
bcressey: nit: do we need this comment?

sources/ghostdog/src/infiniband.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+impl InfinibandPort {
+    /// is_sm_enabled checks the specific capability bit for Subnet Management means its enabled
+    // The actual bit means `isSMDisabled`, so check if its not zero
+    pub(crate) fn is_sm_enabled(&self) -> bool {
+        // SM is enabled if the eleventh bit ("SM disabled") is clear
+        (*self.capability_mask & (1 << 10)) == 0
+    }
+}
```
bcressey, line 127: It'd be good to have unit tests for this also

sources/ghostdog/src/infiniband.rs
yeazelm: It doesn't have to be, but the implementation is case sensitive. I can `.to_lowercase()` everything if that is more desirable.

arnaldo2792 (approved):

sources/ghostdog/src/main.rs
yeazelm: I tried this a bit and felt it actually made the code more complicated, instead of a single place to get the string right, there was business logic sprinkled into the writing function and yet the caller still needed to provide a `format!("\"-g {}\"", *port.port_guid).as_str()` since this is specific to this call. I think this function currently hits a local maximum of flexibility vs complexity. I think there is room for a fully formed "environment variable writer" that takes a struct that properly captures the variables and their prefixes and values, but that is a little overkill right now for what we need.

sources/ghostdog/src/main.rs
```diff
@@ -476,4 +501,77 @@ mod test {
...
+        let temp_dir = TempDir::new().unwrap();
+        let config_path = temp_dir.path().join("config.txt");
+        let empty_content = "";
+
+        // Write empty content to file
+        write_config_string(config_path.clone(), empty_content)?;
+
+        // Verify the content was written correctly
+        let read_content = fs::read_to_string(&config_path).unwrap();
+        assert_eq!(read_content, ""); // Note: writeln! adds a newline
```
bcressey, line 552: nit: can move this comment to the next test:
```suggestion
        assert_eq!(read_content, "");
```

bcressey (approved):
♾️ 🩹

ytsssun (approved):

PR #501, opened 2025-05-12
nvidia-k8s-device-plugin: add ldcache parsing for aarch64 patch

sky1122:
**Description of changes:**
add patch to fix ldcache parsing for aarch64

k8s-device-plugin carries its own nvidia-container-toolkit for now and uses nvidia-ctk to generate the CDI specifications.

The architecture flag for aarch64 is currently missing from the supported architecture flags list. This omission causes the getEntries function to exclude all libraries found on aarch64 hosts. As a result helper programs like nvidia-ctk are unable to generate CDI specifications for the aarch64 architecture.

**Testing done:**
- testing with part of this [PR](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/459)

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-k8s-device-plugin/0001-fix-ldcache-parsing-for-aarch64.patch
+++ b/packages/nvidia-k8s-device-plugin/0001-fix-ldcache-parsing-for-aarch64.patch
@@ -1,0 +1,49 @@
+From be4ba83b821eea9050eefdb7e67df2d757c3795a Mon Sep 17 00:00:00 2001
+From: Jingwei Wang <jweiw@amazon.com>
+Date: Wed, 23 Apr 2025 17:17:35 +0000
+Subject: [PATCH] fix ldcache parsing for aarch64
+
+k8s-device-plugin carries its own nvidia-container-toolkit and uses
+nvidia-ctk to generate the CDI specifications.
+
+The architecture flag for aarch64 is currently missing from the
+supported architecture flags list. This omission causes the getEntries
+function to exclude all libraries found on aarch64 hosts. As a result
+helper programs like nvidia-ctk are unable to generate CDI
+specifications for the aarch64 architecture.
+
+This fix adds the missing aarch64 architecture flag, using the same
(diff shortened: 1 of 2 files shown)


sky1122:
successful build on my build machine, but fail on here. Looking why the patch applied errored.

sky1122:
one comment added, that's why patch fail. Will re-generate the patch.

sky1122:
forget to commit the changes, will forced pushed again

sky1122:
when doing the git changes lost one `spec` file changes, will force pushed again

sky1122:
force pushed to adopt the new changes for patch

arnaldo2792:
The previous patch to generate the devices was removed. Instead of generating the specs for the CDI device through the device plugin, we will use the `generate-cdi-specs.service` systemd unit to provide them. This aligns with what the GPU operator does to provide the same CDI specifications.

arnaldo2792 (approved):

rpkelly (approved):

packages/nvidia-k8s-device-plugin/0001-fix-ldcache-parsing-for-aarch64.patch
```diff
@@ -0,0 +1,49 @@
+From be4ba83b821eea9050eefdb7e67df2d757c3795a Mon Sep 17 00:00:00 2001
```
ytsssun, line 1: Nice work on this one! Q - Do we have plan to upstream this?

ytsssun (approved):

packages/nvidia-k8s-device-plugin/0001-fix-ldcache-parsing-for-aarch64.patch
arnaldo2792: They are already working on fixes :tada: !

https://github.com/NVIDIA/nvidia-container-toolkit/pull/1046

PR #502, opened 2025-05-13
schnauzer: add helper function to get the first value in a array

sky1122:
**Description of changes:**
add a handlebar helper function to get the first value in the list

**Testing done:**
`cargo test`, `make`

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/src/helpers/stdlib/mod.rs
+++ b/sources/api/schnauzer/src/helpers/stdlib/mod.rs
@@ -1257,3 +1257,100 @@ mod test_toml_encode {
         assert_eq!(result, expected);
     }
 }
+
+/// `get_array_first_value` accepts arrary input or a string and get the string value or
+/// the first value in the list
+///
+/// # Example
+///
+/// Consider an array of values: `[ "a", "b", "c" ]` stored in a setting such as
+/// `settings.somewhere.foo-list`. In our template we can write:
+/// `{{ string_or_first_value settings.somewhere.foo-list }}`
+///
+/// This will render `"a"`.
+///
+/// Similarly, for a string: `"string"`, the template {{ get_array_first_value settings.somewhere.foo-list }}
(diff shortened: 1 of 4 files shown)


sky1122:
force-pushed to fix the lint

sources/api/schnauzer/src/helpers/stdlib/mod.rs
```diff
@@ -1257,3 +1257,100 @@ mod test_toml_encode {
...
+/// # Example
+///
+/// Consider an array of values: `[ "a", "b", "c" ]` stored in a setting such as
+/// `settings.somewhere.foo-list`. In our template we can write:
+/// `{{ string_or_first_value settings.somewhere.foo-list }}`
+///
+/// This will render `"a"`.
+///
+/// Similarly, for a string: `"string"`, the template {{ get_array_first_value settings.somewhere.foo-list }}
+/// will render `"string"`.
```
cbgbt, line 1273: Do we need a new helper? I wonder if we can get this behavior with existing helpers (e.g. check under `stdlib/reflective.rs`):

For example:

```handlebars
{{#if (is_array settings.somewhere.foo-list) }}
{{ settings.somewhere.foo-list.[0] }}
{{else if (is_string settings.somewhere.foo-list) }}
{{ settings.somewhere.foo-list }}
{{/if}}
```

sky1122:
@cbgbt suggestion is the use case, so close this PR.

sources/api/schnauzer/src/helpers/stdlib/mod.rs
cbgbt: The change was otherwise well implemented!

If you end up having space to look into a problem with lasting impact, https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/503 describes an issue in this library that would be extremely helpful if implemented.

PR #514, opened 2025-05-15
Update `libnvme` and `nvme-cli`

piyush-jena:
**Description of changes:**
- Update `libnvme` v1.11.1  &rarr; v1.13
- Update `nvme-cli` v2.11  &rarr; v2.13
- Removed patch as it was merged in this PR: https://github.com/linux-nvme/nvme-cli/pull/2559

**Testing done:**
```
bash-5.1# nvme version
nvme version 2.13 (git 2.13)
libnvme version 1.13 (git 1.13)
bash-5.1# nvme amzn stats /dev/nvme0n1
Total Ops:
  Read: 8337
  Write: 99
Total Bytes:
  Read: 531169280
  Write: 8503296
Total Time (us):
  Read: 75410559
  Write: 148461

EBS Volume Performance Exceeded (us):
  IOPS: 0
  Throughput: 0

EC2 Instance EBS Performance Exceeded (us):
  IOPS: 0
  Throughput: 0

Queue Length (point in time): 0

Read IO Latency Histogram
=================================
Lower       Upper        IO Count
=================================
[0        - 1       ] => 0
[1        - 2       ] => 0
[2        - 4       ] => 0
[4        - 8       ] => 0
[8        - 16      ] => 0
[16       - 32      ] => 0
[32       - 64      ] => 0
[64       - 128     ] => 326
[128      - 256     ] => 726
[256      - 512     ] => 2688
[512      - 1024    ] => 3236
[1024     - 2048    ] => 557
[2048     - 4096    ] => 96
[4096     - 8192    ] => 6
[8192     - 16384   ] => 1
[16384    - 32768   ] => 2
[32768    - 65536   ] => 29
[65536    - 131072  ] => 599
[131072   - 262144  ] => 71
[262144   - 524288  ] => 0
[524288   - 1048576 ] => 0
[1048576  - 2097152 ] => 0
[2097152  - 4194304 ] => 0
[4194304  - 8388608 ] => 0
[8388608  - 16777216] => 0
[16777216 - 33554432] => 0
[33554432 - 67108864] => 0
[67108864 - 18446744073709551615] => 0
=================================

Write IO Latency Histogram
=================================
Lower       Upper        IO Count
=================================
[0        - 1       ] => 0
[1        - 2       ] => 0
[2        - 4       ] => 0
[4        - 8       ] => 0
[8        - 16      ] => 0
[16       - 32      ] => 0
[32       - 64      ] => 0
[64       - 128     ] => 0
[128      - 256     ] => 0
[256      - 512     ] => 1
[512      - 1024    ] => 43
[1024     - 2048    ] => 26
[2048     - 4096    ] => 29
[4096     - 8192    ] => 0
[8192     - 16384   ] => 0
[16384    - 32768   ] => 0
[32768    - 65536   ] => 0
[65536    - 131072  ] => 0
[131072   - 262144  ] => 0
[262144   - 524288  ] => 0
[524288   - 1048576 ] => 0
[1048576  - 2097152 ] => 0
[2097152  - 4194304 ] => 0
[4194304  - 8388608 ] => 0
[8388608  - 16777216] => 0
[16777216 - 33554432] => 0
[33554432 - 67108864] => 0
[67108864 - 18446744073709551615] => 0
=================================
```

**Terms of contribution:**


packages/libnvme/libnvme.spec
```diff
@@ -21,14 +21,15 @@ Requires: %{_cross_os}libnvme
 %{summary}.
 
 %prep
-%autosetup -n libnvme-%{version} -p1
+%autosetup -n libnvme-%{version}
```
koooosh: I don't believe we should remove the `-p1` arg since I see it in other package specs even if they have no patches.

packages/nvme-cli/nvme-cli.spec
```diff
@@ -19,7 +18,7 @@ Requires: %{_cross_os}libnvme
 %{summary}.
 
 %prep
-%autosetup -n nvme-cli-%{version} -p1
+%autosetup -n nvme-cli-%{version}
```
koooosh: I don't believe we should remove the `-p1` arg since I see it in other package specs even if they have no patches.

koooosh (commented):
LGTM, aside from small comments. Glad the patch was merged upstream so we can drop it now :)

koooosh:
Also, I recommend making the PR title more appropriate, something like: Update `libnvme` and `nvme-cli` packages

koooosh (approved):

bcressey (approved):

packages/libnvme/libnvme.spec
bcressey: I agree with this, it's a bit like leaving a trailing comma after the last item in a list in code: it cuts down on incidental changes if we have to add a patch later since we'd have to put back `-p1` on this line. It also avoids wasting time if you add a patch but don't notice that the `%autosetup` line doesn't actually apply it 😀

piyush-jena:
rebased and added `-p1` as suggested above.

vigh-m (approved):

PR #518, opened 2025-05-19
enable cryptsetup and tpm2 functionality for systemd

### LINKED ISSUE #506: enable systemd-cryptsetup in systemd also
**What I'd like:** Perhaps the final piece so I can drop my systemd fork would be to enable systemd-cryptsetup in the upstream now that all the other pieces are in place, seeing as you decided to include libssl.so in the libcrypto packaging finally @bcressey 🙏 

**Any alternatives you've considered:** Keep carrying my massive fork of systemd to enable systemd-cryptsetup.

bcressey:
**Issue number:**

Closes #506

**Description of changes:**
Refactor the AWS-LC and cryptsetup packages to make the command line tools optional, and to narrow the dependencies so that `libssl.so` isn't installed when `libcrypto.so` alone is sufficient.

Patch systemd so that we can enable openssl without build failures, and without depending on `libssl.so` at runtime.

`systemd-cryptsetup` and the related programs and targets are added to a new subpackage. The main package now depends additional shared libraries at runtime, as a result of new functionality in `systemd-repart` and `systemd-creds`, plus the new programs in the cryptsetup subpackage.

**Testing done:**
Repeated the testing from https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/444 

This adds around 5 MiB of libraries to the base install:
```
16K	libaio.so.1.0.2
3.3M	libcrypto.so
576K	libcryptsetup.so.12.10.0
364K	libdevmapper.so.1.02
456K	libtss2-esys.so.0.0.1
132K	libtss2-mu.so.0.0.1
36K	libtss2-rc.so.0.0.0
148K	libtss2-sys.so.1.0.1
20K	libtss2-tcti-device.so.0.0.0
24K	libtss2-tctildr.so.0.0.0
```

The largest of these, `libcrypto.so`, is already present in various forms as it's embedded in the FIPS versions of Go binaries, and in Rust binaries by way of the `rustls` dependency.

**Terms of contribution:**


packages/libcryptsetup/libcryptsetup.spec
```diff
@@ -80,13 +89,15 @@ Requires: %{_cross_os}libuuid-devel
...
-%{_cross_sbindir}/integritysetup
-%{_cross_sbindir}/veritysetup
 %{_cross_libdir}/*.so.*
 %{_cross_tmpfilesdir}/cryptsetup.conf
 %exclude %{_cross_mandir}
 
+%files tools
+%{_cross_sbindir}/cryptsetup
+%{_cross_sbindir}/integritysetup
+%{_cross_sbindir}/veritysetup
```
arnaldo2792, line 103: Are these going to actually be used? We are not going to be using these instead of the ones in the SDK right?

arnaldo2792 (approved):

packages/libcryptsetup/libcryptsetup.spec
mikn: I don't think there is a case for integritysetup or veritysetup to be used "in bottlerocket", at least I cannot come up with one. We use the `cryptsetup` binary in the bottlerocket OS layer as an udev rule to encrypt discovered ephemeral devices however.

Something like this:
```
# Handle unformatted disk
ACTION=="add|change", SUBSYSTEM=="block", ENV{DEVTYPE}=="disk", \
  ENV{BOTTLEROCKET_DEVICE_TYPE}=="ephemeral", \
  ENV{ID_FS_USAGE}!="?*", \
  RUN+="/usr/bin/logger -t crypt-additional 'Found unformatted device %k'", \
  RUN+="/usr/sbin/cryptsetup -v --batch-mode luksFormat --type luks2 --pbkdf argon2id --pbkdf-memory 2097152 --pbkdf-parallel 4 --pbkdf-force-iterations 4 --hash sha512 /dev/%k /usr/share/crypt-data/bottlerocket-data.key", \
  RUN+="/usr/bin/systemd-cryptenroll --tpm2-device=auto --tpm2-pcrs=0 --wipe-slot=0 /dev/%k --unlock-key-file=/usr/share/crypt-data/bottlerocket-data.key", \
  ENV{SYSTEMD_WANTS}+="crypt-ephemeral-tpm@%k.service"
```

mikn:
Thank you @bcressey for putting in all this work, this looks like a great solution! :pray:

mikn (approved):

packages/libcryptsetup/libcryptsetup.spec
bcressey: `cryptsetup` is needed to create an encrypted volume at runtime, which some downstream users are doing (per @mikn).

It'd be a bit challenging to use it in the SDK since that would imply a well-known secret, which would need to be available at runtime in order to unlock the volume. It's probably going to be easier to defer the creation of such block devices until runtime, and then decide whether to create the filesystem on an encrypted or unencrypted device based on whether or not we have a TPM.

So this ends up more like our `mkfs.xfs` usage at runtime for the data partition, where we need to know our final filesystem size to create it with the optimal parameters.

`veritysetup` and `integritysetup` are more niche, but also fairly small.

yeazelm (approved):
Nice!

PR #520, opened 2025-05-20
settings-sdk: bump bottlerocket-settings-models to 0.10.0

sky1122:
**Description of changes:**
bump bottlerocket-settings-models to 0.10.0

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -211,13 +211,13 @@ base64 = "0.22"
 
 [workspace.dependencies.bottlerocket-modeled-types]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.9.0"
-version = "0.8.0"
+tag = "bottlerocket-settings-models-v0.10.0"
+version = "0.9.0"
 
 [workspace.dependencies.bottlerocket-settings-models]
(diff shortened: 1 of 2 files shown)


sources/Cargo.toml
```diff
@@ -211,13 +211,13 @@ base64 = "0.22"
 
 [workspace.dependencies.bottlerocket-modeled-types]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.9.0"
-version = "0.8.0"
+tag = "bottlerocket-settings-models-v0.10.0"
+version = "0.9.0"
```
KCSesh, line 215: I believe this should match the settings model version, and it was a miss the last time this was updated. 
```suggestion
version = "0.10.0"
```

ytsssun: @KCSesh @sky1122 it should be `0.9.0` indeed. https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/develop/Cargo.toml#L60

arnaldo2792 (approved):

ytsssun (approved):

sources/Cargo.toml
KCSesh: Oh my mistake - on our last release `modeled-types` was not actually updated - just the tag - so this is correct.

KCSesh (approved):

PR #526, opened 2025-05-27
fix various build warnings

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket-sdk/pull/276

**Description of changes:**
When building with LTO enabled, certain warnings must not be ignored as they indicate that the program is not compatible with LTO. As part of the enablement work, I reviewed warnings from all packages and addressed many of them.

I focused on three main areas:
1. actual LTO incompatibilities - `libaio`, `libcrypto`, and `libisal`. I disabled LTO for the first two, and added recent upstream patches for ISA-L to address the last one.
2. new warnings under LTO - the visibility across compilation units can improve error detection, but can also lead to false positives. `mdadm` and `procps` both needed fixes; I addressed `mdadm` by updating to 4.4, and `procps` by adding an upstream patch.
3. RPM build warnings - these were of a nuisance than an actual problem, but still worth cleaning up as they are easy to fix. `procps` needed the most work to avoid duplicate build ID warnings.

**Testing done:**
Verified that `pgrep`, `pkill`, `pidwait`, `snice`, and `skill` worked as expected after getting the symlink treatment:
```
# ls -latr /usr/bin/{pkill,pgrep,pidwait,snice,skill}
lrwxrwxrwx. 1 root root     5 May 27 04:50 /usr/bin/snice -> skill
-rwxr-xr-x. 1 root root 32200 May 27 04:50 /usr/bin/skill
-rwxr-xr-x. 1 root root 36440 May 27 04:50 /usr/bin/pkill
lrwxrwxrwx. 1 root root     5 May 27 04:50 /usr/bin/pidwait -> pkill
lrwxrwxrwx. 1 root root     5 May 27 04:50 /usr/bin/pgrep -> pkil
```

Confirmed that `mdadm` sets up a RAID 0 array, and that `mdmonitor.service` still runs, though it exits immediately with a non-error "mdadm: No array with redundancy detected, stopping" report.

Otherwise, inspected build logs and found that the warnings went away following the fixes.

**Terms of contribution:**


yeazelm (approved):
Nice! I like the clean up for not needing to exclude the `.la` files!

KCSesh (approved):

PR #528, opened 2025-05-27
systemd: backport fix for daemon-reload and stop using isolate

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket-sdk/pull/276

**Description of changes:**
Guard against a potential regression from adding `-ftrivial-auto-var-init=zero` to the default compiler flags by backporting a `systemd` patch to avoid a 1M stack allocation. As a bonus, this should speed up `systemctl daemon-reload` which is now one of the most expensive parts of our boot sequence.

Replace uses of `systemctl isolate` with `systemctl start`. Upstream calls `isolate` ["a bad idea"](https://github.com/systemd/systemd/issues/26364#issuecomment-1424900066) and suggests `start` instead, which works fine for our purposes.

Note that we rely on the implicit reload that happens when the target is changed to find the units for host and bootstrap containers. This could be an explicit reload, but that would leave the target at its default value - either `preconfigured.target` or `fipscheck.target` - which changes the output that `systemd-analyze` reports.

**Testing done:**
Verified that the system comes up correctly, and that both bootstrap containers and host containers run as expected. 

**Terms of contribution:**

### DIFF
--- a/packages/release/activate-configured.service
+++ b/packages/release/activate-configured.service
@@ -1,12 +1,12 @@
 [Unit]
-Description=Isolates configured.target
+Description=Activate configured.target
 After=preconfigured.target
 Requires=preconfigured.target
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/systemctl set-default configured
-ExecStart=/usr/bin/systemctl isolate default --no-block
+ExecStart=/usr/bin/systemctl set-default configured.target
+ExecStart=/usr/bin/systemctl start configured.target --no-block
 RemainAfterExit=true
 StandardError=journal+console
 
--- a/packages/release/activate-multi-user.service
+++ b/packages/release/activate-multi-user.service
@@ -1,12 +1,12 @@
 [Unit]
-Description=Isolates multi-user.target
+Description=Activate multi-user.target
 After=configured.target reboot-if-required.service
 Requires=configured.target reboot-if-required.service
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/systemctl set-default multi-user
-ExecStart=/usr/bin/systemctl isolate default --no-block
+ExecStart=/usr/bin/systemctl set-default multi-user.target
+ExecStart=/usr/bin/systemctl start multi-user.target --no-block
 RemainAfterExit=true
 StandardError=journal+console
 
--- a/packages/release/activate-preconfigured.service
+++ b/packages/release/activate-preconfigured.service
@@ -1,13 +1,13 @@
 [Unit]
-Description=Isolates preconfigured.target
+Description=Activate preconfigured.target
 DefaultDependencies=no
 After=fipscheck.target
 Requires=fipscheck.target
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/systemctl set-default preconfigured
-ExecStart=/usr/bin/systemctl isolate default --no-block
+ExecStart=/usr/bin/systemctl set-default preconfigured.target
+ExecStart=/usr/bin/systemctl start preconfigured.target --no-block
 RemainAfterExit=true
 StandardOutput=tty
 StandardError=inherit
--- a/packages/systemd/1003-serialize-don-t-allocate-1M-on-the-stack-just-like-t.patch
+++ b/packages/systemd/1003-serialize-don-t-allocate-1M-on-the-stack-just-like-t.patch
@@ -1,0 +1,82 @@
+From 714fc5c244cd9e83e6c82b87ff032de41fe86faf Mon Sep 17 00:00:00 2001
+From: Lennart Poettering <lennart@poettering.net>
+Date: Mon, 11 Sep 2023 12:21:14 +0200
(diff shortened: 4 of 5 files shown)


arnaldo2792 (approved):
Nice! Did it fail to compile with the new flags?

bcressey:
> Nice! Did it fail to compile with the new flags?

No, it worked fine. I happened to have the backported patch in one of my local branches, and found [the upstream issue](https://github.com/systemd/systemd/issues/26950) again while looking for reports of problems with `ftrivial-auto-var-init=zero`.

KCSesh:
>Note that we rely on the implicit reload that happens when the target is changed to find the units for host and bootstrap containers. This could be an explicit reload, but that would leave the target at its default value - either preconfigured.target or fipscheck.target - which changes the output that systemd-analyze reports.

Does this mean with `systemctl isolate` we essentially had 2 implicit reloads?

KCSesh:
For testing the move from `isolate` to `start` did you also cover:
* After boot up, launching a custom host container with apiclient
* A bad setting that caused boot failure

KCSesh:
Ran some of the tests myself, and it looks good. This is very nice!

KCSesh (approved):

bcressey:
> After boot up, launching a custom host container with apiclient
```
$ apiclient apply <<EOF
> [settings.host-containers.sleep]
> enabled = true
> source = "863599026182.dkr.ecr.us-west-2.amazonaws.com/thar-be-a-host-container:latest"
> EOF

$ sudo chroot /.bottlerocket/rootfs/ systemctl status host-containers@sleep
● host-containers@sleep.service - Host container: sleep
     Loaded: loaded (/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/host-containers@.service; enabled; preset: enabled)
    Drop-In: /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/service.d
             └─00-aws-config.conf
     Active: active (running) since Mon 2025-06-02 20:35:16 UTC; 3min 10s ago

$ apiclient exec sleep bash
bash-5.2# head -n2 /etc/os-release
NAME="Amazon Linux"
VERSION="2023"
```

> A bad setting that caused boot failure
```
         Starting Bottlerocket userdata configuration system...
[    6.921543] early-boot-config[5361]: Error PATCHing '/settings?tx=bottlerocket-launch': Status 400 when PATCHing /settings?tx=bottlerocket-launch: Json deserialize error: unknown field `broken`, expected one of `cluster-name`, `cluster-certificate`, `api-server`, `node-labels`, `node-taints`, `static-pods`, `authentication-mode`, `bootstrap-token`, `standalone-mode`, `eviction-hard`, `eviction-soft`, `eviction-soft-grace-period`, `eviction-max-pod-grace-period`, `kube-reserved`, `system-reserved`, `allowed-unsafe-sysctls`, `server-tls-bootstrap`, `cloud-provider`, `registry-qps`, `registry-burst`, `event-qps`, `event-burst`, `kube-api-qps`, `kube-api-burst`, `container-log-max-size`, `container-log-max-files`, `container-log-max-workers`, `container-log-monitor-interval`, `cpu-cfs-quota-enforced`, `cpu-manager-policy`, `cpu-manager-reconcile-period`, `cpu-manager-policy-options`, `topology-manager-scope`, `topology-manager-policy`, `pod-pids-limit`, `image-gc-high-threshold-percent`, `image-gc-low-threshold-percent`, `provider-id`, `log-level`, `credential-providers`, `server-certificate`, `server-key`, `shutdown-grace-period`, `shutdown-grace-period-for-critical-pods`, `memory-manager-reserved-memory`, `memory-manager-policy`, `reserved-cpus`, `max-pods`, `cluster-dns-ip`, `cluster-domain`, `node-ip`, `pod-infra-container-image`, `hostname-override`, `hostname-override-source`, `seccomp-default`, `device-ownership-from-security-context`, `single-process-oom-kill` at line 1 column 2545
[FAILED] Failed to start Bottlerocket userdata configuration system.
See 'systemctl status early-boot-config.service' for details.
[DEPEND] Dependency failed for Bottlerocket initial configuration complete.
[DEPEND] Dependency failed for Activate configured.target.
[DEPEND] Dependency failed for User-specified setting generators.
[DEPEND] Dependency failed for Applies settings to create config files.
[DEPEND] Dependency failed for Sets the hostname.
[DEPEND] Dependency failed for Bootstrap Commands.
[DEPEND] Dependency failed for Send signal to CloudFormation Stack.
[DEPEND] Dependency failed for Generate additional settings for Kubernetes.
```

bcressey:
> Does this mean with `systemctl isolate` we essentially had 2 implicit reloads?

No - in both cases we're relying on `systemctl set-default` to do the daemon reload as a side effect. `isolate` and `start` are both working from systemd's current view of the world.

An explicit reload would be more clear, but it's also rather expensive (200 ms) and isn't worth it. I called out the implicit reload because it's weird to add a patch to speed up daemon reload when there's no daemon reload in sight. 😀

Issue #536, opened 2025-06-02
Upgrade systemd to the latest stable major version available


vigh-m:
The current systemd version provided by Bottlerocket is v252 while the latest upstream available is v257.
The latest version would allow new features to be enabled as well as keep the Bottlerocket updated for longer.

This issue will serve for tracking as well as discussions for this upgrade.

PR #537, opened 2025-06-02
systemd: namespace current systemd version

vigh-m:
**Issue number:**

#536

**Description of changes:**

This change only renames `systemd` to `systemd-252` without making any changes to the existing functionality. No changes to the underlying usage should occur.

The change are in anticipation of building `systemd-257` alongside `systemd-252` without disturbing any of the variant builds.

**Testing done:**

_Basic testing_
Build and tested a k8s-1.32 AMI and tested it with a sample workload

_Testing with all AMIs_

Tested the following variants with a basic launch and sample workload. No issues seen

- [x] aws-ecs-1
- [x] aws-ecs-2
- [x] aws-k8s-1.28
- [x] aws-k8s-1.29
- [x] aws-k8s-1.30
- [x] aws-k8s-1.31
- [x] aws-k8s-1.32

**Terms of contribution:**


packages/dbus-broker/Cargo.toml
```diff
@@ -23,4 +23,4 @@ sha512 = "711e3db030cd8d3953e966c11d2afa4c543548f882d3364fc0b79b7cfb3b8d5ecb65b1
 glibc = { path = "../glibc" }
 libexpat = { path = "../libexpat" }
 libselinux = { path = "../libselinux" }
-systemd = { path = "../systemd" }
+systemd-252 = { path = "../systemd-252" }
```
bcressey, line 26: It'd be worth digging into each of these and determining the nature of the build-time dependency.

The [libsystemd docs](https://www.freedesktop.org/software/systemd/man/latest/libsystemd.html#Interface%20stability) say:
> Strict backwards-compatibility is maintained for the API (application programming interface) and ABI (application binary interface). Symbol versioning is used, with symbols only added and never removed.

If the dependency is purely on `libsystemd.so` then this seems OK.

arnaldo2792: Correct, the dependency is only on `libsystemd.so`:

https://github.com/bus1/dbus-broker/blob/56a4e39a045aa548918cccaf24487ccfd2a8e0d7/meson.build#L77

Mostly to talk to the bus and support socket activation:

https://github.com/bus1/dbus-broker/blob/56a4e39a045aa548918cccaf24487ccfd2a8e0d7/src/launch/main.c#L104
https://github.com/bus1/dbus-broker/blob/56a4e39a045aa548918cccaf24487ccfd2a8e0d7/src/launch/launcher.c#L214

bcressey: Might not have been clear from my comment, but I was asking for review of all uses of `systemd-devel`, not just the `dbus-broker` usage specifically.

vigh-m:
I looked into the `systemd-devel` dependencies. In the process I also found the [interface portability and stability promise](https://systemd.io/PORTABILITY_AND_STABILITY/) by systemd which was an interesting read.

### `dbus-broker`

@arnaldo2792 has already shown that `dbus-broker` only uses `libsystemd`. 
The linked dependencies also show this:
```bash
[root@admin]# ldd .bottlerocket/rootfs/usr/bin/dbus-broker | grep systemd
[root@admin]# ldd .bottlerocket/rootfs/usr/bin/dbus-broker-launch | grep systemd
.bottlerocket/rootfs/usr/bin/dbus-broker-launch: /lib64/libsystemd.so.0: version `LIBSYSTEMD_243' not found (required by .bottlerocket/rootfs/usr/bin/dbus-broker-launch)
.bottlerocket/rootfs/usr/bin/dbus-broker-launch: /lib64/libsystemd.so.0: version `LIBSYSTEMD_237' not found (required by .bottlerocket/rootfs/usr/bin/dbus-broker-launch)
.bottlerocket/rootfs/usr/bin/dbus-broker-launch: /lib64/libsystemd.so.0: version `LIBSYSTEMD_221' not found (required by .bottlerocket/rootfs/usr/bin/dbus-broker-launch)
	libsystemd.so.0 => /lib64/libsystemd.so.0 (0x00007f78ac9d8000)
```

### `docker-engine`:

moby/moby depends on [go-systemd](https://pkg.go.dev/github.com/coreos/go-systemd/v22#section-readme) which includes `sd-id128` and `sd_journal` as cgo code which are libsystemd libraries.

```bash
[root@admin]# ldd .bottlerocket/rootfs/usr/bin/docker | grep systemd
[root@admin]# ldd .bottlerocket/rootfs/usr/bin/dockerd | grep systemd
	libsystemd.so.0 => /lib64/libsystemd.so.0 (0x00007f41a78af000)
[root@admin]# ldd .bottlerocket/rootfs/usr/bin/docker-proxy | grep systemd
[root@admin]#
```

### `wicked`

wicked autoconf has dependencies defined for `libsystemd` and `sd-daemon` and the code only includes `sd-daemon.h` which is a libsystem library

```bash
[root@admin]# ldd .bottlerocket/rootfs/usr/sbin/wicked | grep systemd
	libsystemd.so.0 => /lib64/libsystemd.so.0 (0x00007fe869b9f000)
[root@admin]# ldd .bottlerocket/rootfs/usr/sbin/wickedd | grep systemd
	libsystemd.so.0 => /lib64/libsystemd.so.0 (0x00007fe43653a000)
[root@admin]# ldd .bottlerocket/rootfs/usr/sbin/wickedd-nanny | grep systemd
	libsystemd.so.0 => /lib64/libsystemd.so.0 (0x00007fa6878af000)
```

### `mdadm`
`mdadm` (and `mdmon`) is the only one which does not depend on `libsystemd` but instead depends on `libudev`.

```bash
[root@admin]# ldd .bottlerocket/rootfs/usr/sbin/mdadm | grep systemd
[root@admin]# ldd .bottlerocket/rootfs/usr/sbin/mdadm | grep udev
	libudev.so.1 => /lib64/libudev.so.1 (0x00007fb1580a0000)
[root@admin]# ldd .bottlerocket/rootfs/usr/sbin/mdmon | grep systemd
[root@admin]# ldd .bottlerocket/rootfs/usr/sbin/mdmon | grep udev
	libudev.so.1 => /lib64/libudev.so.1 (0x00007f4d0428a000)
```
 From the `libudev` [manpage](https://www.man7.org/linux/man-pages/man3/libudev.3.html):
```
This library is supported, but should not be used in new projects. 
Please see [sd-device(3)](https://www.man7.org/linux/man-pages/man3/sd-device.3.html)
for an equivalent replacement with a more modern API.
```
While this isn't the same guarantees as `libsystemd` it sounds like there is no development planned for `libudev`. As long as future `systemd` versions provide `libudev.so.1` we should be safe to update. 
Additionally I don't see many references to `libudev` in the systemd changelog.

bcressey:
Thanks for the research; that all sounds good to me. `mdadm` supports a `NO_LIBUDEV` build flag, which we can explore if the dependency ever turns out to be a problem, but otherwise I'm inclined to keep it as-is.

packages/systemd-252/systemd-252.spec
```diff
@@ -126,11 +128,16 @@ Requires: %{_cross_os}libtss2
 Requires: %{_cross_os}libuuid
 Requires: %{_cross_os}libxcrypt
 
+Provides: %{_cross_os}systemd = %{package_priority_epoch}:
+Conflicts: %{_cross_os}systemd
```
bcressey, line 132: ~~nit: every occurrence of `%{_cross_os}systemd` (except the one in the actual `Name` field) can be replaced with `%{name}` to simplify, which may be worth doing if we're going to copy this spec~~

Never mind, this is fine since the goal is to have unversioned provides.

bcressey (approved):

packages/systemd-252/systemd-252.spec
vigh-m: Since `Name` is now `%{_cross_os}systemd-252`,  `%{name}` will not point to `%{_cross_os}systemd`, so I don't think this will work. I can create a new variable in the next PR where we add `systemd-257`

ytsssun (approved):
LGTM

KCSesh (approved):

PR #531, opened 2025-05-30
Third party package updates

piyush-jena:
**Issue number:**

**Description of changes:**
- fix a spacing error in CHANGELOG
- update `aws-iam-authenticator` to v0.7.2
- update `ecr-credential-provider-1.31` to v1.31.6
- update `ecr-credential-provider-1.32` to v1.32.3
- update `kubernetes-1.28 - 1.32` to latest

**Testing done:**
-  [x] `aws-iam-authenticator`
The node joins the cluster in the tests below which is a sufficient test. Skipping!

-  [x] `ecr-credential-provider`
1. Run a container from a private registry. It will fail unless you are using an image from the same account.
```
[fedora@ip-xxxxx Repositories]$ kubectl get pods -A
NAMESPACE     NAME                              READY   STATUS         RESTARTS   AGE
default       hello-nodejs                      0/1     ErrImagePull   0          3s
kube-system   aws-node-lntkq                    2/2     Running        0          27m
kube-system   coredns-5449774944-lpqqk          1/1     Running        0          30m
kube-system   coredns-5449774944-xwfn6          1/1     Running        0          30m
kube-system   kube-proxy-gbrwj                  1/1     Running        0          27m
kube-system   metrics-server-76dfb8cb4b-5b4jn   1/1     Running        0          30m
kube-system   metrics-server-76dfb8cb4b-gc8rw   1/1     Running        0          30m
```
2. Base64 encode aws credentials of the above account in this format
```
[default]
use_fips_endpoint=false

[profile ecr]
use_fips_endpoint=false
aws_access_key_id=xxxxx
aws_secret_access_key=yyyyy
aws_session_token=zzzzz
```
3. Apply the above settings in the bottlerocket instance
```
apiclient apply <<EOF
[settings.kubernetes.credential-providers.ecr-credential-provider]
enabled = true
cache-duration = "30m"
image-patterns = [
  "*.dkr.ecr.us-east-2.amazonaws.com",
  "*.dkr.ecr.us-west-2.amazonaws.com"
]

[settings.aws]
profile = "ecr"
config = "{base64 encoding of the above}"
EOF
```
4. Run the same container from a private registry. It should work now.
```
[fedora@ip-xxxxx Repositories]$ kubectl delete pod hello-nodejs
pod "hello-nodejs" deleted
[fedora@ip-xxxxx Repositories]$ kubectl apply -f hello-nodejs.yaml 
pod/hello-nodejs created
[...]

PR #545, opened 2025-06-10
Prepare for v9.0.0 release

ytsssun:
**Description of changes:**
Prepare the CHANGELOG and bump the core kit version for 9.0.0

**Terms of contribution:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,46 @@
+# v9.0.0 (2025-06-10)
+
+## OS Changes
+* Add support for default configuration file in `xfsprogs` ([#521])
+* Add support for more AWS regions in `schnauzer` and `host-ctr` ([#535])
+* Backport systemd fix to speed up `systemctl daemon-reload` ([#528])
+* Replace `systemctl isolate` with `systemctl start` ([#528])
+* Add capability markers `ioctl_skip_cloexec` and `userspace_initial_context` to SELinux policy ([#534])
+* Add `zramctl` to `util-linux` package ([#543])
+### Third Party Package Updates
+- Update to latest versions of `aws-iam-authenticator`, `ecr-credential-provider`, and `kubernetes` packages ([#531])
+- Update multiple core libraries: SELinux components, `libseccomp`, `libinih`, `libffi`, `libnftnl`, `libelf`, `liburcu`, `libglib`, and `libcap` ([#515])
+- Update core system utilities: `ethtool`, `grep`, `iproute`, `strace`, `makedumpfile`, `nvme-cli`, and `libnvme` ([#532])
+
+## Build Changes
+* Fix clippy warnings for Rust 1.87.0 ([#525])
+* Build `glibc` with frame pointers ([#527])
+* Fix various build warnings and LTO compatibility issues in packages ([#526])
+* Update `bottlerocket-sdk` from 0.61.0 to 0.62.0 ([#542])
+
+## Orchestrator changes
+
+### Kubernetes
+* Drop `kubernetes-1.26` and `ecr-credential-provider-1.26` ([#523])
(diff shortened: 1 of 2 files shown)


vigh-m (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+* Build `glibc` with frame pointers ([#527])
+* Fix various build warnings and LTO compatibility issues in packages ([#526])
+* Update `bottlerocket-sdk` from 0.61.0 to 0.62.0 ([#542])
+
+## Orchestrator changes
+
+### Kubernetes
+* Drop `kubernetes-1.26` and `ecr-credential-provider-1.26` ([#523])
+* Make `soci-snapshotter` a socket-activated systemd service ([#529])
+* Add latest instance types to eni-max-pods mapping ([#538])
```
koooosh: Non-blocking nit: `eni-max-pods` instead of eni-max-pods

koooosh (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
+# v9.0.0 (2025-06-10)
+
+## OS Changes
+* Third party package updates ([#515], [#531], [#532])
+* Add support for default configuration file in `xfsprogs` ([#521])
+* Add support for more AWS regions in `schnauzer` and `host-ctr` ([#535])
+* Backport systemd fix to speedup `daemon-reload` and replace `isolate` with `start` commands ([#528])
```
bcressey:
```suggestion
* Backport systemd fix to speed up `systemctl daemon-reload` ([#528])
* Replace `systemctl isolate` with `systemctl start` ([#528])
```

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
+# v9.0.0 (2025-06-10)
+
+## OS Changes
+* Third party package updates ([#515], [#531], [#532])
```
bcressey: It'd be good to find ways to summarize this in a little more detail - some of these "third party package updates" could be more significant than the other changes. Perhaps a new section where a longer list of packages won't be as distracting?

piyush-jena:
Should we be adding the following?
`* Update EKS 1.32 to latest ([#531])`
`* Update ecr-credential provider  1.31-1.32 to latest ([#531])`

CHANGELOG.md
ytsssun: I can try this if it is better. 
```md
### Third Party Package Updates
- Update to latest versions of `aws-iam-authenticator`, `ecr-credential-provider`, and `kubernetes` packages ([#531])
- Update multiple core libraries: SELinux components, `libseccomp`, `libinih`, `libffi`, `libnftnl`, `libelf`, `liburcu`, `libglib`, and `libcap` ([#515])
- Update core system utilities: `ethtool`, `grep`, `iproute`, `strace`, `makedumpfile`, `nvme-cli`, and `libnvme` ([#532])
```

@vigh-m also suggest about using the [collapsed component](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections) via `<details>` block.

CHANGELOG.md
ytsssun: I did not add the `` because that is what we use for past eni-max-pods updates. I can add that so that we keep using the `eni-max-pods` format.

ytsssun:
> Should we be adding the following? `* Update EKS 1.32 to latest ([#531])` `* Update ecr-credential provider 1.31-1.32 to latest ([#531])`

In previous release we explicitly call out the update when we don't have a summary entry like "third party update". I think if we figure out a way to call out the packages per @bcressey 's suggestion, we can only call it out once.

ytsssun:
Force pushed to address comments ^

bcressey (approved):

PR #546, opened 2025-06-13
libcrypto: update rpm spec to apply all patches

vigh-m:
Also refresh all the patches from upstream and change the starting number to be in line with other packages

**Description of changes:**
- Update `libcrypto` patches
- Update `%prep` to use the `%autosetup` macro and apply all the patches
- Re-number the patches to start from 1000 inline with other packages
- Add a script to generate patches from upstream

**Testing done:**

Built and tested a k8s-1.33 ami and ran a sample pod

**Terms of contribution:**


packages/libcrypto/libcrypto.spec
```diff
@@ -18,11 +18,20 @@ Source0: https://github.com/aws/aws-lc/archive/AWS-LC-FIPS-%{version}/aws-lc-AWS
 # git clone https://github.com/aws/aws-lc
 # cd aws-lc
 # git checkout origin/fips-2024-09-27
-# git format-patch --no-numbered AWS-LC-FIPS-3.0.0..
+# git format-patch --start-number=1001 AWS-LC-FIPS-3.0.0..
```
bcressey: Let's improve this in a couple of ways to make it more repeatable by others:
1. `git format-patch --start-number=1001 --no-numbered --no-signature` to avoid adding the Git versions to the end, and to capture the "no numbered" argument you're using.
2. Add an `awk` command that removes the problematic lines from `Patch1009`.

arnaldo2792: We should keep the `--no-numbered` flag, otherwise folks whose `git` isn't configured to use `--no-numbered` by default may regenerate the patches with the numbered subjects.

vigh-m:
🆙 Regenerated the patches without git signature and patch numbers. 

Working on the `awk` bit to resolve the problems with `Patch1009`

packages/libcrypto/libcrypto.spec
vigh-m: Makes sense. I'll follow up with the changes

vigh-m: Yup! I've update the command and added the flags Ben suggested as well

vigh-m:
⬆️ Added a small script to generate the patches and remove the problem lines from `Patch1009`

vigh-m:
⬆️ Pushed changes to the script to remove a bit more of the commit message for a cleaner patch. The `awk` command is also simpler now

bcressey (approved):
Nice!

ytsssun (approved):

KCSesh (approved):

arnaldo2792 (approved):

PR #548, opened 2025-06-13
configuration-files: add option for write-once on boot

### LINKED ISSUE #547: Allow configuration-files to only render on boot
**What I'd like:**

I would like for the ability to designate certain configuration-files in a Bottlerocket variant to be written only on first boot. Settings that are rendered in these configuration files should still change and that change should be reflected in Bottlerocket's data store, but actual writing of the rendered setting to these configuration files shouldn't occur until reboot.

This could look something like:

```
[configuration-files.motd]
path = "/etc/motd"
template-path = "/usr/share/templates/motd"
write-mode = "once"
```

Keep the status quo.

ginglis13:
**Issue number:**

Closes #547 

**Description of changes:**

Add a new option for configuration-files to be written only once (on boot). Changes to settings rendered in these files will reflect in Bottlerocket's datastore but will not take effect until reboot.

**Testing done:**

Built an aws-dev variant with `motd`'s configuration file set to write once:

```
[configuration-files.motd]
path = "/etc/motd"
template-path = "/usr/share/templates/motd"
write-mode = "once"
```

Launch an instance with userdata:

```
[settings]
motd = "only once"
```

Setting was rendered:

```
bash-5.1# apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "82b61407-dirty",
    "pretty_name": "Bottlerocket OS 1.40.0 (aws-dev)",
    "variant_id": "aws-dev",
    "version_id": "1.40.0"
  }
}
bash-5.1# apiclient get settings.motd
{
  "settings": {
    "motd": "only once"
  }
}
bash-5.1# cat /etc/motd
only once
```

Modify setting, ensure `/etc/motd` was not changed, reboot, ensure `/etc/motd` has the new setting value:

```
bash-5.1# apiclient set settings.motd="not until reboot"
bash-5.1# apiclient get settings.motd
{
  "settings": {
    "motd": "not until reboot"
  }
}
# Actual configuration file still reflects original value
bash-5.1# cat /etc/motd
only once

bash-5.1# apiclient reboot

...

bash-5.1# apiclient get settings.motd
{
  "settings": {
    "motd": "not until reboot"
  }
}
# Actual configuration file still reflects original value
bash-5.1# cat /etc/motd
not until reboot
```

**Terms of contribution:**

### DIFF
--- a/sources/api/thar-be-settings/src/config.rs
+++ b/sources/api/thar-be-settings/src/config.rs
@@ -61,6 +61,13 @@ pub async fn render_config_files(
     // Go write all the configuration files from template
     let mut rendered_configs = Vec::new();
     for (name, metadata) in config_files {
+        if !metadata.should_render() {
+            info!(
+                "File {} already exists and overwrite-path-if-present=false, skipping render of config file",
+                &metadata.path,
+            );
+            continue;
+        }
         debug!("Rendering {}", &name);
 
         let try_rendered =
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -31,7 +31,7 @@ use bottlerocket_release::BottlerocketRelease;
 use bottlerocket_settings_models::model_derive::model;
 use bottlerocket_settings_plugin::BottlerocketSettings;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::{collections::HashMap, path::Path};
 
 use bottlerocket_settings_models::modeled_types::SingleLineString;
 
@@ -77,6 +77,16 @@ struct ConfigurationFile {
     template_path: SingleLineString,
     #[serde(skip_serializing_if = "Option::is_none")]
     mode: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    overwrite_path_if_present: Option<bool>,
+}
+
+impl ConfigurationFile {
+    /// Checks whether the configuration file should be written if the path is present.
+    /// and overwrite_path_if_present is set.
+    pub fn should_render(&self) -> bool {
+        self.overwrite_path_if_present != Some(false) || !Path::new(&self.path.as_ref()).exists()
+    }
 }
 
 ///// Metadata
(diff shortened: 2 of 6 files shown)


ginglis13:
^ force push adds a unit test for new write-mode behavior

sources/api/thar-be-settings/src/config.rs
```diff
@@ -142,18 +153,26 @@ pub struct RenderedConfigFile {
...
             path: PathBuf::from(&path),
             rendered,
             mode: mode.to_owned(),
+            write_mode: write_mode.to_owned(),
         }
     }
 
-    /// Writes the rendered template at the proper location
+    /// Writes the rendered template at the proper location. If the template is configured to write once
+    /// and the destination path for the rendered template already exists, skip the write.
```
bcressey: This comment change would be better applied to `write_config_files`, since `write_to_disk` does not have this logic.

sources/models/src/lib.rs
```diff
@@ -77,6 +77,15 @@ struct ConfigurationFile {
     template_path: SingleLineString,
     #[serde(skip_serializing_if = "Option::is_none")]
     mode: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    write_mode: Option<ConfigurationFileWriteMode>,
```
bcressey: `write-mode` is confusingly similar to `mode`.

Maybe one of these to remove that ambiguity, and to phrase it in terms of the configuration file being operated upon, rather than the behavior of the program doing the writes?
```
clobber-path-if-exists = [ true | false ]
overwrite-path-if-present = [ true | false ]
```

ginglis13: Gonna go with `overwrite-path-if-present` 👍

ginglis13:
^ force push refactors `write_mode` to `overwrite_path_if_present`. added a case for `None` to unit test. manually retested, details below

<details>

```
bash-5.1# cat /var/lib/bottlerocket/datastore/v1.40.0_eIQo42DbfcwvvoUX/live/configuration-files/motd/overwrite-path-if-present
false
bash-5.1# apiclient get settings.motd
{
  "settings": {
    "motd": "do not overwrite"
  }
}
bash-5.1# cat /etc/motd
do not overwrite
bash-5.1# apiclient set settings.motd="new motd"
bash-5.1# apiclient get settings.motd
{
  "settings": {
    "motd": "new motd"
  }
}
bash-5.1# cat /etc/motd
do not overwrite
```

reboot, check reset
```
apiclient reboot

...

bash-5.1# cat /etc/motd
new motd
bash-5.1# apiclient get settings.motd
{
  "settings": {
    "motd": "new motd"
  }
}
```
</details>

sources/api/thar-be-settings/src/config.rs
```diff
@@ -89,9 +91,17 @@ pub async fn render_config_files(
     Ok(rendered_configs)
 }
 
-/// Write all the configuration files to disk
+/// Write all the configuration files to disk. If overwrite_path_if_present is false
+/// and the file already exists, the write will be skipped.
 pub fn write_config_files(rendered_configs: &[RenderedConfigFile]) -> Result<()> {
     for cfg in rendered_configs {
+        if cfg.overwrite_path_if_present == Some(false) && Path::new(&cfg.path).exists() {
+            info!(
```
cbgbt: nit: we've defined the default value here (`true`) when the type is defined in the `models` crate.

I'd prefer to see us have an accessor `should_overwrite_path_if_present() -> bool` on the `ConfigFile` model that clarifies the default value next to the model itself, otherwise you risk having it diverge if the abstraction is re-used.

cbgbt (approved):

sources/api/thar-be-settings/src/config.rs
ginglis13: At this point the config has already been rendered to a `RenderedConfigFile` - I could either

1. add `should_overwrite_path_if_present() -> bool` on `model::ConfigurationFile` and call this method when rendering ConfigurationFiles to RenderedConfigFile, skipping rendering when true

I'm not a fan of this option as the option conveys action writing the config file, not rendering the config file

2. add `should_overwrite_path_if_present() -> bool` on `RenderedConfigFile` - I prefer this option since we'll be using the method to skip writes - not skip renders, like in (1.)

ginglis13:
^ force push adds `should_overwrite_path_if_present() -> bool` to `RenderedConfigFile`

sources/api/thar-be-settings/src/config.rs
cbgbt: Hmm, I see what you mean and I agree that the semantics imply what you are saying.

On the flip side, rendering templates is not free and we should avoid doing it if possible. I'd prefer to avoid the rendering.

Perhaps there's a rephrasing of the option that makes the reality match the description more cleanly... But barring that, I think it's OK to log that we've skipped rendering because the target exists.

ginglis13: I'm happy to move this change in the direction of skipping a render. On second thought, I think the semantics of "If I should not overwrite the path if present, then I will skip rendering the template" has the same outcome that the option implies, so I'm okay to shift this logic to `model::ConfigurationFile` and skip the rendering

ginglis13:
^ force push:

* moves `should_overwrite_path_if_present() -> bool` to `model::ConfigurationFile`
* uses this method in processing templates to render, rather than templates to write, in `render_config_files` and skips rendering `if !metadata.should_overwrite_path_if_present()`
* moves units tests over to models
* removes `overwrite_path_if_present` from the RenderedConfig struct

cbgbt (approved):

sources/api/thar-be-settings/src/config.rs
```diff
@@ -61,6 +61,13 @@ pub async fn render_config_files(
     // Go write all the configuration files from template
     let mut rendered_configs = Vec::new();
     for (name, metadata) in config_files {
+        if !metadata.should_overwrite_path_if_present() {
+            info!(
+                "File {} already exists and overwrite-path-if-present=false, skipping render of config file ",
```
bcressey: nit:
```suggestion
                "File {} already exists and overwrite-path-if-present=false, skipping render of config file",
```

sources/models/src/lib.rs
```diff
@@ -77,6 +77,16 @@ struct ConfigurationFile {
...
     mode: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    overwrite_path_if_present: Option<bool>,
+}
+
+impl ConfigurationFile {
+    /// Checks whether the configuration file should be written if the path is present.
+    /// and overwrite_path_if_present is set.
+    pub fn should_overwrite_path_if_present(&self) -> bool {
+        self.overwrite_path_if_present != Some(false) || !Path::new(&self.path.as_ref()).exists()
```
bcressey: nit: calling this `should_render` might be more clear - it's kind of weird to return `true` from a function called `should_overwrite_path_if_present` when the path is not present

bcressey (approved):

ginglis13:
^ rename method to `should_render`, fix whitespace issue

bcressey (approved):

cbgbt (approved):

PR #558, opened 2025-06-24
Add SBOM generation to Core Kit packages

rpkelly:
**Issue number:**

Closes #

**Description of changes:**
Adds macros required to generate SBOMs to each package in the core kit. Must not be merged until the updated SDK containing the new macro is also merged.

**Testing done:**
ran `make` locally and validated the presence of the new SBOM files.

**Terms of contribution:**

### DIFF
--- a/packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
+++ b/packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
@@ -88,6 +88,8 @@ gofips build "${ECS_CNI_BUILD_ARGS[@]}" -o fips/ecs-ipam ./plugins/ipam
 go build "${ECS_CNI_BUILD_ARGS[@]}" -o ecs-bridge ./plugins/ecs-bridge
 gofips build "${ECS_CNI_BUILD_ARGS[@]}" -o fips/ecs-bridge ./plugins/ecs-bridge
 
+%cross_generate_sbom
+
 %install
 install -d %{buildroot}%{_cross_libexecdir}
 install -D -p -m 0755 ecs-bridge %{buildroot}%{_cross_libexecdir}/cni/ecs/ecs-bridge
@@ -112,9 +114,13 @@ done
 
 %cross_scan_attribution go-vendor vendor
 
+%cross_install_sbom
+
 %files
 %{_cross_attribution_file}
 %{_cross_attribution_vendor_dir}
+%{_cross_sbom_package_dir}/%{name}-spdx.json
+%{_cross_sbom_package_dir}/%{name}-cyclonedx.json
 %license LICENSE
 
 %files bin
--- a/packages/aws-otel-collector/aws-otel-collector.spec
+++ b/packages/aws-otel-collector/aws-otel-collector.spec
@@ -51,6 +51,8 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
 go build -ldflags "${GOLDFLAGS}" -o aws-otel-collector ./cmd/awscollector
 gofips build -ldflags "${GOLDFLAGS}" -o fips/aws-otel-collector ./cmd/awscollector
 
+%cross_generate_sbom
+
 %install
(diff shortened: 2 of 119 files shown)


cbgbt:
Is it possible for us to make this opt-out rather than opt-in?

For example, we have [a flag that can be modified at the package spec level to disable FIPS checks](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/529ff58452faa9064b727a7e011dcbaca97c1d6d/macros/shared#L338)

Can we modify the existing macro structure to generate SBOM by default, and allow packages to disable it?

cbgbt:
I spoke to @rpkelly offline who mentioned that he'd look into an "opt-out-by-default" experience. He also mentioned, though, that we'd need to update the `%files` section of most RPMs regardless.

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -112,9 +114,13 @@ done
 
 %cross_scan_attribution go-vendor vendor
 
+%cross_install_sbom
```
bcressey, line 117: This should be wired into `%__arch_install_post` like `%cross_check_fips` and `%cross_generate_attribution`  so that not every package has to explicitly add them.

packages/amazon-ecs-cni-plugins/amazon-ecs-cni-plugins.spec
```diff
@@ -88,6 +88,8 @@ gofips build "${ECS_CNI_BUILD_ARGS[@]}" -o fips/ecs-ipam ./plugins/ipam
 go build "${ECS_CNI_BUILD_ARGS[@]}" -o ecs-bridge ./plugins/ecs-bridge
 gofips build "${ECS_CNI_BUILD_ARGS[@]}" -o fips/ecs-bridge ./plugins/ecs-bridge
 
+%cross_generate_sbom
```
bcressey, line 91: It looks like this could go into `%__spec_build_post`

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -64,11 +66,15 @@ install -d %{buildroot}{%{_cross_bindir},%{_cross_fips_bindir}}
...
 
+%cross_install_sbom
+
 %files
 %{_cross_attribution_file}
 %{_cross_unitdir}/aws-otel-collector.service
 %{_cross_tmpfilesdir}/aws-otel-collector-tmpfiles.conf
 %{_cross_factorydir}%{_cross_sysconfdir}/aws-otel-collector.yaml
+%{_cross_sbom_package_dir}/%{name}-spdx.json
+%{_cross_sbom_package_dir}/%{name}-cyclonedx.json
```
bcressey, line 77: I would recommend standardizing on just the directory, like we do with `%{_cross_attribution_vendor_dir}`:
```suggestion
%{_cross_sbom_package_dir}
```

packages/aws-otel-collector/aws-otel-collector.spec
```diff
@@ -64,11 +66,15 @@ install -d %{buildroot}{%{_cross_bindir},%{_cross_fips_bindir}}
...
 install -p -m 0755 fips/aws-otel-collector %{buildroot}%{_cross_fips_bindir}
 
+%cross_install_sbom
+
 %files
 %{_cross_attribution_file}
 %{_cross_unitdir}/aws-otel-collector.service
 %{_cross_tmpfilesdir}/aws-otel-collector-tmpfiles.conf
 %{_cross_factorydir}%{_cross_sysconfdir}/aws-otel-collector.yaml
+%{_cross_sbom_package_dir}/%{name}-spdx.json
```
bcressey, line 76: It's going to be gnarly, but to avoid breaking compatibility with all existing specs when the macros to generate and install SBOM files are enabled by default, you could follow the pattern of `%_debuginfo_template` and auto-generate a "-sbom" subpackage which then automatically includes these files.

That subpackage could then add a `Supplements: %{name}` field to set a reverse dependency on the main package to trigger its installation.

PR #560, opened 2025-06-25
Deprecate wicked network stack

### LINKED ISSUE #559: Deprecate the wicked networking stack
**What I'd like:**

Deprecate and remove the wicked networking stack from Bottlerocket core-kit.

Wicked is now legacy and only used by deprecated variants (`aws-ecs-1` and `aws-k8s-1.27*`). All modern Bottlerocket variants have migrated to systemd-networkd as discussed in #2449.

This deprecation should align with the timeline for dropping the legacy variants that still depend on wicked.

- Keep wicked indefinitely: Not preferred as it adds maintenance burden for legacy functionality.

ytsssun:
**Issue number:**

Closes #559

**Note:** We cannot merge this until the depending variants (`aws-ecs-1`, `aws-k8s-1.27*`, etc) are removed.

**Description of changes:**

This PR removes the wicked networking stack.

- Remove wicked package and dependencies from build system
- Remove wicked-specific code from netdog (wicked module, CLI commands)
- Update network configuration generation for systemd-networkd only
- Drop the depending libraries:
  - libiw
  - libdbus
- Remove wicked references from SELinux policy files

This simplifies our networking stack by using systemd-networkd (already included with systemd), reducing system footprint and eliminating potential conflicts between network management systems.

**Testing done:**

__1. Netdog parsing and configuration generation:__ Multiple `net.toml` configurations were tested on metal-dev variant to verify netdog's parsing capabilities and systemd-networkd code paths:

- Basic interface configuration with DHCP:

```toml
version = 3

[enp0s16]
dhcp4 = true
primary = true
```

- Bond configuration with monitoring parameters:

```toml
version = 3

[bond0]
kind = "bond"
mode = "active-backup"
interfaces = ["enp0s16"]
dhcp4 = true
primary = true

[bond0.monitoring]
miimon-frequency-ms = 100
miimon-updelay-ms = 200
miimon-downdelay-ms = 200
```

- VLAN configuration:

```toml
version = 3

[enp0s16]
dhcp4 = true

[vlan100]
kind = "vlan"
device = "enp0s16"
id = 100
dhcp4 = true
primary = true
```

All configurations were successfully parsed by netdog and properly converted to systemd-networkd configuration files. Node booted and all reached multi-user target with correct network access.

__2. Network connectivity verification:__

Confirmed systemd-networkd is active and managing network configuration:

```javascript
bash-5.2# systemctl status systemd-networkd
● systemd-networkd.service - Network Configuration
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/systemd-networkd.service; enabled; preset: enabled)
     Active: active (running) since Tue 2025-06-24 20:06:00 UTC; 2min 39s ago
     Status: "Processing requests..."
```

Verified DHCP IP assignment and interface status:

```javascript
bash-5.2# ip addr
2: enp0s16: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:12:34:56 brd ff:ff:ff:ff:ff:ff
    inet 10.0.2.15/24 metric 1024 brd 10.0.2.255 scope global enp0s16
       valid_lft forever preferred_lft forever
```

Confirmed routing table and gateway configuration:

```javascript
bash-5.2# ip route
default via 10.0.2.2 dev enp0s16 proto dhcp src 10.0.2.15 metric 1024 
10.0.2.0/24 dev enp0s16 proto kernel scope link src 10.0.2.15 metric 1024 
```

Tested external connectivity:

```javascript
bash-5.2# ping -c 4 8.8.8.8
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=255 time=8.32 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=255 time=8.03 ms
64 bytes from 8.8.8.8: icmp_seq=3 ttl=255 time=8.26 ms
64 bytes from 8.8.8.8: icmp_seq=4 ttl=255 time=9.93 ms

--- 8.8.8.8 ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3006ms
```

Verified systemd-networkd configuration files generated correctly:

```javascript
bash-5.2# cat /etc/systemd/network/10-enp0s16.network
[Match]
Name=enp0s16
[Link]
RequiredForOnline=true
[Network]
DHCP=ipv4
KeepConfiguration=dhcp
[DHCPv4]
UseMTU=true
```

__3. Variant compatibility testing:__

- Successfully launched and tested `aws-k8s-1.28` variant with no networking issues.

**Terms of contribution:**


ytsssun:
Fix format with `cargo fmt`.

packages/netdog/netdog.spec
```diff
@@ -21,8 +21,7 @@ Source20: 00-resolved.conf
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}hostname-reverse-dns
 Requires: (%{_cross_os}hostname-imds if %{_cross_os}variant-platform(aws))
-Requires: (%{_cross_os}netdog-systemd-networkd if %{_cross_os}image-feature(systemd-networkd))
-Requires: (%{_cross_os}netdog-wicked if %{_cross_os}image-feature(no-systemd-networkd))
+Requires: %{_cross_os}netdog-systemd-networkd
```
vigh-m: Do we still need to build a `netdog-systemd-networkd` subpackage? Can it be simplified to just be `netdog`?

sources/netdog/src/dns.rs
```diff
@@ -277,82 +200,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "wicked")]
-    fn dns_from_lease_file() {
-        let lease_path = test_data().join("leaseinfo.eth0.dhcp.ipv4");
```
yeazelm, line 282: There is a `test_data` directory that should have all the wicked files removed as well.

packages/netdog/netdog.spec
ytsssun: Good point. We can indeed drop that subpackage now that there is only systemd-networkd backed network stack. 

I kept that initially for minimal change to the codebase, also considering that we might add new network stack in the future. 

But I overall agree with this as it makes the code even cleaner.

sources/netdog/src/dns.rs
ytsssun: Thanks for pointing that out. I just removed the test_data that is related to the wicked.

ytsssun:
Rebased and addressed comments from @yeazelm and @vigh-m:
1. Drop the extra files `test_data` that are for wicked unit tests.
2. Drop netdog-systemd-networkd subpackage and consolidate it within the single netdog package.

yeazelm (approved):
We have a few lingering references to wicked in the comments but at this point they all tend to describe behavior choices made because `wicked` was a supported backend. I think this looks great and we can always tidy up the docs later but for now they help clarify why behavior might be a certain way (like bond/vlan support or static routing with dhcp).

KCSesh:
Changes LGTM - we could consider adding the removal of `libiw` and `libdbus` in this change.

KCSesh (approved):

ytsssun:
Pushed change to drop the libraries that only wicked depends on:
- libiw
- libdbus

Also included the change from https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/607/files which remove the reference of wicked from the SELinux policies.

bcressey (approved):

KCSesh (approved):

yeazelm (approved):

PR #561, opened 2025-06-27
schnauzer: use multicall binary for v1 and v2

cbgbt:
**Description of changes:**

    This change aims to reduce the disk utilization of schnauzer by
    joining the v1 and v2 binaries into a single coreutils-style
    multicall binary.

**Testing done:**
* Built core-kit
* Built aws-k8s-1.32 variant and confirmed symlink existence and dispatch behavior

**Terms of contribution:**


yeazelm (approved):

sources/api/schnauzer/tests/multicall_integration.rs
```diff
@@ -0,0 +1,230 @@
...
+fn test_direct_schnauzer_v2_invocation() {
+    // Given A multicall binary copied with schnauzer-v2 name
+    let test = MulticallTest::new();
+    let v2_binary_path = test.create_copy("schnauzer-v2");
+
+    // When Invoking the renamed binary with --help
+    let output = test.run_command(&v2_binary_path, &["--help"]);
+
+    // Then v2 should handle --help differently than v1
+    assert!(output.status.success() || output.status.code() == Some(2));
```
bcressey: Since `USAGE_ERROR_EXIT_CODE` is `2`, this ends up checking essentially the same conditions as "v1".

sources/api/schnauzer/tests/multicall_integration.rs
```diff
@@ -0,0 +1,230 @@
...
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    assert!(stderr.contains("Usage:"));
+}
+
+/// Test that the multicall binary can be invoked directly as schnauzer-v2
+#[test]
+fn test_direct_schnauzer_v2_invocation() {
+    // Given A multicall binary copied with schnauzer-v2 name
+    let test = MulticallTest::new();
+    let v2_binary_path = test.create_copy("schnauzer-v2");
```
bcressey: I'm not sure these "direct" tests add much over the `test_symlink_dispatch` test case. Seems like it's just "copy of binary" vs "symlink to binary".

sources/api/schnauzer/tests/multicall_integration.rs
```diff
@@ -0,0 +1,230 @@
...
+    let schnauzer_link = test.create_symlink("schnauzer");
+    let schnauzer_v2_link = test.create_symlink("schnauzer-v2");
+
+    // When Executing both symlinks with --help
+    let output_v1 = test.run_command(&schnauzer_link, &["--help"]);
+    let output_v2 = test.run_command(&schnauzer_v2_link, &["--help"]);
+
+    // Then Both should execute and produce different outputs
+    assert!(output_v1.status.code().is_some());
+    assert!(output_v2.status.code().is_some());
```
bcressey: This isn't a very useful check; the status code will always be present unless the process gets killed by a signal.

sources/api/schnauzer/tests/multicall_integration.rs
```diff
@@ -0,0 +1,230 @@
...
+    let output = test.run_command(&v2_binary_path, &["--help"]);
+
+    // Then v2 should handle --help differently than v1
+    assert!(output.status.success() || output.status.code() == Some(2));
+    assert!(test.is_v2_output(&output));
+}
+
+/// Test symlink behavior - the core multicall functionality
+#[test]
+fn test_symlink_dispatch() {
```
bcressey: AFAICT this is the only useful integration test, the others are just repeating this.

sources/api/schnauzer/tests/multicall_integration.rs
```diff
@@ -0,0 +1,230 @@
...
+        "schnauzer-v2 should route to v2"
+    );
+
+    // The outputs should be different, indicating different code paths
+    let stderr_v1 = String::from_utf8_lossy(&output_v1.stderr);
+    let stderr_v2 = String::from_utf8_lossy(&output_v2.stderr);
+    assert_ne!(
+        stderr_v1, stderr_v2,
+        "v1 and v2 should produce different outputs"
+    );
```
bcressey, line 90: The additional checks are only useful because `is_v2_output` just negates the result of `is_v1_output`. 

If `is_v2_output` actually looked for v2-specific output, like v1 does, then this test could more concisely be written as:
```
    assert!(
        test.is_v1_output(&output_v1),
        "schnauzer should route to v1"
    );

    assert_ne!(
        test.is_v2_output(&output_v1),
        "v1 output should not be classified as v2"
    );

    assert!(
        test.is_v2_output(&output_v2),
        "schnauzer-v2 should route to v2"
    );

    assert_ne!(
        test.is_v1_output(&output_v2),
        "v2 output should not be classified as v1"
    );
```

sources/api/schnauzer/src/v1.rs
```diff
@@ -157,6 +157,168 @@ pub fn build_template_registry() -> Result<handlebars::Handlebars<'static>> {
...
 }
 
+pub mod cli {
+    //! CLI module for schnauzer v1
+    use snafu::{ensure, OptionExt, ResultExt};
+    use std::collections::HashMap;
+    use std::string::String;
+    use std::{env, process};
+
+    // Setting generators do not require dynamic socket paths at this moment.
```
bcressey: nit: I don't understand this comment's relevance even though it was in the original also

bcressey (commented):
Code changes LGTM. Most of my concerns are around the integ test, which appears robust on the surface but ends up not checking very much. Better to trim it down so people don't get a false sense of confidence when modifying the code.

cbgbt:
^ removes redundancies in tests suggested by @bcressey

cbgbt:
^ rebase atop develop

bcressey (approved):

PR #562, opened 2025-06-27
packages: update kmod to v34.2

piyush-jena:
**Description of changes:**
1. Updates `kmod` package to v34.2.
2. Modify spec file to support `meson` build.
  - The first patch adds support for static builds.
  - The second patch creates the pkgconfig files in `/usr/lib/pkgconfig` instead of `/usr/share/pkgconfig`

**Testing done:**
1. `aws-dev` and `vmware-dev` instances boot.
2. Made sure that kmod is statically linked
```
  #12 9.747 + readelf -d /home/builder/rpmbuild/BUILD/bottlerocket-kmod-34.2-build/BUILDROOT/x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/kmod
  #12 9.747 ~/rpmbuild/BUILD/bottlerocket-kmod-34.2-build/kmod-34.2
  #12 9.753 
  #12 9.753 There is no dynamic section in this file.
  #12 9.754 + exit 1
```
    

**Terms of contribution:**

### DIFF
--- a/packages/kmod/0001-meson-add-support-for-static-builds.patch
+++ b/packages/kmod/0001-meson-add-support-for-static-builds.patch
@@ -1,0 +1,34 @@
+From b5d3e27e19101712bab37695283c2d8bc3960b28 Mon Sep 17 00:00:00 2001
+From: Piyush Jena <jepiyush@amazon.com>
+Date: Tue, 8 Jul 2025 22:11:12 +0000
+Subject: [PATCH] meson: add support for static builds
+
+Signed-off-by: Piyush Jena <jepiyush@amazon.com>
+---
+ meson.build | 6 ++++++
+ 1 file changed, 6 insertions(+)
+
+diff --git a/meson.build b/meson.build
+index a0cf675..4b16d6a 100644
+--- a/meson.build
++++ b/meson.build
+@@ -465,10 +465,16 @@ kmod_sources = files(
+   'tools/static-nodes.c',
+ )
+ 
++kmod_link_args = []
++if get_option('default_library') == 'static'
++  kmod_link_args += ['-static']
++endif
++
+ kmod = executable(
+   'kmod',
+   kmod_sources,
+   link_with : [libshared, libkmod_internal],
++  link_args : kmod_link_args,
+   gnu_symbol_visibility : 'hidden',
+   build_by_default : get_option('tools'),
+   install : get_option('tools'),
+-- 
+2.47.1
+
--- a/packages/kmod/0002-meson-create-pkgconfig-files-in-default-path.patch
+++ b/packages/kmod/0002-meson-create-pkgconfig-files-in-default-path.patch
@@ -1,0 +1,25 @@
+From ade8d073f25fd5bbf2ceb2077dec0dd442672144 Mon Sep 17 00:00:00 2001
+From: Piyush Jena <jepiyush@amazon.com>
+Date: Tue, 8 Jul 2025 22:12:47 +0000
+Subject: [PATCH] meson: create pkgconfig files in default path
+
+Signed-off-by: Piyush Jena <jepiyush@amazon.com>
+---
+ meson.build | 1 -
+ 1 file changed, 1 deletion(-)
+
+diff --git a/meson.build b/meson.build
+index 4b16d6a..3b59d5e 100644
+--- a/meson.build
++++ b/meson.build
+@@ -501,7 +501,6 @@ endif
+ pkg.generate(
+   name : 'kmod',
+   description : 'Tools to deal with kernel modules',
+-  install_dir : datadir / 'pkgconfig',
+   unescaped_variables : _kmod_unescaped_variables,
+   variables : _kmod_variables,
+ )
+-- 
+2.47.1
+
--- a/packages/kmod/kmod.spec
+++ b/packages/kmod/kmod.spec
@@ -1,5 +1,5 @@
 Name: %{_cross_os}kmod
-Version: 33
+Version: 34.2
 Release: 1%{?dist}
 Summary: Tools for kernel module loading and unloading
 License: GPL-2.0-or-later AND LGPL-2.1-or-later
@@ -7,6 +7,8 @@ URL: http://git.kernel.org/?p=utils/kernel/kmod/kmod.git;a=summary
 Source0: https://www.kernel.org/pub/linux/utils/kernel/kmod/kmod-%{version}.tar.xz
 Source1: https://www.kernel.org/pub/linux/utils/kernel/kmod/kmod-%{version}.tar.sign
 Source2: gpgkey-EAB33C9690013C733916AC839BA2A5A630CBEA53.asc
+Patch1001: 0001-meson-add-support-for-static-builds.patch
+Patch1002: 0002-meson-create-pkgconfig-files-in-default-path.patch
 BuildRequires: %{_cross_os}glibc-devel
 BuildRequires: %{_cross_os}libz-devel
 BuildRequires: %{_cross_os}libzstd-devel
@@ -19,6 +21,8 @@ Requires: %{_cross_os}libzstd
 %package devel
 Summary: Files for development using the tools for kernel module loading and unloading
 Requires: %{name}
+Requires: %{_cross_os}libz-devel
+Requires: %{_cross_os}libzstd-devel
 
 %description devel
 %{summary}.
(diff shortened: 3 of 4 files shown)


packages/kmod/kmod.spec
```diff
@@ -29,44 +32,87 @@ Requires: %{name}
...
-  --with-zstd \
-  --without-openssl \
-  --disable-manpages
-
-%make_build LDFLAGS="-all-static"
-
-popd
+%define cross_meson_dynamic \
+  %write_cross_compilation_conf \
+  %set_cross_build_flags \
```
bcressey: Can you do this without copy/pasting the `%{cross_meson}` macro? You should be able to extend it like:
```
%{cross_meson} \
  --default-library=static \
  --prefer-static \
  "${CONFIGURE_OPTS[@]}"
```

If it complains about `--default-library` being defined twice, you can maybe redefine it like:
```
%define cross_meson_static %(echo %{cross_meson} | sed 's,--default-library=shared,--default-library=static,')
```

piyush-jena: The first suggestion worked. 
```
%{cross_meson} \
  --default-library=static \
  --prefer-static \
  "${CONFIGURE_OPTS[@]}"
  ```
It seems like this overrides it. I tested on `aws-dev`. I will work on further testing.

**Comment by piyush-jena** [comment]
^ In the above force push I removed the redirection of sbin binaries to bin directory. This was introduced in a recent change post meson build. I tested it and it doesn't affect us adversely in any way.

File: packages/kmod/kmod.spec [hunk 2]
```diff
@@ -30,61 +33,52 @@ cp COPYING COPYING.LGPL
...
+%global _cross_vpath_builddir dynamic-build
+%cross_meson_install
 
 pushd static-build
-install -p tools/kmod %{buildroot}%{_cross_bindir}
-
-install -d %{buildroot}%{_cross_sbindir}
-ln -s ../bin/kmod %{buildroot}%{_cross_sbindir}/modprobe
+install -d %{buildroot}%{_cross_bindir}
+ln -s kmod %{buildroot}%{_cross_bindir}/modprobe
```
bcressey: This doesn't seem to be installing the static build, it's just creating a symlink to `kmod` from the previous install.

packages/kmod/kmod.spec
```diff
@@ -30,61 +33,52 @@ cp COPYING COPYING.LGPL
...
+  "${CONFIGURE_OPTS[@]}"
+export LDFLAGS="-all-static"
+%cross_meson_build
 
 %install
-pushd dynamic-build
-%make_install
-popd
+%global _cross_vpath_builddir dynamic-build
+%cross_meson_install
```
bcressey: This would make the dynamic part look more like the static part below:
```
pushd dynamic-build
%ninja_install
popd
```

packages/kmod/kmod.spec
```diff
@@ -30,61 +33,52 @@ cp COPYING COPYING.LGPL
...
-%{_cross_bindir}/rmmod
-%{_cross_sbindir}/modprobe
 %{_cross_libdir}/*.so.*
-%exclude %{_cross_datadir}/bash-completion
+%{_cross_sbindir}/depmod
+%{_cross_sbindir}/insmod
+%{_cross_sbindir}/lsmod
+%{_cross_sbindir}/modinfo
+%{_cross_sbindir}/modprobe
+%{_cross_sbindir}/rmmod
```
bcressey: I'd prefer not moving these to `/usr/sbin`, the future is `/usr/bin` 😀

packages/kmod/0001-meson-provide-the-linker-path-to-dependent-libraries.patch
```diff
@@ -0,0 +1,99 @@
...
+   cdata.set10('ENABLE_' + opt.to_upper(), have)
+@@ -334,6 +349,11 @@ have = dep.found()
+ 
+ if have
+   module_signatures = 'PKCS7 legacy'
++  if libdir_path != ''
++    private_libs += ['-L' + libdir_path, cc.find_library('crypto', required: true)]
++  else
++    private_libs += cc.find_library('crypto', required: true)
++  endif
```
bcressey: `kmod` shouldn't depend on `libcrypto` so this is confusing

bcressey:
> Due to the new changes, the generated libkmod.pc file had the dependent libraries under Requires.private field which forced us to add zlib-devel and zstd-devel as requirements in systemd even though they are not its direct requirements. The patch adds the path to these libraries to the libkmod.pc itself under Libs.private field.

I'm not sure about this rationale. I agree we want to avoid propagating indirect "devel" dependencies to other packages, but currently I've found the best way to do that is just to add those requires to the devel package:
```
%package devel
Summary: Files for development using the tools for kernel module loading and unloading
Requires: %{name}
Requires: %{_cross_os}libz-devel
Requires: %{_cross_os}libzstd-devel
```

(The better approach is to auto-generate these dependencies, but that's not implemented yet.)

piyush-jena:
^ Force push takes care of the above comments.

packages/kmod/kmod.spec
```diff
@@ -1,12 +1,15 @@
...
 Release: 1%{?dist}
 Summary: Tools for kernel module loading and unloading
 License: GPL-2.0-or-later AND LGPL-2.1-or-later
 URL: http://git.kernel.org/?p=utils/kernel/kmod/kmod.git;a=summary
 Source0: https://www.kernel.org/pub/linux/utils/kernel/kmod/kmod-%{version}.tar.xz
 Source1: https://www.kernel.org/pub/linux/utils/kernel/kmod/kmod-%{version}.tar.sign
 Source2: gpgkey-EAB33C9690013C733916AC839BA2A5A630CBEA53.asc
+Patch1001: 0001-meson-add-support-for-static-builds.patch
+Patch1002: 0002-meson-create-pkgconfig-files-in-default-path.patch
+BuildRequires: meson
```
bcressey: nit: can omit this

bcressey (approved):

packages/kmod/0001-meson-add-support-for-static-builds.patch
```diff
@@ -0,0 +1,42 @@
...
+
+diff --git a/meson.build b/meson.build
+index a0cf675..a3f4cf1 100644
+--- a/meson.build
++++ b/meson.build
+@@ -175,6 +175,7 @@ module_signatures = ''
+
+ features = []
+ dep_map = {}
++private_libs = []
```
ytsssun: Is this used anywhere? If not drop it.

ytsssun (approved):

piyush-jena:
The force push above fixes the remaining comments.

KCSesh:
Nice work!

KCSesh (approved):

piyush-jena:
verified that `vmware` instance and ec2 (`t2.micro`) instance boot.

packages/kmod/0002-meson-create-pkgconfig-files-in-default-path.patch
```diff
@@ -0,0 +1,25 @@
+From ade8d073f25fd5bbf2ceb2077dec0dd442672144 Mon Sep 17 00:00:00 2001
+From: Piyush Jena <jepiyush@amazon.com>
+Date: Tue, 8 Jul 2025 22:12:47 +0000
+Subject: [PATCH] meson: create pkgconfig files in default path
```
evelikov, line 4: Coming from the kmod static-tools PR (thanks for that)... A noob question:

Why is the default install location altered? AFAICT one should be installing kmod.pc to the architecture agnostic datadir.

From a casual search: my Arch Linux setup has ~30 projects/packages installing files to `datadir/pkgconfig` - is bottlerocked-os patching dozens of projects just to alter the path?

packages/kmod/kmod.spec
```diff
@@ -30,43 +34,42 @@ cp COPYING COPYING.LGPL
...
-pushd dynamic-build
-
-%cross_configure \
-  --with-noarch-pkgconfigdir=%{_cross_pkgconfigdir} \
-  --with-zlib \
-  --with-zstd \
-  --without-openssl \
-  --disable-manpages
+%cross_meson \
+  --sbindir=%{_cross_bindir} \
```
evelikov, line 51: Out of curiosity: does bindir and sbindir differ on bottlerocket-os? Are there any plans to merge them?

packages/kmod/0002-meson-create-pkgconfig-files-in-default-path.patch
arnaldo2792: @piyush-jena , can you help with this question? Probably we should update the message in the patch to explain why we do this.

packages/kmod/kmod.spec
arnaldo2792: They [do differ] but merging both is not part of our roadmap for the rest of the year. However, we are interested on following what other Linux [distros do].

[do differ]: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/112df59c28da0e426ebb1c93d49f96f63e9f194c/packages/filesystem/filesystem.spec#L28

[distros do]: https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin

evelikov: Glad to hear it's in the todo/wishlist. Fwiw multiple distributions are going that route - be that in part or in full. To name a few Arch, Debian, Fedora, Gentoo, openSuse, etc.

PR #569, opened 2025-07-08
Add configuration template for soci-snapshotter

### LINKED ISSUE #568: Add configuration template file for soci-snapshotter
**What I'd like:**

I'd like the soci-snapshotter package to include a configuration template file that renders settings for soci-snapshotter. This should render the settings added in https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/91. Since the `decompress_streams` setting for SOCI's parallel pull mode won't be configurable by the Bottlerocket settings API, I'd also like for the selected decompression tool to be configured based on architecture built for, with option for downstream variant builders to override this default via package epoch.

Keep the status quo.

ginglis13:
**Issue number:**

Closes #568 

**Description of changes:**

* Add a configuration template for soci-snapshotter that renders settings for the snapshotter as well as configures defaults
* Add igzip and pigz subpackages to soci-snapshotter for setting the `decompress_streams` setting for parallel pull mode and symlink these to a known location
* Drop packaging the SOCI CLI
* Add a subpackage for kubelet drop-in configuration

See the configuration docs on the soci-snapshotter repo: https://github.com/awslabs/soci-snapshotter/blob/main/docs/parallel-mode.md#configuration

**Testing done:**

* Included a local `bottlerocket-settings-sdk` with https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/91, built a core kit with these changes, and launched custom k8s-1.33 variants that package SOCI as well as set up config file rendering.
* Included user data:
```
[settings.container-runtime-plugins.soci-snapshotter]
pull-mode = "parallel-pull-unpack"
[settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack]
concurrent-download-chunk-size = "8mb"
discard-unpacked-layers = true
max-concurrent-downloads = 10
max-concurrent-downloads-per-image = 5
max-concurrent-unpacks = 5
max-concurrent-unpacks-per-image = 3
```

* Ran a pod and ensured that SOCI logs show the unpack and snapshot setup
* Ensured defaults for `/usr/bin/soci-gunzip` symlink to `igzip` on x86 and `unpigz` on aarch64:

```
# x86
bash-5.1# ls -l /usr/bin/soci-gunzip
lrwxrwxrwx. 1 root root 53 Jul  3 01:09 /usr/bin/soci-gunzip -> /x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/igzip
# aarch64
bash-5.1# ls -l /usr/bin/soci-gunzip
lrwxrwxrwx. 1 root root 53 Jul  3 01:09 /usr/bin/soci-gunzip -> /aarch64-bottlerocket-linux-gnu/sys-root/usr/bin/unpigz
```

More details on https://github.com/bottlerocket-os/bottlerocket/pull/4593#issue-3241247044

**Terms of contribution:**

### DIFF
--- a/packages/soci-snapshotter/soci-config-toml
+++ b/packages/soci-snapshotter/soci-config-toml
@@ -1,0 +1,70 @@
+[required-extensions]
+container-registry = "v1"
+container-runtime-plugins = "v1"
+std = { version = "v1", helpers = ["default", "join_array"]}
++++
+[cri_keychain]
+# Make SOCI snapshotter act as a proxy ImageService
+# and cache credentials from requests to pull images.
+enable_keychain = true
+# This tells the SOCI snapshotter where containerd's ImageService is located.
+# The SOCI snapshotter will forward requests here after caching credentials.
+image_service_path="/run/containerd/containerd.sock"
+
+# Use the containerd content store such that SOCI artifacts are maintained through
+# containerd's garbage collector
+[content_store]
+type="containerd"
+
+{{#if settings.container-registry.mirrors}}
+{{#each settings.container-registry.mirrors}}
+[plugins."io.containerd.grpc.v1.cri".registry.mirrors."{{registry}}"]
+endpoint = [{{join_array ", " endpoint }}]
+{{/each}}
+{{/if}}
+
+{{#if settings.container-registry.credentials}}
+{{#each settings.container-registry.credentials}}
+{{#if (eq registry "docker.io" )}}
+[registry.configs."registry-1.docker.io".auth]
+{{else}}
+[registry.configs."{{registry}}".auth]
+{{/if}}
+{{#if username}}
+username = "{{{username}}}"
+{{/if}}
+{{#if password}}
+password = "{{{password}}}"
+{{/if}}
+{{#if auth}}
+auth = "{{{auth}}}"
+{{/if}}
+{{#if identitytoken}}
+identitytoken = "{{{identitytoken}}}"
+{{/if}}
+{{/each}}
+{{/if}}
+
+[pull_modes.soci_v1]
+enable = false
+[pull_modes.soci_v2]
+enable = false
+
+[pull_modes.parallel_pull_unpack]
+enable = true
+{{#if settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack}}
+{{#with settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack}}
+max_concurrent_downloads = {{default -1 max-concurrent-downloads}}
+max_concurrent_downloads_per_image = {{default 3 max-concurrent-downloads-per-image}}
+{{#if concurrent-download-chunk-size}}
+{{#if (eq concurrent-download-chunk-size "unlimited" )}}
+concurrent_download_chunk_size = ""
+{{else}}
+concurrent_download_chunk_size = "{{default "" concurrent-download-chunk-size}}"
+{{/if}}
+{{/if}}
+max_concurrent_unpacks = {{default -1 max-concurrent-unpacks}}
+max_concurrent_unpacks_per_image = {{default 1 max-concurrent-unpacks-per-image}}
+discard_unpacked_layers = {{default false discard-unpacked-layers}}
+{{/with}}
+{{/if}}
--- a/packages/soci-snapshotter/soci-config-toml
+++ b/packages/soci-snapshotter/soci-config-toml
@@ -68,3 +68,8 @@ max_concurrent_unpacks_per_image = {{default 1 max-concurrent-unpacks-per-image}
 discard_unpacked_layers = {{default false discard-unpacked-layers}}
 {{/with}}
 {{/if}}
+
+# Use a symlink to configure soci to use either igzip or unpigz depending on which was packaged.
+[pull_modes.parallel_pull_unpack.decompress_streams."gzip"]
+path = "/usr/bin/soci-gunzip"
+args = ["-d", "-c"]
--- a/packages/soci-snapshotter/soci-snapshotter.service
+++ b/packages/soci-snapshotter/soci-snapshotter.service
@@ -11,3 +11,4 @@ EnvironmentFile=/etc/network/proxy.env
 ExecStart=/usr/bin/soci-snapshotter-grpc --address fd:// --config /etc/soci-snapshotter/config.toml --root /var/lib/soci-snapshotter
 Restart=always
 RestartSec=5
+SyslogIdentifier=soci-snapshotter
--- a/packages/soci-snapshotter/soci-snapshotter.service
+++ b/packages/soci-snapshotter/soci-snapshotter.service
@@ -8,6 +8,6 @@ Before=containerd.service
 [Service]
 Type=notify
 EnvironmentFile=/etc/network/proxy.env
-ExecStart=/usr/bin/soci-snapshotter-grpc --address fd://
+ExecStart=/usr/bin/soci-snapshotter-grpc --address fd:// --config /etc/soci-snapshotter/config.toml --root /var/lib/soci-snapshotter
 Restart=always
 RestartSec=5
--- a/packages/soci-snapshotter/soci-snapshotter.spec
+++ b/packages/soci-snapshotter/soci-snapshotter.spec
@@ -25,7 +25,7 @@ Requires: %{name}(binaries)
 %{summary}.
 
 %package bin
-Summary: Remote management agent binaries
+Summary: A remote snapshotter for containerd
 Provides: %{name}(binaries)
 Requires: (%{_cross_os}image-feature(no-fips) and %{name})
 Conflicts: (%{_cross_os}image-feature(fips) or %{name}-fips-bin)
@@ -34,7 +34,7 @@ Conflicts: (%{_cross_os}image-feature(fips) or %{name}-fips-bin)
 %{summary}.
 
 %package fips-bin
-Summary: Remote management agent binaries, FIPS edition
+Summary: A remote snapshotter for containerd, FIPS edition
 Provides: %{name}(binaries)
 Requires: (%{_cross_os}image-feature(fips) and %{name})
 Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
--- a/packages/soci-snapshotter/soci-snapshotter.spec
+++ b/packages/soci-snapshotter/soci-snapshotter.spec
@@ -54,19 +54,15 @@ export LD_VERSION="-X github.com/awslabs/soci-snapshotter/version.Version=v%{gov
 export LD_REVISION="-X github.com/awslabs/soci-snapshotter/version.Revision=%{gitrev}"
 
 go build -C cmd -ldflags="${GOLDFLAGS} ${LD_VERSION} ${LD_REVISION}" -o "../out/soci-snapshotter-grpc" ./soci-snapshotter-grpc
-go build -C cmd -ldflags="${GOLDFLAGS} ${LD_VERSION} ${LD_REVISION}" -o "../out/soci" ./soci
 
 gofips build -C cmd -ldflags="${GOLDFLAGS} ${LD_VERSION} ${LD_REVISION}" -o "../out/fips/soci-snapshotter-grpc" ./soci-snapshotter-grpc
-gofips build -C cmd -ldflags="${GOLDFLAGS} ${LD_VERSION} ${LD_REVISION}" -o "../out/fips/soci" ./soci
 
 %install
 install -d %{buildroot}%{_cross_bindir}
 install -d %{buildroot}%{_cross_fips_bindir}
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0755 out/soci-snapshotter-grpc %{buildroot}%{_cross_bindir}
-install -p -m 0755 out/soci %{buildroot}%{_cross_bindir}
 install -p -m 0755 out/fips/soci-snapshotter-grpc %{buildroot}%{_cross_fips_bindir}
-install -p -m 0755 out/fips/soci %{buildroot}%{_cross_fips_bindir}
 install -D -p -m 0644 %{S:101} %{buildroot}%{_cross_unitdir}
 install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_unitdir}
 
@@ -81,10 +77,8 @@ install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_unitdir}
 
 %files bin
 %{_cross_bindir}/soci-snapshotter-grpc
-%{_cross_bindir}/soci
 
 %files fips-bin
 %{_cross_fips_bindir}/soci-snapshotter-grpc
-%{_cross_fips_bindir}/soci
 
 %changelog
--- a/packages/soci-snapshotter/soci-snapshotter.spec
+++ b/packages/soci-snapshotter/soci-snapshotter.spec
@@ -14,6 +14,7 @@ Source0: https://github.com/awslabs/soci-snapshotter/archive/v%{gover}/soci-snap
 Source1: bundled-soci-snapshotter-%{gover}.tar.gz
 Source2: bundled-cmd.tar.gz
 Source3: soci-config-toml
+Source4: k8s-snapshotter-conf
 Source100: etc-soci-snapshotter.mount.in
 Source101: soci-snapshotter.service
 Source102: soci-snapshotter.socket
@@ -22,6 +23,7 @@ Source1000: clarify.toml
 BuildRequires: %{_cross_os}glibc-devel
 BuildRequires: %{_cross_os}libz-devel
 Requires: %{name}(binaries)
+Requires: (%{name}-k8s if %{_cross_os}variant-runtime(k8s))
 Requires: %{name}(optimized-gunzip)
 
 %description
@@ -69,6 +71,13 @@ Provides: %{name}(optimized-gunzip) = 0:
 %description igzip
 %{summary}.
 
+%package k8s
+Summary: Drop-ins to override the kubelet's configuration
+Provides: %{name}(k8s)
+
+%description k8s
+%{summary}.
+
 %prep
 %setup -n %{gorepo}-%{gover} -q
 %setup -T -D -n %{gorepo}-%{gover} -b 1 -q
@@ -99,6 +108,7 @@ install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_unitdir}
 
 install -d %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:3} %{buildroot}%{_cross_templatedir}/soci-config-toml
+install -p -m 0644 %{S:4} %{buildroot}%{_cross_templatedir}/k8s-snapshotter-conf
 
 %cross_scan_attribution --clarify %{S:1000} go-vendor vendor
 
@@ -128,5 +138,8 @@ posix.symlink("%{_cross_bindir}/unpigz", "%{_cross_bindir}/soci-gunzip")
 
 %files igzip
 # No files provided by igzip but required for packaging.
+
+%files k8s
+%{_cross_templatedir}/k8s-snapshotter-conf
 
 %changelog
--- a/packages/soci-snapshotter/soci-snapshotter.spec
+++ b/packages/soci-snapshotter/soci-snapshotter.spec
@@ -22,6 +22,7 @@ Source1000: clarify.toml
 BuildRequires: %{_cross_os}glibc-devel
 BuildRequires: %{_cross_os}libz-devel
 Requires: %{name}(binaries)
+Requires: %{name}(optimized-gunzip)
 
 %description
 %{summary}.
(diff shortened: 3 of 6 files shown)


packages/soci-snapshotter/soci-snapshotter.service
```diff
@@ -3,10 +3,11 @@ Description=soci snapshotter containerd plugin
 Documentation=https://github.com/awslabs/soci-snapshotter
 After=configured.target
 Wants=configured.target
-Before=containerd.service
+Requires=containerd.service
```
yeazelm: Do we need Requires or After here?

packages/soci-snapshotter/soci-config-toml
```diff
@@ -0,0 +1,77 @@
...
+{{#if settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-unpacks}}
+max_concurrent_unpacks = {{settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-unpacks}}
+{{/if}}
+{{#if settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-unpacks-per-image}}
+max_concurrent_unpacks_per_image = {{settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-unpacks-per-image}}
+{{/if}}
+{{#if settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.discard-unpacked-layers}}
+discard_unpacked_layers = {{settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.discard-unpacked-layers}}
+{{/if}}
+[pull_modes.parallel_pull_unpack.decompress_streams."gzip"]
```
yeazelm, line 73: Can we visually separate this so its clear you get this every time instead of the preceding if-blocked configurations?

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -67,24 +114,49 @@ install -p -m 0755 out/soci-snapshotter-grpc %{buildroot}%{_cross_bindir}
...
 install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_unitdir}
 
+# Install SOCI config
+install -d %{buildroot}%{_cross_templatedir}
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/soci-snapshotter-grpc
+install -p -m 0644 %{S:3} %{buildroot}%{_cross_templatedir}/soci-config-toml
+
 %cross_scan_attribution --clarify %{S:1000} go-vendor vendor
 
+%post igzip -p <lua>
```
yeazelm, line 115: I know we do this a bit in other places but I'm still foggy on why its required. Can you enlighten me for my own learning?

packages/soci-snapshotter/soci-snapshotter.service
ginglis13: I had originally used `Requires` when settings `content_store="containerd"` for soci because of a dependency requirement but it looks like upstream that requirement was removed with this commit: https://github.com/awslabs/soci-snapshotter/commit/0fe54ed469bed4ee4a8866a9775a746a94282948

I'll revert this back to `Before` following the upstream example service: https://github.com/awslabs/soci-snapshotter/blob/8b2bf80a852b68e7bdc24fcb5d35772566376ada/soci-snapshotter.service#L33

packages/soci-snapshotter/soci-snapshotter.spec
ginglis13: It might seem like we could create this symlink in a `%install` section per-subpackage, but RPM only supports one `%install` section per package. We can't do something like `%install igzip` and `%install pigz` , and we can't put this in the overall package `%install` since we can't construct a symlink `%{_cross_bindir}/soci-gunzip` that links to both `unpigz` and `igzip`.

I considered constructing these respective symlinks in [libisal](https://github.com/bottlerocket-os/bottlerocket-core-kit/tree/develop/packages/libisal) and [pigz](https://github.com/bottlerocket-os/bottlerocket-core-kit/tree/develop/packages/pigz), but I'm not sure that would be appropriate for other dependencies of those packages.

ginglis13:
^ force push:
* reverts change to dependency ordering in soci-snapshotter.service
* uses the `defaults` helper in soci-config-toml. Defaults obtained from running soci-snapshotter v0.11.1 in parallel pull mode:
```
{"concurrent_download_chunk_size":-1,"discard_unpack_layers":false,"level":"info","max_concurrent_downloads":-1,"max_concurrent_downloads_per_image":3,"max_concurrent_unpacks":-1,"max_concurrent_unpacks_per_image":1,"msg":"Parallel image pull enabled","time":"2025-07-10T22:54:53.794332469Z"}
```

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -42,6 +42,24 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
 %description fips-bin
 %{summary}.
 
+%package cli-bin
```
arnaldo2792: You should provide a package `cli`, that requires `%{name}-cli(binaries)` so that the downstreams only require `%{cross_os}-soci-cli` and through RPM magic, they get the correct version (FIPS or No FIPS).

You will have to adjust the subpackages dependencies and conflicts accordingly.

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -42,6 +42,24 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
...
+Summary: soci CLI binary,
```
arnaldo2792: It can be just `Soci CLI`, following what was done for `Docker`. But given the description of the main package, you could do `Remove management agent CLI`

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -13,6 +13,10 @@ URL: https://github.com/awslabs/soci-snapshotter
 Source0: https://github.com/awslabs/soci-snapshotter/archive/v%{gover}/soci-snapshotter-%{gover}.tar.gz
 Source1: bundled-soci-snapshotter-%{gover}.tar.gz
 Source2: bundled-cmd.tar.gz
+Source3: soci-config-toml
+# Mount for writing soci configuration
+# Generated with systemd-escape -p --suffix=mount "/etc/soci-snapshotter-grpc"
+Source100: "etc-soci\\x2dsnapshotter\\x2dgrpc.mount"
```
arnaldo2792: Is this because of the dashes? If that's the case you can keep `etc-soci-snapshotter-grpc` as the name of the file in the sources and escape it at install time.

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -85,17 +89,26 @@ install -p -m 0755 out/soci-snapshotter-grpc %{buildroot}%{_cross_bindir}
 install -p -m 0755 out/soci %{buildroot}%{_cross_bindir}
 install -p -m 0755 out/fips/soci-snapshotter-grpc %{buildroot}%{_cross_fips_bindir}
 install -p -m 0755 out/fips/soci %{buildroot}%{_cross_fips_bindir}
+install -D -p -m 0644 %{S:100} %{buildroot}%{_cross_unitdir}
```
arnaldo2792: You can escape the name of the file here and install `S:100` with the escaped name

packages/soci-snapshotter/soci-config-toml
```diff
@@ -0,0 +1,60 @@
+[required-extensions]
+container-registry = "v1"
+container-runtime-plugins = "v1"
+std = { version = "v1", helpers = ["default", "join_array"]}
++++
+[cri_keychain]
+# This tells the SOCI snapshotter to act as a proxy ImageService
+# and to cache credentials from requests to pull images.
```
arnaldo2792:
```
Make SOCI act as a proxy ImageService...
```

Rather than using `This`. The same voice for all the others.

packages/soci-snapshotter/soci-config-toml
```diff
@@ -0,0 +1,60 @@
+[required-extensions]
```
arnaldo2792, line 1: For the lack of a better place to put this, we should fix the SOCI snapshotter service and make it use the HTTP proxy env var, right?

See: https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/564

ginglis13:
^ force push modifies config and state directories to remove "-grpc" suffix

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -13,13 +13,18 @@ URL: https://github.com/awslabs/soci-snapshotter
 Source0: https://github.com/awslabs/soci-snapshotter/archive/v%{gover}/soci-snapshotter-%{gover}.tar.gz
 Source1: bundled-soci-snapshotter-%{gover}.tar.gz
 Source2: bundled-cmd.tar.gz
+Source3: soci-config-toml
+# Mount for writing soci configuration
+# Generated with systemd-escape -p --suffix=mount "/etc/soci-snapshotter"
+Source100: "etc-soci\\x2dsnapshotter.mount"
```
bcressey: What we've done elsewhere - `release.spec` has lots of examples, and `host-ctr.spec` - is to name the mount unit something more natural, like `etc-soci-snapshotter.mount.in`, and then rename it via `systemd-escape` in the spec file.

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -42,6 +47,48 @@ Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
 %description fips-bin
 %{summary}.
 
+%package cli-bin
+Summary: soci CLI binary,
```
bcressey: I'd prefer to remove the CLI if it's not extremely useful for troubleshooting.

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -13,13 +13,18 @@ URL: https://github.com/awslabs/soci-snapshotter
...
+# Generated with systemd-escape -p --suffix=mount "/etc/soci-snapshotter"
+Source100: "etc-soci\\x2dsnapshotter.mount"
 Source101: soci-snapshotter.service
 Source102: soci-snapshotter.socket
 Source1000: clarify.toml
 
 BuildRequires: %{_cross_os}glibc-devel
 BuildRequires: %{_cross_os}libz-devel
 Requires: %{name}(binaries)
+Requires: %{name}(optimized-gunzip)
```
bcressey, line 27: Following up to my comment [here](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/576#discussion_r2205883876), we could add:
```
Requires: (%{name}-k8s if %{_cross_os}variant-runtime(k8s))
```

To automatically install a `k8s` subpackage containing the containerd (CRI) and k8s drop-ins when installing SOCI on a k8s variant.

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -67,24 +114,49 @@ install -p -m 0755 out/soci-snapshotter-grpc %{buildroot}%{_cross_bindir}
...
+posix.symlink("%{_cross_bindir}/unpigz", "%{_cross_bindir}/soci-gunzip")
+
 %files
 %license LICENSE NOTICE.md
 %{_cross_unitdir}/soci-snapshotter.service
 %{_cross_unitdir}/soci-snapshotter.socket
+%{_cross_unitdir}/etc-soci\\x2dsnapshotter.mount
 %{_cross_attribution_vendor_dir}
 %{_cross_attribution_file}
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/soci-snapshotter-grpc
```
bcressey: We don't seem to be using this factory directory for anything - it would be useful if we wanted to provide a default SOCI config for downstreams that don't have the settings or templates.

ginglis13: Gotcha, in that case I will omit it for now.

packages/soci-snapshotter/soci-snapshotter.spec
ginglis13: I'll declare a macro like

```
%global soci_mount_path %(systemd-escape --path %{_sysconfdir}/soci-snapshotter)
```

to use in `%install` and `%files`

packages/soci-snapshotter/soci-config-toml
ginglis13: Good catch; I'll open a separate PR to do so to keep this scoped to configuration.

ginglis13: See https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/584

ginglis13:
^ force push labels the soci config mount as `etc_secret_t` (see #579)

ginglis13:
^ rebase off develop, remove the `soci` CLI subpacakge

packages/soci-snapshotter/soci-config-toml
```diff
@@ -0,0 +1,65 @@
...
+[pull_modes.soci_v1]
+enable = false
+[pull_modes.soci_v2]
+enable = false
+
+[pull_modes.parallel_pull_unpack]
+enable = true
+max_concurrent_downloads = {{default -1 settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-downloads}}
+max_concurrent_downloads_per_image = {{default 3 settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-downloads-per-image}}
+concurrent_download_chunk_size = "{{default -1 settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.concurrent-download-chunk-size}}"
```
bcressey: [Empty string means default chunk size](https://github.com/awslabs/soci-snapshotter/blob/8d8d1200b8eba2daf31a95ce8585a44a553d91cc/config/parallel.go#L84):
```suggestion
concurrent_download_chunk_size = "{{default "" settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.concurrent-download-chunk-size}}"
```

That same function returns an error for sizes less than zero, so I'm not confident `-1` would do what we expect.

packages/soci-snapshotter/soci-config-toml
```diff
@@ -0,0 +1,65 @@
...
+max_concurrent_unpacks = {{default -1 settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-unpacks}}
+max_concurrent_unpacks_per_image = {{default 1 settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.max-concurrent-unpacks-per-image}}
+discard_unpacked_layers = {{default false settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.discard-unpacked-layers}}
```
bcressey: AFAIK we haven't tried this before, but handlebars should let us do something like this:
```
[pull_modes.parallel_pull_unpack]
enable = true
{{#if settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack}}
{{#with settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack}}
...
max_concurrent_unpacks = {{default -1 max-concurrent-unpacks}}
max_concurrent_unpacks_per_image = {{default 1 max-concurrent-unpacks-per-image}}
discard_unpacked_layers = {{default false discard-unpacked-layers}}
{{/with}}
{{/if}
```

packages/soci-snapshotter/soci-config-toml
ginglis13: I can update this to be empty string to catch that conditional. Ultimately that should resolve to `-1` as the [defaultConcurrentDownloadChunkSize](https://github.com/awslabs/soci-snapshotter/blob/8d8d1200b8eba2daf31a95ce8585a44a553d91cc/config/defaults.go#L127-L129) is [`Unbounded`](https://github.com/awslabs/soci-snapshotter/blob/8d8d1200b8eba2daf31a95ce8585a44a553d91cc/config/defaults.go#L52-L55), which is `-1`

ginglis13:
^ force push adds a `k8s` subpackage that packages a drop-in for kubelet to set imageServiceEndpoint to soci's socket, addresses other feedback for the soci config template.

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -54,37 +91,55 @@ export LD_VERSION="-X github.com/awslabs/soci-snapshotter/version.Version=v%{gov
...
-install -p -m 0755 out/soci %{buildroot}%{_cross_bindir}
 install -p -m 0755 out/fips/soci-snapshotter-grpc %{buildroot}%{_cross_fips_bindir}
-install -p -m 0755 out/fips/soci %{buildroot}%{_cross_fips_bindir}
+install -p -m 0644 %{S:100} %{buildroot}%{_cross_unitdir}/%{soci_mount_path}.mount
 install -D -p -m 0644 %{S:101} %{buildroot}%{_cross_unitdir}
 install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_unitdir}
 
+# Install SOCI config
+install -d %{buildroot}%{_cross_templatedir}
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/soci-snapshotter-grpc
```
bcressey: Not used?
```suggestion
```

bcressey (approved):

ginglis13:
^drop an unnecessary `install`, add a commit to fix the package summaries, remove `[skip ci]` from commit msgs

ginglis13:
^ force push fixes a rebase issue between [soci-snapshotter: configure architecture-specific compression](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/569/commits/798d28626c92e0268594d3f8ed914d8305542d25) and [soci-snapshotter: add drop-in for kubelet configuration](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/569/commits/c2bb649a6a0bb6e626164547a6875cec36ecbfd8)

yeazelm (approved):

ginglis13:
^ force pushes bring us back to the changes in https://github.com/bottlerocket-os/bottlerocket-core-kit/commit/688be9688459d78a25c88d2bb8a765cc9b6edbb8.

ginglis13:
^ adjust for "unlimited" setting value for chunk size (see https://github.com/bottlerocket-os/bottlerocket-settings-sdk/commit/e6665a1280f96ef0b3bc77b9d0075379d8320701 from https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/91

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -2,6 +2,7 @@
 %global gover 0.11.1
 %global rpmver %{gover}
 %global gitrev 28781de6731978b2e2f0f43573a345e9fa14dbd1
+%global soci_mount_path %(systemd-escape --path %{_sysconfdir}/soci-snapshotter)
```
arnaldo2792: I just realized that we have a strong dependency on the system package in the Bottlerocket SDK container :sweat_smile: (not introduced by this change tho)

ginglis13:
^ force push fixes `.mount` file for soci configuration to match `Where` it is mounted and wraps settings `concurrent-download-chunk-size` in `{{if}}` block

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -63,17 +65,26 @@ install -d %{buildroot}%{_cross_fips_bindir}
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0755 out/soci-snapshotter-grpc %{buildroot}%{_cross_bindir}
 install -p -m 0755 out/fips/soci-snapshotter-grpc %{buildroot}%{_cross_fips_bindir}
+
+SOCIMOUNTPATH=$(systemd-escape --path /etc/soci-snapshotter)
+install -p -m 0644 %{S:100} %{buildroot}%{_cross_unitdir}/${SOCIMOUNTPATH}.mount
```
arnaldo2792, line 104: nit: is it possible to make this a `global` and use it in the `%files` section?

packages/soci-snapshotter/soci-snapshotter.spec
```diff
@@ -69,6 +71,13 @@ Provides: %{name}(optimized-gunzip) = 0:
 %description igzip
 %{summary}.
 
+%package k8s
+Summary: Drop-in configuration for kubelet to set imageServiceEndpoint to soci-snapshotter socket
```
arnaldo2792:
```suggestion
Summary: Drop-ins to override the kubelet's configurations
```

In case more soci-snapshotter overrides are provided by this sub-package other than the socket override

arnaldo2792 (approved):

packages/soci-snapshotter/soci-snapshotter.spec
ginglis13: I had originally declared this as global like

```
%global soci_mount_path %(systemd-escape --path %{_sysconfdir}/soci-snapshotter)
```

but escape char `\` would be remove, resulting in `soci_mount_path=etc-socix2dsnapshotter`. Wrapping in quotes would retain the escape char (`soci_mount_path="etc-soci\x2dsnapshotter"`) but then we'd be installing a file with quotes in its name.

This iteration more closely resembles some of our prior art in the release spec: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/9668a6ae28b4e9dadd58e63e2ec675859227833f/packages/release/release.spec#L219-L246

ginglis13:
^ adjust description for `k8s` drop in subpackage

KCSesh (approved):

PR #571, opened 2025-07-08
ephemeral storage: allow soci-snapshotter path

### LINKED ISSUE #570: Add /var/lib/soci-snapshotter-grpc to the list of allowed paths for ephemeral storage
**What I'd like:**

I'd like `/var/lib/soci-snapshotter-gprc` to be included in the allowlist of paths that can bind to ephemeral storage

Keep the status quo

ginglis13:
**Issue number:**

Closes #570 

**Description of changes:**

Add `/var/lib/soci-snapshotter-grpc` to the allowlisted paths that can bind to ephemeral storage.

**Testing done:**

On a custom dev `aws-k8s-1.33` variant that includes `soci-snapshotter`, launched with userdata:

```
commands = [
    ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/var/lib/containerd", "/var/lib/kubelet", "/var/log/pods", "/var/lib/soci-snapshotter-grpc"]
]
essential = true
mode = "always"
```

lsblk output:
```
...
nvme3n1      259:16   0  6.8T  0 disk
`-md127        9:127  0 54.6T  0 raid0 /var/lib/soci-snapshotter-grpc
                                       /var/log/pods
                                       /var/lib/kubelet
                                       /var/lib/containerd
                                       /mnt/.ephemeral
...
```

And successfully ran pods.

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/ephemeral_storage.rs
+++ b/sources/api/apiserver/src/server/ephemeral_storage.rs
@@ -297,6 +297,7 @@ pub fn allowed_bind_dirs(variant: &str) -> BindDirs {
     if variant.contains("k8s") {
         allowed_exact.insert("/var/lib/kubelet");
         allowed_exact.insert("/var/log/pods");
+        allowed_exact.insert("/var/lib/soci-snapshotter");
     }
     if variant.contains("ecs") {
         allowed_exact.insert("/var/lib/docker");


piyush-jena (approved):

arnaldo2792 (approved):

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -297,6 +297,7 @@ pub fn allowed_bind_dirs(variant: &str) -> BindDirs {
     if variant.contains("k8s") {
         allowed_exact.insert("/var/lib/kubelet");
         allowed_exact.insert("/var/log/pods");
+        allowed_exact.insert("/var/lib/soci-snapshotter-grpc");
```
bcressey: The `-grpc` suffix is kind of ugly so if it's trivial to reconfigure this path to be `/var/lib/soci-snapshotter` then we should do that. Otherwise LGTM.

bcressey (approved):

sources/api/apiserver/src/server/ephemeral_storage.rs
ginglis13: Should be trivial, just have to shuffle around some changes to account for it, but in service startup I'll modify the snapshotter root to be `/var/lib/soci-snapshotter`:
```
  -root string
    	path to the root directory for this snapshotter (default "/var/lib/soci-snapshotter-grpc/")
```

ginglis13: Updated to `/var/lib/soci-snapshotter` See https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/569/commits/7e90f546ff03a3aba1647348122d15b98602b147 from #569 for corresponding change in the service

arnaldo2792 (approved):

bcressey (approved):

PR #573, opened 2025-07-09
expand support for efficient relabeling in SELinux policy

### LINKED ISSUE #544: Allow control_t to relabel from/to data_t
**What I'd like:**
We would like to use [filesystem relabeling](https://kubernetes.io/blog/2023/04/18/kubernetes-1-27-efficient-selinux-relabeling-beta/) using `*context` mount options to avoid the expensive recursive relabeling performed by containerd. However, although we have patched the EBS CSI driver (which runs with `privileged: true`) to allow this, we are blocked by SELinux policies defined in bottlerocket-core, i.e.

`audit: type=1400 audit([1748640838](tel:1748640838).961:3103): avc:  denied  { relabelto } for  pid=2712664 comm="mount" scontext=system_u:system_r:control_t:s0-s0:c0.c1023 tcontext=system_u:object_r:data_t:s0:c215,c217 tclass=filesystem permissive=0`

This surfaces up to the kubelet with this red-herring error:
```
Warning  FailedMount             56s (x25 over 68s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-3c92bfbe-f519-4cef-a817-6a99ecccde58" : rpc error: code = Internal desc = could not format "/dev/nvme2n1" and mount it at "/var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/3e117e7b9801a4f361b6e7b0aa83121b79e994f55ab40db9ef0c8b3a7810b848/globalmount": mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t ext4 -o context="system_u:object_r:data_t:s0:c215,c217",defaults /dev/nvme2n1 /var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/3e117e7b9801a4f361b6e7b0aa83121b79e994f55ab40db9ef0c8b3a7810b848/globalmount
Output: mount: /var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/3e117e7b9801a4f361b6e7b0aa83121b79e994f55ab40db9ef0c8b3a7810b848/globalmount: cannot mount /dev/nvme2n1 read-only.
```

It appears that bottlerocket wants to support this use case based on [this change](https://github.com/bottlerocket-os/bottlerocket/commit/4bcf847465d7ae6878e33391ee63acda0f1543ce), but the policy seems to be insufficiently broad. Is there any rationale for not supporting relabels from/to `data_t`?

(Our reference environment is an EKS cluster on Kubernetes 1.31 running Bottlerocket 1.36. I was unable to find any relevant changes to policy in newer releases)

Override the desired volume label to be in `local_t` rather than `data_t` using `selinuxoptions`. This does not appear to be possible as the kubelet always (?) uses the OS default of `data_t`.

bcressey:
**Issue number:**
Fixes #544

**Description of changes:**
Add `data_t` to the labels that containers are allowed to use in the `context=*` option when mounting filesystems. This enables CSI drivers to utilize [efficient SELinux relabeling](https://kubernetes.io/blog/2023/04/18/kubernetes-1-27-efficient-selinux-relabeling-beta/), while also providing the pod's MCS pair to configure the volume with `ReadWriteOncePod` semantics.

**Testing done:**
I verified that [a valkey deployment](https://gist.github.com/bcressey/c49ea00bfca33cd633f5ab4d3a5ae265) with two PVCs - one with the `ReadWriteOncePod` access mode, one with `ReadWriteOnce` - was able to start successfully, and the `context` option was set as expected.

I enabled the `SELinuxMount` feature gate in the `kubelet` config to verify that the same policy change will cover this behavior when that feature gate is enabled by default in the future.

The `context=*` option is applied at mount time by the CSI driver:
```
bash-5.1# mount | grep globalmount
/dev/nvme5n1 on /var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/fd9c17b3da5c184ccd2bce72d1db0e6e963a5c32ef5fba2f7dbd4bfb0597ba46/globalmount type ext4 (rw,relatime,context="system_u:object_r:data_t:s0:c1,c22,c333")
/dev/nvme4n1 on /var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/9d555ea631f871502d4769e525bec811bafa2e70d5f4f870870538112d317b5d/globalmount type ext4 (rw,relatime,context="system_u:object_r:data_t:s0:c1,c22,c333")
```

I also created test cases for privileged and unprivileged container mount operations to validate that the MCS constraints are working as expected.

Process labels:
* `system_u:system_r:control_t:s0-s0:c0.c1023` (privileged)
* `system_u:system_r:container_t:s0:c0.c2` (unprivileged)

The privileged container can override the effective context, for example by mounting a `c3.c5` filesystem on a `c0.c2` directory (or vice versa). The unprivileged container can only adjust the context to the same or more open permissions.

| Context Option | Mount Point Label | Privileged | Unprivileged |
| - | - | - | - |
| `system_u:object_r:local_t:s0` | `system_u:object_r:local_t:s0` | ✅ | ✅ |
| `system_u:object_r:data_t:s0` | `system_u:object_r:local_t:s0` | ✅ | ✅ |
| `system_u:object_r:data_t:s0:c0.c2` | `system_u:object_r:local_t:s0` | ✅ | ✅ |
| `system_u:object_r:data_t:s0:c3.c5` | `system_u:object_r:local_t:s0` | ✅ | ❌ |
| `system_u:object_r:local_t:s0` | `system_u:object_r:data_t:s0:c0.c2` | ✅ | ✅ |
| `system_u:object_r:data_t:s0` | `system_u:object_r:data_t:s0:c0.c2` | ✅ | ✅ |
| `system_u:object_r:data_t:s0:c0.c2` | `system_u:object_r:data_t:s0:c0.c2` | ✅ | ✅ |
| `system_u:object_r:data_t:s0:c3.c5` | `system_u:object_r:data_t:s0:c0.c2` | ✅ | ❌ |
| `system_u:object_r:local_t:s0` | `system_u:object_r:data_t:s0:c3.c5` | ✅ | ❌ |
| `system_u:object_r:data_t:s0` | `system_u:object_r:data_t:s0:c3.c5` | ✅ | ❌ |
| `system_u:object_r:data_t:s0:c0.c2` | `system_u:object_r:data_t:s0:c3.c5` | ✅ | ❌ |
| `system_u:object_r:data_t:s0:c3.c5` | `system_u:object_r:data_t:s0:c3.c5` | ✅ | ❌ |

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/mcs.cil
+++ b/packages/selinux-policy/mcs.cil
@@ -50,12 +50,13 @@
           (or (eq t2 all_s)
               (eq t2 unconstrained_o)))))
 
-; Restrict file writes unless one of these conditions is met:
+; Restrict file writes (and mounts) unless one of these conditions is
+; met:
 ; * the source dominates the target
 ; * the source context is for a privileged subject (e.g. `control_t`)
 ; * the target context is for an unconstrained object
 
-(mlsconstrain (files (mutate))
+(mlsconstrain (files (mutate mount))
   (or (dom h1 h2)
       (or (eq t1 privileged_s)
           (eq t2 unconstrained_o))))
--- a/packages/selinux-policy/rules.cil
+++ b/packages/selinux-policy/rules.cil
@@ -271,8 +271,12 @@
 (allow container_s constrained_o (files (relabel)))
 
 ; Containers are also allowed to "relabel" filesystems at mount time
-; using "*context=" options, but only to and from "local_t".
+; using "*context=" options, but only to and from the "local_t" and
+; "data_t" types. This enables CSI drivers to mount filesystems on
+; behalf of containers with the appropriate label, so the container
+; runtime does not need to relabel all files.
 (allow container_s local_t (filesystems (relabel)))
+(allow container_s data_t (filesystems (relabel)))
 
 ; Untrusted components are not allowed to relabel most files, and
 ; cannot use "*context=" options to impose arbitrary labels.


bcressey:
> The unprivileged container can only adjust the context to the same or more open permissions.

In practical terms, this restriction will prevent unprivileged CSI drivers from leveraging the "efficient relabel" mechanism, since it means that the driver won't be able to set a context that it can't also read and write to.

For example, a CSI driver couldn't mount a filesystem on a `local_t` directory - which the directories under `/var/lib/kubelet/plugins/kubernetes.io/csi/` will have as a label - with some other pod's MCS pair.

Likewise, a CSI driver couldn't mount a filesystem on a `data_t:s0:cX,cY` directory - which its own root filesystem will have as a label - with a different MCS pair, and then bind-mount it over to the CSI directory for propagation back to the host.

I don't consider this a blocker since CSI drivers will need `CAP_SYS_ADMIN` and an unconfined seccomp policy to perform mounts. If those security options must be set, adding a custom SELinux label as well shouldn't be too difficult.

Also, letting the CSI driver mount a filesystem that it doesn't have permissions to unmount seems like the sort of thing that can lead to trouble. Fixing that - by letting it unmount filesystems that it can't read or write to - doesn't sound very secure. There are also some filesystems, like overlayfs, that check the mounter's credentials as well as the process's, which wouldn't work.

bcressey:
Additional notes in case it's helpful, for reviewers or for future me.

These are the AVC logs for a successful mount:
```
[  461.748781] audit: type=1400 audit(1751918282.853:2683): avc:  granted  { mounton } for  pid=11558 comm="mount" path="/var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/1ed3cd355fd6c6da01382aa62d9bd994d22ada1db8e156fb871f6314ee93fa2c/globalmount" dev="nvme2n1" ino=135162362 scontext=system_u:system_r:control_t:s0-s0:c0.c1023 tcontext=system_u:object_r:local_t:s0 tclass=dir
[  461.752802] audit: type=1400 audit(1751918282.853:2684): avc:  granted  { relabelfrom } for  pid=11558 comm="mount" scontext=system_u:system_r:control_t:s0-s0:c0.c1023 tcontext=system_u:object_r:local_t:s0 tclass=filesystem
[  461.752809] audit: type=1400 audit(1751918282.853:2685): avc:  granted  { relabelto } for  pid=11558 comm="mount" scontext=system_u:system_r:control_t:s0-s0:c0.c1023 tcontext=system_u:object_r:data_t:s0:c1,c22,c333 tclass=filesystem
[  461.752812] audit: type=1400 audit(1751918282.853:2686): avc:  granted  { relabelfrom } for  pid=11558 comm="mount" scontext=system_u:system_r:control_t:s0-s0:c0.c1023 tcontext=system_u:object_r:data_t:s0:c1,c22,c333 tclass=filesystem
[  461.752814] audit: type=1400 audit(1751918282.853:2687): avc:  granted  { mount } for  pid=11558 comm="mount" name="/" dev="nvme3n1" ino=2 scontext=system_u:system_r:control_t:s0-s0:c0.c1023 tcontext=system_u:object_r:data_t:s0:c1,c22,c333 tclass=filesystem
```

1. First comes a "mounton" permission check for the target directory.
2. Next, [if the context sid is specified](https://elixir.bootlin.com/linux/v6.12.33/source/security/selinux/hooks.c#L814), we call [may_context_mount_sb_relabel](https://elixir.bootlin.com/linux/v6.12.33/source/security/selinux/hooks.c#L429). This checks whether the subject can "relabelfrom" the existing label, and "relabelto" the provided context.
3. Since we're not specifying the "rootcontext" option, we [set the rootcontext sid to the same value as context](https://elixir.bootlin.com/linux/v6.12.33/source/security/selinux/hooks.c#L828).
4. Because [the rootcontext sid is now specified](https://elixir.bootlin.com/linux/v6.12.33/source/security/selinux/hooks.c#L834), we call [may_context_mount_inode_relabel](https://elixir.bootlin.com/linux/v6.12.33/source/security/selinux/hooks.c#L446). This checks whether the subject can "relabelfrom" the provided context.
5. Later on, the final "mount" permission check happens.

The interaction of (3) and (4) means that the subject needs both "relabelto" and "relabelfrom" permissions for any potential MCS level.

arnaldo2792 (approved):

ytsssun:
Thanks for this effort! The change LGTM. The supported mount options for the unprivileged CSI drivers seems already a good scope. And we can always use privileged CSI drivers for more aggressive relabeling. I just have some clarification questions:
1. The efficient relabeling feature claims that using mount options with SELinux labels prevents recursive relabeling by the container runtime. Did your testing verify this performance benefit with volumes containing a large number of files? For example, by toggling `seLinuxMount: true` in the CSI driver DS, did you observe container creation time difference?

2. The official Kubernetes documentation indicates that efficient relabeling is currently limited to volumes with `ReadWriteOncePod` access mode, but your tests show it working with `ReadWriteOnce` as well. What happens if two pods on the same node with different SELinux levels try to mount the same `ReadWriteOnce` volume? Would the second mount fail, would the SELinux context be ignored, or would the pod fail to access the volume due to permission issues?

bcressey:
> Did your testing verify this performance benefit with volumes containing a large number of files?

I haven't tested that specifically. We've observed in the past that allowing the CSI driver to set `local_t` as the context label avoided this overhead though, so I felt it was sufficient to verify that the driver could now set `data_t`.

> What happens if two pods on the same node with different SELinux levels try to mount the same ReadWriteOnce volume?

These scenarios are discussed in KEP 1710's [conflicts with other pods](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1710-selinux-relabeling/README.md#conflicts-with-other-pods) section:
> In general, a Pod that wants the volume mounted with a different -o context option than it's currently used must wait for all pods that use the volume to get deleted and the volume unmounted.

ytsssun (approved):

rpkelly (approved):

KCSesh (approved):

PR #582, opened 2025-07-14
Soci snapshotter change service

### LINKED ISSUE #580: Add service to ensure clean containerd snapshotter state when setting `settings.container-runtime.snapshotter`
**What I'd like:**

Bottlerocket will allow users to modify the snapshotter configured for containerd via `settings.container-runtime.snapshotter`. Switching snapshotters requires the clean state of containerd, meaning no active
running containers or existing images. Bottlerocket should only make the change to configured snapshotter at boot of an instance and not do so at runtime.

Currently, there's an issue where the combination of using `discard_unpacked_layers=true` on a snapshotter and switching snapshotters can lead to dangling references of container layer and snapshot content. Layer blobs are missing from the containerd content store because they were previously discarded after the unpack step in snapshot creation, causing failure to unpack an image that has been pulled but its layers discarded.

To avoid this problematic state, Bottlerocket should add a new service that performs cleanup operations when it detects a change between the selected snapshotter configuration and the currently active snapshotter on boot. This service should do the following:

1. If active-snapshotter != selected-snapshotter
2. Execute `find /var/lib/containerd -mindepth 1 -delete` to clean containerd state
3. Execute `find /var/lib/containerd /var/lib/soci-snapshotter -mindepth 1 -delete` to clean soci-snapshotter state
4. Set the selected-snapshotter to be the active snapshotter

This cleanup ensures that when users switch snapshotters via the settings API, the system starts with a clean state that prevents layer reference issues and ensures reliable container image operations with the
newly configured snapshotter.

Keep the status quo.

ginglis13:
**Issue number:**

Closes #580

**Description of changes:**

Bottlerocket will allow users to modify the snapshotter configured for containerd via settings.container-runtime.snapshotter. Switching snapshotters requires the clean state of containerd, meaning no active
running containers or existing images. Bottlerocket should only make the change to configured snapshotter at boot of an instance and not do so at runtime.

This commit  adds a service to configure snapshotter selection change on reboot of an instance. When settings.container-runtime.snapshotter is modified, it will only take effect at reboot. Change in containerd's underlying snapshotter can lead to dangling content, so when such a change is detected, reset respective snapshotter state directories /var/lib/containerd and /var/lib/soci-snapshotter-grpc.

**Testing done:**

In each case, launch the same pod listed below. Chosen variant: `aws-k8s-1.33-x86_64`.

<details>

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: djl-deployment
  labels:
    app: djl-soci
spec:
  replicas: 1
  selector:
    matchLabels:
      app: djl-soci
  template:
    metadata:
      labels:
        app: djl-soci
    spec:
      containers:
      - name: djl-container
        #image: deepjavalibrary/djl-serving:0.26.0-deepspeed
        image: 763104351884.dkr.ecr.us-west-2.amazonaws.com/djl-inference:0.22.1-deepspeed0.9.2-cu118
        command: ["sleep"]
        args: ["inf"]
```

</details>

- [x] Reboot an instance with no change in snapshotter selection
  * 1x for soci and 1x for overlay configured
  * ensure all pods come back up
- [x] Boot with soci. Run some pods with example images. change snapshotter. reboot
  * ensure pods come back up to `Running`
  * ensure `ls -lR /var/lib/soci-snapshotter` is empty
- [x] Boot with overlay. Run some pods. change snapshotter. reboot.
  * ensure pods come back up to `Running`
  * ensure content via `ls -lR /var/lib/soci-snapshotter`
* **RESULTS: https://gist.github.com/ginglis13/cfc7d1d907e65d889af9b55923476e12**

**Terms of contribution:**

### DIFF
--- a/packages/containerd-1.7/containerd.service
+++ b/packages/containerd-1.7/containerd.service
@@ -3,6 +3,7 @@ Description=containerd container runtime
 Documentation=https://containerd.io
 After=network-online.target configured.target
 Wants=network-online.target configured.target
+Requires=configure-snapshotter.service
 
 [Service]
 Slice=runtime.slice
--- a/packages/containerd-2.0/containerd.service
+++ b/packages/containerd-2.0/containerd.service
@@ -3,6 +3,7 @@ Description=containerd container runtime
 Documentation=https://containerd.io
 After=network-online.target configured.target
 Wants=network-online.target configured.target
+Requires=configure-snapshotter.service
 
 [Service]
 Slice=runtime.slice
--- a/packages/release/configure-snapshotter.service
+++ b/packages/release/configure-snapshotter.service
@@ -1,0 +1,30 @@
+[Unit]
+Description=Configure Snapshotter
+Before=containerd.service
+
+[Service]
+Type=oneshot
+EnvironmentFile=-/etc/containerd/selected-snapshotter
+EnvironmentFile=-/var/cache/containerd/active-snapshotter
+# Skip cleanup if either snapshotter variable is empty
+ExecCondition=[ -n "${ACTIVE_SNAPSHOTTER}" ]
+ExecCondition=[ -n "${SELECTED_SNAPSHOTTER}" ]
+# Check if the active snapshotter has changed
+ExecCondition=[ "${SELECTED_SNAPSHOTTER}" != "${ACTIVE_SNAPSHOTTER}" ]
+# Don't error if the directories don't exist.
+ExecStart=-/usr/bin/find /var/lib/soci-snapshotter -mindepth 1 -delete -true
+ExecStart=-/usr/bin/find /var/lib/containerd -mindepth 1 -delete -true
+ExecStart=/usr/bin/truncate -s0 /var/cache/containerd/active-snapshotter
+ExecStart=/usr/bin/echo 'ACTIVE_SNAPSHOTTER="${SELECTED_SNAPSHOTTER}"'
+
+# Set the ACTIVE_SNAPSHOTTER regardless of if conditions are met for cleanup.
+# This mitigates the behavior that an unmet ExecCondition will truncate the active-snapshotter EnvironmentFile.
+ExecStopPost=/usr/bin/truncate -s0 /var/cache/containerd/active-snapshotter
+ExecStopPost=/usr/bin/echo 'ACTIVE_SNAPSHOTTER="${SELECTED_SNAPSHOTTER}"'
+
+RemainAfterExit=true
+# Write the active snapshotter.
+StandardOutput=file:/var/cache/containerd/active-snapshotter
+
+[Install]
+WantedBy=multi-user.target
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -17,6 +17,7 @@ Source96: release-repart-local.conf
 Source97: release-sysctl.conf
 Source98: release-systemd-system.conf
 Source99: release-ca-certificates-tmpfiles.conf
+Source100: release-snapshotter-tmpfiles.conf
 
 Source200: motd.template
 Source201: proxy-env
@@ -28,6 +29,7 @@ Source206: aws-config
 Source207: aws-credentials
 Source208: modules-load.template
 Source209: log4j-hotpatch-enabled
+Source210: selected-snapshotter-template
 
 # Core targets, services, and slices.
 Source1001: multi-user.target
@@ -73,6 +75,7 @@ Source1064: deprecation-warning@.timer
 Source1065: check-kernel-integrity.service
 Source1066: check-fips-modules.service
 Source1067: fips-modprobe@.service
+Source1068: configure-snapshotter.service
 
 # Mounts that require build-time edits.
 Source1080: var-lib-kernel-devel-lower.mount.in
@@ -161,6 +164,7 @@ install -d %{buildroot}%{_cross_tmpfilesdir}
 install -p -m 0644 %{S:93} %{buildroot}%{_cross_tmpfilesdir}/release.conf
 install -p -m 0644 %{S:99} %{buildroot}%{_cross_tmpfilesdir}/release-ca-certificates.conf
 install -p -m 0644 %{S:94} %{buildroot}%{_cross_tmpfilesdir}/release-fips.conf
+install -p -m 0644 %{S:100} %{buildroot}%{_cross_tmpfilesdir}/release-snapshotter.conf
 
 install -d %{buildroot}%{_cross_libdir}/systemd/networkd.conf.d
 install -p -m 0644 %{S:95} %{buildroot}%{_cross_libdir}/systemd/networkd.conf.d/80-release.conf
@@ -197,7 +201,7 @@ install -p -m 0644 \
   %{S:1040} %{S:1041} %{S:1042} %{S:1043} %{S:1044} \
   %{S:1045} %{S:1046} %{S:1047} %{S:1048} %{S:1049} \
   %{S:1060} %{S:1061} %{S:1062} %{S:1063} %{S:1064} \
-  %{S:1065} %{S:1066} %{S:1067} \
+  %{S:1065} %{S:1066} %{S:1067} %{S:1068} \
   %{buildroot}%{_cross_unitdir}
 
 install -d %{buildroot}%{_cross_unitdir}/systemd-tmpfiles-setup.service.d
@@ -256,6 +260,7 @@ install -p -m 0644 %{S:206} %{buildroot}%{_cross_templatedir}/aws-config
 install -p -m 0644 %{S:207} %{buildroot}%{_cross_templatedir}/aws-credentials
 install -p -m 0644 %{S:208} %{buildroot}%{_cross_templatedir}/modules-load
 install -p -m 0644 %{S:209} %{buildroot}%{_cross_templatedir}/log4j-hotpatch-enabled
+install -p -m 0644 %{S:210} %{buildroot}%{_cross_templatedir}/selected-snapshotter
 
 install -d %{buildroot}%{_cross_unitdir}/systemd-udev-trigger.service.d/
 install -p -m 0644 %{S:1105} %{buildroot}%{_cross_unitdir}/systemd-udev-trigger.service.d/00-selinux.conf
@@ -342,6 +347,9 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 %{_cross_templatedir}/log4j-hotpatch-enabled
 %{_cross_udevrulesdir}/61-mount-cdrom.rules
 %{_cross_datadir}/logdog.d/logdog.common.conf
+%{_cross_unitdir}/configure-snapshotter.service
+%{_cross_templatedir}/selected-snapshotter
+%{_cross_tmpfilesdir}/release-snapshotter.conf
 
 %files fips
 %{_cross_bootconfigdir}/10-fips.conf
--- a/packages/selinux-policy/rules.cil
+++ b/packages/selinux-policy/rules.cil
@@ -251,7 +251,7 @@
 (allow api_s private_t (files (mutate)))
 (allow clock_s measure_t (files (mutate)))
 (allow network_s lease_t (files (mutate)))
-(allow runtime_s cache_t (files (mutate)))
+(allow trusted_s cache_t (files (mutate)))
 
 ; Other components should not be permitted to modify these files,
 ; or to manage mounts for these directories.
--- a/packages/release/release-snapshotter-tmpfiles.conf
+++ b/packages/release/release-snapshotter-tmpfiles.conf
@@ -1,0 +1,2 @@
+d /var/cache/containerd 0755 - - -
+f /var/cache/containerd/active-snapshotter 0644 - - - ACTIVE_SNAPSHOTTER="overlayfs"
--- a/packages/release/selected-snapshotter-template
+++ b/packages/release/selected-snapshotter-template
@@ -1,0 +1,5 @@
+[required-extensions]
+container-runtime = "v1"
+std = { version = "v1" }
++++
+SELECTED_SNAPSHOTTER="{{#if settings.container-runtime.snapshotter}}{{settings.container-runtime.snapshotter}}{{else}}overlayfs{{/if}}"


packages/os/configure-snapshotter.service
```diff
@@ -0,0 +1,28 @@
...
+# This file might not exist as users don't have to set `settings.container-runtime.snapshotter`
+EnvironmentFile=-/etc/selected-snapshotter
+# Don't error if this doesn't exist yet, which it won't on first boot.
+EnvironmentFile=-/var/cache/containerd/active-snapshotter
+# Skip cleanup if selected snapshotter is empty
+ExecCondition=[ -n "${SELECTED_SNAPSHOTTER}" ]
+# Check if the active snapshotter has changed
+ExecCondition=[ "${SELECTED_SNAPSHOTTER}" != "${ACTIVE_SNAPSHOTTER}" ]
+ExecStart=find /var/lib/soci-snapshotter-grpc -mindepth 1 -delete -true
+ExecStart=find /var/lib/containerd -mindepth 1 -delete -true
```
arnaldo2792: For commands like this, use the full path to the binary.

packages/os/os.spec
```diff
@@ -424,6 +428,11 @@ Conflicts: %{_cross_os}bash
 %description -n %{_cross_os}brush
 %{summary}.
 
+%package -n %{_cross_os}configure-snapshotter
+Summary: Configure snapshotter service
+%description -n %{_cross_os}configure-snapshotter
+Service to configure snapshotter and clean up containerd and soci-snapshotter-grpc directories when snapshotter changes.
```
arnaldo2792: It is convention in bottlerocket to use the `summary` as the description of the package.

packages/os/os.spec
```diff
@@ -89,6 +92,7 @@ Requires: %{_cross_os}storewolf
 Requires: %{_cross_os}sundog
 Requires: %{_cross_os}xfscli
 Requires: %{_cross_os}thar-be-settings
+Requires: %{_cross_os}configure-snapshotter
```
arnaldo2792: Are you adding this here because the snapshotter is independent of the SOCI package or containerd? If that's the case, this will be better at `release.spec` instead of `os.spec`.

`os.spec` is mostly used to compile first party Rust. `release.spec` is where we put things that should exist in all variants.

packages/os/os.spec
```diff
@@ -424,6 +428,11 @@ Conflicts: %{_cross_os}bash
 %description -n %{_cross_os}brush
 %{summary}.
 
+%package -n %{_cross_os}configure-snapshotter
+Summary: Configure snapshotter service
```
arnaldo2792:
```suggestion
Summary: Service and configuration files to set the containerd active snapshotter
```

packages/containerd-1.7/containerd.service
```diff
@@ -3,6 +3,7 @@ Description=containerd container runtime
 Documentation=https://containerd.io
 After=network-online.target configured.target
 Wants=network-online.target configured.target
+Requires=snapshotters-configured.service
```
arnaldo2792: If you want containerd to start after `snapshotters-configured.servce`, you should add the service to the `After=` of this unit.

Also the name of your service is wrong: `configure-snapshotter.service` is what you used in the first commit.

packages/containerd-2.0/containerd.service
```diff
@@ -3,6 +3,7 @@ Description=containerd container runtime
 Documentation=https://containerd.io
 After=network-online.target configured.target
 Wants=network-online.target configured.target
+Requires=snapshotters-configured.service
```
arnaldo2792: Same as elsewhere regarding ordering and the name of the service.

packages/selinux-policy/rules.cil
```diff
@@ -246,6 +246,7 @@
 (allow clock_s measure_t (files (mutate)))
```
arnaldo2792, line 252: For the lack of a better place to put this, your commit message refers to a label that doesn't exist (`process_s`). I think you meant `trusted_s`.

packages/os/os.spec
bcressey: +1, definitely put this in `release.spec`. I cringe whenever anything triggers a rebuild for `os`.

packages/selinux-policy/rules.cil
```diff
@@ -246,6 +246,7 @@
...
 (allow network_s lease_t (files (mutate)))
 (allow runtime_s cache_t (files (mutate)))
+(allow trusted_s cache_t (files (mutate)))
```
bcressey, line 254: You can change the line above from `runtime_s` to `trusted_s` (which includes `runtime_s`).

packages/os/configure-snapshotter.service
```diff
@@ -0,0 +1,28 @@
+[Unit]
+Description=Configure Snapshotter
+Before=containerd.service
+After=settings-applier.service bootstrap-commands.service
+Requires=settings-applier.service bootstrap-commands.service
+
+[Service]
+Type=oneshot
+# This file might not exist as users don't have to set `settings.container-runtime.snapshotter`
+EnvironmentFile=-/etc/selected-snapshotter
```
bcressey: Let's put this under `/etc/containerd/selected-snapshotter`.

packages/os/configure-snapshotter.service
```diff
@@ -0,0 +1,28 @@
...
+# Don't error if this doesn't exist yet, which it won't on first boot.
+EnvironmentFile=-/var/cache/containerd/active-snapshotter
+# Skip cleanup if selected snapshotter is empty
+ExecCondition=[ -n "${SELECTED_SNAPSHOTTER}" ]
+# Check if the active snapshotter has changed
+ExecCondition=[ "${SELECTED_SNAPSHOTTER}" != "${ACTIVE_SNAPSHOTTER}" ]
+# Don't error if /var/lib/soci-snapshotter doesn't exist, which it may not if soci is not selected as snapshotter.
+ExecStart=-/x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/find /var/lib/soci-snapshotter -mindepth 1 -delete -true
+ExecStart=-/x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/find /var/lib/containerd -mindepth 1 -delete -true
```
bcressey: Need to remove the sysroot prefix or else it won't work on aarch64:
```suggestion
# Don't error if the directories don't exist.
ExecStart=-/usr/bin/find /var/lib/soci-snapshotter /var/lib/containerd -mindepth 1 -delete -true
```

packages/os/configure-snapshotter.service
```diff
@@ -0,0 +1,28 @@
+[Unit]
+Description=Configure Snapshotter
+Before=containerd.service
+After=settings-applier.service bootstrap-commands.service
+Requires=settings-applier.service bootstrap-commands.service
```
bcressey: You don't need these dependencies since you're running in `multi-user.target`; they'll be completed by the time this unit is queued to start.

packages/os/configure-snapshotter.service
```diff
@@ -0,0 +1,28 @@
...
+
+[Service]
+Type=oneshot
+# This file might not exist as users don't have to set `settings.container-runtime.snapshotter`
+EnvironmentFile=-/etc/selected-snapshotter
+# Don't error if this doesn't exist yet, which it won't on first boot.
+EnvironmentFile=-/var/cache/containerd/active-snapshotter
```
bcressey: nit: not really true since we create it via `tmpfiles`

ginglis13:
^ force pushes addresses feedback:

* move `configure-snapshotter.service` to the `release` package
* update some incorrect comments
* put selected snapshotter under /etc/containerd/selected-snapshotter
* remove sysroot prefix from `find` absolute path
* remove the dependencies `settings-applier.service` and `bootstrap-commands.service` from `configure-snapshotter.service`
* replace with `runtime_s cache_t` rule with `trusted_s cache_t` rule given `trusted_s` includes `runtime_s`


Additionally, provide `soci-snapshotter` in `release` if variant is k8s.

packages/release/release.spec
```diff
@@ -149,6 +154,11 @@ Requires: %{_cross_os}libkcapi
 %description fips
 %{summary}.
 
+%package -n %{_cross_os}configure-snapshotter
+Summary: Service and configuration files to set the containerd active snapshotter
+%description -n %{_cross_os}configure-snapshotter
+%{summary}.
+
```
bcressey: I don't think we need a separate subpackage for this.

ginglis13:
^ remove configure-snapshotter subpackage

packages/release/configure-snapshotter.service
```diff
@@ -0,0 +1,25 @@
...
+Description=Configure Snapshotter
+Before=containerd.service
+
+[Service]
+Type=oneshot
+# This file might not exist as users don't have to set `settings.container-runtime.snapshotter`
+EnvironmentFile=-/etc/containerd/selected-snapshotter
+EnvironmentFile=-/var/cache/containerd/active-snapshotter
+# Skip cleanup if selected snapshotter is empty
+ExecCondition=[ -n "${SELECTED_SNAPSHOTTER}" ]
```
bcressey: For additional defense - I'm not sure how this condition would arise, but it seems safest to proceed only if we have both values and we know they're different:
```suggestion
# Skip cleanup if either snapshotter variable is empty
ExecCondition=[ -n "${SELECTED_SNAPSHOTTER}" ]
ExecCondition=[ -n "${ACTIVE_SNAPSHOTTER}" ]
```

packages/release/configure-snapshotter.service
```diff
@@ -0,0 +1,25 @@
...
+# Check if the active snapshotter has changed
+ExecCondition=[ "${SELECTED_SNAPSHOTTER}" != "${ACTIVE_SNAPSHOTTER}" ]
+# Don't error if the directories don't exist.
+ExecStart=-/usr/bin/find /var/lib/soci-snapshotter -mindepth 1 -delete -true
+ExecStart=-/usr/bin/find /var/lib/containerd -mindepth 1 -delete -true
+ExecStart=echo 'ACTIVE_SNAPSHOTTER="${SELECTED_SNAPSHOTTER}"'
+
+# Skip cleanup if no change detected
+RemainAfterExit=true
```
bcressey: comment doesn't apply to this directive:
```suggestion
RemainAfterExit=true
```

packages/release/release.spec
```diff
@@ -197,7 +202,7 @@ install -p -m 0644 \
   %{S:1040} %{S:1041} %{S:1042} %{S:1043} %{S:1044} \
   %{S:1045} %{S:1046} %{S:1047} %{S:1048} %{S:1049} \
   %{S:1060} %{S:1061} %{S:1062} %{S:1063} %{S:1064} \
-  %{S:1065} %{S:1066} %{S:1067} \
+  %{S:1065} %{S:1066} %{S:1067} %{S:1068}\
```
bcressey:
```suggestion
  %{S:1065} %{S:1066} %{S:1067} %{S:1068} \
```

bcressey (approved):

packages/release/release.spec
```diff
@@ -104,6 +107,7 @@ Source1500: bootconfig-fips.conf
 
 Requires: %{_cross_os}audit
 Requires: %{_cross_os}chrony
+Requires: %{_cross_os}configure-snapshotter
```
ginglis13: Will remove - now that we're not providing the subpackage this will result in build failures.

ginglis13:
^ drop `[skip ci]`, some fixups for configure-snapshotter.service and the release spec

packages/release/configure-snapshotter.service
```diff
@@ -0,0 +1,25 @@
...
+EnvironmentFile=-/var/cache/containerd/active-snapshotter
+# Skip cleanup if either snapshotter variable is empty
+ExecCondition=[ -n "${SELECTED_SNAPSHOTTER}" ]
+ExecCondition=[ -n "${ACTIVE_SNAPSHOTTER}" ]
+# Check if the active snapshotter has changed
+ExecCondition=[ "${SELECTED_SNAPSHOTTER}" != "${ACTIVE_SNAPSHOTTER}" ]
+# Don't error if the directories don't exist.
+ExecStart=-/usr/bin/find /var/lib/soci-snapshotter -mindepth 1 -delete -true
+ExecStart=-/usr/bin/find /var/lib/containerd -mindepth 1 -delete -true
+ExecStart=echo 'ACTIVE_SNAPSHOTTER="${SELECTED_SNAPSHOTTER}"'
```
ginglis13:
```suggestion
ExecStart=/usr/bin/echo 'ACTIVE_SNAPSHOTTER="${SELECTED_SNAPSHOTTER}"'
```

packages/release/configure-snapshotter.service
ginglis13: I encountered a bug in testing today where snapshotter state wasn't being cleaned up on reboot in between changes to snapshotter selection and I've narrowed it down to this change. It turns out, `ExecCondition`s open the `StandardOutput` file for writing - from the [man page for systemd exec](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html):

> For units with multiple command lines, e.g. Type=oneshot services with multiple ExecStart=, or services with ExecCondition=, ExecStartPre= or ExecStartPost=, the output file is reopened and therefore re-truncated for each command line.

These `ExecCondition`s were not being met on first boot if `$SELECTED_SNAPSHOTTER` is not set (`$ACTIVE_SNAPSHOTTER` was set via the tmpfiles, but settings allow for `settings.container-runtime.snapshotter` to be unset), resulting in `$ACTIVE_SNAPSHOTTER` never being set and `/var/cache/containerd/active-snapshotter` being overwitten. From there on, the evaluation of 

```
ExecCondition=[ -n "${ACTIVE_SNAPSHOTTER}" ]
```

will never hold.

ginglis13:
^ force push modifies `configure-snapshotter.service`:
* writes StandardOutput to `journal+console` rather than `/var/cache/containerd/active-snapshotter`
* Use an `ExecStartPost` to copy `/etc/containerd/selected-snapshotter` to `/var/cache/containerd/active-snapshotter`

ginglis13:
^ force push fixes a bug in the previous fix where `ACTIVE_SNAPSHOTTER` wasn't being set. This reverts back to the `truncate` approach to write to the active snapshotter `EnvironmentFile`, but now runs the echo to set `ACTIVE_SNAPSHOTTER="${SELECTED_SNAPSHOTTER}"` as an `ExecStopPost` so that this is run regardless of if `ExecCondition`s are met

ginglis13:
^ rebase for SDK bump + clippy lint fixes

ginglis13:
^ force push adjusts to writing stdout to `file:/` rather than truncating. Testing details can be found at https://gist.github.com/ginglis13/cfc7d1d907e65d889af9b55923476e12

ginglis13:
^ force push `truncate`s the active snapshotter environment file before writing.

ginglis13:
^ adjust soci state path in commit message

arnaldo2792 (approved):

PR #584, opened 2025-07-15
soci-snapshotter: inherit proxy env vars

### LINKED ISSUE #583: Ensure soci-snapshotter inherits proxy environment variables
**What I'd like:**

soci-snapshotter to inherit proxy environment variables via:

```
EnvironmentFile=/etc/network/proxy.env
```

in its service file.

Keep the status quo

ginglis13:
**Issue number:**

Closes #583

**Description of changes:**

Provide proxy environment variables to soci-snapshotter's systemd so these variables can be used for image pulls.

**Terms of contribution:**

### DIFF
--- a/packages/soci-snapshotter/soci-snapshotter.service
+++ b/packages/soci-snapshotter/soci-snapshotter.service
@@ -7,6 +7,7 @@ Before=containerd.service
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
 ExecStart=/usr/bin/soci-snapshotter-grpc --address fd://
 Restart=always
 RestartSec=5


packages/soci-snapshotter/soci-snapshotter.service
```diff
@@ -7,6 +7,7 @@ Before=containerd.service
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
```
arnaldo2792, line 10: I wonder if we should consider using a system-wide config for the proxy, to reduce the risk of missing it in a service where it must be used.

arnaldo2792 (approved):

bcressey (approved):

PR #590, opened 2025-07-17
add release subpackage to enable swap

bcressey:
**Issue number:**
Related: bottlerocket-os/bottlerocket#4075

**Description of changes:**
In low memory, pre-OOM-kill conditions, an orchestrator agent like kubelet that is responsible for freeing up system resources may be unable to terminate a process that allocates too much memory, if its clean pages are dropped from the cache and must be read back from the filesystem's block device whenever it runs, only to have those pages dropped again when the scheduler selects another process to run.

For storage devices with metered I/O, such Amazon EBS gp2 volumes, this cycle of reading the same filesystem blocks over and over can exhaust the available credits, and leave the system livelocked for for hours or days.

Providing a swap device gives the kernel an alternative to evicting clean pages to free up memory prior to invoking the OOM killer: it can swap out anonymous memory instead.

If the swap device were also stored on a device with I/O limits, the same failure mode might arise when faulting in pages for anonymous memory.

Using zram for the swap device avoids this problem. Instead of being swapped to disk, the anonymous memory pages are compressed, and the savings relative to the uncompressed size are returned to the system. The zram device is not used until physical memory is exhausted; it has an overhead of around 1%, or 1 MiB for the 1 GiB reservation.

Enabling zram-backed swap may become the default for Bottlerocket in the future. For now, add the logic to a new subpackage so variants can opt in by specifying "release-swap" instead of "release" in their package list.

**Testing done:**
Verified that the service started up as expected:
```
systemctl status prepare-swap
● prepare-swap.service - Prepare Swap
     Loaded: loaded (/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/prepare-swap.service; enabled; preset: enabled)
    Drop-In: /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/service.d
             └─00-aws-config.conf
     Active: active (exited) since Thu 2025-07-17 21:30:17 UTC; 2min 27s ago
   Main PID: 1287 (code=exited, status=0/SUCCESS)
        CPU: 19ms

Jul 17 21:30:17 localhost prepare-swap[727]: Setting up swapspace version 1, size = 1024 MiB (1073737728 bytes)
Jul 17 21:30:17 localhost prepare-swap[727]: no label, UUID=00000000-0000-0000-0000-000000000000
Jul 17 21:30:17 localhost systemd[1]: Finished Prepare Swap.
```

**Terms of contribution:**

### DIFF
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -9,15 +9,22 @@ Summary: Bottlerocket release
 License: Apache-2.0 OR MIT
 URL: https://github.com/bottlerocket-os/bottlerocket
 
+# Core config files.
 Source11: nsswitch.conf
+
+# Sysctl drop-ins.
+Source80: release-sysctl.conf
+Source81: release-swap-sysctl.conf
+
+# Other drop-ins.
 Source93: release-tmpfiles.conf
 Source94: release-fips-tmpfiles.conf
 Source95: release-systemd-networkd.conf
 Source96: release-repart-local.conf
-Source97: release-sysctl.conf
 Source98: release-systemd-system.conf
 Source99: release-ca-certificates-tmpfiles.conf
 
+# Templates for the settings API.
 Source200: motd.template
 Source201: proxy-env
 Source202: hostname-env
@@ -63,6 +70,7 @@ Source1046: mask-local-var.service
 Source1047: repart-data-preferred.service
 Source1048: repart-data-fallback.service
 Source1049: prepare-local-fs.service
+Source1050: prepare-swap.service
 
 # Feature-specific units.
 Source1060: capture-kernel-dump.service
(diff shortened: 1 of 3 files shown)


tzneal (approved):
lgtm

packages/release/release.spec
```diff
@@ -149,6 +157,14 @@ Requires: %{_cross_os}libkcapi
 %description fips
 %{summary}.
 
+%package swap
+Summary: Bottlerocket release, with zram-based swap
+Requires: %{name}
+Requires: (%{name}-fips if %{_cross_os}image-feature(fips))
+
+%description swap
+%{summary}.
```
bcressey, line 166: This approach seems OK to me, though it wouldn't scale very well to other optional features.

We could think about a Twoliter feature to let kits declare their own features, so for example we could list features like "swap" in the core kit's manifest, and then add this to the top-level package:
```
Requires: (%{name}-swap if %{_cross_os}kit-feature(%{_cross_os}core-kit.swap-enabled))
```

And then variants would do something like this:
```
[package.metadata.build-variant.kit-features.bottlerocket]
bottlerocket-core-kit.swap-enabled = true
```

Something to think about 🤔

yeazelm (approved):

arnaldo2792 (approved):

PR #603, opened 2025-07-23
chore: bump bottlerocket-settings-models to v0.12.0

### LINKED ISSUE #602: Bump bottlerocket-settings-models to v0.12.0
**What I'd like:**

Bump to the latest release of the bottlerocket-settings-sdk https://github.com/bottlerocket-os/bottlerocket-settings-sdk/releases/tag/bottlerocket-settings-models-v0.12.0

Keep the status quo

ginglis13:
**Issue number:**

Closes #602

**Description of changes:**

Bump to the latest version of the bottlerocket-settings-sdk https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/602

**Testing done:**

`make ARCH=x86_64 && make ARCH=aarch64`

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -211,13 +211,13 @@ base64 = "0.22"
 
 [workspace.dependencies.bottlerocket-modeled-types]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.11.0"
+tag = "bottlerocket-settings-models-v0.12.0"
 version = "0.10.0"
 
 [workspace.dependencies.bottlerocket-settings-models]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.11.0"
-version = "0.11.0"
+tag = "bottlerocket-settings-models-v0.12.0"
+version = "0.12.0"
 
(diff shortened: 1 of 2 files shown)


sumukhballal (approved):

sources/Cargo.toml
```diff
@@ -226,7 +226,7 @@ version = "0.1.0"
 
 [workspace.dependencies.settings-extension-oci-defaults]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
-tag = "bottlerocket-settings-models-v0.11.0"
+tag = "bottlerocket-settings-models-v0.12.0"
```
ytsssun, line 229: I wonder why we didn't explicitly refer to settings-extension and its subpackages like `settings-extension-container-runtime` here. 

And if we did, those would need to be bumped. But I confirmed that from the Cargo.lock introduced in this PR that those versions are correctly updated. Approving.

ytsssun (approved):

sources/Cargo.toml
ginglis13: It looks like we actually depend on the oci-defaults here: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/4bb9a8ad831aec5018ec8903f8377de942326387/sources/api/schnauzer/Cargo.toml#L46

PR #610, opened 2025-07-27
[ecs-agent] Fix ECS_DISABLE_PRIVILEGED environment variable

vermdeep:
**Issue number:**

Closes #

**Description of changes:**
Fixes privileged mode in ECS agent. Currently, incorrect environment variable is being used.
Ref - https://github.com/aws/amazon-ecs-agent?tab=readme-ov-file#environment-variables

**Testing done:** Yes

**Terms of contribution:**

### DIFF
--- a/packages/ecs-agent/ecs-base-conf
+++ b/packages/ecs-agent/ecs-base-conf
@@ -24,7 +24,7 @@ Environment=ECS_TASK_METADATA_RPS_LIMIT="{{ecs_metadata_service_limits settings.
 Environment=ECS_WARM_POOLS_CHECK="{{default "false" settings.autoscaling.should-wait}}"
 
 # Boolean configurations whose values are inverted in the API
-Environment=ECS_PRIVILEGED_DISABLED="{{negate_or_else true settings.ecs.allow-privileged-containers}}"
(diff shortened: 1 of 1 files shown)


arnaldo2792 (approved):

ytsssun:
@vermdeep Thanks for the fix!

ytsssun (approved):

koooosh (approved):
Went through the other vars in this file and noticed that [ECS_BACKEND_HOST](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/ecs-agent/ecs-base-conf#L11) is not mentioned in the ecs-agent reference.

However, this looks to be a missing documentation issue in the reference rather than incorrectly named var here like ECS_DISABLE_PRIVILEGED

koooosh:
Pulled in change and successfully launched an `aws-ecs-2` variant instance with defaults:
```
[ssm-user@control]$ apiclient get settings.ecs.allow-privileged-containers
{
  "settings": {
    "ecs": {
      "allow-privileged-containers": false
    }
  }
}

bash-5.1# cat usr/share/templates/ecs-base-conf | grep PRIVILEGE
Environment=ECS_DISABLE_PRIVILEGED="{{negate_or_else true settings.ecs.allow-privileged-containers}}"
```

PR #615, opened 2025-08-05
docs: remove OCI consideration from BUILDING.md

### LINKED ISSUE #591: BUILDING.md implies docker isn't required when crane is available but this isn't true.
the phrasing in BUILDING.md implies you can pick either `docker` or `crane` to build your artifacts, but the current build system requires docker to function and i assume `crane` just removes the need for docker's `containerd-snapshotter` driver. if so, then this bit
https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/d8bdb3c0df0c345b73e14d2880bee853a3d48535/BUILDING.md?plain=1#L35-L37
could be changed to something like:
```markdown
Building a kit results in building OCI artifacts, there are two ways to build these artifacts: `docker` with `crane`, or with the `containerd-snapshotter` driver enabled. 
  
 We recommend using `crane` (and `krane`) over Docker's experimental `containerd-snapshotter` driver  as it has shown better performance in our testing.
```

should i make a pr with this change?

cbgbt:
**Issue number:**

Closes #591

**Description of changes:**
Twoliter now handles OCI fetching internally -- there is no need for builders to consider the OCI toolchain.

**Terms of contribution:**

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -30,32 +30,9 @@ To get it, run:
 cargo install cargo-make
 ```
 
-### OCI Artifacts
-
-Building a kit results in building OCI artifacts, there are two ways to build these artifacts: `crane` or `docker` with the `containerd-snapshotter` feature enabled.
-
-We recommend using `crane` (and `krane`) over `docker` as it has shown better performance in our testing.
-
 #### Docker
-We recommend [Docker](https://docs.docker.com/install/#supported-platforms) 20.10.10 or later. The default seccomp policy of older versions of Docker do not support the `clone3` syscall in recent versions of Fedora or Ubuntu, on which the Bottlerocket SDK is based.
-Builds rely on Docker's integrated BuildKit support, which has received many fixes and improvements in newer versions.
-
 You'll need to have Docker installed and running, with your user account added to the `docker` group.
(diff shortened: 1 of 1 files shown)


bcressey (approved):

bcressey:
We'll need a similar change over in the kernel kit.

BUILDING.md
```diff
@@ -30,33 +30,6 @@ To get it, run:
 cargo install cargo-make
 ```
 
-### OCI Artifacts
-
-Building a kit results in building OCI artifacts, there are two ways to build these artifacts: `crane` or `docker` with the `containerd-snapshotter` feature enabled.
-
-We recommend using `crane` (and `krane`) over `docker` as it has shown better performance in our testing.
-
-#### Docker
```
jmt-lab, line 39: I can get the above, is there a reason we are removing the docker part entirely, like i don't think we need the daemon.json anymore due to modern docker and our move to krane but docker is still a build requirement that is needed.

koooosh (approved):

BUILDING.md
cbgbt: @jmt-lab totally true, mind taking another look?

arnaldo2792 (approved):

PR #619, opened 2025-08-08
always lowercase the hostnames provided by the hostname helpers

tzneal:
**Issue number:**

Closes #

**Description of changes:**

The DHCP RFC doesn't require that the domain name option 15 be lowercase, which can result in hostname that has uppercase letters. This isn't a valid hostname and causes setting deserialization to fail.

**Testing done:**

- Unit testing
- Built an AMI with this core-kit and a VPC configured with DHCP Options using a capital letter domain name. Nodes went ready and Pods scheduled and went running.

**Terms of contribution:**

### DIFF
--- a/sources/imdsclient/src/lib.rs
+++ b/sources/imdsclient/src/lib.rs
@@ -242,8 +242,8 @@ impl ImdsClient {
         Ok(self.fetch_string(&hostname_target).await?.and_then(|h| {
             h.split_whitespace()
                 .next()
-                .map(|h| h.trim_end_matches('.'))
-                .map(String::from)
+                // trim any trailing dots and lowercase the result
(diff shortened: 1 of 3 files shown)


sources/imdsclient/src/lib.rs
```diff
@@ -842,8 +843,9 @@ mod test {
         let base_uri = format!("http://{}", server.addr());
         let token = "some+token";
         let response_code = 200;
+        // shoudl always lowercase the response
```
yeazelm: nit: 
```suggestion
        // should always lowercase the response
```

yeazelm (approved):

tzneal:
Sorry for the churn, it passes `make twoliter check-fmt` now.

ytsssun (approved):

bcressey (approved):

PR #632, opened 2025-08-19
apiclient: add implicit default behavior for ephemeral-storage bind

### LINKED ISSUE #630: Add default behavior for `apiclient ephemeral-storage bind`
**What I'd like:**

Modify the `apiclient ephemeral-storage bind` command to automatically use platform-specific default directories when no explicit directories are specified.

Currently, users must always specify directories when binding ephemeral storage using the `--dirs` option. Products/services that take a dependency on the allowlist of dirs must keep up-to-date with this list (e.g. Karpenter https://github.com/aws/karpenter-provider-aws/blob/a843836fe85b8e0f82875beacffb68ad226a32a7/pkg/providers/amifamily/bootstrap/bottlerocket.go#L94-L103). If `apiclient ephemeral-storage bind` defaults to using all platform-specific allowlisted directories, then as Bottlerocket extends its allowlist, downstream dependencies of this command won't have to as well; they can simply continue to `apiclient ephemeral-storage bind`

1. Provide this behavior via explicit flag such as `--default-dirs`
2. Or, keep the status quo

ginglis13:
**Issue number:**

Closes #630

**Description of changes:**

Add implicit default behavior where running `apiclient ephemeral-storage bind` without `--dirs` automatically uses platform-specific default directories.

Update help documentation to reflect the implicit default behavior where running `ephemeral-storage bind` without `--dirs` uses platform defaults.

**Testing done:**

Unit tests; Manual testing:

User data:

```
[settings.bootstrap-commands.ephemeral-storage]
commands = [
    ["apiclient", "ephemeral-storage" ,"bind"]
]
essential = true
mode = "always"
```

k8s-1.33 variant `lsblk` output:

```
...
nvme3n1      259:16   0  6.8T  0 disk
`-md127        9:127  0 54.6T  0 raid0 /var/lib/kubelet
                                       /var/lib/soci-snapshotter
                                       /var/log/pods
                                       /var/lib/containerd
                                       /var/lib/host-containerd
                                       /mnt/.ephemeral

...
```

ecs-2 variant `lsblk` output:

```
...
nvme4n1      259:23   0  6.8T  0 disk
`-md127        9:127  0 54.6T  0 raid0 /var/lib/docker
                                       /var/log/ecs
                                       /var/lib/host-containerd
                                       /var/lib/containerd
                                       /mnt/.ephemeral
...
```

**Terms of contribution:**

### DIFF
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -272,8 +272,9 @@ fn usage() -> ! {
 
         ephemeral-storage bind options:
             --dirs DIR [DIR ...]       Directories to bind to configured ephemeral storage
-                                       (e.g. /var/lib/containerd). If no ephemeral disks are found
-                                       this operation does nothing.
+                                       (e.g. /var/lib/containerd). If not specified, uses platform (k8s vs. ECS)
+                                       defaults. If no ephemeral disks are found this operation does nothing.
+
 
         ephemeral-storage list-disks options:
             -f, --format               Format of the disk listing (text or json). Default format is text.
@@ -794,9 +795,13 @@ fn parse_ephemeral_storage_init_args(args: Vec<String>) -> EphemeralStorageSubco
 
 /// Parses arguments for the 'bind' ephemeral-storage subcommand.
 fn parse_ephemeral_storage_bind_args(args: Vec<String>) -> EphemeralStorageSubcommand {
+    // If no arguments, use default directories
     if args.is_empty() {
-        usage_msg("Did not give arguments to bind")
+        return EphemeralStorageSubcommand::Bind(EphemeralStorageBindArgs {
+            targets: Vec::new(),
+        });
     }
+
     let mut targets = Vec::new();
     let mut iter = args.into_iter().peekable();
     while let Some(arg) = iter.next() {
@@ -810,6 +815,7 @@ fn parse_ephemeral_storage_bind_args(args: Vec<String>) -> EphemeralStorageSubco
             x => usage_msg(format!("Unknown argument '{x}'")),
         }
     }
+
     EphemeralStorageSubcommand::Bind(EphemeralStorageBindArgs { targets })
 }
 
--- a/sources/api/apiserver/src/server/ephemeral_storage.rs
+++ b/sources/api/apiserver/src/server/ephemeral_storage.rs
@@ -118,6 +118,17 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
         _ => format!("{RAID_DEVICE_DIR}{RAID_DEVICE_NAME}"),
     };
 
+    let dirs = if dirs.is_empty() {
+        let allowed_dirs = allowed_bind_dirs(variant);
+        allowed_dirs
+            .allowed_exact
+            .into_iter()
+            .map(|s| s.to_string())
+            .collect()
+    } else {
+        dirs
+    };
+
     // Normalize input by trimming trailing "/"
     let dirs: Vec<String> = dirs
         .into_iter()
@@ -431,3 +442,33 @@ pub mod error {
 }
 
 pub type Result<T> = std::result::Result<T, error::Error>;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_bind_with_default_dirs_k8s() {
+        let variant = "aws-k8s-1.33";
+        let allowed_dirs = allowed_bind_dirs(variant);
+
+        for dir in [
+            "/var/lib/kubelet",
+            "/var/lib/containerd",
+            "/var/lib/soci-snapshotter",
+            "/var/log/pods",
+        ] {
+            assert!(allowed_dirs.allowed_exact.contains(dir));
+        }
+    }
+
+    #[test]
+    fn test_bind_with_default_dirs_ecs() {
+        let variant = "aws-ecs-2";
+        let allowed_dirs = allowed_bind_dirs(variant);
+
+        for dir in ["/var/lib/docker", "/var/lib/containerd", "/var/log/ecs"] {
+            assert!(allowed_dirs.allowed_exact.contains(dir));
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -431,3 +442,43 @@ pub mod error {
...
+            .allowed_exact
+            .into_iter()
+            .map(|s| s.to_string())
+            .collect();
+
+        // Verify that k8s variant includes kubelet directory
+        assert!(expected_dirs.contains(&"/var/lib/kubelet".to_string()));
+        assert!(expected_dirs.contains(&"/var/lib/containerd".to_string()));
+        assert!(expected_dirs.contains(&"/var/lib/soci-snapshotter".to_string()));
+        assert!(expected_dirs.contains(&"/var/log/pods".to_string()));
```
arnaldo2792: Why the `to_string()` and then pass the reference of the `String`? Why not just `"/string"`?

ginglis13: `expected_dirs` is a `Vec<String>`. The `contains` method expects a reference to the container's type, so in this case, we must use `&String`, not `&str`, which are technically two different types

piyush-jena (approved):

vigh-m (approved):

sources/api/apiclient/src/main.rs
```diff
@@ -794,9 +795,11 @@ fn parse_ephemeral_storage_init_args(args: Vec<String>) -> EphemeralStorageSubco
 
 /// Parses arguments for the 'bind' ephemeral-storage subcommand.
 fn parse_ephemeral_storage_bind_args(args: Vec<String>) -> EphemeralStorageSubcommand {
+    // If no arguments, use default directories
     if args.is_empty() {
-        usage_msg("Did not give arguments to bind")
+        return EphemeralStorageSubcommand::Bind(EphemeralStorageBindArgs { targets: None });
```
ytsssun: Can we just keep `targets` Vec<String> and just return empty vec here instead of wrapping it with the extra Option?

sources/api/apiclient/src/main.rs
```diff
@@ -272,8 +272,9 @@ fn usage() -> ! {
 
         ephemeral-storage bind options:
             --dirs DIR [DIR ...]       Directories to bind to configured ephemeral storage
-                                       (e.g. /var/lib/containerd). If no ephemeral disks are found
-                                       this operation does nothing.
+                                       (e.g. /var/lib/containerd). If not specified, uses platform
+                                       defaults. If no ephemeral disks are found this operation does nothing.
```
ytsssun: It seems that the "platform defaults" means more for ECS/EKS variants? Could we be more specific here?

If you don't think this is the best place, we might need to call this out somewhere else in the README.

ytsssun: Not blocking: I also realize that we owe an update to the readme https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/a54c56e11dd357e846a2450d89fdbac6401122a0/sources/api/apiclient/src/lib.rs#L1-L3 to include the ephemeral-storage subcommand there?

sources/api/apiclient/src/main.rs
```diff
@@ -1064,7 +1074,8 @@ async fn run() -> Result<()> {
                 .context(error::EphemeralStorageSnafu)?;
             }
             EphemeralStorageSubcommand::Bind(bind_args) => {
-                ephemeral_storage::bind(&args.socket_path, bind_args.targets)
+                let targets = bind_args.targets.unwrap_or_default();
```
ytsssun: Looks like we want empty Vec anyway? To my question above.

ytsssun (approved):

sources/api/apiserver/src/server/ephemeral_storage.rs
KCSesh: You should be able to do something like - to clean it up a bit: 

```
for dir in [
        "/var/lib/kubelet",
        "/var/lib/containerd",
        "/var/lib/soci-snapshotter",
        "/var/log/pods"
] {
    assert!(allowed_dirs.allowed_exact.contains(dir));
}
```

Less String wrangling

ginglis13: @KCSesh this has the same `&str` vs `&String` issue

sources/api/apiclient/src/main.rs
ginglis13: updated the helptext and added a commit w/ README changes for `ephemeral-storage`

sources/api/apiclient/src/main.rs
ginglis13: Good callout, latest push takes this suggestion

ginglis13:
^ force push moves the README changes over to the `.tpl` and includes generated README

sources/api/apiclient/src/main.rs
```diff
@@ -1064,7 +1070,8 @@ async fn run() -> Result<()> {
                 .context(error::EphemeralStorageSnafu)?;
             }
             EphemeralStorageSubcommand::Bind(bind_args) => {
-                ephemeral_storage::bind(&args.socket_path, bind_args.targets)
+                let targets = bind_args.targets;
+                ephemeral_storage::bind(&args.socket_path, targets)
```
ytsssun: Nit: can drop the local variable `target` here? Essentially keep the original code.

ytsssun (approved):

ytsssun (approved):

sources/api/apiserver/src/server/ephemeral_storage.rs
ginglis13: Ah I mistook that you've dropped `expected_dirs` and are running this check directly on `allowed_exact` - yup this works :D

ginglis13:
^ force push takes @KCSesh suggestion for simpler unit test syntax

KCSesh (approved):

ytsssun:
> ^ force push takes @KCSesh suggestion for simpler unit test syntax

Might be worth mentioning is that one trick I learnt from @cbgbt is that, we could use the [#test_case](https://docs.rs/test-case/latest/test_case/) macro for tests like this where you have different test inputs that runs the same function.

PR #641, opened 2025-08-26
kubernetes: update kubelet config to use new static pods setting

rpkelly:
**Description of changes:**
Makes use of the new enable-static-pods setting to allow static pods to be disabled with an explicit false setting. If unset, defaults to true.

**Testing done:**

1. Launched without setting and confirmed Kubelet set static pod path
2. Launched with setting explicitly enabled and confirmed Kubelet set static pod path
3. Launched with setting explicitly disabled and confirmed Kubelet did not set static pod path

**Terms of contribution:**


vigh-m (approved):

ytsssun (approved):

KCSesh:
Nice usage of `if_not_null` 👍

PR #644, opened 2025-08-27
Update ghostdog to handle multiple neuron drivers

yeazelm:
**Issue number:**

Related to https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/238

**Description of changes:**
This adds a `match-driver` command to ghostdog which will eventually replace `match-nvidia-driver` as a generic command for multiple drivers. This allows neuron devices to be chosen in the same way that NVIDIA drivers are based upon the devices found on the PCI bus.

This is useful for inf1 instances where they will need to remain on 2.21.* but newer instances would benefit from a newer driver. This will allow Bottlerocket to choose the best driver for the particular instance it is launched on. This change relies upon a change in the kernel kit to vend this additional driver and call `driverdog` to actually do the loading.

See https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/277 for the other side of this change. 

**Testing done:**
Booted an image with this and https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/277 and booted on both inf1 and trn1 instances. The correct driver is loaded on each.

### on `trn1.2xlarge'
```
[ssm-user@control]$ apiclient exec admin modinfo neuron
filename:       /lib/modules/6.12.40/neuron_latest/neuron.ko
alias:          pci:v00001d0fd00007064sv*sd*bc*sc*i*
version:        2.23.9.0
license:        GPL
description:    Neuron Driver, built from SHA: b8ad53947481bbe9d8e76b117c06b2484b8a949f
import_ns:      DMA_BUF
srcversion:     68886FE4A17A11B0934DAFE
depends:
name:           neuron
retpoline:      Y
vermagic:       6.12.40 SMP preempt mod_unload modversions
```

### on `inf1.xlarge`
```
[ssm-user@control]$ apiclient exec admin modinfo neuron
filename:       /lib/modules/6.12.40/neuron_2_21/neuron.ko
alias:          pci:v00001d0fd00007064sv*sd*bc*sc*i*
version:        2.21.37.0
license:        GPL
description:    Neuron Driver, built from SHA: fd7535995e4b3ae73c543f829b152c3351ef21f6
import_ns:      DMA_BUF
srcversion:     4DE159506B58B4692C035AA
depends:
name:           neuron
retpoline:      Y
```

**Terms of contribution:**

### DIFF
--- a/packages/release/drivers.target
+++ b/packages/release/drivers.target
@@ -1,0 +1,9 @@
+[Unit]
+Description=Driver units
+AllowIsolate=yes
+After=basic.target
+Before=preconfigured.target
+Requires=basic.target
+
+[Install]
+RequiredBy=preconfigured.target multi-user.target
--- a/sources/ghostdog/src/error.rs
+++ b/sources/ghostdog/src/error.rs
@@ -66,6 +66,18 @@ pub(super) enum Error {
         requested: String,
         preferred: String,
     },
+    #[snafu(display("{driver} is not a supported driver"))]
+    UnsupportedDriver { driver: String },
+    #[snafu(display("{flavor} is not a supported choice for {driver}"))]
+    UnsupportedDriverFlavor { driver: String, flavor: String },
+    #[snafu(display("Failed to check if this is an inf1 instance: {}", source))]
+    CheckInf1Failure { source: pciclient::PciClientError },
+    #[snafu(display("Failed to check if this is an inf2+ instance: {}", source))]
+    CheckInf2Failure { source: pciclient::PciClientError },
+    #[snafu(display("Did not detect inf1 hardware"))]
+    NoInf1Present,
+    #[snafu(display("Did not detect inf2+ hardware"))]
+    NoInf2Present,
 }
 
 pub(crate) type Result<T> = std::result::Result<T, Error>;
--- a/sources/ghostdog/src/main.rs
+++ b/sources/ghostdog/src/main.rs
@@ -58,6 +58,7 @@ enum SubCommand {
     EbsDeviceName(EbsDeviceNameArgs),
     EfaPresent(EfaPresentArgs),
     NeuronPresent(NeuronPresentArgs),
+    MatchDriver(MatchDriverArgs),
     MatchNvidiaDriver(MatchNvidiaDriverArgs),
     WriteInfinibandGuid(WriteInfinibandGuidArgs),
 }
@@ -94,6 +95,16 @@ struct EbsDeviceNameArgs {
 struct MatchNvidiaDriverArgs {
     #[argh(positional)]
     driver_name: String,
+}
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "match-driver")]
+/// Returns if devices on the PCI bus support the provided driver.
+struct MatchDriverArgs {
+    #[argh(positional)]
+    driver_name: String,
+    #[argh(positional)]
+    flavor_name: String,
 }
 
 #[derive(FromArgs, PartialEq, Debug)]
@@ -156,6 +167,19 @@ fn main() -> Result<()> {
             let driver_name = driver.driver_name;
             nvidia_driver_supported(&driver_name)?;
         }
+        SubCommand::MatchDriver(driver) => {
+            let driver_name = driver.driver_name;
+            let flavor_name = driver.flavor_name;
+            match driver_name.as_str() {
+                "nvidia" => nvidia_driver_supported(&flavor_name)?,
+                "neuron" => match_neuron_driver(&flavor_name)?,
+                _ => {
+                    return Err(error::Error::UnsupportedDriver {
+                        driver: driver_name.to_string(),
+                    })
+                }
+            }
+        }
         SubCommand::WriteInfinibandGuid(envfile) => {
             find_and_write_infiniband_guid(envfile.env_file)?;
         }
@@ -177,6 +201,24 @@ fn is_neuron_attached() -> Result<()> {
     } else {
         Err(error::Error::NoNeuronPresent)
     }
+}
+
+// Returns true if this is an inf1 instance
+fn is_inf1_instance() -> Result<()> {
+    if pciclient::is_inf1_instance().context(error::CheckInf1FailureSnafu)? {
+        return Ok(());
+    }
+    Err(error::Error::NoInf1Present)
+}
+
+// Returns true if is a Neuron-based instance but not inf1. Inf2 is a stand in name for all hardware after
+// inf1 but leaves this open for more hardware types that may need specific decisions from ghostdog in the
+// future.
+fn is_inf2_instance() -> Result<()> {
+    if pciclient::is_inf2_instance().context(error::CheckInf2FailureSnafu)? {
+        return Ok(());
+    }
+    Err(error::Error::NoInf2Present)
 }
 
 /// Detects if infiniband is present. If not, return early. If it does find Infiniband devices,
@@ -357,6 +399,17 @@ fn nvidia_driver_supported(driver_name: &str) -> Result<()> {
         }
     );
     Ok(())
+}
+
+fn match_neuron_driver(driver_flavor: &str) -> Result<()> {
+    match driver_flavor {
+        "inf1" => is_inf1_instance(),
+        "latest" => is_inf2_instance(),
+        _ => Err(error::Error::UnsupportedDriverFlavor {
+            driver: "neuron".to_string(),
+            flavor: driver_flavor.to_string(),
+        }),
+    }
 }
 
 // Known system partition types for Bottlerocket.
--- a/sources/pciclient/src/lib.rs
+++ b/sources/pciclient/src/lib.rs
@@ -7,7 +7,10 @@ pciclient provides util functions that can:
 */
 mod private;
 
-use private::{call_list_devices, check_efa_attachment, check_neuron_attachment, PciClient};
+use private::{
+    call_list_devices, check_efa_attachment, check_inf1_attachment, check_inf2_attachment,
+    check_neuron_attachment, PciClient,
+};
 
 use bon::Builder;
 use derive_getters::Getters;
@@ -120,6 +123,16 @@ pub fn is_neuron_attached() -> Result<bool> {
     check_neuron_attachment(PciClient {})
 }
 
+/// Call `lspci` and check if there are inf1 devices attached
+pub fn is_inf1_instance() -> Result<bool> {
+    check_inf1_attachment(PciClient {})
+}
+
+/// Call `lspci` and check if there are inf2 or later devices attached
+pub fn is_inf2_instance() -> Result<bool> {
+    check_inf2_attachment(PciClient {})
+}
+
 mod error {
     use snafu::Snafu;
 
@@ -147,4 +160,5 @@ mod error {
 }
 
 pub use error::PciClientError;
+
 pub type Result<T> = std::result::Result<T, PciClientError>;
--- a/sources/pciclient/src/private.rs
+++ b/sources/pciclient/src/private.rs
@@ -15,12 +15,19 @@ use crate::{
 const AMAZON_VENDOR_CODE: &str = "1d0f";
 const EFA_KEYWORD: &str = "efa";
 
+// Neuron devices specific to inf1 instance types
 lazy_static! {
-    static ref NEURON_DEVICES: HashSet<&'static str> = hashset! {
+    static ref INF1_DEVICES: HashSet<&'static str> = hashset! {
         "7064", // inf1 device id 0
         "7065", // inf1 device id 1
         "7066", // inf1 device id 2
         "7067", // inf1 device id 3
+    };
+}
+
+// Neuron device ids outside of inf1-specific devices
+lazy_static! {
+    static ref NEURON_DEVICES: HashSet<&'static str> = hashset! {
         "7164", // trn1 device id 0
         "7264", // inf2 device id 0
         "7364", // trn2 device id 0
@@ -170,6 +177,33 @@ pub(crate) fn check_neuron_attachment<T: CommandExecutor>(command_executor: T) -
         .vendor(AMAZON_VENDOR_CODE.to_string())
         .build();
     let list_device_results = call_list_devices(command_executor, list_devices_param)?;
+    Ok(list_device_results.iter().any(|device_info| {
+        NEURON_DEVICES.contains(&device_info.device.as_str())
+            || INF1_DEVICES.contains(&device_info.device.as_str())
+    }))
+}
+
+/// Call `lspci` and check if there is any inf1-specific Neuron device attached.
+/// Internal usage, adding command_executor as parameter allows us to better unit test.
+/// For external usage, check [`crate::is_inf1_instance`].
+pub(crate) fn check_inf1_attachment<T: CommandExecutor>(command_executor: T) -> Result<bool> {
+    let list_devices_param = ListDevicesParam::builder()
+        .vendor(AMAZON_VENDOR_CODE.to_string())
+        .build();
+    let list_device_results = call_list_devices(command_executor, list_devices_param)?;
+    Ok(list_device_results
+        .iter()
+        .any(|device_info| INF1_DEVICES.contains(&device_info.device.as_str())))
+}
+
+/// Call `lspci` and check if there is any inf2 or later Neuron device attached.
+/// Internal usage, adding command_executor as parameter allows us to better unit test.
+/// For external usage, check [`crate::is_inf2_instance`].
+pub(crate) fn check_inf2_attachment<T: CommandExecutor>(command_executor: T) -> Result<bool> {
+    let list_devices_param = ListDevicesParam::builder()
+        .vendor(AMAZON_VENDOR_CODE.to_string())
+        .build();
+    let list_device_results = call_list_devices(command_executor, list_devices_param)?;
     Ok(list_device_results
         .iter()
         .any(|device_info| NEURON_DEVICES.contains(&device_info.device.as_str())))
@@ -194,7 +228,7 @@ mod test {
     use crate::{ListDevicesOutput, ListDevicesParam, Result};
 
     use super::{
-        call_list_devices, check_efa_attachment, check_neuron_attachment,
+        call_list_devices, check_efa_attachment, check_inf1_attachment, check_neuron_attachment,
         parse_list_devices_output, CommandExecutor,
     };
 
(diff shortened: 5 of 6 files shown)


sources/driverdog/src/main.rs
```diff
@@ -403,6 +432,43 @@ fn get_kernel_version() -> Result<String> {
...
 }
 
+/// Choose the priority order via depmod config drop-ins for the provided module
+fn choose_module(module: Option<String>) -> Result<()> {
+    let module_str = module.ok_or(error::Error::MissingModule)?;
+    match module_str.parse::<OpinionatedModule>() {
+        Ok(OpinionatedModule::Neuron) => {
+            // get data from pciclient for inf1
+            if is_inf1_instance().context(error::CheckInf1FailureSnafu)? {
+                if !Path::new(DEPMOD_D_PATH).is_dir() {
```
yeazelm: We should switch to tmpfiles directive.

sources/pciclient/src/private.rs
```diff
@@ -27,6 +27,15 @@ lazy_static! {
...
 }
 
+lazy_static! {
+    static ref INF1_DEVICES: HashSet<&'static str> = hashset! {
+        "7064", // inf1 device id 0
+        "7065", // inf1 device id 1
+        "7066", // inf1 device id 2
+        "7067", // inf1 device id 3
+    };
+}
```
bcressey, line 35: Rather than duplicating this, I'd rather split the current `NEURON_DEVICES` into two sets, and update the `check_neuron_attachment` function to query both.

sources/driverdog/src/main.rs
```diff
@@ -69,6 +74,7 @@ struct Args {
 enum Subcommand {
     Link(LinkModulesArgs),
     Load(LoadModulesArgs),
+    Prioritize(PrioritizeModuleArgs),
```
bcressey: I'd rather use a `ghostdog` subcommand helper for this, and if possible do it the same way we do NVIDIA driver matching: systemd units with a `ghostdog` ExecCondition that invoke `driverdog` with different configs.

packages/release/drivers.target
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Bottlerocket driver loading complete
+AllowIsolate=yes
+After=basic.target
+Requires=basic.target
+
+[Install]
+RequiredBy=preconfigured.target multi-user.target
```
bcressey, line 9: This doesn't establish an ordering dependency with `preconfigured.target`, and we want it to happen before.

sources/pciclient/src/lib.rs
```diff
@@ -147,4 +152,6 @@ mod error {
 }
 
 pub use error::PciClientError;
+
+use crate::private::check_inf1_attachment;
```
bcressey: This should go up at the top with the other, non-error related `use` statements.

sources/ghostdog/src/main.rs
```diff
@@ -359,6 +404,17 @@ fn nvidia_driver_supported(driver_name: &str) -> Result<()> {
     Ok(())
 }
 
+fn match_neuron_driver(driver_flavor: &str) -> Result<()> {
+    if driver_flavor == "inf1" {
+        return is_inf1_instance();
+    }
+    if driver_flavor == "latest" {
+        return is_after_inf1_instance();
+    }
```
bcressey: Let's also return an error on an unmatched flavor. If we don't return non-zero then the `ExecCondition` logic in units won't work as expected.

sources/Cargo.lock
```diff
@@ -2317,7 +2318,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
```
bcressey: Can we back out the `windows-sys` changes?

sources/driverdog/Cargo.toml
```diff
@@ -11,6 +11,7 @@ exclude = ["README.md"]
 [dependencies]
 argh.workspace = true
 log.workspace = true
+pciclient.workspace = true
```
bcressey: This change is no longer needed?

sources/driverdog/src/main.rs
```diff
@@ -138,13 +153,19 @@ fn link_modules_sets(
         match driver_config {
             DriverType::Copying(config) => copy_modules(config, &kernel_version)?,
             DriverType::Linking(config) => link_modules(config, &kernel_version)?,
+            DriverType::DepmodOnly(_) => {
+                debug!("Depmod only is not supported for linking modules sets")
+            }
```
bcressey: Can we call `configure_modules` here so that `DepmodOnly` can be handled the same way as the other two? And then `load_modules` would have the same boilerplate to extract the list of modules to load.

Conceptually `driverdog` has two stages: "prepare" (which we call "link") and "load", so I'd prefer to do the work in the "prepare" stage rather than deferring it to "load".

Or in other words: the only side effect of `driverdog load-modules` should be that modules might be loaded. Other filesystem operations should happen in `driverdog link-modules`.

sources/driverdog/src/main.rs
```diff
@@ -403,6 +428,65 @@ fn get_kernel_version() -> Result<String> {
...
+        modules_path.exists(),
+        error::MissingModulePathSnafu { path: modules_path }
+    );
+
+    let depmod_conf = Path::new(DEPMOD_D_PATH).join(&module.depmod_filename);
+
+    let config_string = module
+        .kernel_modules
+        .iter()
+        .map(|name| format!("override {} * {}", name, module.lib_modules_path))
```
bcressey: Can we also set the specific kernel version instead of `*` here, since we know what that version is?

sources/driverdog/src/main.rs
```diff
@@ -438,20 +522,22 @@ fn run() -> Result<()> {
...
-        })?)
-    .flatten()
-    {
-        let path = entry.path();
-        let modules_sets: HashMap<String, DriverType> = toml::from_str(
-            &fs::read_to_string(&path).context(error::ReadPathSnafu { path: &path })?,
-        )
-        .context(error::DeserializeSnafu { path: &path })?;
+    if Path::new(driver_config_path).is_dir() {
+        for entry in (driver_config_path
```
bcressey: Why do we need this new conditional logic? Why wouldn't `driver_config_path` exist?

packages/release/drivers.target
yeazelm: Yeah this worked in my tests but you are right that we need `Before=preconfigured.target` to ensure it really does go before.

sources/driverdog/src/main.rs
yeazelm: "link" is really doing too much here now, but I agree that had we named it "prepare", this would feel natural. In this case I do agree we should prep the depmod in "link" and only load in "load". I debated on if we wanted to have two invocations of driverdog for these but I think the additional call is worth it to keep the separation of concerns. I'll update the logic to do so.

sources/driverdog/src/main.rs
yeazelm: Yeah, we can be more specific. We don't ship more than one kernel in an image so it isn't strictly required but this does make it more correct and specific.

sources/driverdog/src/main.rs
yeazelm: This change was a quick one I made when I first added driverdog to all images. We didn't provide `/etc/drivers` since that was only created by NVIDIA kmods. In this case, I have refactored since then to provide files for neuron so we don't strictly need this check anymore, but it does feel overly strict to not allow driverdog to run *at all* without `/etc/drivers` existing on the system. Since `driver_config_path` is always created, driverdog _always_ tries to load these files. I think a more correct way to solve this is to package a tmpfiles directive that creates this directory with driverdog since the tool has some deep assumptions about these configuration files being present to fully function. There isn't really a valid use case to use driverdog without these files being present. I can back it out, or we can ensure the directory is present as part of packaging.

sources/driverdog/Cargo.toml
yeazelm: Correct, I'll clean up the Cargo dependencies from the refactor.

sources/ghostdog/src/main.rs
yeazelm: Make sense!

sources/pciclient/src/lib.rs
yeazelm: I'll fix, not sure why the auto add from my IDE chose here.

sources/Cargo.lock
yeazelm: Yes, I'll unwind this change.

yeazelm:
^ Updated the code to reflect comments. There is now two calls, one for link and one for load in driverdog. The kernel kit PR has been updated as well to include these calls. I re-ran validation that the expected drivers load on each instance type.

sources/driverdog/src/main.rs
yeazelm: I added this logic, the resulting file at `/etc/depmod.d/neuron.conf` now looks like this for inf1:
```
override neuron 6.12.40 neuron_2_21
```

packages/release/drivers.target
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Bottlerocket driver loading complete
```
bcressey: From existing target units, maybe just this:
```suggestion
Description=Driver units
```

We don't want a message like "Started Bottlerocket driver loading complete" on the console.

sources/ghostdog/src/error.rs
```diff
@@ -66,6 +66,16 @@ pub(super) enum Error {
         requested: String,
         preferred: String,
     },
+    #[snafu(display("{driver} is not a supported driver"))]
+    UnsupportedDriver { driver: String },
+    #[snafu(display("{flavor} is not a supported choice for {driver}"))]
```
bcressey, line 71: Do these `{driver}` and `{source}` format parameters work?

I don't recall seeing them elsewhere.

sources/ghostdog/src/main.rs
```diff
@@ -179,6 +203,27 @@ fn is_neuron_attached() -> Result<()> {
...
+fn is_after_inf1_instance() -> Result<()> {
+    if pciclient::is_neuron_attached().context(error::CheckNeuronFailureSnafu)? {
+        if !pciclient::is_inf1_instance().context(error::CheckInf1FailureSnafu)? {
+            return Ok(());
+        } else {
+            return Err(error::Error::NotAfterInf1Instance);
+        }
+    }
+    Err(error::Error::NoNeuronPresent)
+}
```
bcressey: I see why you're doing this, to make the error messages more friendly, but it's going to be hard to maintain that if we ever need a third driver variant.

I would probably just have `is_inf1_instance` call `pciclient::is_inf1_instance()`, and `is_inf1_instance` call `pciclient::is_inf2_instance()`, and let "inf2" be a stand-in for all hardware of that generation until the next breaking change. (IIRC it was like this earlier.) Or `trn1` instead of `inf2`.

As it stands, the function ends up calling `lspci` twice, once with a combined Neuron devices filter, and again with a Inf1-only filter, and that's suboptimal.

sources/pciclient/src/private.rs
```diff
@@ -166,13 +173,27 @@ pub(crate) fn check_efa_attachment<T: CommandExecutor>(command_executor: T) -> R
...
+    Ok(list_device_results.iter().any(|device_info| {
+        NEURON_DEVICES.contains(&device_info.device.as_str())
+            || INF1_DEVICES.contains(&device_info.device.as_str())
+    }))
+}
+
+/// Call `lspci` and check if there is any inf1-specific Neuron device attached.
+/// Internal usage, adding command_executor as parameter allows us to better unit test.
+/// For external usage, check [`crate::is_neuron_attached`].
+pub(crate) fn check_inf1_attachment<T: CommandExecutor>(command_executor: T) -> Result<bool> {
```
bcressey, line 189: The `check_neuron_attachment` change is good for backwards compatibility; `ghostdog neuron-present` should probably continue to work for now. But I'd still add a `check_inf2_attachment` function and related helper, and have `ghostdog` call `is_inf2_attached`, even if it means a third near-duplication of this method.

sources/driverdog/src/main.rs
```diff
@@ -138,13 +153,17 @@ fn link_modules_sets(
...
         }
     } else {
         // Link all the modules sets if no target module was given
         for driver_config in modules_sets.values() {
             match driver_config {
                 DriverType::Copying(config) => copy_modules(config, &kernel_version)?,
                 DriverType::Linking(config) => link_modules(config, &kernel_version)?,
+                DriverType::DepmodOnly(_) => {
+                    debug!("Depmod only is not supported for linking modules sets")
+                }
```
bcressey: I'd prefer to do the depmod configuration at "link" time, and keep the load path consistent across the three driver types.

sources/driverdog/src/main.rs
bcressey: > the tool has some deep assumptions about these configuration files being present to fully function

I'd just do an early return with a "no configurations found in `/etc/drivers`" error message, if the path doesn't exist.

sources/ghostdog/src/error.rs
yeazelm: Clippy has recently been calling out the other style of `#[snafu(display("{} is not a supported choice for {}", flavor, driver))]` in explicit `format!()` calls, I just started getting into a pattern of using this and sure enough it worked without me thinking about it. I can try moving it back but I actually do like this style. We can move everything to it in a follow up commit/PR if we want to keep things consistent. 

I wanted to double check and sure enough, it works as expected:
```
bash-5.1# /usr/bin/ghostdog match-driver foo bar
Error: foo is not a supported driver

bash-5.1# /usr/bin/ghostdog match-driver neuron bar
Error: bar is not a supported choice for neuron

bash-5.1#
```

sources/driverdog/src/main.rs
yeazelm: I'll update this. I was overly focused on the single case and didn't shift this to be consistent.

sources/ghostdog/src/main.rs
yeazelm: I agree. This may yield a bit less data but really streamlines the implementation. I'll shift everything to be specific to inf2. This name doesn't cover everything as of today but leaves it specific enough that any future refactors should be easier to make.

sources/ghostdog/src/error.rs
```diff
@@ -66,6 +66,18 @@ pub(super) enum Error {
         requested: String,
         preferred: String,
```
arnaldo2792, line 67: nit: typo in commit message:

```
This command also supports
```

sources/ghostdog/src/main.rs
```diff
@@ -156,6 +167,19 @@ fn main() -> Result<()> {
...
+        SubCommand::MatchDriver(driver) => {
+            let driver_name = driver.driver_name;
+            let flavor_name = driver.flavor_name;
+            match driver_name.as_str() {
+                "nvidia" => nvidia_driver_supported(&flavor_name)?,
+                "neuron" => match_neuron_driver(&flavor_name)?,
+                _ => {
+                    return Err(error::Error::UnsupportedDriver {
+                        driver: driver_name.to_string(),
+                    })
```
arnaldo2792, line 179: I think you can do:

```suggestion
                    return error::Error::UnsupportedDriver {
                        driver: driver_name.to_string(),
                    }.fail()
```

sources/ghostdog/src/main.rs
```diff
@@ -359,6 +402,20 @@ fn nvidia_driver_supported(driver_name: &str) -> Result<()> {
...
+        return is_inf1_instance();
+    }
+    if driver_flavor == "latest" {
+        return is_inf2_instance();
+    }
+
+    Err(error::Error::UnsupportedDriverFlavor {
+        driver: "neuron".to_string(),
+        flavor: driver_flavor.to_string(),
+    })
```
arnaldo2792: Same as elsewhere, you could just `fail`.

sources/ghostdog/src/main.rs
```diff
@@ -359,6 +402,20 @@ fn nvidia_driver_supported(driver_name: &str) -> Result<()> {
     Ok(())
 }
 
+fn match_neuron_driver(driver_flavor: &str) -> Result<()> {
+    if driver_flavor == "inf1" {
```
arnaldo2792: Why not use `match` for `inf1` and `latest` and fail otherwise in the fallback branch?

sources/ghostdog/src/main.rs
```diff
@@ -179,6 +203,25 @@ fn is_neuron_attached() -> Result<()> {
     }
 }
 
+// Returns true if this is an inf1 instance
+fn is_inf1_instance() -> Result<()> {
+    if pciclient::is_inf1_instance().context(error::CheckInf1FailureSnafu)? {
```
arnaldo2792, line 208: Wouldn't this error be more "correct" in the underlying `check_` function? then in here you can do `pciclient::is_inf1_instance()?` without having to add any context.

sources/ghostdog/src/main.rs
```diff
@@ -179,6 +203,25 @@ fn is_neuron_attached() -> Result<()> {
...
+}
+
+// Returns true if is a Neuron-based instance but not inf1. Inf2 is a stand in name for all hardware after
+// inf1 but leaves this open for more hardware types that may need specific decisions from ghostdog in the
+// future.
+fn is_inf2_instance() -> Result<()> {
+    if pciclient::is_inf2_instance().context(error::CheckInf2FailureSnafu)? {
+        return Ok(());
+    }
+    Err(error::Error::NoInf2Present)
```
arnaldo2792, line 221: Above you implemented this with `if/else`. For consistency, you should use the same here.

bcressey (approved):
LGTM, no notes beyond what @arnaldo2792 has commented on.

sources/ghostdog/src/main.rs
bcressey: Agreed re: consistency but I prefer this style - early return + outdented body for what would be the else clause. So I'd make the other one look like this rather than vice versa.

sources/ghostdog/src/main.rs
yeazelm: This is following the existing style of other functions like `is_efa_attached()` or `is_neuron_attached()`. The intention here is what function failed *is* the context. The failure would ultimately be coming from being unable to use `lspci` and so this helps provide the additional context on what check failed along with the actual failure. So the `check_` functions will emit logs about the `call_list_devices()` call failing and the additional messaging will be why we were calling `call_list_devices()`. I think this might be a bigger refactor of error handling in pciclient which we can do if we think it will improve error messaging, but probably should do it separately from this change.

sources/ghostdog/src/main.rs
yeazelm: This was to keep it _mostly_ consistent with the rest of the if statements in these functions. I think an enum for supported flavors is probably a stronger mechanism anyway and then the match statement can work from that. It will force the compiler to hold us accountable when adding any new flavors. I'll update with this change!

sources/ghostdog/src/main.rs
yeazelm: The rest of ghostdog is using `Err()` style. I can shift to fail() but we should remain consistent in the program. If we like `fail()` more than this style, I can do a change to follow up with the style improvement.

sources/ghostdog/src/main.rs
yeazelm: The flavor changes depending on the original driver. We may want to refactor this to be more robust in types for these flavors but for now its two strings, so I'll keep the strings but in the future if we add more, we should define types to hold all of these.

jpculp (approved):

arnaldo2792 (approved):

PR #647, opened 2025-08-30
apiclient: support `--` separator in `exec` subcommand

### LINKED ISSUE #633: `apiclient exec` always handles `-h`
**Package I'm using:**

`apiclient`

When running
`apiclient exec admin sheltie df -h`

I would expect it to pass `-h` to the df; instead apiclient interprets it.

I also tried `apiclient exec admin -- sheltie df -h` but it does not work either.

I think there's a couple ways on how to solve this:

1. Pass all arguments after the subcommand to the subcommand arg parser. That would be backwards incompatible though as we would not be able to specify the socket after the subcommand.
2. If we find a `--`, pass all subsequent arguments to the subcommand parser.

Link to code: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/api/apiclient/src/main.rs#L308

**What actually happened:**

It prints `apiclient`'s CLI help.

See above.

ytsssun:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/633

**Description of changes:**
For subcommand like `exec` in `apiclient`, it would further execute nested commands. From the GH issue #633, example command was reported not working as expected:
```sh
apiclient exec admin sheltie df -h
```
This was because when in the parsing of the argument, `-h` would always be treated as global arguments and `apiclient` would exit early upon parsing this flag (See https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/api/apiclient/src/main.rs#L311). Note that `usage()` is a helper function that would print the help information of `apiclient` and call process::exit(2). But in the example command, `-h` was intended to be used as an argument to the nested command `df`.

As per discusssion in https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/633#issuecomment-3215790291, this PR extends `apiclient` to support `--` for subcommand `exec`, but only for the `exec` subcommand. This special handling is limited to the `exec` subcommand because it's the only command where passing arbitrary flags to a nested command makes sense.

With this change, arguments passed after `--` for the `exec` subcommand will be treated as arguments to the nested command being executed. This makes it possible to invoke the above command via:
```sh
apiclient exec admin -- sheltie df -h
``` 
 

**Testing done:**
Manually tested with multiple variations of the command to ensure the special handling for `--` works correctly:

<ins>_All of these commands correctly pass -h to df_</ins>
```sh
apiclient exec -- admin sheltie df -h
apiclient exec admin -- sheltie df -h
apiclient exec admin sheltie -- df -h
apiclient exec admin sheltie df -- -h
```

All test cases display the same output, confirming that the special handling of `--` for the `exec` subcommand works as expected.
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       306M  306M     0 100% /
devtmpfs        3.9G     0  3.9G   0% /dev
tmpfs           3.9G     0  3.9G   0% /dev/shm
tmpfs           1.6G  380K  1.6G   1% /run
tmpfs           3.9G  476K  3.9G   1% /etc
tmpfs           3.9G     0  3.9G   0% /etc/cni
tmpfs           3.9G     0  3.9G   0% /tmp
tmpfs           3.9G   16K  3.9G   1% /etc/containerd
tmpfs           3.9G  4.0K  3.9G   1% /etc/soci-snapshotter
tmpfs           3.9G   16K  3.9G   1% /etc/host-containers
tmpfs           3.9G     0  3.9G   0% /etc/kubernetes/pki/private
tmpfs           3.9G  4.0K  3.9G   1% /root/.aws
tmpfs           3.9G     0  3.9G   0% /run/netdog
/dev/xvda3       23M   14M  7.6M  64% /boot
/dev/xvdb1       20G  1.5G   19G   8% /local
/dev/xvda12      35M  1.5M   31M   5% /var/lib/bottlerocket
overlay          20G  1.5G   19G   8% /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/modules
overlay          20G  1.5G   19G   8% /opt/cni
overlay          20G  1.5G   19G   8% /opt/csi
overlay          20G  1.5G   19G   8% /x86_64-bottlerocket-linux-gnu/sys-root/usr/src/kernels
overlay          20G  1.5G   19G   8% /run/host-containerd/io.containerd.runtime.v2.task/default/control/rootfs
overlay          20G  1.5G   19G   8% /run/host-containerd/io.containerd.runtime.v2.task/default/admin/rootfs
```

For non-exec subcommands, the behavior remains unchanged. For example, attempting to use `--` with the `get` subcommand still results in:
```
[ssm-user@control]$ apiclient get -- settings
Unknown argument '--'

Usage: apiclient [SUBCOMMAND] [OPTION]...

Global options:
      -s, --socket-path PATH     Override the server socket path.  Default: /run/api.sock
      --log-level                Desired amount of output; trace|debug|info|warn|error
      -v, --verbose              Sets log level to 'debug'.  This prints extra info,
                                 like HTTP status code to stderr in 'raw' mode.
  ...
```

Also, ran the unit tests which cover these test cases:
```
  test tests::test_exec_parsing::exec_with_separator - Tests "apiclient exec admin -- sheltie df -h"
  test tests::test_exec_parsing::exec_without_separator - Tests "apiclient exec admin sheltie df"
  test tests::test_exec_parsing::exec_with_tty_and_separator - Tests "apiclient exec -t admin -- sheltie df -h"
  test tests::test_issue_case_df_with_h_flag - Tests specific handling of "-h" flag in "apiclient exec admin -- sheltie df -h"
```
All tests pass successfully:
```
  cargo test --no-default-features
```

**Terms of contribution:**

### DIFF
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -300,7 +300,7 @@ fn usage_msg<S: AsRef<str>>(msg: S) -> ! {
 // Arg parsing
 
 /// Parses user arguments into an Args structure.
-fn parse_args(args: env::Args) -> (Args, Subcommand) {
+fn parse_args(args: impl Iterator<Item = String>) -> (Args, Subcommand) {
     let mut global_args = Args::default();
     let mut subcommand = None;
     let mut subcommand_args = Vec::new();
@@ -308,6 +308,22 @@ fn parse_args(args: env::Args) -> (Args, Subcommand) {
     let mut iter = args.into_iter().skip(1);
     while let Some(arg) = iter.next() {
         match arg.as_ref() {
+            // Handle separator - only valid after exec subcommand is set
+            "--" => {
+                if subcommand.is_none() {
+                    usage_msg("'--' separator requires a subcommand to be specified first");
+                }
+                // Only do special handling for exec subcommand
+                if subcommand.as_deref() == Some("exec") {
+                    // For exec, collect all remaining args and stop processing
+                    subcommand_args.extend(iter);
+                    break;
+                } else {
+                    // For other subcommands, treat -- as a regular argument
+                    subcommand_args.push(arg);
+                }
+            }
+
             "-h" | "--help" => usage(),
 
             // Global args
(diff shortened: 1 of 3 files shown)


sources/api/apiclient/src/main.rs
```diff
@@ -1160,3 +1176,40 @@ mod error {
...
+            assert_eq!(exec_args.target, expected_target);
+            let expected_os_strings: Vec<OsString> =
+                expected_command.iter().map(|s| OsString::from(s)).collect();
+            assert_eq!(exec_args.command, expected_os_strings);
+        } else {
+            panic!("Expected Exec subcommand, got: {:?}", subcommand);
+        }
+    }
+
+    #[test_case("apiclient -v exec admin -- sheltie df -h", LevelFilter::Debug; "verbose flag with exec")]
```
arnaldo2792: Can you add a test that checks for `apiclient -- -v exec`?

sources/api/apiclient/src/main.rs
```diff
@@ -1160,3 +1176,40 @@ mod error {
...
+
+    fn parse_command_line(cmd_str: &str) -> (Args, Subcommand) {
+        let args: Vec<String> = cmd_str.split_whitespace().map(|s| s.to_string()).collect();
+        parse_args(args.into_iter())
+    }
+
+    #[test_case("apiclient exec admin -- sheltie df -h", "admin", &["sheltie", "df", "-h"]; "exec with separator")]
+    #[test_case("apiclient exec admin sheltie df", "admin", &["sheltie", "df"]; "exec without separator")]
+    #[test_case("apiclient exec -t admin -- sheltie df -h", "admin", &["sheltie", "df", "-h"]; "exec with tty and separator")]
+    #[test_case("apiclient exec -T admin sheltie df", "admin", &["sheltie", "df"]; "exec with no-tty")]
```
bcressey: I find these test cases unsatisfying for two reasons.

The first is that we don't really check the global arg parsing or the subcommand. If there were a compact way of creating the proper structs, for example via a macro, then it would be easier to scan only the test case line to check whether the right behavior is tested.
```
#[test_case("apiclient -v exec admin -- sheltie df -h", global_args!(LevelFilter::Debug, "..."), exec_cmd!("admin", None, ["sheltie", "df", "-h"]); "exec with separator"]
fn test_exec_parsing(cmd_str: &str, expected_global_args: &Args, expected_subcommand: &Subcommand) {
    let (global_args, subcommand) = parse_command_line(cmd_str);
    assert_eq!(global_args, expected_global_args);
    assert_eq!(subcommand, expected_subcommand);
}
```

The second is that all of the `usage` calls create divergent execution where the process just exits. This means we can't have a unit test for current behavior of `df -h` which is motivating the change.

It's probably OK to ignore the `-h` case specifically, but it would be good to have tests that show correct parsing for the more general case:
1. `-v` passed as a global arg but not an exec command arg
2. `-v` passed as an exec command arg but not a global arg (using `--`)
3. `-v` passed in both positions

The switch to macros - or if that's too annoying, then just pre-created copies of `Args` and `Subcommand::Exec` that we mix and match as expected results - should make this coverage more obvious.

bcressey (commented):
The parsing change looks good to me. Nice!

sources/api/apiclient/src/main.rs
ytsssun: Right now with the limit mentioned in the description (`usage()` would invoke process::exit(), and exit the unit test), it is not easy to test the negative cases. But I tested this with manual test - and it failed as expected.

sources/api/apiclient/src/main.rs
ytsssun: This is a fair feedback. I can refactor the unit test to reflect the global and subcommand args parsing. 

Note that for the reason you mentioned (usage() would exit the program), we cannot really implement the negative unit test cases which would trigger `usage()` without a good amount of mocking of the parse_args function, which IMO is probably equally the amount of work comparing to migrating to argh - which the unit test would be much easier.

ytsssun:
Introduce macro `global_args!` and `exec_cmd!` for better readability.

Add more test cases per @bcressey

bcressey (approved):

bcressey:
Very nice, thanks!

ytsssun:
Just realizing that `cargo fmt` does not really format the test code with `#[test_case]` annotation. Manually format the code and force pushed.

jpculp (approved):

KCSesh (approved):

PR #655, opened 2025-09-09
Backport systemd patch to downgrade log from error to debug

vigh-m:
**Description of changes:**
When restarting or starting services with systemd-257, the following is observed:
```
bash-5.1# systemctl start containerd
Failed to execute /x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/systemd-tty-ask-password-agent: No such file or directory
```
The process completes successfully but the log is still present on the console. This is because, `systemd-tty-ask-password-agent` is only shipped inside the `systemd-console` subpackage and is not present in any standard variant.

The added patch is backported from systemd upstream where this log is downgraded to a Debug when the error is due to `ENOENT` (error no entity). 

**Testing done:**
- Built and tested an AMI with the patch and the error was not seen on restarting services

**Terms of contribution:**

### DIFF
--- a/packages/systemd-257/systemd-257.spec
+++ b/packages/systemd-257/systemd-257.spec
@@ -14,6 +14,10 @@ Source0: https://github.com/systemd/systemd/archive/v%{version}/systemd-%{versio
 Source1: systemd-mount-rate-bootconfig.conf
 Source2: systemd-cgroup-legacy-force-bootconfig.conf
 
+# Backport of upstream patch to change `Failed to execute <filepath> No such file 
+# or directory` error logs to debug 
+Patch1001: 1001-exec-util-make-missing-agents-a-gracefull-handled-issues.patch
+
 # Local patch to add the acquire the id for VMware
 Patch9001: 9001-machine-id-setup-generate-stable-ID-under-VM.patch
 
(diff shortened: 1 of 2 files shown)


packages/systemd-257/systemd-257.spec
```diff
@@ -14,6 +14,10 @@ Source0: https://github.com/systemd/systemd/archive/v%{version}/systemd-%{versio
 Source1: systemd-mount-rate-bootconfig.conf
 Source2: systemd-cgroup-legacy-force-bootconfig.conf
 
+# Backport of upstream patch to change `Failed to execute %s No such file or 
+# directory` error logs to debug
```
bcressey: nit: I would avoid `%` since it has special meaning in a spec file - macros are parsed even inside "comments"

Can just say the patch suppresses an error message if  `systemd-tty-ask-password-agent` is not installed

vigh-m: Sure. The patch is not specific to `systemd-tty-ask-password-agent` which is why I put the `%s` placeholder. I'll reword it.

bcressey (approved):

ytsssun (approved):

PR #656, opened 2025-09-10
install driverdog for nvidia-fips variant

fletcherw:
This will handle new nvidia-fips variants as well as possible future uses of driverdog in other variants.

**Terms of contribution:**

### DIFF
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -78,6 +78,7 @@ Requires: %{_cross_os}bloodhound
 Requires: %{_cross_os}bootstrap-commands
 Requires: %{_cross_os}corndog
 Requires: %{_cross_os}certdog
+Requires: %{_cross_os}driverdog
 Requires: %{_cross_os}ghostdog
 Requires: %{_cross_os}logdog
 Requires: %{_cross_os}metricdog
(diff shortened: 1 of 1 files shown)


packages/os/os.spec
```diff
@@ -104,7 +104,7 @@ Requires: (%{_cross_os}cfsignal if %{_cross_os}variant-platform(aws))
 
 Requires: (%{_cross_os}warm-pool-wait if %{_cross_os}variant-family(aws-k8s))
 
-Requires: (%{_cross_os}driverdog if %{_cross_os}variant-flavor(nvidia))
+Requires: (%{_cross_os}driverdog if %{_cross_os}variant-flavor(nvidia) or %{_cross_os}variant-flavor(nvidia-fips))
```
bcressey: Seems like we can just relax this and include everywhere:
```suggestion
Requires: %{_cross_os}driverdog
```

yeazelm: I have a [draft change](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/644/files) that was moving this up to the sorted list on line 81 and removing the constraint for NVIDIA variants. This would be fine to take by itself right now.

fletcherw: Ok, done.

bcressey (approved):

yeazelm (approved):
Nice!

PR #661, opened 2025-09-18
Add whippet

arnaldo2792:
**Issue number:**

Part of #660

**Description of changes:**

This adds `whippet`, a minimal implementation of a dbus-launcher. The set of features implemented by whippet include:
- Systemd socket activation
- Policy parsing but using TOML

<details>

**Differences between the dbus launcher and whippet**

**No default connect rule for whippet’s user**

One of the key differences between the launcher and whippet is that whippet doesn’t insert a default connect rule, the reason for this is twofold:

* Whippet doesn’t have to interact with the bus, as it doesn’t support the features that require interactions with it
    * Dbus service activation
    * Reload configurations
* Bottlerocket’s default dbus policy allows all system users to connect to the bus
    * So instead of allowing whippet to connect to the bus by default, it only gets the permission if actually allowed by the policy

**Drop support for deprecated or unused rule configurations**

To keep whippet’s implementation as minimal as possible whippet only supports the rule fields that are actually used in Bottlerocket. Additionally, whippet removes fields marked as deprecated in the configuration file and policy rules, this includes

* In configuration file:
    * user: used to drop privileges to the selected user. whippet relies on systemd directives to drop this privilege.
    * auth: [this isn’t implemented](https://github.com/bus1/dbus-broker/blob/b0db0890d1254477cf832e5f9f0a798360c80fd9/src/launch/config.c#L1173) in the dbus-launcher
    * listen: [this isn’t implemented](https://github.com/bus1/dbus-broker/blob/b0db0890d1254477cf832e5f9f0a798360c80fd9/src/launch/config.c#L1170) in the dbus-launcher
* In Rules:
    * send_type/receive_type = method_return or error: the dbus-launcher [warns and ignores these values](https://github.com/bus1/dbus-broker/blob/b0db0890d1254477cf832e5f9f0a798360c80fd9/src/launch/policy.c#L439)
    * Don’t support group, at_console, no_console or mandatory contexts, as they aren’t currently used in Bottlerocket

**Rule priority assignment**

In the dbus-launcher, rules are given a priority that is used to determine which rule wins when sending or receiving a message through the bus. The rules with the highest priorities win.

Whippet implements the priority assignment as close as possible to what the dbus-launcher does. However, there is a limitation on whippet’s implementation. The dbus-launcher assigns priorities to the rules in the order they appear, e.g. these rules:

```xml
<policy context="default">
    <allow user="*" />
    <deny send_path="*" />
</policy>

<policy user="root">
    <allow send_path="*" />
[...]

Issue #660, opened 2025-09-18
Remove expat from core kit


arnaldo2792:
**What I'd like:**
Remove the expat package from the core kit

bcressey:
Done in c042559fa.

PR #662, opened 2025-09-18
Drop expat

arnaldo2792:
**Issue number:**
Part of #660

**Description of changes:**

#661 added whippet, a minimal broker launcher. This series completes the work to remove both the dbus-launcher and libexpat. By dropping the dbus-launcher from the dbus-broker build, the dependency on systemd can be removed as well.

A patch was added to the dbus-broker project to allow building the journal catalogs for the broker in the absence of the dbus-launcher.

The dbus policies for systemd were programmatically generated. I have a personal preference on this format for the policies:

```toml
[default]
rules = [
   { allow = true, user = "*" }
]
```
But the serializers, really love the other format:

```toml
[[default.rules]]
allow = true
user = "*"
```

I didn't want to mess up with the serializers, so I just used the format they used by default. 

**Testing done:**

In combination with #661:
- [X] Confirmed instance boots
- [X] Confirmed instances join clusters
- [X] Confirmed pods run without issues
- [X] Confirmed the policy generated by whippet is identical to the policy generated by the dbus-launcher

**Terms of contribution:**


arnaldo2792:
I'll close this PR in favor of two smaller PRs, as we are not dropping expat yet.

Issue #670, opened 2025-09-25
Add hwloc for OpenMPI enablement


yeazelm:
**What I'd like:**
`nvbandwidth` uses `hwloc-ls` to bind to the correct memory for Open MPI workloads. To get the best performance, this should be available in the OS for EC2 instances like P6e-GB200.36xlarge. 

**Any alternatives you've considered:**
Not adding this allows the tool to "fall back" to the less performant method without binding to shared memory.

PR #672, opened 2025-09-25
Add `hwloc` to core kit

yeazelm:
**Issue number:**

Closes # https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/670

**Description of changes:**
Adds `hwloc` as package to the core kit. Also vend a devel package in case variants want to bind against hwloc as a build dependency rather than use the binaries. 

**Testing done:**
Built images with the binaries included. 

`hwloc-ls` is usable from `sheltie`:
```
bash-5.1# hwloc-ls
Machine (15GB total)
  Package L#0
    NUMANode L#0 (P#0 15GB)
    L3 L#0 (8192KB)
      L2 L#0 (512KB) + L1d L#0 (32KB) + L1i L#0 (32KB) + Core L#0
        PU L#0 (P#0)
        PU L#1 (P#2)
      L2 L#1 (512KB) + L1d L#1 (32KB) + L1i L#1 (32KB) + Core L#1
        PU L#2 (P#1)
        PU L#3 (P#3)
  Block(Disk) "nvme0n1"
  Block(Disk) "nvme2n1"
  Block(Disk) "nvme1n1"
  Net "eth1"
  Net "eth0"
```

Built an image and confirmed hwloc-ls will work from a container.
Specfile:
```
apiVersion: v1
kind: Pod
metadata:
  name: hwloc-pod
spec:
  containers:
  - name: hwloc
    image: ubuntu:24.04 # Has glibc 2.38 or later
    command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep infinity']
    volumeMounts:
    - name: hwloc-binary
      mountPath: /usr/local/bin/hwloc-ls
      readOnly: true
  volumes:
  - name: hwloc-binary
    hostPath:
      path: /bin/hwloc-ls
      type: File
```
Output:
```
 $ kubectl exec -it hwloc-pod -- bash
root@hwloc-pod:/# hwloc-ls
Machine (15GB total)
  Package L#0
    NUMANode L#0 (P#0 15GB)
    L3 L#0 (8192KB)
      L2 L#0 (512KB) + L1d L#0 (32KB) + L1i L#0 (32KB) + Core L#0
        PU L#0 (P#0)
        PU L#1 (P#2)
      L2 L#1 (512KB) + L1d L#1 (32KB) + L1i L#1 (32KB) + Core L#1
        PU L#2 (P#1)
        PU L#3 (P#3)
  Block(Disk) "nvme0n1"
  Block(Disk) "nvme2n1"
  Block(Disk) "nvme1n1"
```

**Terms of contribution:**

### DIFF
--- a/packages/hwloc/Cargo.toml
+++ b/packages/hwloc/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "hwloc"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://www.open-mpi.org/software/ompi/v5.0/"
+
+[[package.metadata.build-package.external-files]]
+url = "https://download.open-mpi.org/release/hwloc/v2.12/hwloc-2.12.2.tar.bz2"
+sha512 = "949d6c9d7b858ee58e477b15e6c06f57812872142fa1c7f3ef20aae2e4ef954135f839e8604404bfd0637fde99729c7d00211c8aee860dfde9ac60bba0e78aef"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
+systemd-252 = { path = "../systemd-252" }
--- a/packages/hwloc/hwloc.spec
+++ b/packages/hwloc/hwloc.spec
@@ -1,0 +1,83 @@
+Name: %{_cross_os}hwloc
+Version: 2.12.2
+Release: 1%{?dist}
+Summary: Portable hardware locality library
+URL: https://www.open-mpi.org/projects/hwloc/
+License: BSD-3-Clause
+Source0: https://download.open-mpi.org/release/hwloc/v2.12/hwloc-%{version}.tar.bz2
+
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}systemd-devel
+
+%description
+%{summary}.
+
+%package tools
+Summary: Command line tools for the hwloc library
+Requires: %{name}
+
+%description tools
+%{summary}.
+
+%package devel
+Summary: hwloc development libraries and headers
+Requires: %{name}
+Requires: %{_cross_os}systemd-devel
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n hwloc-%{version} -p1
+
+%build
+%cross_configure \
+    --disable-cairo \
+    --disable-gl \
+    --disable-libxml2 \
+    --disable-opencl \
+    --disable-pci \
+    --disable-plugins \
+    --exec-prefix=%{_cross_prefix} \
+    --program-prefix=""
+
+%force_disable_rpath
+
+%make_build
+
+%install
+%make_install
+
+%files
+%license COPYING
+%{_cross_attribution_file}
+%{_cross_libdir}/libhwloc.so
+%{_cross_libdir}/libhwloc.so.*
+%exclude %{_cross_bindir}/lstopo
+%exclude %{_cross_bindir}/hwloc-compress-dir
+%exclude %{_cross_bindir}/hwloc-gather-topology
+%exclude %{_cross_datadir}
+%exclude %{_cross_mandir}
+
+%files tools
+%{_cross_bindir}/hwloc-annotate
+%{_cross_bindir}/hwloc-ls
+%{_cross_bindir}/hwloc-ps
+%{_cross_bindir}/hwloc-bind
+%{_cross_bindir}/hwloc-calc
+%{_cross_bindir}/hwloc-diff
+%{_cross_bindir}/hwloc-distrib
+%{_cross_bindir}/hwloc-info
+%{_cross_bindir}/hwloc-patch
+%{_cross_bindir}/lstopo-no-graphics
+# These are not on aarch64
+%if "%{_cross_arch}" == "x86_64"
+%{_cross_sbindir}/hwloc-dump-hwdata
+%{_cross_bindir}/hwloc-gather-cpuid
+%endif
+
+%files devel
+%{_cross_includedir}/hwloc.h
+%{_cross_includedir}/hwloc/*.h
+%{_cross_includedir}/hwloc/autogen/*.h
+%{_cross_pkgconfigdir}/*.pc
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,6 +102,7 @@ dependencies = [
  "glibc",
  "grep",
  "host-ctr",
+ "hwloc",
  "iproute",
  "iptables",
  "iputils",
@@ -406,6 +407,14 @@ name = "host-ctr"
 version = "0.1.0"
 dependencies = [
  "glibc",
+]
+
+[[package]]
+name = "hwloc"
+version = "0.1.0"
+dependencies = [
+ "glibc",
+ "systemd-252",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ members = [
   "packages/host-ctr",
   "packages/iproute",
   "packages/iptables",
+  "packages/hwloc",
   "packages/iputils",
   "packages/kexec-tools",
   "packages/keyutils",
--- a/kits/bottlerocket-core-kit/Cargo.toml
+++ b/kits/bottlerocket-core-kit/Cargo.toml
@@ -49,6 +49,7 @@ findutils = { path = "../../packages/findutils" }
 glibc = { path = "../../packages/glibc" }
 grep = { path = "../../packages/grep" }
 host-ctr = { path = "../../packages/host-ctr" }
+hwloc = { path = "../../packages/hwloc" }
 iproute = { path = "../../packages/iproute" }
 iptables = { path = "../../packages/iptables" }
 iputils = { path = "../../packages/iputils" }


ndbaker1 (approved):

mgsharm (approved):

packages/hwloc/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+package-name = "hwloc"
+releases-url = "https://download.open-mpi.org/release/"
```
bcressey: This takes me to an S3 access denied error page.

packages/hwloc/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+package-name = "hwloc"
```
bcressey: not necessary if the package name doesn't contain special chars:
```suggestion
```

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,78 @@
+Name: %{_cross_os}hwloc
+Version: 2.12.2
+Release: 1%{?dist}
+Summary: Portable hardware locality library
+URL: http://www.open-mpi.org/projects/hwloc/
+License: BSD-2-Clause
```
bcressey: [COPYING](https://github.com/open-mpi/hwloc/blob/master/COPYING) states `BSD-3-Clause`.

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,78 @@
...
+
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}systemd-devel
+
+%description
+%{summary}.
+
+%package devel
+Summary: hwloc development libraries and headers
+Requires: %{name}
```
bcressey, line 24: You can check the `.pc` file but often you need to include other devel dependencies here as runtime requirements:
```suggestion
Requires: %{name}
Requires: %{_cross_os}systemd-devel
```

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,78 @@
...
+%{summary}.
+
+%prep
+%autosetup -n hwloc-%{version} -p1
+
+%build
+%cross_configure \
+    --disable-cairo \
+    --disable-gl \
+    --disable-libudev \
```
bcressey: Isn't `libudev` what the systemd-devel dependency provides?

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,78 @@
...
+%{_cross_bindir}/hwloc-ls
+%{_cross_bindir}/hwloc-ps
+%{_cross_bindir}/hwloc-bind
+%{_cross_bindir}/hwloc-calc
+%{_cross_bindir}/hwloc-diff
+%{_cross_bindir}/hwloc-distrib
+%{_cross_bindir}/hwloc-info
+%{_cross_bindir}/hwloc-patch
+%{_cross_bindir}/lstopo
+%{_cross_bindir}/lstopo-no-graphics
```
bcressey, line 72: These are all binaries, not helper scripts?

Can we omit any of them?

packages/release/release.spec
```diff
@@ -136,6 +136,7 @@ Requires: %{_cross_os}findutils
 Requires: %{_cross_os}glibc
 Requires: %{_cross_os}grep
 Requires: %{_cross_os}grub
+Requires: %{_cross_os}hwloc
```
bcressey: > Add the binary package to release to include in all variants since EFA can also use hwloc for finding colocation-relevant data.

You're saying that EFA (the kernel driver) does an upcall to `hwloc` in the host OS? Or what exactly does that dependency look like?

I'd prefer to make this NVIDIA-only if there's nothing in the host that actually uses this.

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,78 @@
...
+%force_disable_rpath
+
+%make_build
+
+%install
+%make_install
+
+find %{buildroot} -name '*.la' -exec rm -f {} ';'
+find %{buildroot} -name '*.debug' -exec rm -f {} ';'
+
```
bcressey: The `.la` files should be removed by `brp-remove-la-files` automatically.

Why are there `*.debug` files that need to be removed?

packages/release/release.spec
yeazelm: I went to find the specific references to these to trim them down and I'm not finding direct usage of `hwloc-ls`. The errors I first encountered are referencing [this code](https://github.com/open-mpi/ompi/blob/v4.1.8/orte/mca/rtc/hwloc/rtc_hwloc.c#L424-L447) which has a library dependency on hwloc. In light of not finding any specifics, I think we should avoid adding this to images by default, and use this more as a debugging package when needed.

packages/hwloc/hwloc.spec
yeazelm: The primary use case I've seen is using `hwloc-ls` for confirming topology. I think `hwloc-info` might be similar. I added a reply to another comment but I'm not able to find the specific invocation we need for any specific binary so I think we may just keep this as a package that can be included in `-dev` variants. Either way, I think we can drop `lstopo` which is graphically oriented. Outside of the ls/info, the rest are for manipulating the topology files which would be useful if trying to debug this manually but honestly I think this would not be generally how things are done for containers.

packages/hwloc/hwloc.spec
yeazelm: I had these before I got the flags dialed in, they are no longer needed.

yeazelm:
^ Updated to no longer include hwloc in `release` and updated from the rest of the comments from the PR.

packages/hwloc/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://www.open-mpi.org/software/ompi/v5.0/"
+
+[[package.metadata.build-package.external-files]]
+url = "https://download.open-mpi.org/release/hwloc/v2.12/hwloc-2.12.1.tar.bz2"
```
KCSesh: Not a blocker - but they have released: 2.12.2

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,75 @@
+Name: %{_cross_os}hwloc
+Version: 2.12.1
+Release: 1%{?dist}
+Summary: Portable hardware locality library
+URL: http://www.open-mpi.org/projects/hwloc/
```
KCSesh:
```suggestion
URL: https://www.open-mpi.org/projects/hwloc/
```

packages/hwloc/Cargo.toml
KCSesh: Can we use:
`https://www.open-mpi.org/projects/hwloc` 

Their[ github release page](https://github.com/open-mpi/hwloc/releases) seems to link that and gives nice info

packages/hwloc/hwloc.spec
yeazelm: Updated

packages/hwloc/Cargo.toml
yeazelm: Updated

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,75 @@
...
+%build
+%cross_configure \
+    --disable-cairo \
+    --disable-gl \
+    --disable-libxml2 \
+    --disable-opencl \
+    --disable-pci \
+    --disable-plugins \
+    --enable-static \
+    --disable-shared \
```
arnaldo2792: Why enabling `static` libraries but not `shared` libraries?  Is it because `hwloc` will be used within a container?

packages/hwloc/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[package.metadata.build-package]
+releases-url = "https://www.open-mpi.org/software/ompi/v5.0/"
+
+[[package.metadata.build-package.external-files]]
+url = "https://download.open-mpi.org/release/hwloc/v2.12/hwloc-2.12.2.tar.bz2"
+sha512 = "949d6c9d7b858ee58e477b15e6c06f57812872142fa1c7f3ef20aae2e4ef954135f839e8604404bfd0637fde99729c7d00211c8aee860dfde9ac60bba0e78aef"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
```
arnaldo2792, line 19: Can we begin skipping `glibc` as a dependency of the packages @bcressey ? Now that `glibc-devel` is provided by the SDK?

packages/hwloc/hwloc.spec
yeazelm: That was the intention. If we don't need them mounted into containers we can use shared instead and save of space.

yeazelm: The intention was to allow these to be mounted into a container. If we don't expect that to happen, we can switch back to dynamic. It will save space on the filesystem so I'm ok to do that.

packages/hwloc/hwloc.spec
```diff
@@ -0,0 +1,75 @@
...
+
+%install
+%make_install
+
+%files
+%license COPYING
+%{_cross_attribution_file}
+%{_cross_bindir}/hwloc-annotate
+%{_cross_bindir}/hwloc-compress-dir
+%{_cross_bindir}/hwloc-gather-topology
```
bcressey: These are both shell scripts, at least on Fedora.

bcressey (commented):
If we don't have use cases in mind for the tools, I'd prefer to invert the packaging and have `libhwloc`, `libhwloc-devel`, and `libhwloc-tools` - so that the tools are optional when the shared library is needed.

Compare to `libcrypto-tools` and `libcryptsetup-tools`.

packages/hwloc/hwloc.spec
yeazelm: I somehow missed these the last time I looked at the files. I've excluded them and moved the rest into a tools package.

yeazelm:
I rebased the changes and pushed an updated version that provides the shared object and three packages:
hwloc - primary package vends hwloc.so*
hwloc-tools - provides the binaries
hwloc-devel - provides build dependencies

mgsharm (approved):

bcressey (approved):

PR #673, opened 2025-09-26
Update first-party packages

vigh-m:
**Description of changes:**
- Update the dependencies of our first party sources (Rust and Go packages)
- Removed the unused `quick-xml` dependency
- Updated schnauzer toml_encode unit-tests as the render output changed slightly:
  <Details>
  The unit tests gave the following error:
  
  ```
  thread 'helpers::stdlib::test_toml_encode::toml_encode_toml_injection_1' panicked at api/schnauzer/src/helpers/stdlib/mod.rs:1235:9:
  assertion `left == right` failed
    left: "[\"\"\"apiclient set motd=hello', 'echo pwned\"\"\"\"]"
   right: "['''apiclient set motd=hello', 'echo pwned\"''']"
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

  ---- helpers::stdlib::test_toml_encode::toml_encode_toml_injection_3 stdout ----

  thread 'helpers::stdlib::test_toml_encode::toml_encode_toml_injection_3' panicked at api/schnauzer/src/helpers/stdlib/mod.rs:1257:9:
  assertion `left == right` failed
    left: "[\"\"\"apiclient set motd=hello\", \"echo pwned\", 'echo pwned2'\"\"\"]"
   right: "[\"apiclient set motd=hello\\\", \\\"echo pwned\\\", 'echo pwned2'\"]"
  ```
  
  The way the characters are escaped has changed but there was no difference in behaviour on the node when tested. Settings are still rendered correctly.  

  </Details>
- Unable to remove `serde-xml-rs` due to the dependency in `early-boot-config` for vmware
- Clippy changes

**Testing done:**
-  Launched aws-k8s and aws-ecs3 regular and nvidia variants and verified functionality via smoke tests
- AMIs are able to launch and join the cluster without issue
- bootstrap and host containers startup without error
- `/var/lib/bottlerocket/bootconfig.data` changes correctly (prairiedog test( )
- No errors on any systemd services observed
- Tested with custom registry for certdog. I saw 404s as expected.
- `apiclient exec` works as expected:
  ```
  [ssm-user@control]$ apiclient exec admin echo hello
  hello
  ERRO[0000] resize pty                                    error="cannot resize a stopped container"
  ```

**TODOs/Follow ups**
- Unable to update `aws-lc-rs` since `01076072268b5562b22a101aae5ed4bdcbd79090` in `aws-lc` is causing build failures
-  Unable to update to latest `containerd/nri` due to incompatible change. Need to investigate major version bump for containerd deps
- Added `rand` to the allowlist for multiple-packages. `actix-web` has updated to rand v0.9 but we need to evaluate our own sources before we can do the same

**Terms of contribution:**


vigh-m:
⬆️ Pushed fixes for typos and fix for multiple version of the `rand` package

vigh-m:
⬆️ Fixed unit tests and clippy errors

sources/deny.toml
```diff
@@ -66,6 +61,10 @@ skip-tree = [
     # aws-smithy-experimental brings a set of new dependencies that
     # the aws-sdk doesn't support
     { name = "aws-smithy-experimental", version = "=0.1" },
+    # actix-http has upgraded to v0.9 of 'rand' but upstream has called out some
+    # breaking changes which will need to be explored for our first party code
+    # https://rust-random.github.io/book/update-0.9.html
+    { name = "rand" },
```
bcressey: Can we pin this to a specific version of `rand`? Please make sure there's a backlog issue open for this.

vigh-m:
⬆️ 'check-fmt' fixes

sources/api/schnauzer/src/helpers/stdlib/mod.rs
```diff
@@ -1231,7 +1228,7 @@ mod test_toml_encode {
             &json!({"settings": {"foo-string": [ "apiclient set motd=hello', 'echo pwned\""]}}),
         )
         .unwrap();
-        let expected = "['''apiclient set motd=hello', 'echo pwned\"''']";
+        let expected = "[\"\"\"apiclient set motd=hello', 'echo pwned\"\"\"\"]";
```
bcressey: Not sure we want these changes to `test_toml_encode` - the goal is to test very specific literal strings, and this is changing those literals.

What's the specific lint that's failing? Can we disable that?

sources/Cargo.lock
```diff
@@ -2870,41 +2937,43 @@ dependencies = [
...
+ "pin-utils",
  "smallvec",
  "tokio",
  "want",
 ]
 
 [[package]]
 name = "hyper-http-proxy"
-version = "1.0.0"
+version = "1.1.0"
```
bcressey: We should hold back the `hyper-http-proxy` update until we verify that the new functionality to set an authorization header is doing what we expect.

sources/api/schnauzer/src/helpers/stdlib/mod.rs
vigh-m: It wasn't a linter but the unit tests itself that broke.  
The returned output from the `setup_and_render_template()` methods changed from 
`"['''apiclient set motd=hello', 'echo pwned\"''']"` to
`"[\"\"\"apiclient set motd=hello', 'echo pwned\"\"\"\"]"`

There was no difference in behaviour that I could see and testing did not turn up anything.

vigh-m:
⬆️ Added a specific version for `rand` and also pinned `hyper-http-proxy` to prevent update to 1.1.0 until further testing is completed

vigh-m:
Further testing done:
- `apiclient exec` works as expected:
```
[ssm-user@control]$ apiclient exec admin echo hello
hello
ERRO[0000] resize pty                                    error="cannot resize a stopped container"
```

`ERRO[0000] resize pty` is an existing bug present on current releases as well.

vigh-m:
⬆️ Updated the unit-test logic to be more robust

bcressey (approved):

KCSesh (approved):

PR #677, opened 2025-10-01
Provide dbus-launcher as its own package

arnaldo2792:
**Issue number:**

Part of #660 

**Description of changes:**

Make the dbus-broker require the new 'dbus-launcher' capability, so that variants can choose what they use to launch the broker.

Break down the dbus-launcher-broker as its own package, which provides the 'dbus-launcher' capability with a higher epoch number so that it is installed by default.

Added a patch to allow install the generated journal catalogs, even when the `launcher` isn't built, in preparation for the expat removal.

**Testing done:**

I confirmed that without making changes to the existing variants, the `dbus-broker-launcher` package is selected by default as the `dbus-launcher` capability.

**Terms of contribution:**

### DIFF
--- a/packages/dbus-broker/dbus-broker.spec
+++ b/packages/dbus-broker/dbus-broker.spec
@@ -21,11 +21,23 @@ BuildRequires: %{_cross_os}systemd-devel
 Requires: %{_cross_os}libexpat
 Requires: %{_cross_os}libselinux
 Requires: %{_cross_os}systemd
+Requires: %{_cross_os}dbus-broker(launcher)
 
 # Work around an aliasing rules violation.
 Patch0001: 0001-c-utf8-disable-strict-aliasing-optimizations.patch
+# Allow building the journal catalogs when dbus-launcher is excluded
+Patch0002: 0002-meson.build-remove-condition-to-build-the-journal-ca.patch
 
 %description
+%{summary}.
+
+%package launcher
+Summary: Launcher for the D-BUS message broker
+Provides: %{_cross_os}dbus-broker(launcher) = 1:
+Conflicts: %{_cross_os}dbus-broker(launcher)
+Requires: %{name}
+
+%description launcher
 %{summary}.
 
 %prep
@@ -39,6 +51,7 @@ CONFIGURE_OPTS=(
  -Ddocs=false
  -Dlauncher=true
  -Dselinux=true
+ -Dcatalogdir=%{_cross_journalcatalogdir}
 )
 
 %cross_meson "${CONFIGURE_OPTS[@]}"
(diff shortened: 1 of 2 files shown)


packages/dbus-broker/dbus-broker.spec
```diff
@@ -21,13 +22,25 @@ BuildRequires: %{_cross_os}systemd-devel
...
 # Work around an aliasing rules violation.
 Patch0001: 0001-c-utf8-disable-strict-aliasing-optimizations.patch
+# Allow building the journal catalogs when dbus-launcher is excluded
+Patch0002: 0002-meson.build-remove-condition-to-build-the-journal-ca.patch
 
 %description
 %{summary}.
 
+%package launcher
+Summary: A dbus-broker launcher
```
bcressey: nit:
```suggestion
Summary: Launcher for the D-BUS message broker
```

packages/dbus-broker/dbus-broker.spec
```diff
@@ -60,14 +76,17 @@ install -p -m 0644 %{S:13} %{buildroot}%{_cross_sysusersdir}/dbus.conf
...
-%{_cross_unitdir}/dbus-broker.service
 %{_cross_unitdir}/dbus.socket
+%{_cross_unitdir}/dbus-broker.service
 %exclude %{_cross_userunitdir}/dbus-broker.service
 
+%files launcher
+%{_cross_bindir}/dbus-broker-launch
+%{_cross_unitdir}/dbus-broker.service.d/dbus-broker-launcher.conf
+%{_cross_journalcatalogdir}/dbus-broker-launch.catalog
+%{_cross_datadir}/dbus-1/*
```
bcressey: nit: can you list individual files and directories instead of a wildcard?

packages/dbus-broker/dbus-broker.spec
```diff
@@ -39,6 +52,7 @@ CONFIGURE_OPTS=(
  -Ddocs=false
  -Dlauncher=true
  -Dselinux=true
+ -Dcatalogdir=%{_cross_journalcatalogdir}
```
bcressey, line 54: Why is this a required option now when it wasn't before?

packages/dbus-broker/dbus-broker.service
```diff
@@ -15,8 +15,8 @@ LimitNOFILE=16384
 ProtectSystem=full
 PrivateTmp=true
 PrivateDevices=true
-ExecStart=/usr/bin/dbus-broker-launch --scope system
-ExecReload=/usr/bin/busctl call org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus ReloadConfig
+# Must be overriden by a drop-in
+ExecStart=/usr/bin/false
```
bcressey: I don't think we want this coupling long-term - whippet should have its own `whippet.service` unit so that we can delete this one when we drop this launcher.

packages/dbus-broker/dbus-broker.spec
arnaldo2792: It is automatically set when `-Dlauncher=true`. Once we stop building the launcher, this config won't bet set. This is just to prepare the build for when the launcher isn't build anymore.

arnaldo2792:
Forced push includes:
- Updates to the new package's summary based off feedback
- List individual files under `%{_cross_datadir}/dbus-1/`
- Don't provide a drop-in for the dbus-launcher package

yeazelm (approved):

bcressey (approved):

PR #678, opened 2025-10-01
Provide whippet and dbus-launcher configs

arnaldo2792:
**Issue number:**

Part of #660

**Description of changes:**

This series provides the `whippet` policies translated from the configurations provided by systemd for `logind`, `networkd` and `resolved`.

As part of this change, the XML configurations for the `dbus-launcher` are now provided in a subpackage, so that the correct configurations are selected depending on the packages available in the variant. 

**Testing done:**

In combination with #677 and #661:

- I modified the `aws-dev` variant, to include `whippet` in the list of installed packages. I confirmed that the `whippet` configuration files were installed and the `dbus-launcher` configuration files weren't
- Without modifications to the `aws-k8s-1.33` variant, I confirmed that the `dbus-launcher` configuration files were installed, and the `whippet` configuration files weren't

**Terms of contribution:**

### DIFF
--- a/packages/systemd-252/org.freedesktop.login1.toml
+++ b/packages/systemd-252/org.freedesktop.login1.toml
@@ -1,0 +1,509 @@
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+allow = false
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.DBus.Introspectable"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.DBus.Peer"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.DBus.Properties"
+send_member = "Get"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.DBus.Properties"
+send_member = "GetAll"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "GetSession"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "GetSessionByPID"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "GetUser"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "GetUserByPID"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "GetSeat"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "ListSessions"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "ListUsers"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "ListSeats"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "ListInhibitors"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "Inhibit"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "SetUserLinger"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "ActivateSession"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "ActivateSessionOnSeat"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "LockSession"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "UnlockSession"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "LockSessions"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "UnlockSessions"
+allow = true
+
+[[default.rules]]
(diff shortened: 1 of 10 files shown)


vigh-m:
If the configs are the same for both versions of systemd, does it make sense to include them in the `release` package instead.

packages/systemd-252/org.freedesktop.login1.toml
```diff
@@ -0,0 +1,515 @@
+# Generated programatically, serde/toml doesn't allow to force the "compact"
+# format on serialization:
+# [user.root]
+# rules = []
```
bcressey: I don't understand this comment - did the programmatic generator write it and write out the commented lines? What program was used?

packages/systemd-252/systemd-252.spec
```diff
@@ -164,11 +170,40 @@ Provides: %{_cross_os}systemd-networkd = %{package_priority_epoch}:
...
+%description dbus-launcher-config
+%{summary}.
+
+%package whippet-config
+Summary: Whippet configuration files
+Requires: %{name}
+Provides: %{name}(dbus-config) = 1:
+Provides: %{name}-resolved(dbus-config) = 0:
+Provides: %{name}-networkd(dbus-config) = 0:
+Conflicts: (%{_cross_os}dbus-broker-launcher or %{name}-dbus-launcher-config)
```
bcressey: nit: this conflict is doing too much work. It also feels like a bug that `whippet-config` doesn't require `whippet` and `dbus-launcher-config` doesn't require `dbus-broker-launcher`.

You can just ensure that `dbus-broker-launcher` and `whippet`  - the respective dependencies - conflict with each other, and that these config packages conflict with each other, and that each depends on its launcher.
```suggestion
Requires: %{_cross_os}whippet
Conflicts: %{name}-dbus-launcher-config
```

packages/systemd-252/systemd-252.spec
```diff
@@ -164,11 +170,40 @@ Provides: %{_cross_os}systemd-networkd = %{package_priority_epoch}:
...
+Conflicts: %{name}-resolved(dbus-config)
+Conflicts: %{name}-networkd(dbus-config)
+
+%description dbus-launcher-config
+%{summary}.
+
+%package whippet-config
+Summary: Whippet configuration files
+Requires: %{name}
+Provides: %{name}(dbus-config) = 1:
```
bcressey: Why is this provided at priority 1 instead of 0?

packages/systemd-252/systemd-252.spec
```diff
@@ -164,11 +170,40 @@ Provides: %{_cross_os}systemd-networkd = %{package_priority_epoch}:
...
 
 %description resolved
 %{summary}.
 
+%package dbus-launcher-config
+Summary: Dbus launcher configuration files
+Requires: %{name}
+Provides: %{name}(dbus-config) = 1:
+Provides: %{name}-resolved(dbus-config) = 1:
+Provides: %{name}-networkd(dbus-config) = 1:
```
bcressey: Not sure there's much value in these separate capabilities if they're always provided as a group.
```suggestion
Provides: %{name}(dbus-config) = 1:
```

packages/systemd-257/systemd-257.spec
```diff
@@ -773,6 +805,19 @@ install -p -m 0644 %{S:2} %{buildroot}%{_cross_bootconfigdir}/21-cgroup-enable-l
...
+%{_cross_datadir}/dbus-1/system.d/org.freedesktop.systemd1.conf
+%{_cross_datadir}/dbus-1/system.d/org.freedesktop.resolve1.conf
+%{_cross_datadir}/dbus-1/system-services
+%{_cross_datadir}/dbus-1/services/org.freedesktop.systemd1.service
+
+%files whippet-config
+%{_cross_datadir}/whippet/policies.d/org.freedesktop.login1.toml
+%{_cross_datadir}/whippet/policies.d/org.freedesktop.network1.toml
+%{_cross_datadir}/whippet/policies.d/org.freedesktop.systemd1.toml
+%{_cross_datadir}/whippet/policies.d/org.freedesktop.resolve1.toml
```
bcressey: It would be less work overall to just ship both the XML and TOML configs in the same packages that previously just shipped the XML configs. Then which one is used only depends on which launcher binary gets installed.

packages/systemd-252/org.freedesktop.login1.toml
```diff
@@ -0,0 +1,515 @@
...
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Session"
+send_member = "SetDisplay"
+allow = true
+
+[[default.rules]]
+receive_sender = "org.freedesktop.login1"
+allow = true
+
```
bcressey, line 498: nit: programmatic generator should not add superfluous empty lines at the end of the file 😀 
```suggestion
```

packages/systemd-257/org.freedesktop.systemd1.toml
```diff
@@ -0,0 +1,579 @@
+# Generated programatically, serde/toml doesn't allow to force the "compact"
+# format on serialization:
+# [user.root]
+# rules = []
```
bcressey: This is also missing systemd 257 methods like `QueueSignal`.

packages/systemd-257/org.freedesktop.login1.toml
```diff
@@ -0,0 +1,515 @@
+# Generated programatically, serde/toml doesn't allow to force the "compact"
+# format on serialization:
+# [user.root]
+# rules = []
```
bcressey: For systemd 257, this is missing:

* `ListSessionsEx`
* `Sleep`
* `ReleaseSession`
* `SetTTY`

packages/systemd-257/org.freedesktop.login1.toml
```diff
@@ -0,0 +1,515 @@
...
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "GetSession"
+allow = true
+
+[[default.rules]]
+send_destination = "org.freedesktop.login1"
+send_interface = "org.freedesktop.login1.Manager"
+send_member = "GetSessionByPID"
+allow = true
```
bcressey, line 37: Not a blocker but these rules would be a lot easier to read if we allowed a list of methods per (destination, interface).
```
[[default.rules]]
send_destination = "org.freedesktop.login1"
send_interface = "org.freedesktop.login1.Manager"
send_members = [
  "GetSession",
  "GetSessionByPID",
]
allow = true
```

That would significantly reduce the boilerplate in these files.

arnaldo2792:
Forced push includes:
- Simplify how configurations for whippet and the dbus-launcher are provided
- Fixed rules for systemd-257
- Don't include comments in toml files generated by helper policy parser

packages/systemd-252/org.freedesktop.network1.toml
```diff
@@ -0,0 +1,19 @@
+[[default.rules]]
+send_destination = "org.freedesktop.network1"
+allow = true
```
bcressey, line 3: nit: the `.rules` on the end of each list element is kind of repetitive - if it were easy to eliminate via serde then I'd vote to scrap it

packages/systemd-252/systemd-252.spec
```diff
@@ -348,10 +347,13 @@ rm -f %{buildroot}%{_cross_libdir}/systemd/{system,user}/graphical.target
 ln -s  ../proc-sys-fs-binfmt_misc.mount \
   %{buildroot}%{_cross_unitdir}/sysinit.target.wants/proc-sys-fs-binfmt_misc.mount
 
+install -d %{buildroot}%{_cross_datadir}/whippet/policies.d
+install -p -m 0644 %{S:100} %{S:101} %{S:102} %{S:103} %{buildroot}%{_cross_datadir}/whippet/policies.d
 
 # Remove any README files.
 find %{buildroot} -type f -name README -print -delete
 
+
```
bcressey: nit: no random newlines please
```suggestion
```

bcressey (approved):

yeazelm (approved):

arnaldo2792:
I did one more test, along with all the other changes linked. I confirmed the correct set of rules are shipped with their corresponding systemd version:

```bash
bash-5.1# grep -R ReleaseSession /usr/share/ | wc -l
2
bash-5.1# grep -R ListSessionsEx /usr/share/ | wc -l
2
bash-5.1# grep -R SetTTY /usr/share/ | wc -l
2
bash-5.1# grep -R \"Sleep\" /usr/share/ | wc -l
2
bash-5.1#
```

PR #686, opened 2025-10-06
feat: add support for MTU in Netdog

### LINKED ISSUE #3341: Configurable MTU
**What I'd like:**
I'd like to be able to configure the MTU of EC2 instances running bottlerocket. 

I'm using cilium on AWS EKS with cluster mesh across multiple AWS VPCs through a Transit Gateway. Unfortunately, Transit Gateways support a max of 8500 MTU and the EKS optimized AMIs use jumbo frames.

None

Related discussion: https://github.com/bottlerocket-os/bottlerocket/discussions/3338

mikn:
**Issue number:** https://github.com/bottlerocket-os/bottlerocket/issues/3341

Closes https://github.com/bottlerocket-os/bottlerocket/issues/3341

**Description of changes:** This introduces the ability to set MTU as part of netdog configuration in net.toml.

**Testing done:** Only the testing contained in the PR so far.

**Terms of contribution:**

### DIFF
--- a/sources/netdog/src/net_config/v3.rs
+++ b/sources/netdog/src/net_config/v3.rs
@@ -108,7 +108,7 @@ impl Validate for NetConfigV3 {
 #[cfg(test)]
 mod tests {
     use crate::net_config::test_macros::{
-        basic_tests, bonding_tests, dhcp_tests, static_address_tests, vlan_tests,
+        basic_tests, bonding_tests, dhcp_tests, mtu_tests, static_address_tests, vlan_tests,
     };
 
     basic_tests!(3);
@@ -116,4 +116,5 @@ mod tests {
     static_address_tests!(3);
     vlan_tests!(3);
     bonding_tests!(3);
+    mtu_tests!(3);
 }
--- a/sources/netdog/src/networkd/devices/vlan.rs
+++ b/sources/netdog/src/networkd/devices/vlan.rs
@@ -24,6 +24,7 @@ pub(crate) struct NetworkDVlan {
     // entry for this VLAN
     pub(crate) device: InterfaceName,
     pub(crate) id: VlanId,
+    pub(crate) mtu: Option<u32>,
 }
 
 impl NetDevFileCreator for NetworkDVlan {
@@ -40,6 +41,7 @@ impl NetDevFileCreator for NetworkDVlan {
             routes: _,
             device: _, // Device isn't used in .netdev files
             id,
+            mtu: _, // MTU is configured in .network files, not .netdev
         } = self;
 
         let mut netdev = NetDevBuilder::new_vlan(name.clone());
@@ -63,6 +65,7 @@ impl NetworkFileCreator for NetworkDVlan {
             routes,
             device: _, // device and id aren't used in .network files
             id: _,
+            mtu,
         } = self;
 
         let mut network = NetworkBuilder::new_vlan(name.clone());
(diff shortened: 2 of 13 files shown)


sources/netdog/src/networkd/devices/vlan.rs
```diff
@@ -63,13 +65,17 @@ impl NetworkFileCreator for NetworkDVlan {
...
         } = self;
 
         let mut network = NetworkBuilder::new_vlan(name.clone());
         network.with_dhcp(dhcp4.clone(), dhcp6.clone());
         maybe_add_some!(network, with_static_config, static4);
         maybe_add_some!(network, with_static_config, static6);
         maybe_add_some!(network, with_routes, routes);
+        if let Some(m) = mtu {
+            network.with_mtu(*m);
+        }
```
bcressey, line 78: Same suggestion for the other occurrences of this pattern:
```suggestion
    maybe_add_some!(network, with_mut, mtu);
```

sources/netdog/src/net_config/v3.rs
```diff
@@ -108,12 +108,13 @@ impl Validate for NetConfigV3 {
...
-        basic_tests, bonding_tests, dhcp_tests, static_address_tests, vlan_tests,
+        basic_tests, bonding_tests, dhcp_tests, mtu_tests, static_address_tests, vlan_tests,
     };
 
     basic_tests!(3);
     dhcp_tests!(3);
     static_address_tests!(3);
     vlan_tests!(3);
     bonding_tests!(3);
+    mtu_tests!(3);
```
bcressey, line 119: We'll need to increment the schema version for the MTU-related changes:
```suggestion
    mtu_tests!(4);
```

bcressey (commented):
It's been a while since I've reviewed this code, and the original maintainer has moved on, but my sense is we're meant to add new versions of every struct that gets updated to include an MTU, and tie that all together in a `NetConfigV4` struct.

```
pub(crate) struct NetConfigV4 {
      pub(crate) net_devices: IndexMap<InterfaceId, NetworkDeviceV2>,
}

pub(crate) enum NetworkDeviceV2 {
    Interface(NetInterfaceV3),
    BondDevice(NetBondV2),
    VlanDevice(NetVlanV2),
}
```

Otherwise the versioning scheme in `net.toml` will break down; an older `netdog` that supports `version = 3` won't be able to parse the new MTU fields if present.

Ideally also common shapes and implementations would be shifted into shared modules rather than copy-pasted.

It's rather painful when the goal is just to add one new field, which is one reason why `net.toml` has been so slow to evolve. The pain involved also creates an incentive to wait and group a lot of new features together as part of each version increment.

mikn:
> It's been a while since I've reviewed this code, and the original maintainer has moved on, but my sense is we're meant to add new versions of every struct that gets updated to include an MTU, and tie that all together in a `NetConfigV4` struct.
> 
> ```
> pub(crate) struct NetConfigV4 {
>       pub(crate) net_devices: IndexMap<InterfaceId, NetworkDeviceV2>,
> }
> 
> pub(crate) enum NetworkDeviceV2 {
>     Interface(NetInterfaceV3),
>     BondDevice(NetBondV2),
>     VlanDevice(NetVlanV2),
> }
> ```
> 
> Otherwise the versioning scheme in `net.toml` will break down; an older `netdog` that supports `version = 3` won't be able to parse the new MTU fields if present.
> 
> Ideally also common shapes and implementations would be shifted into shared modules rather than copy-pasted.
> 
> It's rather painful when the goal is just to add one new field, which is one reason why `net.toml` has been so slow to evolve. The pain involved also creates an incentive to wait and group a lot of new features together as part of each version increment.

It was something I was hoping wasn't true as I was implementing this 🫠 But yes I'll sort it out.

PR #690, opened 2025-10-10
fix: only fetch pending settings if they'll be logged

fletcherw:
check_pending_settings logs the fetched settings at debug level. So, if our log level is lower, don't bother fetching since it will have no visible output.


bcressey (approved):
Nice find!

KCSesh (approved):

koooosh (approved):

PR #706, opened 2025-11-05
add libudev for bootstrapping purposes

bcressey:
**Issue number:**

**Description of changes:**
Add a bootstrap build of systemd for the purpose of providing libudev to packages that are dependencies of the final systemd package. This is required to build `libdevmapper` and `libcryptsetup` with udev support, which improves the coordination between the kernel and userspace when adding devices.

Also add a bonus patch to quiet the message about FIPS mode.

**Testing done:**
`dmsetup` shows that the udev synchronization commands are available:
```
# dmsetup udevcookies
Cookie       Semid      Value      Last semop time           Last change time
```

**Terms of contribution:**

### DIFF
--- a/packages/libudev/Cargo.toml
+++ b/packages/libudev/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "libudev"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/systemd/systemd-stable/releases"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/systemd/systemd-stable/archive/v252.39/systemd-stable-252.39.tar.gz"
+sha512 = "0a24faf0dd6da9d5b55dafab332bdd40bfb90f643bd6caa3b6ecfcb8feaeabe51b1cf007f033e8d9ea3f52efdb4722ed63816b31646d70b99c22eb1eb2705450"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
+kmod = { path = "../kmod" }
+libacl = { path = "../libacl" }
+libattr = { path = "../libattr" }
+libcap = { path = "../libcap" }
+libseccomp = { path = "../libseccomp" }
+libselinux = { path = "../libselinux" }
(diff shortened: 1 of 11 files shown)


bcressey:
Without these changes, then encrypting the `BOTTLEROCKET-DATA` partition causes the system to block waiting for `/dev/mapper/BOTTLEROCKET-DATA` to appear, even though the device exists. This fixes the udev integration which fixes the systemd integration so everything "just works".

packages/libudev/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/systemd/systemd-stable/releases"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/systemd/systemd-stable/archive/v252.22/systemd-stable-252.22.tar.gz"
```
vigh-m: We have a PR open to bump `systemd-stable` to the latest minor version. Would it make sense to take that update here from the beginning?

vigh-m (approved):

packages/systemd-252/systemd-252.spec
```diff
@@ -132,6 +132,7 @@ Requires: %{_cross_os}libuuid
 Requires: %{_cross_os}libxcrypt
 
 Provides: %{_cross_os}systemd = %{package_priority_epoch}:
+Provides: %{_cross_os}libudev = %{package_priority_epoch}:
```
arnaldo2792, line 135: Is this needed to help RPM choose this package instead of the bootstrap package?

arnaldo2792 (approved):

packages/systemd-252/systemd-252.spec
bcressey: > Is this needed to help RPM choose this package instead of the bootstrap package?

`libdevmapper` ends up with a requirement on `libudev`, so this helps it pick the one that goes with the selected systemd package.

`libcryptsetup` does not actually link with `libudev`, from what `readelf` tells me, so I didn't add a requirement there.

The conflicts I added to the bootstrap package (`filesystem`, `release`, `systemd`) should be enough to block it from being installed in any actual image, or if the final package is available.

PR #711, opened 2025-11-06
Add Docker v29 packages

koooosh:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/712

**Description of changes:**

- Namespace existing `docker-engine` and `docker-cli` pkgs to `docker-engine-25` and `docker-cli-25` respectively
  - Set `package_priority_epoch` to 1 so `docker-engine/cli` resolves to v25 first 
- Add `docker-engine-29` and `docker-cli-29` pkgs, currently on v29.0.0-rc.2
  - `package_priority_epoch` is set to 0
  - Main changes are in `docker-engine-29`
    - Both patches needed to be updated to port to v29
    - Add `nftables` as build dependencies
    - Module/directory path changes in v29

**Testing done:**

- `docker*-25`: ECS conformance tests pass on `aws-ecs-2` variant on both archs
- `docker*-29`: Relevant ECS conformance tests pass on variant using these new pkgs on both archs

**Terms of contribution:**


bcressey (approved):

bcressey:
Your description says "Add systemd-257 and nftables as build dependencies" but `docker-engine-29` still depends on `systemd-252`.

That seems correct to me, since we want to build with the oldest compatible version we can.

packages/docker-cli-25/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "docker-cli"
+name = "docker-cli-25"
```
arnaldo2792, line 2: Just FYI, you will have to update the exiting ECS variants.

arnaldo2792 (approved):

koooosh:
Force push uses the older systemd version as the build dep for `docker-engine` as suggested by @bcressey

packages/docker-cli-25/Cargo.toml
koooosh: We verified rpm resolves to the older docker version without editing the variant defs. This is due to the `Provides: %{_cross_os}docker-%{gorepo} = %{package_priority_epoch}:` line.

packages/docker-engine-29/daemon-json
```diff
@@ -0,0 +1,28 @@
+[required-extensions]
+container-registry = "v1"
+oci-defaults = { version = "v1", helpers = ["oci_defaults"] }
+std = { version = "v1", helpers = ["join_array"] }
++++
+{
+  "log-driver": "journald",
+  "live-restore": true,
+  "max-concurrent-downloads": 10,
```
KCSesh, line 9: Not a blocker - but we do have a setting for this. We should consider wrapping this later, especially for docker-25. 

I imagine some of these settings may not be needed when we use containerd snapshotter in docker-29.

But otherwise, LGTM.

KCSesh (approved):

PR #714, opened 2025-11-06
Add `apiclient network configure` subcommand

### LINKED ISSUE #713: Add network configure API to apiclient
**What I'd like:**
Add `apiclient network configure <URI> command` to configure network settings at runtime. Supports `file://` URIs and base64-encoded content. Can be written to `/.bottlerocket/net.toml` for netdog to apply at next boot.

1. We could extend settings API with 1:1 mappings to what net.toml supports. But that will be more involved and it would be challenging if we have backward incompatible changes in net.toml version schemas.
2. Bootstrap container approach that puts net.toml files to /var/lib/netdog/net.toml and let netdog pick it up. But this approach is no longer supported with the introduction of https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/638

ytsssun:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/713

**Description of changes:**
Adds `apiclient network configure <URI>` command to configure network settings at runtime. The implementation includes:

  - API Server: `New /network/configure` POST endpoint that validates and writes net.toml content to `/.bottlerocket/net.toml`
  - API Client: New `network configure` subcommand supporting `file://` and `base64:` URI schemes
  - Input validation: UTF-8 validation, base64 decoding, and file access with proper error handling
  - Documentation: Updated README template and module documentation

**Testing done:**

**1. Tested the change via `bootstrap-commands` using `base64` scheme**
Started `metal-dev` with command
```bash
./tools/start-local-vm --variant metal-dev --arch $(uname -m) --inject-file net.toml --inject-file user-data.toml
```

My initial `net.toml` only has configurations for a single interface
```
version = 3

[enp0s16]
dhcp4 = true
primary = true
```

My `user-data.toml` uses `bootstrap-commands`
```
[settings.bootstrap-commands.setup-network-inline]
commands = [
  ["apiclient", "network", "configure", "base64:dmVyc2lvbiA9IDMKCltlbnAwczE2XQpkaGNwNCA9IHRydWUKZGhjcDYgPSBmYWxzZQpwcmltYXJ5ID0gdHJ1ZQoKW2VucDBzMTddCmRoY3A0ID0gdHJ1ZQpkaGNwNiA9IGZhbHNlCg=="],
  ["apiclient", "reboot"]
]
mode = "once"
essential = true
```
Where the base64 encoded net.toml is 
```bash
echo "dmVyc2lvbiA9IDMKCltlbnAwczE2XQpkaGNwNCA9IHRydWUKZGhjcDYgPSBmYWxzZQpwcmltYXJ5ID0gdHJ1ZQoKW2VucDBzMTddCmRoY3A0ID0gdHJ1ZQpkaGNwNiA9IGZhbHNlCg==" | 
base64 -d

version = 3

[enp0s16]
dhcp4 = true
dhcp6 = false
primary = true

[enp0s17]
dhcp4 = true
dhcp6 = false
```

Confirmed that the node automatically rebooted and the net.toml was written to the desired path /.bottlerocket/net.toml
```bash
bash-5.2# cat /.bottlerocket/net.toml 
version = 3

[enp0s16]
dhcp4 = true
dhcp6 = false
primary = true

[enp0s17]
dhcp4 = true
dhcp6 = false
```

And both interfaces were brought up
```bash
ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: enp0s16: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 52:54:00:12:34:56 brd ff:ff:ff:ff:ff:ff
3: enp0s17: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 52:54:00:12:34:57 brd ff:ff:ff:ff:ff:ff
4: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN mode DEFAULT group default 
    link/ether 02:42:f1:1e:86:ce brd ff:ff:ff:ff:ff:ff
bash-5.2# cat /etc/systemd/network/10-enp0s16.network
```

**2. Tested the change via `bootstrap-container` using `file://` scheme**
Used similar setup as 1. with `metal-dev`. Only difference is the `user-data.toml`
```toml
[settings.bootstrap-containers.setup-network-file]
source = "public.ecr.aws/bottlerocket/bottlerocket-bootstrap:v0.2.8"
mode = "once"
user-data = "IyEvdXNyL2Jpbi9lbnYgYmFzaApzZXQgLWV1CgojIENoZWNrIHJlYm9vdCBtYXJrZXIgZmlyc3QgLSBza2lwIGlmIGFscmVhZHkgcHJvY2Vzc2VkCklOU1RBTkNFX1JFQk9PVEVEPS8uYm90dGxlcm9ja2V0L2Jvb3RzdHJhcC1jb250YWluZXJzL2N1cnJlbnQvcmVib290ZWQKaWYgWyAtZiAiJElOU1RBTkNFX1JFQk9PVEVEIiBdOyB0aGVuCiAgICBlY2hvICJDb25maWd1cmF0aW9uIGFscmVhZHkgcHJvY2Vzc2VkIgogICAgZXhpdCAwCmZpCgplY2hvICJTZXR0aW5nIHVwIG5ldHdvcmsgY29uZmlndXJhdGlvbiB2aWEgZmlsZTovLyBzY2hlbWUuLi4iCgojIENyZWF0ZSB0aGUgZXhhY3QgbmV0d29yayBjb25maWd1cmF0aW9uCmNhdCA+IC90bXAvbmV0d29yay1jb25maWcudG9tbCA8PCAnSU5ORVJFT0YnCnZlcnNpb24gPSAzCgpbZW5wMHMxNl0KZGhjcDQgPSB0cnVlCmRoY3A2ID0gZmFsc2UKcHJpbWFyeSA9IHRydWUKCltlbnAwczE3XQpkaGNwNCA9IHRydWUKZGhjcDYgPSBmYWxzZQpJTk5FUkVPRgoKZWNobyAiTmV0d29yayBjb25maWd1cmF0aW9uIGNyZWF0ZWQ6IgpjYXQgL3RtcC9uZXR3b3JrLWNvbmZpZy50b21sCgplY2hvICJBcHBseWluZyBuZXR3b3JrIGNvbmZpZ3VyYXRpb24gdXNpbmcgZmlsZTovLyBzY2hlbWUuLi4iCmFwaWNsaWVudCBuZXR3b3JrIGNvbmZpZ3VyZSBmaWxlOi8vL3RtcC9uZXR3b3JrLWNvbmZpZy50b21sCgplY2hvICJOZXR3b3JrIGNvbmZpZ3VyYXRpb24gYXBwbGllZCBzdWNjZXNzZnVsbHkiCgojIE1hcmsgYXMgcHJvY2Vzc2VkIHRvIHByZXZlbnQgcmVib290IGxvb3AKdG91Y2ggIiRJTlNUQU5DRV9SRUJPT1RFRCIKCmVjaG8gIlJlYm9vdGluZyB0byBhcHBseSBuZXR3b3JrIGNoYW5nZXMuLi4iCmFwaWNsaWVudCByZWJvb3QK"
```

Where the encoded user-data here is:
```bash
#!/usr/bin/env bash
set -eu

# Check reboot marker first - skip if already processed
INSTANCE_REBOOTED=/.bottlerocket/bootstrap-containers/current/rebooted
if [ -f "$INSTANCE_REBOOTED" ]; then
    echo "Configuration already processed"
    exit 0
fi

echo "Setting up network configuration via file:// scheme..."

# Create the exact network configuration
cat > /tmp/network-config.toml << 'INNEREOF'
version = 3

[enp0s16]
dhcp4 = true
dhcp6 = false
primary = true

[enp0s17]
dhcp4 = true
dhcp6 = false
INNEREOF

echo "Network configuration created:"
cat /tmp/network-config.toml

echo "Applying network configuration using file:// scheme..."
apiclient network configure file:///tmp/network-config.toml

echo "Network configuration applied successfully"

# Mark as processed to prevent reboot loop
touch "$INSTANCE_REBOOTED"

echo "Rebooting to apply network changes..."
apiclient reboot
```

With this approach a network-config.toml file was created within the container at `/tmp/network-config.toml` and we then used `apiclient network configure file:///tmp/network-config.toml` to configure the net.toml.

Note that for bootstrap-containers, a **marker file** is needed to prevent reboot loop. The reboot happens during the execution of the bootstrap script so the subsequent process that turns the `once` mode to `off` will never happen. 

`systemd-inhibit` [command](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/os/bootstrap-commands.service#L12) we used for bootstrap-commands is something we can look into which could be a way to avoid the usage of marker file. 

@KCSesh - Contribution: 

**1. Re-Rean the `bootstrap-commands` using `base64` scheme**

**2. Re-ran: Tested the change via `bootstrap-container` using `file://` scheme**
Note: Running this test multiple times required: `--force-extract`, to clean PRIVATE partition marker file

**3. Validated STDIN**

```
apiclient network configure <<EOF
version = 2
[eth0]
dhcp4 = true
primary = true
EOF
```

**4. Tested invalid input**
<details>

Invalid Version:
```
bash-5.2# journalctl -u bootstrap-commands
Nov 11 23:35:45 localhost systemd[1]: Starting Bootstrap Commands...
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1351]: 23:35:46 [INFO] Processing bootstrap command 'setup-network-invalid-version' ...
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1351]: 23:35:46 [WARN] Bootstrap command failed to execute 'apiclient' failed - stderr: Failed to configure network: Failed to POST network configuration to '/actions/network/configure': Status 400 when POSTing /actions/network/configure: Invalid network configuration: Failed to parse network config from stdin: Invalid network configuration: Unknown network config version: 99
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1351]: Bootstrap command 'setup-network-invalid-version' failed.
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal systemd-inhibit[1345]: /usr/bin/bootstrap-commands failed with exit status 1.
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal systemd[1]: bootstrap-commands.service: Main process exited, code=exited, status=1/FAILURE
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal systemd[1]: bootstrap-commands.service: Failed with result 'exit-code'.
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal systemd[1]: Failed to start Bootstrap Commands.

bash-5.2# journalctl -u apiserver
...
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal apiserver[1251]: 23:35:46 [INFO] Invoking netdog commit to validate and write network configuration
Nov 11 23:35:46 ip-10-0-2-15.us-west-2.compute.internal apiserver[1251]: 23:35:46 [ERROR] Netdog commit failed: Network configuration validation failed: Failed to parse network config from stdin: Invalid network configuration: Unknown network config version: 99

```

Unknown field: 

```
bash-5.2# journalctl -u bootstrap-commands
Nov 11 23:37:54 localhost systemd[1]: Starting Bootstrap Commands...
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]: 23:37:54 [INFO] Processing bootstrap command 'setup-network-unknown-field' ...
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]: 23:37:54 [WARN] Bootstrap command failed to execute 'apiclient' failed - stderr: Failed to configure network: Failed to POST network configuration to '/actions/network/configure': Status 400 when POSTing /actions/network/configure: Invalid network configuration: Failed to parse network config from stdin: Failed to parse network config: TOML parse error at line 2, column 7
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]:   |
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]: 2 | hello=hi
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]:   |       ^
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]: invalid string
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]: expected `"`, `'`
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal bootstrap-commands[1336]: Bootstrap command 'setup-network-unknown-field' failed.
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal systemd-inhibit[1330]: /usr/bin/bootstrap-commands failed with exit status 1.
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal systemd[1]: bootstrap-commands.service: Main process exited, code=exited, status=1/FAILURE
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal systemd[1]: bootstrap-commands.service: Failed with result 'exit-code'.
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal systemd[1]: Failed to start Bootstrap Commands.

bash-5.2# journalctl -u apiserver
...
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal apiserver[1238]: 23:37:54 [INFO] Invoking netdog commit to validate and write network configuration
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal apiserver[1238]: 23:37:54 [ERROR] Netdog commit failed: Network configuration validation failed: Failed to parse network config from stdin: Failed to parse network config: TOML parse error at line 2, column 7
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal apiserver[1238]:   |
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal apiserver[1238]: 2 | hello=hi
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal apiserver[1238]:   |       ^
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal apiserver[1238]: invalid string
Nov 11 23:37:54 ip-10-0-2-15.us-west-2.compute.internal apiserver[1238]: expected `"`, `'`
bash-5.2#

```

Invalid stdin:

```
# Starting point: 
bash-5.2# cat /.bottlerocket/net.toml
version = 3

[enp0s16]
dhcp4 = true
dhcp6 = false
primary = true

[enp0s17]
dhcp4 = true
dhcp6 = false

# Apply incorrect net.toml:
bash-5.2# apiclient network configure <<EOF
version = 3

[enp0s17]
dhcp5 = true
dhcp6 = false
EOF

[  191.895006] apiserver[1238]: 16:06:56 [ERROR] netdog commit failed: network configuration validation failed: Failed to parse network config from stdin: Failed to parse network config: data did not match any variant of untagged enum NetworkDeviceV1
Failed to configure network: Failed to POST network configuration to '/actions/network/configure': Status 400 when POSTing /actions/network/configure: Invalid network configuration: Failed to parse network config from stdin: Failed to parse network config: data did not match any variant of untagged enum NetworkDeviceV1

# No change to file: 
bash-5.2# cat /.bottlerocket/net.toml
version = 3

[enp0s16]
dhcp4 = true
dhcp6 = false
primary = true

[enp0s17]
dhcp4 = true
dhcp6 = false
bash-5.2#
```

</details>

**5. Verified new tmp approach, with admin-container and inotify on metal:**
```
[root@admin]# apiclient network configure <<EOF
> version = 3
> [enp0s16]
> dhcp4 = true
> dhcp6 = false
> primary = true
> EOF
IN_CREATE: /.bottlerocket/rootfs/.bottlerocket/.tmpZkeAI5
IN_OPEN: /.bottlerocket/rootfs/.bottlerocket/.tmpZkeAI5
IN_MODIFY: /.bottlerocket/rootfs/.bottlerocket/.tmpZkeAI5
IN_OPEN: /.bottlerocket/rootfs/.bottlerocket/.tmpZkeAI5
IN_MODIFY: /.bottlerocket/rootfs/.bottlerocket/.tmpZkeAI5
IN_CLOSE_WRITE: /.bottlerocket/rootfs/.bottlerocket/.tmpZkeAI5
IN_MOVED_FROM: /.bottlerocket/rootfs/.bottlerocket/.tmpZkeAI5
IN_MOVED_TO: /.bottlerocket/rootfs/.bottlerocket/net.toml
IN_CLOSE_WRITE: /.bottlerocket/rootfs/.bottlerocket/net.toml
```
**Terms of contribution:**

### DIFF
--- a/sources/api/apiclient/README.md
+++ b/sources/api/apiclient/README.md
@@ -5,7 +5,7 @@ Current version: 0.1.0
 ## apiclient binary
 
 The `apiclient` binary provides high-level methods to interact with the Bottlerocket API.
-There's a [set](#set-mode) subcommand for changing settings, an [update](#update-mode) subcommand for updating the host, and an [exec](#exec-mode) subcommand for running commands in host containers.
+There's a [set](#set-mode) subcommand for changing settings, a [network configure](#network-configure) subcommand for configuring network settings, an [update](#update-mode) subcommand for updating the host, and an [exec](#exec-mode) subcommand for running commands in host containers.
 There's also a low-level [raw](#raw-mode) subcommand for direct interaction with the HTTP API.
 
 It talks to the Bottlerocket socket by default.
@@ -73,6 +73,52 @@ You can use JSON form to set it:
 
 ```shell
 apiclient set --json '{"motd": "42"}'
+```
+
+### Network configure
+
+This allows you to configure network settings by providing a network configuration file. The configuration will be applied at the next boot.
+
+The `network configure` command accepts input from different sources using URI schemes:
+
+#### File URI input
+
+You can specify a local file path using the `file://` URI scheme:
+
+```shell
+apiclient network configure file:///path/to/net.toml
+```
+
+#### Base64 encoded input
+
+For inline configuration, you can provide base64-encoded content:
+
+```shell
+apiclient network configure base64:dmVyc2lvbiA9IDIKCltldGgwXQpkaGNwNCA9IHRydWU=
+```
+
+This is particularly useful for automation and configuration management where you want to embed the network configuration directly in scripts or user data.
+
+#### Configuration format
+
+The `net.toml` file uses the same format that netdog supports. Here's an example:
+
+```toml
+version = 2
+
+[eth0]
+dhcp4 = true
+dhcp6 = false
+
+[eth1]
+static4 = ["192.168.1.100/24"]
+route = [{to = "default", via = "192.168.1.1"}]
+```
+
+After configuring the network settings, you'll need to reboot the system for the changes to take effect:
+
+```shell
+apiclient reboot
 ```
 
 ### Update mode
@@ -347,7 +393,7 @@ The results from each item in the report will be one of:
 ## apiclient library
 
 The apiclient library provides high-level methods to interact with the Bottlerocket API.  See
-the documentation for submodules [`apply`], [`exec`], [`get`], [`reboot`], [`report`], [`set`],
+the documentation for submodules [`apply`], [`exec`], [`get`], [`network`], [`reboot`], [`report`], [`set`],
 and [`update`] for high-level helpers.
 
 For more control, and to handle APIs without high-level wrappers, there are also 'raw' methods
--- a/sources/api/apiclient/README.tpl
+++ b/sources/api/apiclient/README.tpl
@@ -5,7 +5,7 @@ Current version: {{version}}
 ## apiclient binary
 
 The `apiclient` binary provides high-level methods to interact with the Bottlerocket API.
-There's a [set](#set-mode) subcommand for changing settings, an [update](#update-mode) subcommand for updating the host, and an [exec](#exec-mode) subcommand for running commands in host containers.
+There's a [set](#set-mode) subcommand for changing settings, a [network configure](#network-configure) subcommand for configuring network settings, an [update](#update-mode) subcommand for updating the host, and an [exec](#exec-mode) subcommand for running commands in host containers.
 There's also a low-level [raw](#raw-mode) subcommand for direct interaction with the HTTP API.
 
 It talks to the Bottlerocket socket by default.
@@ -73,6 +73,52 @@ You can use JSON form to set it:
 
 ```shell
 apiclient set --json '{"motd": "42"}'
+```
+
+### Network configure
+
+This allows you to configure network settings by providing a network configuration file. The configuration will be applied at the next boot.
+
+The `network configure` command accepts input from different sources using URI schemes:
+
+#### File URI input
+
+You can specify a local file path using the `file://` URI scheme:
+
+```shell
+apiclient network configure file:///path/to/net.toml
+```
+
+#### Base64 encoded input
+
+For inline configuration, you can provide base64-encoded content:
+
+```shell
+apiclient network configure base64:dmVyc2lvbiA9IDIKCltldGgwXQpkaGNwNCA9IHRydWU=
+```
+
+This is particularly useful for automation and configuration management where you want to embed the network configuration directly in scripts or user data.
+
+#### Configuration format
+
+The `net.toml` file uses the same format that netdog supports. Here's an example:
+
+```toml
+version = 2
+
+[eth0]
+dhcp4 = true
+dhcp6 = false
+
+[eth1]
+static4 = ["192.168.1.100/24"]
+route = [{to = "default", via = "192.168.1.1"}]
+```
+
+After configuring the network settings, you'll need to reboot the system for the changes to take effect:
+
+```shell
+apiclient reboot
 ```
 
 ### Update mode
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -44,7 +44,7 @@ impl Default for Args {
 }
 
 /// Stores the usage mode specified by the user as a subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum Subcommand {
     Apply(ApplyArgs),
     Exec(ExecArgs),
@@ -60,13 +60,13 @@ enum Subcommand {
 }
 
 /// Stores user-supplied arguments for the 'apply' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct ApplyArgs {
     input_sources: Vec<String>,
 }
 
 /// Stores user-supplied arguments for the 'exec' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct ExecArgs {
     command: Vec<OsString>,
     target: String,
@@ -74,7 +74,7 @@ struct ExecArgs {
 }
 
 /// Stores user-supplied arguments for the 'get' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum GetArgs {
     Prefixes {
         include: Vec<String>,
@@ -85,7 +85,7 @@ enum GetArgs {
 }
 
 /// Stores user-supplied arguments for the 'raw' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct RawArgs {
     method: String,
     uri: String,
@@ -93,11 +93,11 @@ struct RawArgs {
 }
 
 /// Stores user-supplied arguments for the 'lockdown' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct LockdownArgs {}
 
 /// Stores user-supplied arguments for the 'reboot' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct RebootArgs {}
 
 /// Stores a vector of user-supplied key-value pairs for the 'set' subcommand.
@@ -107,14 +107,14 @@ pub struct SetKeyPairSettings {
 }
 
 /// Stores user-supplied arguments for the 'set' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum SetArgs {
     Simple(Vec<String>),
     Json(serde_json::Value),
 }
 
 /// Stores the 'update' subcommand specified by the user.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum UpdateSubcommand {
     Check(UpdateCheckArgs),
     Apply(UpdateApplyArgs),
@@ -122,7 +122,7 @@ enum UpdateSubcommand {
 }
 
 /// The available 'report' subcommands.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum ReportSubcommand {
     Cis(CisReportArgs),
     CisK8s(CisReportArgs),
@@ -130,35 +130,35 @@ enum ReportSubcommand {
 }
 
 /// Stores common user-supplied arguments for the cis report subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct CisReportArgs {
     level: Option<i32>,
     format: Option<String>,
 }
 
 /// Stores common user-supplied arguments for the fips report subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct FipsReportArgs {
     format: Option<String>,
 }
 
 /// Stores user-supplied arguments for the 'update check' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct UpdateCheckArgs {}
 
 /// Stores user-supplied arguments for the 'update apply' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct UpdateApplyArgs {
     check: bool,
     reboot: bool,
 }
 
 /// Stores user-supplied arguments for the 'update cancel' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct UpdateCancelArgs {}
 
 /// Stores the 'ephemeral-storage' subcommand specified by the user.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum EphemeralStorageSubcommand {
     Init(EphemeralStorageInitArgs),
     Bind(EphemeralStorageBindArgs),
@@ -168,7 +168,7 @@ enum EphemeralStorageSubcommand {
 }
 
 /// Stores user-supplied arguments for the 'ephemeral-storage init' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct EphemeralStorageInitArgs {
     disks: Option<Vec<String>>,
     ebs_volumes: Option<Vec<String>>,
@@ -177,24 +177,24 @@ struct EphemeralStorageInitArgs {
 }
 
 /// Stores user-supplied arguments for the 'ephemeral-storage bind' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct EphemeralStorageBindArgs {
     targets: Vec<String>,
 }
 /// Stores user-supplied arguments for the 'ephemeral-storage list-disks/list-ebs-volumes/list-dirs' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct EphemeralStorageFormatArgs {
     format: Option<String>,
 }
 
 /// Stores the 'network' subcommand specified by the user.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum NetworkSubcommand {
     Configure(NetworkConfigureArgs),
 }
 
 /// Stores user-supplied arguments for the 'network configure' subcommand.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct NetworkConfigureArgs {
     input_source: Option<String>,
 }
@@ -1604,12 +1604,10 @@ mod tests {
         assert_eq!(global_args.socket_path, "/run/api.sock");
 
         // Test network configure subcommand with no input source (stdin)
-        match subcommand {
-            Subcommand::Network(NetworkSubcommand::Configure(configure_args)) => {
-                assert_eq!(configure_args.input_source, None);
-            }
-            _ => panic!("Expected Network::Configure subcommand, got: {subcommand:?}"),
-        }
+        let expected = Subcommand::Network(NetworkSubcommand::Configure(NetworkConfigureArgs {
+            input_source: None,
+        }));
+        assert_eq!(subcommand, expected);
     }
 
     #[test_case("apiclient network configure file:///tmp/net.toml",
@@ -1644,15 +1642,10 @@ mod tests {
         assert_eq!(global_args.log_level, expected_args.log_level);
         assert_eq!(global_args.socket_path, expected_args.socket_path);
 
-        // Test network configure subcommand and extract input source
-        match subcommand {
-            Subcommand::Network(NetworkSubcommand::Configure(configure_args)) => {
-                assert_eq!(
-                    configure_args.input_source,
-                    Some(expected_input_source.to_string())
-                );
-            }
-            _ => panic!("Expected Network::Configure subcommand, got: {subcommand:?}"),
-        }
+        // Test network configure subcommand matches expected
+        let expected = Subcommand::Network(NetworkSubcommand::Configure(NetworkConfigureArgs {
+            input_source: Some(expected_input_source.to_string()),
+        }));
+        assert_eq!(subcommand, expected);
     }
 }
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -7,7 +7,8 @@
 // to the API, which is intended to be reusable by other crates.
 
 use apiclient::{
-    apply, ephemeral_storage, exec, get, lockdown, reboot, report, set, update, SettingsInput,
+    apply, ephemeral_storage, exec, get, lockdown, network, reboot, report, set, update,
+    SettingsInput,
 };
 use log::{info, log_enabled, trace, warn};
 use model::ephemeral_storage::{Filesystem, Preference};
@@ -49,6 +50,7 @@ enum Subcommand {
     Exec(ExecArgs),
     Get(GetArgs),
     Lockdown(LockdownArgs),
+    Network(NetworkSubcommand),
     Raw(RawArgs),
     Reboot(RebootArgs),
     Set(SetArgs),
@@ -185,6 +187,18 @@ struct EphemeralStorageFormatArgs {
     format: Option<String>,
 }
 
+/// Stores the 'network' subcommand specified by the user.
+#[derive(Debug)]
+enum NetworkSubcommand {
+    Configure(NetworkConfigureArgs),
+}
+
+/// Stores user-supplied arguments for the 'network configure' subcommand.
+#[derive(Debug)]
+struct NetworkConfigureArgs {
+    input_source: Option<String>,
+}
+
 /// Informs the user about proper usage of the program and exits.
 fn usage() -> ! {
     let msg = &format!(
@@ -203,6 +217,8 @@ fn usage() -> ! {
                                        or from stdin.
             get                        Retrieve and print settings.
             set                        Changes settings and applies them to the system.
+            network configure          Configures network settings from net.toml files at
+                                       given URIs.
             update check               Prints information about available updates.
             update apply               Applies available updates.
             update cancel              Deactivates an applied update.
@@ -247,6 +263,13 @@ fn usage() -> ! {
 
                                        If neither prefixes nor URI are specified, get will show
                                        settings and OS info.
+
+        network configure options:
+            [ URI ]                    URI to a network configuration file (TOML format)
+                                       to apply. Supports file:// and base64: URI schemes.
+                                       If no URI is specified, reads from stdin.
+                                       Configuration is written to /.bottlerocket/net.toml and
+                                       validated at next boot by netdog.
 
         set options:
             KEY=VALUE [KEY=VALUE ...]  The settings you want to set.  For example:
@@ -376,8 +399,8 @@ fn parse_args(args: impl Iterator<Item = String>) -> (Args, Subcommand) {
             }
 
             // Subcommands
-            "raw" | "apply" | "exec" | "get" | "lockdown" | "reboot" | "report" | "set"
-            | "update" | "ephemeral-storage"
+            "raw" | "apply" | "exec" | "get" | "lockdown" | "network" | "reboot" | "report"
+            | "set" | "update" | "ephemeral-storage"
                 if subcommand.is_none() && !arg.starts_with('-') =>
             {
                 subcommand = Some(arg)
@@ -395,6 +418,7 @@ fn parse_args(args: impl Iterator<Item = String>) -> (Args, Subcommand) {
         Some("exec") => (global_args, parse_exec_args(subcommand_args)),
         Some("get") => (global_args, parse_get_args(subcommand_args)),
         Some("lockdown") => (global_args, parse_lockdown_args(subcommand_args)),
+        Some("network") => (global_args, parse_network_args(subcommand_args)),
         Some("reboot") => (global_args, parse_reboot_args(subcommand_args)),
         Some("report") => (global_args, parse_report_args(subcommand_args)),
         Some("set") => (global_args, parse_set_args(subcommand_args)),
@@ -563,6 +587,42 @@ fn parse_lockdown_args(args: Vec<String>) -> Subcommand {
         usage_msg(format!("Unknown arguments: {}", args.join(", ")));
     }
     Subcommand::Lockdown(LockdownArgs {})
+}
+/// Parses the desired subcommand of 'network'.
+fn parse_network_args(args: Vec<String>) -> Subcommand {
+    let mut subcommand = None;
+    let mut subcommand_args = Vec::new();
+
+    for arg in args.into_iter() {
+        match arg.as_ref() {
+            // Subcommands
+            "configure" if subcommand.is_none() => subcommand = Some(arg),
+
+            // Other arguments are passed to the subcommand parser
+            _ => subcommand_args.push(arg),
+        }
+    }
+
+    match subcommand.as_deref() {
+        Some("configure") => parse_network_configure_args(subcommand_args),
+        _ => usage_msg("Missing or unknown subcommand for 'network'"),
+    }
+}
+
+/// Parses arguments for the 'network configure' subcommand.
+fn parse_network_configure_args(args: Vec<String>) -> Subcommand {
+    let mut input_source = None;
+
+    for arg in args.into_iter() {
+        if input_source.is_some() {
+            usage_msg("apiclient network configure takes only one input source URI.")
+        }
+        input_source = Some(arg);
+    }
+
+    Subcommand::Network(NetworkSubcommand::Configure(NetworkConfigureArgs {
+        input_source,
+    }))
 }
 
 /// Parses arguments for the 'reboot' subcommand.
@@ -1084,6 +1144,17 @@ async fn run() -> Result<()> {
                 .context(error::LockdownSnafu)?;
         }
 
+        Subcommand::Network(subcommand) => match subcommand {
+            NetworkSubcommand::Configure(configure_args) => {
+                let content = network::get_content(configure_args.input_source)
+                    .await
+                    .context(error::NetworkGetContentSnafu)?;
+                network::configure(&args.socket_path, content)
+                    .await
+                    .context(error::NetworkConfigureSnafu)?;
+            }
+        },
+
         Subcommand::Reboot(_reboot) => {
             reboot::reboot(&args.socket_path)
                 .await
@@ -1252,7 +1323,9 @@ async fn main() {
 }
 
 mod error {
-    use apiclient::{apply, ephemeral_storage, exec, get, lockdown, reboot, report, set, update};
+    use apiclient::{
+        apply, ephemeral_storage, exec, get, lockdown, network, reboot, report, set, update,
+    };
     use snafu::Snafu;
 
     #[derive(Debug, Snafu)]
@@ -1272,6 +1345,11 @@ mod error {
 
         #[snafu(display("Failed to lockdown: {}", source))]
         Lockdown { source: lockdown::Error },
+        #[snafu(display("Failed to get network configuration content: {}", source))]
+        NetworkGetContent { source: network::Error },
+
+        #[snafu(display("Failed to configure network: {}", source))]
+        NetworkConfigure { source: network::Error },
 
         #[snafu(display("Failed to reboot: {}", source))]
         Reboot { source: reboot::Error },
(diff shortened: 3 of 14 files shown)


sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,221 @@
...
+//! /.bottlerocket/net.toml for validation and application at next boot.
+
+use base64::{engine, Engine};
+use reqwest::Url;
+use snafu::{futures::TryFutureExt, OptionExt, ResultExt};
+use std::path::Path;
+use tokio::io::AsyncReadExt;
+
+/// Configures network settings from the given input source (URI or stdin), writing the content
+/// directly to /.bottlerocket/net.toml via the API server.
```
arnaldo2792: Since this is the client, it doesn't care (and know) where the configuration file is written to.

```suggestion
/// Configures network settings from the given input source (URI or stdin)
```

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,221 @@
...
+pub async fn configure<P>(socket_path: P, input_source: String) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    // Get the network configuration content from the input source
```
arnaldo2792: Comments are appreciated, but please review the comments and make sure they explain "why" you did what you did instead of the "what" and "how".

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,221 @@
...
+{
+    let input_source = input_source.into();
+
+    // Read from stdin if "-" was given.
+    if input_source == "-" {
+        let mut output = String::new();
+        tokio::io::stdin()
+            .read_to_string(&mut output)
+            .context(error::StdinReadSnafu)
+            .await?;
```
arnaldo2792: Is there any benefit of using futures here? There aren't any other tasks executed in the background and this call is blocking right away at this exact moment.

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,221 @@
...
+    })?;
+
+    if uri.scheme() == "file" {
+        // Turn the URI to a file path, and return the file content.
+        let path = uri.to_file_path().ok().context(error::FileUriSnafu {
+            input_source: &input_source,
+        })?;
+        tokio::fs::read_to_string(path)
+            .context(error::FileReadSnafu { input_source })
+            .await
```
arnaldo2792: Same as elsewhere, if this is blocking right away, why making it an `async` call?

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,221 @@
...
+            source: Box<crate::Error>,
+        },
+
+        #[snafu(display("Failed to read given file '{}': {}", input_source, source))]
+        FileRead {
+            input_source: String,
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Given invalid file URI '{}'", input_source))]
```
arnaldo2792:
```suggestion
#[snafu(display("Invalid file URI '{}'", input_source))]
```

sources/api/apiserver/src/server/mod.rs
```diff
@@ -651,6 +652,30 @@ async fn reboot() -> Result<HttpResponse> {
...
+
+    // Define the target directory and file path
+    let config_dir = Path::new("/.bottlerocket");
+    let config_file = config_dir.join("net.toml");
+
+    // Create the directory if it doesn't exist
+    if !config_dir.exists() {
+        std::fs::create_dir_all(config_dir)
+            .context(error::NetworkConfigDirSnafu { path: config_dir })?;
+    }
```
arnaldo2792: This program shouldn't be creating this directory, as it is expected to be managed by the system. Even if the program tries to create the directory, it will fail as the rootfs is read-only.

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let response = response.error_for_status().context(error::ReqwestSnafu {
+            uri: &input_source,
+            method: "GET",
+        })?;
+
+        response.text().await.context(error::ReqwestSnafu {
+            uri: input_source,
+            method: "GET",
+        })
```
arnaldo2792: You may want to have a different error for this, as it is passed the point of a failed `GET` request and it is at the point of parsing the text that was returned by the request.

sources/api/apiclient/src/main.rs
```diff
@@ -225,6 +241,13 @@ fn usage() -> ! {
                                        If neither prefixes nor URI are specified, get will show
                                        settings and OS info.
 
+        network configure options:
+            [ URI ]                    The URI to a network configuration file (TOML format) that you
+                                       want to apply. Supports file:// and base64: URI schemes.
```
arnaldo2792: According to the implementation, it also supports `http` and `https`.

ytsssun: Ahh, great catch! `http` and `https` should not be in the scope. I have added explicit deny for this case. Right now we should only support file and base64. 

We will expand the support for broader input sources later, which I will follow up with a GH issue.

sources/api/apiclient/src/network.rs
ytsssun: Updated, thanks!

sources/api/apiclient/src/network.rs
ytsssun: We should explicitly forbid this case as other schemes are not supported.

sources/api/apiserver/src/server/mod.rs
ytsssun: Yeah I somehow had an old commit here.. You are right the mount point `/.bottlerocket` is managed by the mount unit we added for PRIVATE partition. `apiserver` should never bother creating the directory.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -651,6 +652,30 @@ async fn reboot() -> Result<HttpResponse> {
     Ok(HttpResponse::NoContent().finish())
 }
 
+/// Configures network settings by writing content to /.bottlerocket/net.toml
+async fn configure_network(body: web::Bytes) -> Result<HttpResponse> {
```
cbgbt: One thing made clear by the client docs but not this source code is that we are writing the network configuration but that this API route expects for the caller to reboot for the changes to take effect.

sources/api/apiclient/src/network.rs
cbgbt: It somewhat makes sense because `apiclient` is a lib crate and could be used in another asynchronous application in a concurrent fashion.

But I think exposing reading from stdin in the pub `network::configure` function is a bit odd -- it would be a very strange way to call the function from a consuming application.

Probably the public API should allow configuration with a string and the "stdin or file" element of `apiclient` should be kept private and only invoked via `main()`.

To be more explicit with my feedback, I think `network::configure()` should remain `pub` but should probably not call `get_content()` itself, and instead assume that the caller will do so.

sources/api/apiclient/src/network.rs
ytsssun: That's fair, the code itself is pretty self-explanatory. Dropping this comment.

sources/api/apiclient/src/network.rs
ytsssun: > It somewhat makes sense because apiclient is a lib crate and could be used in another asynchronous application in a concurrent fashion.

I was about to make the change to use synchronous call, but this actually makes sense to me. 

The `stdin` was not expected, I somehow had an old commit here.. 

But the idea to keep `configure()` and `get_content()` separate makes sense to me so that the caller will make two calls. I will keep the functions async in that case. And just update `configure()` to not call `get_content` instead.

sources/api/apiserver/src/server/mod.rs
ytsssun: Fair ask, Will update the comment to mention that a reboot is expected to allow it to take effect.

sources/api/apiclient/src/network.rs
ytsssun: Resolving as per @cbgbt if using this function via a crate, it could be used in async fashion, in which we should not block the runtime.

ytsssun:
Force pushed to address comments from @cbgbt and @arnaldo2792

sources/api/apiclient/README.md
```diff
@@ -75,6 +75,53 @@ You can use JSON form to set it:
...
+
+```shell
+apiclient network configure base64:dmVyc2lvbiA9IDIKCltldGgwXQpkaGNwNCA9IHRydWU=
+```
+
+This is particularly useful for automation and configuration management where you want to embed the network configuration directly in scripts or user data.
+
+#### Configuration format
+
+The net.toml file uses the same format that netdog supports. Here's an example:
```
arnaldo2792: nit: for consistency `net.toml` and not net.toml

sources/api/apiclient/README.tpl
```diff
@@ -75,6 +75,53 @@ You can use JSON form to set it:
 apiclient set --json '{"motd": "42"}'
 ```
 
+### Network configure
+
+This allows you to configure network settings by providing a net.toml file which netdog uses for network configuration.
+The configuration is written to `/.bottlerocket/net.toml` and will be validated and applied at the next boot by netdog.
+
+The network configure command accepts input from different sources using URI schemes:
```
arnaldo2792:
```suggestion
The `network configure` command accepts input from different sources using URI schemes:
```

sources/api/apiclient/src/main.rs
```diff
@@ -1486,4 +1568,91 @@ mod tests {
...
+            (
+                Subcommand::Network(NetworkSubcommand::Configure(actual)),
+                Subcommand::Network(NetworkSubcommand::Configure(expected)),
+            ) => {
+                assert_eq!(actual.input_source, expected.input_source);
+            }
+            _ => panic!(
+                "Expected Network::Configure subcommand: {expected_subcommand:?}, got: {subcommand:?}"
+            ),
+        }
```
cbgbt: The macro above is pretty effortful in making it easy to test new subcommands, but then we have to maintain this match tree.

Can we just `#[derive(PartialEq)]` on `Subcommand`?

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,258 @@
...
+        .context(error::ConfigureRequestSnafu { uri, method })?;
+
+    Ok(())
+}
+
+/// Retrieves the network configuration content from the given source URI.
+///
+/// Supports file:// and base64: URI schemes. Use this function to get content
+/// before calling `configure()`.
+pub async fn get_content<S>(input_source: S) -> Result<String>
```
cbgbt: minor nit: This should probably just be `pub(crate)`

sources/api/apiclient/README.md
```diff
@@ -75,6 +75,53 @@ You can use JSON form to set it:
 apiclient set --json '{"motd": "42"}'
 ```
 
+### Network configure
+
+This allows you to configure network settings by providing a net.toml file which netdog uses for network configuration.
+The configuration is written to `/.bottlerocket/net.toml` and will be validated and applied at the next boot by netdog.
```
cbgbt: Similar to what Arnaldo mentioned, I might drop references to `net.toml` from the client configuration, and just mention that it will be applied at next boot. Protects us from stale documentation if the underlying implementation changes later.

cbgbt (approved):
Nice work! just a few suggestions.

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,258 @@
+//! This module provides functionalities to configure network settings.
+//! It accepts the same net.toml format supported by netdog, but allows sourcing it from
+//! file:// and base64: URIs.
```
arnaldo2792:
```suggestion
//! This module implements network configuration API calls.
//! It supports sourcing `net.toml` configuration files from the filesystem or base64 encoded strings.
```

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,258 @@
...
+
+use base64::{engine, Engine};
+use snafu::{OptionExt, ResultExt};
+use std::path::Path;
+use url::Url;
+
+/// Configures network settings by sending the provided content to the API server.
+///
+/// The configuration will be applied at the next boot - a reboot is required for changes to take effect.
```
arnaldo2792, line 12: nit: I wouldn't put this here as this is relevant to the user (which won't read this), and rather should be in the website docs or the main `apiclient` documentation.

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,258 @@
...
+        .await
+        .context(error::ConfigureRequestSnafu { uri, method })?;
+
+    Ok(())
+}
+
+/// Retrieves the network configuration content from the given source URI.
+///
+/// Supports file:// and base64: URI schemes. Use this function to get content
+/// before calling `configure()`.
```
arnaldo2792:
```suggestion
/// Supports file:// and base64: URI schemes
```

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,258 @@
...
+    async fn test_http_uri_rejected() {
+        // Given an HTTP URI
+        // When attempting to get content from the HTTP URI
+        let result = get_content("http://example.com/net.toml").await;
+
+        // Then the operation should fail with an unsupported URI scheme error
+        assert!(result.is_err());
+        let error_msg = result.unwrap_err().to_string();
+        assert!(error_msg.contains("Unsupported URI scheme 'http'"));
+        assert!(error_msg.contains("Only file:// and base64: schemes are supported"));
```
arnaldo2792: neat!

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_https_uri_rejected() {
```
arnaldo2792: Can you use your awesome `test_case` here and have the one test for the different use cases?

sources/api/apiclient/src/network.rs
cbgbt: I personally found it helpful, without this comment I spent time looking around after seeing the apiserver implementation to try to understand if we were missing the code to actually apply the config after saving it.

But I've also been on the unpopular side of this same tension many times, so I'm curious to see how others think about it.

sources/api/apiclient/src/network.rs
ytsssun: I cannot make it `pub(crate)` as the compiler sees the binary crate (`main.rs`) and lib crate (`lib.rs`) separate crates. Making this pub(crate) would make it invisible from `main.rs`. I think we should still keep this function within network.rs, in which case we have to keep get_content() as public function...

Alternative is to move get_content() to main.rs, but the functionality boundary is less clear get_content() was used to fetch the content. 

We have similar logic in `apply` where we use `get` as a private function. Another alternative is to revert to the approach where we call get_content() within configure() so that it remains private. 

My preference in order is:
1. Keep it `pub`.
2. Make it `pub(crate)` and move it to `main.rs`
3. Move back to the state where we make it private and call it within `configure()`

sources/api/apiclient/src/network.rs
ytsssun: I don't have strong opinion here and I could see the tradeoffs. But this change is made to address comments here - https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/714#discussion_r2505679409

So I would like to keep it as is now to avoid extra churns.

sources/api/apiclient/src/main.rs
ytsssun: Done. I had to add `#derive(PartialEq)` to all the relevant structs referred by `Subcommand`.

sources/api/apiclient/src/network.rs
ytsssun: Done.

sources/api/apiclient/README.md
ytsssun: Updated. I think it makes sense to not mention the exact path it will be written to. But mentioning `net.toml` and its format is probably needed. See the section below (#### Configuration format).

The user needs to know how to compose the `net.toml` as the source.

sources/api/apiclient/src/main.rs
ytsssun: LMK what you think, I think that might not be worth it?

I now feel I should just reduce the test scope down to `test_network_configure_parsing`? I don't feel so good about adding `#derive(PartialEq)` to all the relevant structs, especially this is just for unit test.

ytsssun:
Force pushed to address comments from @cbgbt and @arnaldo2792 :
1. Remove comments that is redundant where the code is self-explanatory.
2. Documents wording update.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -687,6 +688,31 @@ async fn reboot() -> Result<HttpResponse> {
...
+    // Ensure the directory exists. The target directory (/.bottlerocket) is managed by the system,
+    // we should throw error immediately if that does not exist.
+    ensure!(
+        config_dir.exists(),
+        error::NetworkConfigDirMissingSnafu { path: config_dir }
+    );
+
+    // Write the configuration content to the file
+    std::fs::write(&config_file, content)
+        .context(error::NetworkConfigWriteSnafu { path: &config_file })?;
```
bcressey: Given the high potential for bricking the system, I'd prefer we have a `netdog validate` command that ensures it's a valid, deserializable configuration.

sources/api/apiclient/src/main.rs
```diff
@@ -564,6 +587,46 @@ fn parse_lockdown_args(args: Vec<String>) -> Subcommand {
...
+    for arg in args.into_iter() {
+        if input_source.is_some() {
+            usage_msg("apiclient network configure takes only one input source URI.")
+        }
+        input_source = Some(arg);
+    }
+
+    let input_source = input_source.unwrap_or_else(|| {
+        usage_msg("apiclient network configure requires an input source URI (file:// or base64:)")
+    });
```
bcressey: It'd be nice to default to stdin here, as `apiclient apply` does, so we can run:
```
apiclient network configure <<EOF
version = 2
[eth0]
dhcp4 = true
primary = true
EOF
```

sources/api/apiclient/src/network.rs
cbgbt: I think keeping it `pub` is fine.

sources/api/apiclient/src/main.rs
cbgbt: I think adding `#[derive(PartialEq)]` for relevant structs is a good idea, especially for tests.

It's pretty common to derive `Debug`, `Clone`, `Copy` (if applicable), `PartialEq`, and `Eq` if you can.

sources/netdog/src/cli/commit.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::io::Read;
+use std::path::Path;
+
+/// Validate and commit network configuration from stdin
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "commit")]
+pub struct CommitArgs {}
+
+pub fn run(_args: CommitArgs) -> Result<()> {
+    // Read content from stdin
```
cbgbt: Mind purging these "what" comments?

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,262 @@
...
+pub async fn get_content<S>(input_source: Option<S>) -> Result<String>
+where
+    S: Into<String>,
+{
+    get_content_with_reader(input_source, tokio::io::stdin()).await
+}
+
+/// Internal function that accepts a custom reader
+async fn get_content_with_reader<S, R>(input_source: Option<S>, mut reader: R) -> Result<String>
+where
```
cbgbt: Mind updating this docstring to be more descriptive? Ideally should declare the function's purpose in isolation.

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+/// Retrieves the network configuration content from the given source URI.
+///
+/// Supports file:// and base64: URI schemes. If no input source is provided, reads from stdin.
+pub async fn get_content<S>(input_source: Option<S>) -> Result<String>
+where
+    S: Into<String>,
+{
+    get_content_with_reader(input_source, tokio::io::stdin()).await
```
cbgbt: It feels somewhat like `get_content` should be deciding whether or not to use the `input_source` or `stdin` rather than passing both to the delegate function and allowing it to decide.

KCSesh:
^ This is a decent diff URL:
https://github.com/bottlerocket-os/bottlerocket-core-kit/compare/d5a84546284dd157011259800a3debbeac32ecde..4facba9d46b4f48afc45a1d75471fbda13139cf4

Above I added 2 minor changes:
* Add `netdog commit` which gets invoked by `apiserver` after `apiclient` does initial parse logic. Netdog commit now does the write to `/.bottlerocket/net.toml` 
* Add `stdin` support to `apiclient network configure`

I retested both valid and invalid net.toml files on `metal-dev`  (will update the description).

KCSesh:
^ The above push is just a rebase to get back on/near the tip of `develop`

KCSesh:
^ Fix comment typo, renamed 1 error, and created a const.

sources/netdog/src/cli/commit.rs
```diff
@@ -0,0 +1,66 @@
...
+    // Write to file after successful validation
+    let config_dir = Path::new("/.bottlerocket");
+    let config_file = config_dir.join("net.toml");
+
+    if !config_dir.exists() {
+        return error::NetConfigDirMissingSnafu { path: config_dir }.fail();
+    }
+
+    std::fs::write(&config_file, content)
+        .context(error::NetConfigWriteSnafu { path: &config_file })?;
```
bcressey: It'd be good to follow the pattern of writing to a temp file in the same directory, then rename it to the final file. Helps with power safety. See `persist_sysctl_config()` in `corndog` for an example.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -687,6 +688,44 @@ async fn reboot() -> Result<HttpResponse> {
...
+
+    // Validate and write net.toml using netdog commit
+    let validation_result = std::process::Command::new(NETDOG_BIN)
+        .arg("commit")
+        .stdin(std::process::Stdio::piped())
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .spawn()
+        .and_then(|mut child| {
+            use std::io::Write;
```
bcressey:
```suggestion
```

sources/api/apiclient/src/network.rs
```diff
@@ -0,0 +1,282 @@
...
+}
```
bcressey, line 37: It might be more readable to have this fork into two functions, like `get_content_with_stdin` or `get_content_from_source`.

Then `get_content_from_source` could further split into `get_content_from_file`, `get_content_from_base64`.

I found it hard to quickly answer questions like "can I read base64-encoded data from a file or URI, or only from the command line?" - because of all the branching in `get_content_with_reader`.

(It'd also be great to support the base64: syntax for `apiclient apply`, for overall UI consistency.)

bcressey (approved):

sources/api/apiserver/src/server/mod.rs
KCSesh: Thanks!

sources/api/apiclient/src/network.rs
KCSesh: Sounds good - made the change for now, and I'll add a tracker in apiclient apply to support base64

sources/netdog/src/cli/commit.rs
KCSesh: Nice!

cbgbt (commented):
Apologies, I reviewed this yesterday but never hit submit.

Do you think you could fix the PartialEq comment left earlier?

sources/netdog/src/cli/commit.rs
KCSesh: Removed the what commets from commit.rs

sources/api/apiclient/src/network.rs
KCSesh: Updated, and it was also refactored a bit on the previous push!

sources/api/apiclient/src/network.rs
KCSesh: Previous push, should handle this

KCSesh:
@cbgbt 
>Do you think you could fix the PartialEq comment left earlier?

Sorry missed this - fixed in the latest push

Along with the other nits (update docstring and remove what comments from netdog commit)

cbgbt (approved):

sources/api/apiclient/src/main.rs
KCSesh: Fixed in the latest push!

bcressey (approved):

PR #716, opened 2025-11-06
add apiclient support for canonical settings

bcressey:
**Issue number:**

**Description of changes:**
Bottlerocket settings can be significant for remote attestation purposes. For example, it may be desirable to assert that host containers are disabled or that kernel lockdown is enforced. Hence, we'd like to measure the current settings into one of the TPM2 PCRs.

In order for such a measurement to be useful, there are two requirements:
1. Settings must be in some canonical form.
2. Settings that are random or unique make the measurement unpredictable, and must be excluded.

This change adds `apiclient` support for canonicalizing settings using the same `olpc-cjson` crate we use to canonicalize TUF repo metadata. It also supports excluding prefixes so that problematic values, like `settings.updates.random-seed`, can be stripped from the output.

**Testing done:**
Tested both features:
```
bash-5.2# apiclient get settings --exclude settings.updates.seed --exclude settings.network.hostname --exclude settings.host-containers.admin.user-data --exclude settings.aws.config --canonicalize
{"settings":{"aws":{"profile":"default","region":"us-west-2"},"boot":{"kernel":{"SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE":["1"],"SYSTEMD_DEFAULT_MOUNT_RATE_LIMIT_BURST":["25"],"module_blacklist":["i8042"]}},"cloudformation":{"logical-resource-id":"","should-signal":false,"stack-name":""},"host-containers":{"admin":{"enabled":false,"source":"328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.12.2","superpowered":true},"control":{"enabled":true,"source":"328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-control:v0.8.7","superpowered":false}},"kernel":{"lockdown":"none"},"metrics":{"metrics-url":"https://metrics.bottlerocket.aws/v1/metrics","send-metrics":false,"service-checks":["apiserver","chronyd","containerd","host-containerd","docker"]},"motd":"Welcome to Bottlerocket!","ntp":{"options":["iburst"],"time-servers":["169.254.169.123","2.amazon.pool.ntp.org"]},"oci-hooks":{"log4j-hotpatch-enabled":false},"updates":{"ignore-waves":false,"metadata-base-url":"https://updates.bottlerocket.aws/2020-07-07/aws-dev/aarch64/","targets-base-url":"https://updates.bottlerocket.aws/targets/","version-lock":"latest"}}}
```

**Terms of contribution:**


sources/api/apiclient/src/get.rs
```diff
@@ -24,13 +29,79 @@ where
...
+/// Returns the number of keys remaining in the object after removal.
+fn remove_prefixes_recursive(
+    value: &mut serde_json::Value,
+    parent_path: &str,
+    prefixes: &[String],
+) -> usize {
+    let serde_json::Value::Object(map) = value else {
+        // Non-object values count as 1 (they exist)
+        return 1;
+    };
```
cbgbt, line 67: Discussed with @bcressey, just documenting here for backreference:

Q: What about null values? Should we remove them? Should they contribute to the count?
A: The server doesn't permit the storage of null values, so we shouldn't receive any. To specially treat them here implies we should always specially treat them on the client side, and there doesn't seem to be a better way to handle them generally than passing them through.

cbgbt (approved):
lgtm!

sources/api/apiclient/src/get.rs
```diff
@@ -24,13 +29,79 @@ where
...
+    Ok(merged)
+}
+
+/// Removes keys matching any of the given prefixes from the JSON value.
+/// Uses dotted key notation (e.g., "settings.network" matches "settings.network.hostname").
+/// Prefixes without trailing dots match any key starting with that prefix.
+/// Prefixes with trailing dots only match keys with that exact prefix followed by more path segments.
+fn remove_prefixes(value: &mut serde_json::Value, prefixes: &[String]) -> Result<()> {
+    remove_prefixes_recursive(value, "", prefixes);
+    Ok(())
```
ytsssun: Nit: looks like there is no Err case. Should we just not return anything instead?

sources/api/apiclient/src/get.rs
```diff
@@ -24,13 +29,79 @@ where
...
+
+    // Recursively process remaining nested objects and remove if empty
+    for (key, full_path) in keys_to_recurse {
+        if let Some(nested_value) = map.get_mut(&key) {
+            let remaining = remove_prefixes_recursive(nested_value, &full_path, prefixes);
+            if remaining == 0 {
+                map.remove(&key);
+            }
+        }
+    }
```
ytsssun, line 102: nit: we could make this a one pass that is included in the for-loop above. 

I am also fine with this second pass for better readability. This may be something we need to worry about if the number of settings grows more significantly.

sources/api/apiclient/src/get.rs
```diff
@@ -24,13 +29,79 @@ where
...
+        map.remove(&key);
+    }
+
+    // Recursively process remaining nested objects and remove if empty
+    for (key, full_path) in keys_to_recurse {
+        if let Some(nested_value) = map.get_mut(&key) {
+            let remaining = remove_prefixes_recursive(nested_value, &full_path, prefixes);
+            if remaining == 0 {
+                map.remove(&key);
+            }
```
ytsssun: Looks like the only effective value here is whether the map is empty or not. Do we want to use boolean as return value instead? 

So that we can do something like
```rust
  let should_remove = remove_prefixes_recursive(nested, path, prefixes);
  if should_remove {
      map.remove(&key);
  }
```

and return map.is_empty() at the end.

For the leaf-node check at the beginning (L64-L67) we can do
```
// Boolean - Natural
  let serde_json::Value::Object(map) = value else {
      return false;  // "Don't remove non-objects"
  };
```

This is also non-blocking as I think the readability is already pretty good.

ytsssun (approved):

ytsssun:
Left a few non-blocking nits, but the change LGTM.

sources/api/apiclient/src/get.rs
bcressey: Doing this in a later loop helps avoid needing a mutable borrow of `map` while iterating through an immutable borrow. I'd like to keep it for clarity for now.

I agree that we'll be in trouble if we end up with a large number of settings, or lots of excluded prefixes to match against.

ytsssun (approved):
Will approve once the unit tests are fixed.

sources/api/apiclient/src/get.rs
```diff
@@ -24,13 +29,78 @@ where
...
+
+/// Recursively removes keys from JSON, tracking the dotted path as we traverse.
+/// Returns the number of keys remaining in the object after removal.
+fn remove_prefixes_recursive(
+    value: &mut serde_json::Value,
+    parent_path: &str,
+    prefixes: &[String],
+) -> bool {
+    let serde_json::Value::Object(map) = value else {
+        // Non-object values count as 1 (they exist)
```
ytsssun: This comment needs to be updated now that the return value is `false`.

sources/api/apiclient/src/get.rs
```diff
@@ -24,13 +29,78 @@ where
...
+/// Removes keys matching any of the given prefixes from the JSON value.
+/// Uses dotted key notation (e.g., "settings.network" matches "settings.network.hostname").
+/// Prefixes without trailing dots match any key starting with that prefix.
+/// Prefixes with trailing dots only match keys with that exact prefix followed by more path segments.
+fn remove_prefixes(value: &mut serde_json::Value, prefixes: &[String]) {
+    remove_prefixes_recursive(value, "", prefixes);
+}
+
+/// Recursively removes keys from JSON, tracking the dotted path as we traverse.
+/// Returns the number of keys remaining in the object after removal.
```
ytsssun: Sorry I missed this one too.. wording should be updated to honor the boolean return type.

Change otherwise LGTM

ytsssun (approved):

arnaldo2792 (approved):

PR #717, opened 2025-11-06
add rottweiler to manage storage encryption

bcressey:
**Issue number:**

**Description of changes:**
With Secure Boot enabled and dm-verity enforced, we usually consider the kernel and root filesystem to be trusted. Encrypting BOTTLEROCKET-DATA (`/local`) and BOTTLEROCKET-PRIVATE (`/.bottlerocket`) helps expand that trust boundary to include mutable storage as well, which is useful for remote attestation.

We can encrypt the entire data partition as we usually expect it to be empty and create the filesystem on first boot. However, the private partition cannot be fully encrypted; GRUB needs to read the bootconfig file, and there's potentially also a `net.toml` for network configuration. That means two types of encryption to support - LUKS for full disk encryption, fscrypt for directories.

`systemd` has excellent support for encrypted block devices, but not for directories except when `systemd-homed` is in use. [Managing encrypted filesystems with dirlock](https://lwn.net/Articles/1038859/) provides an overview of some tools in the space, with `dirlock` as an interesting project. However, most of the tools are opinionated (and flexible!) in ways that don't mesh perfectly with our needs. In particular, any sort of automatic fallback to passwords is problematic; if the storage cannot be automatically unlocked, then it can't be trusted and the system should fail closed.

`rottweiler` exists to provide a uniform wrapper around the underlying tools: `systemd-creds`, `cryptsetup`, and the kernel's fscrypt API. It requires a TPM2 device, and uses `systemd-creds` to generate wrapped keys that are sealed to specific PCRs. 

**Testing done:**
Enabled the [encrypted storage](https://github.com/bottlerocket-os/twoliter/pull/589) feature. Booted (and rebooted) images with encrypted block devices and directories.

That exercises these code paths:
```
# /local
rottweiler check block-device ${BOTTLEROCKET_DATA} unencrypted
rottweiler generate-key bottlerocket-data
rottweiler encrypt block-device ${BOTTLEROCKET_DATA} bottlerocket-data
rottweiler attach block-device ${BOTTLEROCKET_DATA} bottlerocket-data
rottweiler detach block-device ${BOTTLEROCKET_DATA}

# /.bottlerocket/datastore
rottweiler check directory ${DATASTORE_DIR} unencrypted
rottweiler generate-key datastore
rottweiler encrypt directory ${DATASTORE_DIR} datastore
```

I've also tested the lock and unlock paths:
```
$ rottweiler lock directory /.bottlerocket/datastore
<succeeds>

# file names are encrypted
$ ls -1 /.bottlerocket/datastore/
8Dpj4eE5GltMa8f00gibrGAoPQGRwM6dpl1KVTJwAHDUs_XVrilBeA
DiPPFeTtYxFN6kBKUkbnRijqId6IlecykRDKSuaFXoT7VOvnX2nswA
UJS1XpONYwoewy4cD9S68nrKCrozqkr33B_G8RnF-Ma6c-GeAbRKKA
eouecYx9ArlXpKsgZrYaDy7bxWrM_AlcXDVmsP0JEGgjjBr-pfgXJw
jM5WLUKW_72Ir3AAqm_Fx9dlekNo4NdO8PVlKhPbukAuu-suNmbaSw
nqDxnwOaczPa3kkYp_14aawGEpszfOYNuidNNRMBNKv25AiuvEcVAA

# files are inaccessible
$ cat /.bottlerocket/datastore/DiPPFeTtYxFN6kBKUkbnRijqId6IlecykRDKSuaFXoT7VOvnX2nswA
cat: /.bottlerocket/datastore/DiPPFeTtYxFN6kBKUkbnRijqId6IlecykRDKSuaFXoT7VOvnX2nswA: Required key not available

# directory can't be unlocked if the PCRs aren't in the expected state
$ rottweiler unlock directory /.bottlerocket/datastore datastore
Error: /usr/bin/systemd-creds failed with exit code 1: Failed to unseal secret using TPM2: Operation not permitted
```

**Terms of contribution:**

### DIFF
--- a/packages/rottweiler/rottweiler.spec
+++ b/packages/rottweiler/rottweiler.spec
@@ -1,0 +1,34 @@
+%global _cross_first_party 1
+%undefine _debugsource_packages
+
+Name: %{_cross_os}rottweiler
+Version: 0.1.0
+Release: 1%{?dist}
+Summary: Bottlerocket storage encryption helper
+License: Apache-2.0 OR MIT
+URL: https://github.com/bottlerocket-os/bottlerocket
+
+BuildRequires: %{_cross_os}glibc-devel
+Requires: %{_cross_os}cryptsetup
+Requires: %{_cross_os}systemd-cryptsetup
+Requires: %{_cross_os}tpm2-tools
+
+%description
+%{summary}.
+
+%prep
+%setup -T -c
+%cargo_prep
+
+%build
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p rottweiler
+
+%install
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 %{__cargo_outdir}/rottweiler %{buildroot}%{_cross_bindir}
+ln -s rottweiler %{buildroot}%{_cross_bindir}/rw
+
+%files
+%{_cross_bindir}/rottweiler
+%{_cross_bindir}/rw
--- a/sources/bottlerocket-image-features/README.md
+++ b/sources/bottlerocket-image-features/README.md
@@ -1,0 +1,49 @@
+# bottlerocket-image-features
+
+Current version: 0.1.0
+
+## Introduction
+
+*bottlerocket-image-features* is a library for parsing Bottlerocket image feature flags
+from the system configuration file.
+
+### Overview
+
+This crate provides functionality to read and parse the `/usr/share/bottlerocket/image-features.env`
+file, which contains feature flags that control various aspects of the Bottlerocket system.
+
+### Features
+
+Currently supported feature flags:
+
+- `IN_PLACE_UPDATES` - Controls whether in-place updates are enabled (default: true)
+- `ENCRYPTED_STORAGE` - Controls whether encrypted storage is enabled (default: false)
+
+### Usage
+
+```rust
+use bottlerocket_image_features::parse_image_features;
+
+let features = parse_image_features()?;
+if features.in_place_updates {
+    println!("In-place updates are enabled");
+}
+```
+
+### File Format
+
+The image features file uses a simple key=value format with support for:
+- Comments (lines starting with `#`)
+- Empty lines (ignored)
+- Quoted or unquoted values
+- Boolean values ("true" or "false")
+
+Example:
+```
+# Image feature configuration
+IN_PLACE_UPDATES="true"
+```
+
+## Colophon
+
+This text was generated using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/lib.rs`.
--- a/sources/bottlerocket-image-features/src/lib.rs
+++ b/sources/bottlerocket-image-features/src/lib.rs
@@ -1,0 +1,152 @@
+/*!
+# Introduction
+
+*bottlerocket-image-features* is a library for parsing Bottlerocket image feature flags
+from the system configuration file.
+
+## Overview
+
+This crate provides functionality to read and parse the `/usr/share/bottlerocket/image-features.env`
+file, which contains feature flags that control various aspects of the Bottlerocket system.
+
+## Features
+
+Currently supported feature flags:
+
+- `IN_PLACE_UPDATES` - Controls whether in-place updates are enabled (default: true)
+- `ENCRYPTED_STORAGE` - Controls whether encrypted storage is enabled (default: false)
+
+## Usage
+
+```rust,ignore
+use bottlerocket_image_features::parse_image_features;
+
+let features = parse_image_features()?;
+if features.in_place_updates {
+    println!("In-place updates are enabled");
+}
+```
+
+## File Format
+
+The image features file uses a simple key=value format with support for:
+- Comments (lines starting with `#`)
+- Empty lines (ignored)
+- Quoted or unquoted values
+- Boolean values ("true" or "false")
+
+Example:
+```text
+# Image feature configuration
+IN_PLACE_UPDATES="true"
+```
+*/
+
+use serde::Deserialize;
+use snafu::prelude::*;
+use std::fs;
+
+type Result<T> = std::result::Result<T, snafu::Whatever>;
+
+const IMAGE_FEATURES_FILE: &str = "/usr/share/bottlerocket/image-features.env";
+
+#[derive(Deserialize)]
+pub struct ImageFeatures {
+    #[serde(default = "default_true")]
+    pub in_place_updates: bool,
+    #[serde(default)]
+    pub encrypted_storage: bool,
+}
+
+fn default_true() -> bool {
+    true
+}
+
+/// Parse image features from the env file
+pub fn parse_image_features() -> Result<ImageFeatures> {
+    if !std::path::Path::new(IMAGE_FEATURES_FILE).exists() {
+        return Ok(ImageFeatures {
+            in_place_updates: true,
+            encrypted_storage: false,
+        });
+    }
+
+    let content = fs::read_to_string(IMAGE_FEATURES_FILE)
+        .with_whatever_context(|_| format!("failed to read {}", IMAGE_FEATURES_FILE))?;
+
+    parse_image_features_from_str(&content)
+}
+
+/// Parse image features from a string (useful for testing)
+pub fn parse_image_features_from_str(content: &str) -> Result<ImageFeatures> {
+    let pairs: Vec<(String, String)> = content
+        .lines()
+        .filter_map(|line| {
+            if line.starts_with('#') || line.trim().is_empty() {
+                return None;
+            }
+            let mut parts = line.splitn(2, '=');
+            let key = parts.next()?;
+            let mut value = parts.next()?;
+            if value.starts_with('"') {
+                value = &value[1..];
+            }
+            if value.ends_with('"') {
+                value = &value[..value.len() - 1];
+            }
+            Some((key.to_owned(), value.to_owned()))
+        })
+        .collect();
+
+    envy::from_iter(pairs).with_whatever_context(|_| "failed to parse image features")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_env_line_with_quotes() {
+        let content = r#"IN_PLACE_UPDATES="false""#;
+        let features = parse_image_features_from_str(content).unwrap();
+        assert_eq!(features.in_place_updates, false);
+    }
+
+    #[test]
+    fn test_parse_env_line_without_quotes() {
+        let content = "IN_PLACE_UPDATES=false";
+        let features = parse_image_features_from_str(content).unwrap();
+        assert_eq!(features.in_place_updates, false);
+    }
+
+    #[test]
+    fn test_parse_env_ignores_comments() {
+        let content = r#"# This is a comment
+IN_PLACE_UPDATES="true"
+# Another comment"#;
+        let features = parse_image_features_from_str(content).unwrap();
+        assert_eq!(features.in_place_updates, true);
+    }
+
+    #[test]
+    fn test_parse_env_ignores_empty_lines() {
+        let content = r#"
+IN_PLACE_UPDATES="false"
+
+"#;
+        let features = parse_image_features_from_str(content).unwrap();
+        assert_eq!(features.in_place_updates, false);
+    }
+
+    #[test]
+    fn test_default_true() {
+        assert_eq!(default_true(), true);
+    }
+
+    #[test]
+    fn test_default_when_missing() {
+        let content = "";
+        let features = parse_image_features_from_str(content).unwrap();
+        assert_eq!(features.in_place_updates, true);
+    }
+}
--- a/sources/rottweiler/src/block_device.rs
+++ b/sources/rottweiler/src/block_device.rs
@@ -1,0 +1,77 @@
+use snafu::prelude::*;
+use std::io::Read;
+use std::path::{Path, PathBuf};
+
+use crate::{key, system};
+
+type Result<T> = std::result::Result<T, snafu::Whatever>;
+
+/// Encrypt a block device with LUKS2 using the specified key
+pub fn encrypt(path: PathBuf, key_id: String) -> Result<()> {
+    let device = path
+        .to_str()
+        .with_whatever_context(|| format!("path is not valid UTF-8: '{}'", path.display()))?;
+
+    let key_bytes = key::load(key_id)?;
+
+    system::cryptsetup_luks_format(device, &key_bytes)
+}
+
+/// Attach (unlock) an encrypted block device, creating a device mapper entry
+pub fn attach(path: PathBuf, key_id: String) -> Result<()> {
+    let volume_name = filename(&path)?;
+
+    let source_device = path
+        .to_str()
+        .with_whatever_context(|| format!("path is not valid UTF-8: '{}'", path.display()))?;
+
+    let key_bytes = key::load(key_id)?;
+
+    system::systemd_cryptsetup_attach(volume_name, source_device, &key_bytes)
+}
+
+/// Detach (lock) an encrypted block device, removing the device mapper entry
+pub fn detach(path: PathBuf) -> Result<()> {
+    let volume_name = filename(&path)?;
+
+    system::systemd_cryptsetup_detach(volume_name)
+}
+
+/// Resize a LUKS2 encrypted block device to match the underlying device size
+pub fn resize(path: PathBuf, key_id: String) -> Result<()> {
+    let volume_name = filename(&path)?;
+
+    let key_bytes = key::load(key_id)?;
+
+    system::cryptsetup_resize(volume_name, &key_bytes)
+}
+
+const LUKS2_MAGIC: &[u8; 6] = b"LUKS\xba\xbe";
+const LUKS2_VERSION: u16 = 2;
+
+/// Check if a block device is LUKS2 encrypted by reading its header
+pub fn is_encrypted(path: PathBuf) -> Result<bool> {
+    let mut file = std::fs::File::open(&path)
+        .with_whatever_context(|_| format!("failed to open '{}'", path.display()))?;
+
+    let mut header = [0u8; 8];
+    file.read_exact(&mut header)
+        .with_whatever_context(|_| format!("failed to read header from '{}'", path.display()))?;
+
+    if &header[..6] == LUKS2_MAGIC {
+        let version = u16::from_be_bytes([header[6], header[7]]);
+        if version == LUKS2_VERSION {
+            return Ok(true);
+        }
+    }
+
+    Ok(false)
+}
+
+/// Extract filename from path as UTF-8 string
+fn filename(path: &Path) -> Result<&str> {
+    path.file_name()
+        .with_whatever_context(|| format!("failed to extract filename from '{}'", path.display()))?
+        .to_str()
+        .with_whatever_context(|| format!("filename is not valid UTF-8: '{}'", path.display()))
+}
--- a/sources/rottweiler/src/fscrypt.rs
+++ b/sources/rottweiler/src/fscrypt.rs
@@ -1,0 +1,295 @@
+use std::os::fd::AsRawFd;
+use std::path::Path;
+
+use hkdf::Hkdf;
+use nix::{ioctl_read, ioctl_readwrite};
+use sha2::Sha512;
+use snafu::prelude::*;
+use zeroize::ZeroizeOnDrop;
+
+use crate::mount_point::MountPoint;
+
+type Result<T> = std::result::Result<T, snafu::Whatever>;
+
+/// Public key for fscrypt operations (contains only the key identifier)
+pub struct FscryptPublicKey {
+    identifier: [u8; 16],
+}
+
+impl FscryptPublicKey {
+    /// Read the encryption policy from a directory and extract the key identifier
+    pub fn from_directory(path: &Path) -> Result<Self> {
+        let policy = FscryptGetPolicyExArg::from_path(path)?;
+        Ok(Self {
+            identifier: policy.key_identifier(),
+        })
+    }
+
+    /// Remove the encryption key from the kernel keyring, locking the directory
+    pub fn lock_directory(&self, path: &Path) -> Result<()> {
+        let mount_point = MountPoint::from_path(path).with_whatever_context(|_| {
+            format!("Failed to find mount point for '{}'", path.display())
+        })?;
+        let mount_fd = mount_point.open()?;
+        let mut remove_key = FscryptRemoveKey::new(self.identifier);
+        unsafe { remove_encryption_key_all_users(mount_fd.as_raw_fd(), &mut remove_key) }
+            .with_whatever_context(|_| {
+                format!("Failed to remove encryption key for '{}'", path.display())
+            })?;
+        Ok(())
+    }
+
+    /// Set the encryption policy on a directory
+    pub fn encrypt_directory(&self, path: &Path) -> Result<()> {
+        let mut policy = FscryptPolicyV2::new(self.identifier);
+        let dir_fd = std::fs::File::open(path)
+            .with_whatever_context(|_| format!("Failed to open directory '{}'", path.display()))?;
+        let argptr = &mut policy as *mut FscryptPolicyV2 as *mut FscryptPolicyV1Ioctl;
+        unsafe { set_encryption_policy(dir_fd.as_raw_fd(), argptr) }.with_whatever_context(
+            |_| format!("Failed to set encryption policy for '{}'", path.display()),
+        )?;
+        Ok(())
+    }
+}
+
+/// Private key for fscrypt operations (contains the key identifier and raw key material)
+#[derive(ZeroizeOnDrop)]
+pub struct FscryptPrivateKey {
+    identifier: [u8; 16],
+    raw: Vec<u8>,
+}
+
+impl FscryptPrivateKey {
+    /// Create a private key from raw key bytes, deriving the key identifier
+    pub fn from_bytes(raw: &[u8]) -> Result<Self> {
+        let identifier = calculate_key_identifier(raw)?;
+        Ok(Self {
+            identifier,
+            raw: raw.to_vec(),
+        })
+    }
+
+    /// Add the encryption key to the kernel keyring, unlocking the directory
+    pub fn unlock_directory(&self, path: &Path) -> Result<()> {
+        let mount_point = MountPoint::from_path(path).with_whatever_context(|_| {
+            format!("Failed to find mount point for '{}'", path.display())
+        })?;
+        let mount_fd = mount_point.open()?;
+        let mut add_key = FscryptAddKey::new(&self.raw)?;
+        let argptr = &mut add_key as *mut FscryptAddKey as *mut FscryptAddKeyIoctl;
+        unsafe { add_encryption_key(mount_fd.as_raw_fd(), argptr) }.with_whatever_context(
+            |_| format!("Failed to add encryption key for '{}'", path.display()),
+        )?;
+        Ok(())
+    }
+}
+
+impl From<FscryptPrivateKey> for FscryptPublicKey {
+    fn from(private: FscryptPrivateKey) -> Self {
+        Self {
+            identifier: private.identifier,
+        }
+    }
+}
+
+const FSCRYPT_KEY_SPEC_TYPE_IDENTIFIER: u32 = 2;
+const FSCRYPT_MODE_AES_256_XTS: u8 = 1;
+const FSCRYPT_MODE_AES_256_CTS: u8 = 4;
+const FSCRYPT_POLICY_FLAGS_PAD_32: u8 = 3;
+const FSCRYPT_POLICY_V2: u8 = 2;
+const FSCRYPT_MAX_KEY_SIZE: usize = 64;
+
+/// Ioctl struct for FS_IOC_SET_ENCRYPTION_POLICY (ioctl 19).
+/// Corresponds to kernel's fscrypt_policy_v1. Used for ioctl definition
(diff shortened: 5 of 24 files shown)


packages/rottweiler/rottweiler.spec
```diff
@@ -0,0 +1,31 @@
...
+%description
+%{summary}.
+
+%prep
+%setup -T -c
+%cargo_prep
+
+%build
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p rottweiler
```
arnaldo2792, line 25: I assume the `cross_scan_attribution` isn't needed because of this?

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/cfe9fded8063b79402e517077b11f77596bdcd73/packages/os/os.spec#L750

sources/rottweiler/src/key.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let mut random = fs::File::open(DEV_RANDOM)
+        .with_whatever_context(|_| format!("failed to open {}", DEV_RANDOM))?;
+
+    random
+        .read_exact(&mut random_bytes)
+        .with_whatever_context(|_| "failed to read random bytes")?;
+
+    let encrypted = system::systemd_creds_encrypt(&key_id, &random_bytes)?;
+
```
cbgbt, line 41: We should potentially consider [zeroize](https://docs.rs/zeroize/latest/zeroize/)ing the random bytes here to force them out of process memory once the encryption is complete.

sources/rottweiler/src/key.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/// Load and decrypt a TPM2-encrypted key from the keystore
+pub fn load(key_id: String) -> Result<Vec<u8>> {
+    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
+
+    let encrypted = fs::read(&key_path)
+        .with_whatever_context(|_| format!("failed to read key from '{}'", key_path.display()))?;
+
+    system::systemd_creds_decrypt(&key_id, &encrypted)
```
cbgbt, line 58: Could return a `Zeroizing<T>` wrapper here.

sources/rottweiler/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+            name,
+            "--with-key=tpm2",
+            &pcrs,
+        ],
+        Some(plaintext),
+    )
+}
+
+/// Decrypt data using systemd-creds with TPM2
+pub fn systemd_creds_decrypt(name: &str, ciphertext: &[u8]) -> Result<Vec<u8>> {
```
cbgbt: Likewise, could opt to return a `Zeroizing<T>` type here to ensure the returned creds are removed from memory as soon as possible.

sources/rottweiler/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+    execute(
+        CRYPTSETUP,
+        &["resize", "--key-file=-", volume_name],
+        Some(key_data),
+    )?;
+    Ok(())
+}
+
+/// Execute a command with optional stdin input and return stdout
+fn execute(cmd: &str, args: &[&str], input: Option<&[u8]>) -> Result<Vec<u8>> {
```
cbgbt: We should consider zero-izing the stdin and stdout buffers.

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+        let argptr = &mut policy as *mut FscryptPolicyV2 as *mut FscryptPolicyV1Ioctl;
+        unsafe { set_encryption_policy(dir_fd.as_raw_fd(), argptr) }.with_whatever_context(
+            |_| format!("Failed to set encryption policy for '{}'", path.display()),
+        )?;
+        Ok(())
+    }
+}
+
+/// Private key for fscrypt operations (contains the key identifier and raw key material)
+pub struct FscryptPrivateKey {
```
cbgbt, line 66: Could derive the `zeroize` type here.

sources/rottweiler/src/key.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// Generate a random encryption key and encrypt it with TPM2 PCRs 7+14
+pub fn generate(key_id: String) -> Result<()> {
+    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
+
+    // Skip generation if key already exists
+    if key_path.exists() {
+        return Ok(());
+    }
+
```
cbgbt: Do we care about potential TOCTOU issues here?

sources/rottweiler/src/mount_point.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::path::{Path, PathBuf};
+
+type Result<T> = std::result::Result<T, snafu::Whatever>;
+
+/// Convert device number from stat encoding to kernel encoding
+///
+/// stat() returns dev_t as u64 using old encoding: (major & 0xff) << 8 | (minor & 0xff) | ((minor & 0xfff00) << 12)
+/// Kernel uses new encoding internally: (major << 20) | minor, where major fits in 12 bits and minor in 20 bits
+/// This function extracts major/minor from old encoding and re-encodes to match kernel's format.
+fn dev_to_kernel_encoding(stat_dev: u64) -> u32 {
```
cbgbt: Pure function probably warrants a few tests.

sources/rottweiler/src/key.rs
```diff
@@ -0,0 +1,51 @@
...
+type Result<T> = std::result::Result<T, snafu::Whatever>;
+
+const DEV_RANDOM: &str = "/dev/random";
+const KEYSTORE: &str = "/.bottlerocket/keystore";
+const KEY_SIZE: usize = 64;
+
+/// Generate a random encryption key and encrypt it with TPM2 PCRs 7+14
+pub fn generate(key_id: String) -> Result<()> {
+    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
```
cbgbt, line 24: We could consider sanitizing `key_id` for path traversal elements, but as I understand it, the input here is almost certainly coming from a static systemd service, so I don't think that there's necessarily a problem here.

cbgbt (commented):
Still working on a more thorough review, but here are some initial ideas from stepping through the code with Q.

sources/rottweiler/src/block_device.rs
```diff
@@ -0,0 +1,93 @@
...
+    let volume_name = path
+        .file_name()
+        .with_whatever_context(|| {
+            format!("failed to extract volume name from '{}'", path.display())
+        })?
+        .to_str()
+        .with_whatever_context(|| {
+            format!("volume name is not valid UTF-8: '{}'", path.display())
+        })?;
+
```
cbgbt: Consider extracting to a function, given the repetition.

sources/rottweiler/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/// Format a device with LUKS2 using the provided key
+pub fn cryptsetup_luks_format(device: &str, key_data: &[u8]) -> Result<()> {
+    // Use minimal PBKDF2 iterations (1000, per NIST SP 800-132) since we're using a high-entropy
+    // TPM2-sealed key. This matches systemd's behavior and avoids unnecessary key stretching.
+    execute(
+        CRYPTSETUP,
+        &[
+            "luksFormat",
```
cbgbt, line 74: I'd wondered if detecting the physical sector size of the underlying device is a good idea. Instance types with 4K physical sectors could perform better if we format to the same sector size.

On the contrary, though, if you re-use an EBS snapshot you could end up in a situation where the new host has a smaller physical sector size. My understanding is that having larger LUKS sectors than the physical disk sectors could lead to data corruption on power loss.

sources/rottweiler/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+    if !std::path::Path::new(IMAGE_FEATURES_FILE).exists() {
+        return Ok(ImageFeatures {
+            in_place_updates: true,
+        });
+    }
+
+    let content = fs::read_to_string(IMAGE_FEATURES_FILE)
+        .with_whatever_context(|_| format!("failed to read {}", IMAGE_FEATURES_FILE))?;
+
+    let pairs: Vec<(String, String)> = content
```
cbgbt: We should probably add unit tests for this parser.

sources/rottweiler/src/mount_point.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Find the mount point for a given path by walking up the directory tree
+    pub fn from_path(path: &Path) -> Result<Self> {
+        let mut current = path.canonicalize().with_whatever_context(|_| {
+            format!("failed to canonicalize path '{}'", path.display())
+        })?;
+
+        let current_stat = std::fs::metadata(&current).with_whatever_context(|_| {
+            format!("failed to get metadata for '{}'", current.display())
+        })?;
+        let current_dev = dev_to_kernel_encoding(current_stat.dev());
```
cbgbt: Do we actually want to do this encoding transformation? We're always comparing devices as returned by rust's stdlib, so the kernel encoding should not matter.

I'm pretty sure we can remove the `dev_to_kernel_encoding` function.

sources/rottweiler/src/mount_point.rs
```diff
@@ -0,0 +1,64 @@
...
+type Result<T> = std::result::Result<T, snafu::Whatever>;
+
+/// Convert device number from stat encoding to kernel encoding
+///
+/// stat() returns dev_t as u64 using old encoding: (major & 0xff) << 8 | (minor & 0xff) | ((minor & 0xfff00) << 12)
+/// Kernel uses new encoding internally: (major << 20) | minor, where major fits in 12 bits and minor in 20 bits
+/// This function extracts major/minor from old encoding and re-encodes to match kernel's format.
+fn dev_to_kernel_encoding(stat_dev: u64) -> u32 {
+    // Extract major and minor from old stat encoding
+    let major = ((stat_dev >> 8) & 0xff) as u32; // 8 bits max
```
cbgbt: I think the major number can actually be > 8 bits now. NVME devices have a major of 259.

But I left another comment pointing out that we probably don't actually need to change the device number encoding.

cbgbt (commented):
Still working through directory encryption!

sources/rottweiler/src/mount_point.rs
bcressey: > I'm pretty sure we can remove the dev_to_kernel_encoding function.

You're right. This is an artifact of the BPF LSM hook that I'm planning to add later, to block changes to a mount point.

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    /// Set the encryption policy on a directory
+    pub fn encrypt_directory(&self, path: &Path) -> Result<()> {
+        let mut policy = FscryptPolicyV2::new(self.identifier);
+        let dir_fd = std::fs::File::open(path)
+            .with_whatever_context(|_| format!("Failed to open directory '{}'", path.display()))?;
+        let argptr = &mut policy as *mut FscryptPolicyV2 as *mut FscryptPolicyV1Ioctl;
+        unsafe { set_encryption_policy(dir_fd.as_raw_fd(), argptr) }.with_whatever_context(
+            |_| format!("Failed to set encryption policy for '{}'", path.display()),
+        )?;
```
cbgbt, line 59:
```suggestion
        // SAFETY: The ioctl requires a valid file descriptor and a pointer to a policy struct.
        // - `dir_fd` is a valid open file descriptor for the directory
        // - The kernel reads the version field first (via get_user), then copies the full struct
        // - Both FscryptPolicyV2 and FscryptPolicyV1Ioctl are repr(C) and start with version field
        // - The cast is safe because the kernel interprets the data based on the version field
 
        unsafe { set_encryption_policy(dir_fd.as_raw_fd(), argptr) }.with_whatever_context(
            |_| format!("Failed to set encryption policy for '{}'", path.display()),
        )?;
```

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /// Remove the encryption key from the kernel keyring, locking the directory
+    pub fn lock_directory(&self, path: &Path) -> Result<()> {
+        let mount_point = MountPoint::from_path(path).with_whatever_context(|_| {
+            format!("Failed to find mount point for '{}'", path.display())
+        })?;
+        let mount_fd = mount_point.open()?;
+        let mut remove_key = FscryptRemoveKey::new(self.identifier);
+        unsafe { remove_encryption_key_all_users(mount_fd.as_raw_fd(), &mut remove_key) }
```
cbgbt, line 39:
```suggestion
        // SAFETY: The ioctl requires a valid file descriptor and a pointer to fscrypt_remove_key_arg.
        // - `mount_fd` is a valid open file descriptor
        // - `remove_key` is properly initialized with repr(C) layout matching the kernel struct
        // - The kernel copies data with copy_from_user/copy_to_user, so no lifetime issues
        unsafe { remove_encryption_key_all_users(mount_fd.as_raw_fd(), &mut remove_key) }
```

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+    pub fn unlock_directory(&self, path: &Path) -> Result<()> {
+        let mount_point = MountPoint::from_path(path).with_whatever_context(|_| {
+            format!("Failed to find mount point for '{}'", path.display())
+        })?;
+        let mount_fd = mount_point.open()?;
+        let mut add_key = FscryptAddKey::new(&self.raw)?;
+        let argptr = &mut add_key as *mut FscryptAddKey as *mut FscryptAddKeyIoctl;
+        unsafe { add_encryption_key(mount_fd.as_raw_fd(), argptr) }.with_whatever_context(
+            |_| format!("Failed to add encryption key for '{}'", path.display()),
+        )?;
```
cbgbt, line 96:
```suggestion
        // SAFETY: The ioctl requires a valid file descriptor and a pointer to fscrypt_add_key_arg.
        // - `mount_fd` is a valid open file descriptor for the mount point
        // - `add_key` is properly initialized with repr(C) layout matching the kernel struct
        // - The cast is safe: FscryptAddKey contains FscryptAddKeyIoctl fields plus raw[] array
        // - The kernel uses copy_from_user to read the struct including the raw key bytes
        unsafe { add_encryption_key(mount_fd.as_raw_fd(), argptr) }.with_whatever_context(
            |_| format!("Failed to add encryption key for '{}'", path.display()),
        )?;
```

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+        })
+    }
+
+    fn from_path(path: &Path) -> Result<Self> {
+        let dir = std::fs::File::open(path)
+            .with_whatever_context(|_| format!("Failed to open directory '{}'", path.display()))?;
+        let mut arg = Self::new()?;
+        let argptr = &mut arg as *mut FscryptGetPolicyExArg as *mut FscryptGetPolicyExArgIoctl;
+        unsafe { get_encryption_policy_ex(dir.as_raw_fd(), argptr) }
+            .with_whatever_context(|_| "Failed to get encryption policy")?;
```
cbgbt, line 214:
```suggestion
        // SAFETY: The ioctl requires a valid file descriptor and a pointer to fscrypt_get_policy_ex_arg.
        // - `dir` is a valid open file descriptor for the directory
        // - `arg` is properly initialized with policy_size and version fields
        // - The kernel first copies policy_size from userspace, then copies the full policy back
        // - Both structs are repr(C) and share the same initial layout (policy_size field)
        unsafe { get_encryption_policy_ex(dir.as_raw_fd(), argptr) }
            .with_whatever_context(|_| "Failed to get encryption policy")?;
```

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+        Ok(arg)
+    }
+
+    fn key_identifier(&self) -> [u8; 16] {
+        unsafe { self.policy.v2.master_key_identifier }
```
cbgbt:
```suggestion
        // SAFETY: Accessing union field is safe because:
        // - We initialized the policy union with version = FSCRYPT_POLICY_V2
        // - The kernel's fscrypt_ioctl_get_policy_ex writes the policy based on what's stored
        // - The kernel implementation (fs/crypto/policy.c) guarantees the v2 variant is active
        //   when the stored policy version is v2
        // - We only call this after successfully retrieving the policy via ioctl
        unsafe { self.policy.v2.master_key_identifier }
```

But we could also modify the implementation to make it impossible to call this in Rust without first using a separate type. [Sample of the idea](https://gist.github.com/cbgbt/fb7366b36ceb5f554f972bef3be0eca4)

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+/// V2 encryption policy struct. Corresponds to kernel's fscrypt_policy_v2.
+/// Contains the actual policy data passed to FS_IOC_SET_ENCRYPTION_POLICY.
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct FscryptPolicyV2 {
+    version: u8,
+    contents_encryption_mode: u8,
+    filenames_encryption_mode: u8,
+    flags: u8,
+    __reserved: [u8; 4],
```
cbgbt, line 138: It seems like kernel 6.12 has eaten one byte of the reserved space here to add a field called `log2_data_unit_size`.

But`0` is still a reasonable default for that (use filesystem block size), so it doesn't have an impact on our implementation. I'm OK to leave the reserved space as-is.

sources/rottweiler/src/fscrypt.rs
```diff
@@ -0,0 +1,291 @@
...
+ioctl_read!(set_encryption_policy, b'f', 19, FscryptPolicyV1Ioctl);
+
+/// Calculate the fscrypt key identifier from raw key bytes using HKDF-SHA512
+fn calculate_key_identifier(key: &[u8]) -> Result<[u8; 16]> {
+    let hkdf = Hkdf::<Sha512>::new(None, key);
+    let mut output = [0u8; 16];
+    hkdf.expand(b"fscrypt\x00\x01", &mut output)
+        .with_whatever_context(|_| "Failed to derive key identifier")?;
+    Ok(output)
+}
```
cbgbt, line 330: I got Q to generate const assertions that the FFI types here match the layout of the kernel ones:


```rust
// Compile-time validation of struct layouts against kernel definitions
// These assertions ensure our repr(C) structs match the kernel's C structs exactly.

const _: () = {
    use std::mem::{align_of, offset_of, size_of};

    // struct fscrypt_policy_v1: 1 + 1 + 1 + 1 + 8 = 12 bytes
    const _: () = assert!(size_of::<FscryptPolicyV1Ioctl>() == 12);
    const _: () = assert!(align_of::<FscryptPolicyV1Ioctl>() == 1);

    // struct fscrypt_policy_v2: 1 + 1 + 1 + 1 + 1 + 3 + 16 = 24 bytes
    const _: () = assert!(size_of::<FscryptPolicyV2>() == 24);
    const _: () = assert!(align_of::<FscryptPolicyV2>() == 1);
    const _: () = assert!(offset_of!(FscryptPolicyV2, version) == 0);
    const _: () = assert!(offset_of!(FscryptPolicyV2, contents_encryption_mode) == 1);
    const _: () = assert!(offset_of!(FscryptPolicyV2, filenames_encryption_mode) == 2);
    const _: () = assert!(offset_of!(FscryptPolicyV2, flags) == 3);
    const _: () = assert!(offset_of!(FscryptPolicyV2, __reserved) == 4);
    const _: () = assert!(offset_of!(FscryptPolicyV2, master_key_identifier) == 8);

    // struct fscrypt_key_specifier: 4 + 4 + 32 = 40 bytes
    const _: () = assert!(size_of::<FscryptKeySpecifier>() == 40);
    const _: () = assert!(align_of::<FscryptKeySpecifier>() == 4);
    const _: () = assert!(offset_of!(FscryptKeySpecifier, type_) == 0);
    const _: () = assert!(offset_of!(FscryptKeySpecifier, __reserved) == 4);
    const _: () = assert!(offset_of!(FscryptKeySpecifier, u) == 8);

    // struct fscrypt_add_key_arg (without raw[]): 40 + 4 + 4 + 32 = 80 bytes
    const _: () = assert!(size_of::<FscryptAddKeyIoctl>() == 80);
    const _: () = assert!(align_of::<FscryptAddKeyIoctl>() == 4);
    const _: () = assert!(offset_of!(FscryptAddKey, key_spec) == 0);
    const _: () = assert!(offset_of!(FscryptAddKey, raw_size) == 40);
    const _: () = assert!(offset_of!(FscryptAddKey, key_id) == 44);
    const _: () = assert!(offset_of!(FscryptAddKey, __reserved) == 48);
    const _: () = assert!(offset_of!(FscryptAddKey, raw) == 80);

    // struct fscrypt_add_key_arg with raw[64]: 80 + 64 = 144 bytes
    const _: () = assert!(size_of::<FscryptAddKey>() == 144);
    const _: () = assert!(align_of::<FscryptAddKey>() == 4);

    // struct fscrypt_remove_key_arg: 40 + 4 + 20 = 64 bytes
    const _: () = assert!(size_of::<FscryptRemoveKey>() == 64);
    const _: () = assert!(align_of::<FscryptRemoveKey>() == 4);
    const _: () = assert!(offset_of!(FscryptRemoveKey, key_spec) == 0);
    const _: () = assert!(offset_of!(FscryptRemoveKey, removal_status_flags) == 40);
    const _: () = assert!(offset_of!(FscryptRemoveKey, __reserved) == 44);

    // struct fscrypt_get_policy_ex_arg: 8 + max(1, 12, 24) = 32 bytes
    const _: () = assert!(size_of::<FscryptGetPolicyExArg>() == 32);
    const _: () = assert!(align_of::<FscryptGetPolicyExArg>() == 8);

    // __u8[9] for ioctl definition: 8 + 1 = 9 bytes (packed)
    const _: () = assert!(size_of::<FscryptGetPolicyExArgIoctl>() == 9);
    const _: () = assert!(align_of::<FscryptGetPolicyExArgIoctl>() == 1);
};

```

cbgbt (commented):
Done! Nice work.

bcressey:
Added commits to fix all suggestions - thanks @cbgbt!

sources/rottweiler/src/system.rs
bcressey: > On the contrary, though, if you re-use an EBS snapshot you could end up in a situation where the new host has a smaller physical sector size. My understanding is that having larger LUKS sectors than the physical disk sectors could lead to data corruption on power loss.

Since the encryption key is sealed to the host's TPM, it won't be available on a different host, meaning it won't be possible to mount the LUKS volume elsewhere.

bcressey:
Added two new commits:
1. The new `measure` command works around a problem I was having with `tpm2_pcrevent`, and also cleans up a lot of clunky mechanics around setting up the input for that command in the absence of a shell.
2. The SELinux policy change came over here since it's small, and since `rottweiler` is managing its keystore directory on the private filesystem.

bcressey:
Added a commit to add parsing for the encrypted storage image feature too.

bcressey:
Added a fix to shift dependencies to `rottweiler.spec`, now that it wraps all of the underlying system commands.

sources/rottweiler/src/system.rs
```diff
@@ -158,3 +160,33 @@ fn get_tpm2_pcrs() -> Result<String> {
         "4+7+9+11+14".to_string()
     })
 }
+
+/// Get canonicalized settings from apiclient, excluding random-seed and hostname
```
cbgbt: It feels somewhat brittle to explicitly exclude these here, but I don't think we really have a feasible alternative.

I'd be OK with opening an issue to discuss, but perhaps some system where the variant owner can record specific settings in a file to mark them as being transient/host-identifying.

cbgbt (approved):

sources/rottweiler/src/key.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    // Skip generation if key already exists
```
arnaldo2792, line 26:
```suggestion

```

sources/rottweiler/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+/// PCR decoder ring:
+/// - PCR 4: hashes of shim, grub, and vmlinuz
+/// - PCR 7: Secure Boot policy
+/// - PCR 9: kernel command line (includes dm-verity root hash of userspace)
+/// - PCR 11: boot phase (sysinit, preconfigured, configured, ready, shutdown, final)
+/// - PCR 14: machine-owner keys
+///
+/// If in-place updates are disabled, then the measurements in PCR 4 (kernel) and
+/// PCR 9 (userspace) shouldn't change.
+fn get_tpm2_pcrs() -> Result<String> {
```
arnaldo2792, line 155: Were these PCRs arbitrarily selected?

sources/bottlerocket-image-features/src/lib.rs
```diff
@@ -53,6 +53,8 @@ const IMAGE_FEATURES_FILE: &str = "/usr/share/bottlerocket/image-features.env";
 pub struct ImageFeatures {
```
arnaldo2792, line 54: Did you mean to `fixup` this commit?

packages/rottweiler/rottweiler.spec
```diff
@@ -9,6 +9,8 @@ License: Apache-2.0 OR MIT
 URL: https://github.com/bottlerocket-os/bottlerocket
```
arnaldo2792, line 9: Same with this? `fixup!` the commit?

sources/bottlerocket-image-features/README.md
```diff
@@ -17,6 +17,7 @@ file, which contains feature flags that control various aspects of the Bottleroc
 Currently supported feature flags:
```
arnaldo2792, line 17: Same as elsewhere about the `fixup` commits

packages/rottweiler/rottweiler.spec
bcressey: > I assume the `%cross_scan_attribution` isn't needed because of this?

Yes, we'll create the right attribution files based on the scan in `os`.

sources/rottweiler/src/system.rs
bcressey: > Were these PCRs arbitrarily selected?

No 😀 A lot of thought went into them. There's some documentation over in #721.

7 is Secure Boot; if the system isn't in Secure Boot mode, or if that ever changes for some reason, we don't want the sealed key to be available.

11 is extended whenever we finish a boot phase. Including it means that the sealed key can only ever be retrieved during the same boot phase that we first generated it.

14 is managed by shim and records machine-owner keys. There shouldn't ever be any, but if there are, it effectively means the set of keys trusted for Secure Boot has changed, which is unexpected and basically the same as Secure Boot being disabled under our threat model.

sources/bottlerocket-image-features/src/lib.rs
bcressey: > Did you mean to fixup this commit?

I'm planning to squash all commits together prior to merge. I wanted to make it easier for reviewers to see incremental changes in case they were partway through a review.

sources/rottweiler/src/system.rs
KCSesh: Yeah I think I would want to see this built into apiclient as a subcommand? 
Maybe something like:

```
apiclient get-canonicalize settings
```
🤷 

It would be nice to have a test to help capture this, but I agree it makes sense for now giving no great alternatives.

sources/rottweiler/src/system.rs
```diff
@@ -0,0 +1,192 @@
...
+
+/// Get canonicalized settings from apiclient, excluding random-seed and hostname
+pub fn apiclient_get_settings() -> Result<Zeroizing<Vec<u8>>> {
+    execute(
+        APICLIENT,
+        &[
+            "get",
+            "settings",
+            "--exclude",
+            "settings.updates.random-seed",
```
KCSesh: I am trying to think through the possibilities here across bootstrap commands/containers or host-containers setting values on boot. 

What was your methodology here for this list?

KCSesh: Am I missing an alias? 

```suggestion
            "settings.updates.seed",
```

KCSesh: Seems this was just fixed 👍

bcressey:
Two fixes:
1. `settings.updates.seed` is the correct setting for update waves
2. when encrypting a directory, remove it first. The fscrypt API to set the encryption policy will fail if the directory is not empty, but this way the behavior is the same as when encrypting a block device (existing content is wiped).

arnaldo2792 (approved):

sources/rottweiler/src/system.rs
bcressey: > What was your methodology here for this list?

I targeted settings that were different across two runs of the same AMI.

sources/rottweiler/src/system.rs
bcressey: Opened https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/739 for this as a follow-up feature.

cbgbt (approved):

KCSesh (approved):

bcressey:
⬆️ (hopefully) final force push to squash commits and finalize commit messages

sources/rottweiler/src/fscrypt.rs
KCSesh: Good idea!

KCSesh (approved):
Finalized changes LGTM

sources/rottweiler/src/system.rs
bcressey: > Am I missing an alias?

No, this was a bug. Too many random numbers on my mind 😀

PR #722, opened 2025-11-07
feat: add image signing verification for ECR images signed by AWS Signer

ginglis13:
**Issue number:**

Related: https://github.com/bottlerocket-os/bottlerocket/issues/4684

**Description of changes:**

In containerd 2.0, support for [image verification plugins](https://github.com/containerd/containerd/blob/main/docs/image-verification.md) landed as part of the new [containerd image transfer service](https://github.com/containerd/containerd/blob/main/docs/transfer.md). This PR includes an image verifier binary and its required dependencies to verify ECR images signed via the [AWS Signer service](https://docs.aws.amazon.com/signer/latest/developerguide/Welcome.html).

New packages:
* [notation](https://github.com/notaryproject/notation): a CLI for signing and verifying artifacts.
* [ecr-credential-helper](https://github.com/awslabs/amazon-ecr-credential-helper): a Docker credential helper for authenticating with ECR
* [aws-signer-notation-plugin](https://github.com/aws/aws-signer-notation-plugin): a plugin for notation to interact with AWS signer. As packaged, it will write the AWS signer commercial and gov root CAs to a location expected by the notation CLI
* notation-image-verifier: a first party source that sets points notation to the config, library, and cache directories packaged via the core-kit, wraps the notation CLI and runs `notation verify` on the incoming image.

**Testing done:**

I built the core-kit with https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/106 and changes to my local Bottlerocket aws-k8s-1.34 variant to include these new packages / settings / configuration-files. I then tested, in both `us-west-2` and `us-gov-west-1`:

1. Configure my node with `settings.image-verifier-plugins.notation.trustpolicy=""ewog..."` which was base64 encoded from a trustpolicy.json like:

```json
{
   "version":"1.0",
   "trustPolicies":[
      {
         "name":"aws-signer-tp",
         "registryScopes":[
            "*"
         ],
         "signatureVerification":{
            "level":"strict",
            "override": {
                "revocation": "skip"
            }
         },
         "trustStores":[
            "signingAuthority:aws-signer-ts",
            "signingAuthority:aws-us-gov-signer-ts"
         ],
         "trustedIdentities":[
            "arn:aws:signer:us-west-2:<redacted>:/signing-profiles/notation_test"
         ]
      }
   ]
}
```

2. ctr pull an unsigned image from a private registry in my account. Pull fails due to lack of signature.
3. ctr pull a signed image from a private registry in my account. Pull succeeds.
4. Repeat 2 and 3 with public registries, and all of this on aws-k8s-1.34-fips

Example of failure:

```
bash-5.1# ctr i pull docker.io/library/hello-world:latest
ctr: image verifier bindir blocked pull of docker.io/library/hello-world:latest with digest sha256:56433a6be3fda188089fb548eae3d91df3ed0d6589f7c2656121b911198df065 for reason: verifier notation-image-verifier rejected image (exit code 1): verifying image: docker.io/library/hello-world:latest
image verification failed: log: failed to create directory: mkdir /root/.ecr: read-only file systemWarning: Always verify the artifact using digest(@sha256:...) rather than a tag(:latest) because resolved digest may not point to the same signed artifact, as tags are mutable.
Error: signature verification failed: no signature is associated with "docker.io/library/hello-world@sha256:56433a6be3fda188089fb548eae3d91df3ed0d6589f7c2656121b911198df065", make sure the artifact was signed successfully
```

Pull succeeding has no additional output to normal pull without image verification.

Processes also are inheriting the `runtime_t` selinux label:

```
bash-5.1# ps -eZ | grep notation
system_u:system_r:runtime_t:s0     2922 ?        00:00:00 notation-image-
system_u:system_r:runtime_t:s0     2928 ?        00:00:00 notation
bash-5.1# ps -eZ | grep containerd
system_u:system_r:runtime_t:s0     1574 ?        00:00:07 containerd
system_u:system_r:runtime_t:s0     1626 ?        00:00:01 containerd
system_u:system_r:runtime_t:s0     1936 ?        00:00:00 containerd-shim
system_u:system_r:runtime_t:s0     2414 ?        00:00:00 containerd-shim
```

**Terms of contribution:**


packages/aws-signer-notation-plugin/aws-signer-notation-root.crt
```diff
@@ -0,0 +1,15 @@
+-----BEGIN CERTIFICATE-----
```
ginglis13, line 1: these certs are linked in this guide: https://docs.aws.amazon.com/signer/latest/developerguide/image-signing-prerequisites.html

sources/notation-image-verifier/main.go
```diff
@@ -0,0 +1,46 @@
...
+	if *digest == "" || *name == "" {
+		// containerd only captures the stdout of image verification binaries, hence no
+		// write to stderr here (and elsewhere in the program).
+		fmt.Printf("Usage: %s -digest <digest> -name <name>\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	fmt.Printf("verifying image: %s\n", *name)
+
+	// Override the default notation paths to what's packaged in packages/notation
```
ginglis13, line 58: I'll update this with a direct link to https://notaryproject.dev/docs/user-guides/how-to/directory-structure/ in the comment

packages/notation/notation.spec
```diff
@@ -0,0 +1,92 @@
...
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
+%{_cross_attribution_vendor_dir}
+%{_cross_templatedir}/notation-trust-policy-json
+%{_cross_tmpfilesdir}/notation.conf
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/notation
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/notation/plugins
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/notation/truststore/x509/signingAuthority
```
ginglis13, line 81: See https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trust-store

packages/aws-signer-notation-plugin/aws-signer-notation-plugin.spec
```diff
@@ -0,0 +1,98 @@
...
+%cross_scan_attribution go-vendor vendor
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
+%{_cross_attribution_vendor_dir}
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/notation/truststore/x509/signingAuthority/aws-signer-ts/
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/notation/truststore/x509/signingAuthority/aws-us-gov-signer-ts/
+%{_cross_factorydir}%{_cross_sysconfdir}/notation/truststore/x509/signingAuthority/aws-signer-ts/aws-signer-notation-root.crt
+%{_cross_factorydir}%{_cross_sysconfdir}/notation/truststore/x509/signingAuthority/aws-us-gov-signer-ts/aws-us-gov-signer-notation-root.crt
```
ginglis13: See the notation spec https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trust-store

ginglis13:
^ force push adds
```
C /etc/notation - - - - /usr/share/factory/etc/notation
```

to the notation tmpfiles to copy over the factory files.

packages/notation/notation.spec
```diff
@@ -0,0 +1,92 @@
...
+Release: 1%{?dist}
+Summary:  A CLI tool to sign and verify artifacts.
+License: Apache-2.0
+URL: https://%{goimport}
+Source0: https://%{goimport}/archive/v%{gover}/%{gorepo}-v%{gover}.tar.gz
+Source1: bundled-%{gorepo}-v%{gover}.tar.gz
+Source2: notation-trust-policy-json
+Source3: notation-tmpfiles.conf
+
+BuildRequires: git
```
arnaldo2792: You don't need git for this build, right? No `git` repos are initialized, and there aren't patches applied.

packages/notation/notation-tmpfiles.conf
```diff
@@ -0,0 +1,2 @@
+d /var/cache/notation - - - -
+C /etc/notation - - - - /usr/share/factory/etc/notation
```
arnaldo2792: If the source factory is the exact same path as the target path, you can avoid it.

packages/ecr-credential-helper/ecr-credential-helper.spec
```diff
@@ -0,0 +1,81 @@
...
+%prep
+%autosetup -n %{gorepo}-%{gover} -p1
+%cross_go_setup %{gorepo}-%{gover} %{goproject} %{goimport}
+
+
+%build
+# cross_go_configure cd's to the correct GOPATH location
+%cross_go_configure %{goimport}
+
+export GO_MAJOR="1.24"
```
arnaldo2792: Why do you need this? isn't `GO_MAJOR` already `1.24`?

packages/aws-signer-notation-plugin/aws-signer-notation-plugin.spec
```diff
@@ -0,0 +1,98 @@
...
+go build -ldflags "${GOLDFLAGS}" -o notation-com.amazonaws.signer.notation.plugin ./cmd
+gofips build -ldflags "${GOLDFLAGS}" -o fips/notation-com.amazonaws.signer.notation.plugin ./cmd
+
+%install
+install -d %{buildroot}{%{_cross_bindir},%{_cross_fips_bindir},%{_cross_templatedir}}
+
+# Place the binaries where notation expects them.
+install -d %{buildroot}%{_cross_bindir}/notation-plugins
+install -d %{buildroot}%{_cross_bindir}/notation-plugins/plugins
+install -d %{buildroot}%{_cross_bindir}/notation-plugins/plugins/com.amazonaws.signer.notation.plugin
```
arnaldo2792: You could create a `global` to reduce some of the verbosity of the path.

packages/aws-signer-notation-plugin/aws-signer-notation-plugin.spec
```diff
@@ -0,0 +1,98 @@
...
+
+install -d %{buildroot}%{_cross_fips_bindir}/notation-plugins
+install -d %{buildroot}%{_cross_fips_bindir}/notation-plugins/plugins
+install -d %{buildroot}%{_cross_fips_bindir}/notation-plugins/plugins/com.amazonaws.signer.notation.plugin
+
+install -p -m 0755 notation-com.amazonaws.signer.notation.plugin %{buildroot}%{_cross_bindir}/notation-plugins/plugins/com.amazonaws.signer.notation.plugin/notation-com.amazonaws.signer.notation.plugin
+install -p -m 0755 fips/notation-com.amazonaws.signer.notation.plugin %{buildroot}%{_cross_fips_bindir}/notation-plugins/plugins/com.amazonaws.signer.notation.plugin/notation-com.amazonaws.signer.notation.plugin
+
+# Add the notation config truststore directories.
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/notation/truststore/x509/signingAuthority/aws-signer-ts/
```
arnaldo2792: Same for this, a `global` could help with the verbosity.

packages/aws-signer-notation-plugin/aws-signer-notation-root.crt
arnaldo2792: Is there a way to know when we have to rotate the certificates? Maybe a script or something we can run?

packages/notation-image-verifier/notation-image-verifier.spec
```diff
@@ -0,0 +1,35 @@
+%global _cross_first_party 1
+%global workspace_name notation-image-verifier
+
+Name: %{_cross_os}%{workspace_name}
+Version: 0.1.0
+Release: 1%{?dist}
+Summary: Binary for containerd image verification plugin
```
arnaldo2792:
```suggestion
Summary: A notation-based containerd image verification plugin
```

packages/notation-image-verifier/notation-image-verifier.spec
```diff
@@ -0,0 +1,35 @@
...
+
+%description
+%{summary}.
+
+%prep
+%setup -T -c
+cp -r %{_builddir}/sources/%{workspace_name}/* .
+
+%build
+export GO_MAJOR="1.24"
```
arnaldo2792: Same as elsewhere, `GO_MAJOR` is already `1.24`. Since we control the sources of this package, you could just use 1.25 right away.

packages/notation-image-verifier/notation-image-verifier.spec
```diff
@@ -0,0 +1,35 @@
...
+
+%set_cross_go_flags
+go build -ldflags="${GOLDFLAGS}" -o notation-image-verifier .
+
+%install
+install -d %{buildroot}%{_cross_bindir}/image-verifiers
+install -d %{buildroot}%{_cross_bindir}/image-verifiers/bin
+install -p -m 0755 notation-image-verifier %{buildroot}%{_cross_bindir}/image-verifiers/bin
```
arnaldo2792: You still need to add the `cross_attribution` macro here.

sources/Cargo.toml
```diff
@@ -91,6 +91,7 @@ datastore = { version = "0.1", path = "api/datastore" }
 dogtag = { version = "0.1", path = "dogtag" }
 early-boot-config-provider = { version = "0.1", path = "early-boot-config/early-boot-config-provider" }
 generate-readme = { version = "0.1", path = "generate-readme" }
+image-verifier = { version = "0.1", path = "image-verifier"  }
```
arnaldo2792: You should only add this here if the package was a Rust program

```suggestion
```

packages/os/os.spec
```diff
@@ -82,6 +82,7 @@ Requires: %{_cross_os}corndog
 Requires: %{_cross_os}certdog
 Requires: %{_cross_os}driverdog
 Requires: %{_cross_os}ghostdog
+Requires: %{_cross_os}notation-image-verifier
```
arnaldo2792: You may not want this here, if anything it should go in `release`. But I'll say make this an opt-in and we will include it in the variant definition.

packages/notation/notation-trust-policy-json
```diff
@@ -0,0 +1,7 @@
+[required-extensions]
+image-verifier-plugins = "v1"
+std = { version = "v1", helpers = ["base64_decode"] }
++++
+{{~#if settings.image-verifier-plugins.notation.trustpolicy~}}
+{{base64_decode settings.image-verifier-plugins.notation.trustpolicy}}
+{{~/if~}}
```
bcressey, line 7: Should we emit a `{}` empty policy if we don't have a trust policy, or there's no need?

packages/notation/notation.spec
```diff
@@ -0,0 +1,92 @@
...
+%global gover 1.3.2
+%global rpmver %{gover}
+%global gitrev 001cc919603c1dc16c6aad387c94b4209cb9c901
+
+%global _dwz_low_mem_die_limit 0
+
+Name: %{_cross_os}%{gorepo}
+Version: %{rpmver}
+Release: 1%{?dist}
+Summary:  A CLI tool to sign and verify artifacts.
```
bcressey: nit:

```suggestion
Summary: A CLI tool to sign and verify artifacts.
```

packages/notation/notation.spec
```diff
@@ -0,0 +1,92 @@
...
+
+%description fips-bin
+%{summary}.
+
+%prep
+%setup -n %{gorepo}-%{gover} -q
+%setup -T -D -n %{gorepo}-%{gover} -b 1 -q
+
+%build
+
```
bcressey: nit:
```suggestion
```

sources/notation-image-verifier/main.go
```diff
@@ -0,0 +1,46 @@
...
+	os.Setenv("NOTATION_CONFIG", "/etc/notation")
+	os.Setenv("NOTATION_CACHE", "/var/cache/notation")
+	os.Setenv("NOTATION_LIBEXEC", "/usr/bin/notation-plugins")
+
+	// The notation CLI looks in $HOME/.docker/config.json for credentials.
+	// Bottlerocket doesn't set a $HOME.
+	os.Setenv("HOME", "/root")
+
+	cmd := exec.Command("notation", "verify", *name)
```
arnaldo2792: Is there a problem if you use the full path to `notation`? At least that's what we do in rust programs when we call other helper binaries.

packages/aws-signer-notation-plugin/aws-signer-notation-plugin.spec
```diff
@@ -0,0 +1,98 @@
...
+
+go build -ldflags "${GOLDFLAGS}" -o notation-com.amazonaws.signer.notation.plugin ./cmd
+gofips build -ldflags "${GOLDFLAGS}" -o fips/notation-com.amazonaws.signer.notation.plugin ./cmd
+
+%install
+install -d %{buildroot}{%{_cross_bindir},%{_cross_fips_bindir},%{_cross_templatedir}}
+
+# Place the binaries where notation expects them.
+install -d %{buildroot}%{_cross_bindir}/notation-plugins
+install -d %{buildroot}%{_cross_bindir}/notation-plugins/plugins
```
arnaldo2792: I saw you have control over this path, it will be best to follow fedora's guidelines and install them in `libexec`.

sources/notation-image-verifier/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+func main() {
+	var digest = flag.String("digest", "", "image digest to verify")
+	var name = flag.String("name", "", "image name to verify")
+	var _ = flag.String("stdin-media-type", "", "image media type")
+	flag.Parse()
+
+	if *digest == "" || *name == "" {
+		// containerd only captures the stdout of image verification binaries, hence no
+		// write to stderr here (and elsewhere in the program).
```
arnaldo2792: So are you saying `containerd` only cares about what's written to the `stdout`? does it expect any format in the output or does it capture `stdout` for logging purposes?

packages/ecr-credential-helper/ecr-credential-helper.spec
```diff
@@ -0,0 +1,81 @@
...
+Requires: (%{_cross_os}image-feature(fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
+
+%description fips-bin
+%{summary}.
+
+%prep
+%autosetup -n %{gorepo}-%{gover} -p1
+%cross_go_setup %{gorepo}-%{gover} %{goproject} %{goimport}
+
```
bcressey, line 51:
```suggestion
```

sources/notation-image-verifier/main.go
```diff
@@ -0,0 +1,46 @@
...
+		fmt.Printf("Usage: %s -digest <digest> -name <name>\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	fmt.Printf("verifying image: %s\n", *name)
+
+	// Override the default notation paths to what's packaged in packages/notation
+	os.Setenv("NOTATION_CONFIG", "/etc/notation")
+	os.Setenv("NOTATION_CACHE", "/var/cache/notation")
+	os.Setenv("NOTATION_LIBEXEC", "/usr/bin/notation-plugins")
```
arnaldo2792: As mentioned elsewhere, it will be best to refer to `libexec`

sources/notation-image-verifier/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// The notation CLI looks in $HOME/.docker/config.json for credentials.
+	// Bottlerocket doesn't set a $HOME.
+	os.Setenv("HOME", "/root")
```
arnaldo2792: TIL!

Cargo.toml
```diff
@@ -101,6 +101,7 @@ members = [
   "packages/netdog",
```
arnaldo2792, line 103: nit: for consistency, you could separate the `source` and `package` commits.

packages/containerd-2.1/image-verification-toml
```diff
@@ -0,0 +1,11 @@
+[required-extensions]
+image-verifier-plugins = "v1"
+std = { version = "v1" }
++++
+{{~#if settings.image-verifier-plugins.notation.trustpolicy~}}
+[plugins]
+  [plugins."io.containerd.image-verifier.v1.bindir"]
+    bin_dir = "/usr/bin/image-verifiers/bin"
```
arnaldo2792: If we control this, I would say `libexec` is the right place for the plugins

packages/aws-signer-notation-plugin/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "aws-signer-notation-plugin"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[[package.metadata.build-package.external-files]]
```
bcressey, line 14: nit: missing `releases-url` section

sources/notation-image-verifier/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+	cmd := exec.Command("notation", "verify", *name)
+	output, err := cmd.CombinedOutput()
+
+	if err != nil {
+		fmt.Printf("image verification failed: %s\n", string(output))
+		os.Exit(1)
+	}
```
bcressey: Is there a concern that we're not validating the digest also? Should we try some approach where we append `@<digest>` to `name` if it doesn't already end in the digest? Or parse the output to see if it indicates the same digest?

packages/notation/notation-tmpfiles.conf
```diff
@@ -0,0 +1,2 @@
+d /var/cache/notation - - - -
```
bcressey: What ends up in the cache? If it's not vital, we can clean it up with an `R` directive on each boot before creating it. One less bit of state to worry about across upgrades and downgrades.

packages/notation-image-verifier/notation-image-verifier.spec
```diff
@@ -0,0 +1,35 @@
+%global _cross_first_party 1
+%global workspace_name notation-image-verifier
+
+Name: %{_cross_os}%{workspace_name}
+Version: 0.1.0
+Release: 1%{?dist}
+Summary: Binary for containerd image verification plugin
+License: Apache-2.0 OR MIT
+URL: https://github.com/bottlerocket-os/bottlerocket
+BuildRequires: %{_cross_os}glibc-devel
```
bcressey, line 10: This package should probably require the notation CLI and perhaps the signer plugin also.

packages/os/os.spec
```diff
@@ -83,6 +82,7 @@ Requires: %{_cross_os}corndog
...
```
bcressey: I'd probably move this Requires to containerd-2.1:
1. It's the only one configured to use it.
2. We're a bit short on space for older variants.

packages/notation/notation-trust-policy-json
ginglis13: `notation` will fail with an empty policy. IMO there's no need - we're already making it a requirement to provide a valid trustpolicy to use the notation image verifier.

The alternative would be we provide a default trustpolicy like

```
{
  "version": "1.0",
  "trustPolicies": [
    {
      "name": "default",
      "registryScopes": ["*"],
      "signatureVerification": {
        "level": "skip"
      }
    }
  ]
}
```

that would essentially be a no-op for signature verification, WDYT?

packages/notation/notation-tmpfiles.conf
ginglis13: The cache has Certificate Revocation List hashes: https://notaryproject.dev/docs/user-guides/how-to/directory-structure/#structure

I'll add the `R` directive 👍

packages/ecr-credential-helper/ecr-credential-helper.spec
```diff
@@ -0,0 +1,81 @@
...
+install -d %{buildroot}/root/.docker
+install -m 0644 %{S:2} %{buildroot}/root/.docker/config.json
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
+%{_cross_attribution_vendor_dir}
+%dir /root/.docker
+/root/.docker/config.json
```
ginglis13: Working through the initial feedback first, but I'm going to look deeper into notation's config.json - might be a more appropriate place to do this https://notaryproject.dev/docs/user-guides/how-to/directory-structure/#general-configuration

sources/notation-image-verifier/main.go
ginglis13: It captures stdout for logging - stdout is where it expects the "reason" for image verification rejection: https://github.com/containerd/containerd/blob/main/docs/image-verification.md#image-pull-judgement

sources/notation-image-verifier/main.go
ginglis13: good call, I'm going to update to the absolute path

I'll keep this unresolved as I keep getting through the lower hanging fruit feedback - I'll have to consider the FIPS case for that path.

sources/notation-image-verifier/main.go
ginglis13: I'll update this to instead construct the image reference we pass to `notation` like `<uri>:<digest>`. the `notation` CLI does warn against using named tags, since tags can be immutable. I'll have to parse the incoming `name` from containerd a bit to construct this.

I'll also pull some of this logic out of `main` and add some unit testing.

packages/notation-image-verifier/notation-image-verifier.spec
ginglis13: Since `notation-image-verifier` doesn't pull in any 3rd party modules, there are no dependencies to vendor. the `cross_scan_attribution` macro need a vendor directory to process. We could create an empty one and produce an empty attribution, but I'm just going to omit that instead.

ginglis13:
^ force push addresses feedback, the more notable changes being:

* move the aws-signer-notation-plugin and notation-image-verifier binaries to libexecdir
* notation-image-verifier now constructs an image reference from the incoming image sha to pass to notation

sources/notation-image-verifier/main.go
ginglis13: NVM - I'll point to `/usr/bin/notation`, we have an overlay that already handles this :D https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/39a939781bd126aeaf3ef7e3b3185f6c18a8b439/packages/release/usr-bin.mount.in#L11

ginglis13:
I've rerun manual tests on a custom `aws-k8s-1.34` variant along with the latest in https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/106. 

User Data:
```
[settings.image-verifier-plugins.notation]
trustpolicy = "ewogIC..."

[settings.image-verifier-plugins]
enabled = true
```

Successfully pulled my image in private registry, failed to pull `docker.io/library/hello-world:latest` as expected

packages/containerd-2.1/containerd-2.1.spec
```diff
@@ -38,13 +39,13 @@ Source1000: clarify.toml
...
 # Patch to support moving from containerd-1.7 to 2.x
 Patch1001: 1001-Revert-Don-t-allow-io_uring-related-syscalls-in-the-.patch
-Patch1002: 1002-Ensure-errContentRangeIgnored-error-when-range-get-r.patch
 
 BuildRequires: git
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}runc
 Requires: %{name}(optimized-gunzip)
 Requires: %{name}(binaries)
+Requires: %{_cross_os}notation-image-verifier
```
bcressey: If the only settings plugin that will include these new settings is `aws-ecs-3`, then we should omit this and just add it at the variant level. If you're planning to include everywhere containerd 2.1 is included, then this is fine.

Maybe these new files should go into the `notation-image-verifier` package? It doesn't seem like `containerd-2.1` has to do anything special.

packages/notation-image-verifier/notation-image-verifier.spec
```diff
@@ -0,0 +1,66 @@
...
+cp -r %{_builddir}/sources/%{workspace_name}/* .
+
+%build
+%set_cross_go_flags
+go build -ldflags="${GOLDFLAGS}" -o notation-image-verifier .
+gofips build -ldflags="${GOLDFLAGS}" -o fips/notation-image-verifier .
+
+%install
+install -d %{buildroot}%{_cross_libexecdir}/image-verifiers
+install -d %{buildroot}%{_cross_libexecdir}/image-verifiers/bin
```
bcressey: nit: can condense into one line:

```suggestion
install -d %{buildroot}%{_cross_libexecdir}/image-verifiers/bin
```

(same with fips)

packages/notation/notation-tmpfiles.conf
```diff
@@ -0,0 +1,2 @@
+R /var/cache/notation - - - -
```
bcressey, line 1: Does this directory get automatically created?

sources/notation-image-verifier/main.go
```diff
@@ -0,0 +1,73 @@
...
+func constructImageReference(name, digest string) string {
+	// Remove tag if present (everything after the last colon that's not part of a port)
+	uri := name
+	if lastColon := strings.LastIndex(name, ":"); lastColon != -1 {
+		// Check if this colon is part of a hostname:port (contains slash after colon)
+		if !strings.Contains(name[lastColon:], "/") {
+			uri = name[:lastColon]
+		}
+	}
+	return uri + "@" + digest
```
bcressey, line 40: With one of my own images:
```
# Ideally we are checking a signature for this digest.
❯ crane digest 760038019737.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-sdk:v0.665.0
sha256:4af3f570d6d60ca101f0aa05b927377191ac256a2397759cc64a64c8db325e48

# And indeed that seems to exist:
❯ crane digest 760038019737.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-sdk:v0.665.0@sha256:4af3f570d6d60ca101f0aa05b927377191ac256a2397759cc64a64c8db325e48
sha256:4af3f570d6d60ca101f0aa05b927377191ac256a2397759cc64a64c8db325e

# But I have an older image in the same repository:
❯ crane digest 760038019737.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-sdk:v0.60.0
sha256:3fb0141c578f86d3ae5143f7d21bed41874045dfcf6cd6360e1331faf1c35896

# And I can use that ref to refer to the older image:
❯ crane digest 760038019737.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-sdk:v0.665.0@sha256:3fb0141c578f86d3ae5143f7d21bed41874045dfcf6cd6360e1331faf1c35896
sha256:3fb0141c578f86d3ae5143f7d21bed41874045dfcf6cd6360e1331faf1c35896
```

Is Notation better at handling this? I worry that it'd fall back to checking whether any image in the registry had a signature for the digest, instead of the tag.

If the only thing that's safe is to accept images without a tag and with the digest value at the end, then we should enforce that.

bcressey: Chatted with @ginglis13 offline. If we trust that containerd has already resolved the ref to a digest, then the alternate forms of the image ref don't really matter. We just have to construct a ref that refers unambiguously to the provided digest, then look for a matching signature.

packages/notation/notation-tmpfiles.conf
ginglis13: it gets created at runtime by notation: https://github.com/notaryproject/notation/blob/e49f07cd19d47fb5c79a0c56321c70bf13d525ef/internal/revocation/revocation.go#L34

sources/notation-image-verifier/main.go
ginglis13: Some manual testing and looking at containerd sources confirms that containerd will resolve the digest from the incoming ref first before invoking verifier binaries. The notation-image-verifier only uses the incoming resolved digest to verify signatures.

packages/containerd-2.1/containerd-2.1.spec
ginglis13: It's safest for now to pull this into the notation-image-verifier and include that in the aws-ecs-3 variant.

ginglis13:
^ last 2 force pushes handle the case of `registry/repo:tag@digest` as the input image name to the `notation-image-verifier` and package the containerd drop-in via the `notation-image-verifier` package. I've also dropped the changed to `containerd-2.1`, since those changes were just to pull in the verifier binary as a direct dependency and package the template file for the verification plugins containerd config drop-in.

ginglis13:
^ force push packages a `notation` config.json at `/etc/notation`, configuring ecr-credential-helper to be used. Also moves the notation package to depend on ecr-credential-helper.

packages/notation/notation.spec
```diff
@@ -0,0 +1,94 @@
...
+Source1: bundled-%{gorepo}-v%{gover}.tar.gz
+Source2: notation-trust-policy-json
+Source3: notation-tmpfiles.conf
+Source4: notation-config.json
+
+BuildRequires: %{_cross_os}glibc-devel
+Requires: %{name}(binaries)
+Requires: %{_cross_os}ecr-credential-helper
+
+Provides: %{_cross_os}notation
```
bcressey: nit: can cut this, it should be provided by the package's name:

```suggestion
```

bcressey (approved):

bcressey (approved):

arnaldo2792 (approved):

ginglis13:
^ force push:
* reverts back to using `/root/.docker` for configuring `"credsStore": "ecr-login"` - helpful when using the docker CLI directly on an ecs- variant.
* adds `secret_t` mounts for `/root/.docker` for the docker config and `/root/.ecr` for the ecr-credential-helper cache

bcressey (approved):

arnaldo2792 (approved):

PR #725, opened 2025-11-07
rdma-core: bump version to 60.0 and enable pci in hwloc

KCSesh:
**Description of changes:**
* rdma-core: bump version to 60.0
* Also enable pci in hwloc

**Testing done:**
* Performed confirmation tests to ensure EKS compatibility.
* Verified booted AMI

**Terms of contribution:**

### DIFF
--- a/packages/hwloc/hwloc.spec
+++ b/packages/hwloc/hwloc.spec
@@ -36,7 +36,6 @@ Requires: %{_cross_os}systemd-devel
     --disable-gl \
     --disable-libxml2 \
     --disable-opencl \
-    --disable-pci \
     --disable-plugins \
     --exec-prefix=%{_cross_prefix} \
     --program-prefix=""
--- a/packages/rdma-core/rdma-core.spec
+++ b/packages/rdma-core/rdma-core.spec
@@ -1,7 +1,7 @@
-%global abiver 57
+%global abiver 59
 
 Name: %{_cross_os}rdma-core
-Version: 57.0
+Version: 60.0
 Release: 1%{?dist}
 Summary: RDMA core userspace infrastructure, including core libraries and util programs.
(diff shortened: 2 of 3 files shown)


packages/rdma-core/rdma-core.spec
```diff
@@ -1,7 +1,7 @@
-%global abiver 57
+%global abiver 59
```
piyush-jena, line 1: Should this be 60?

KCSesh: No, it should be 59:

https://github.com/linux-rdma/rdma-core/blob/stable-v60/CMakeLists.txt#L82-L86

piyush-jena (approved):

vigh-m (approved):

piyush-jena (approved):

packages/hwloc/hwloc.spec
```diff
@@ -36,7 +36,6 @@ Requires: %{_cross_os}systemd-devel
     --disable-gl \
     --disable-libxml2 \
     --disable-opencl \
-    --disable-pci \
```
bcressey, line 39: nit: if this is the full set of features, and we're moving one from disabled to enabled, we can add a `--enable-pci` if that's supported.

bcressey (approved):

yeazelm (approved):

packages/hwloc/hwloc.spec
KCSesh: Seems no `--enable-pci` exists. 

I do notice:

```
  --enable-32bits-pci-domain
                          Enable 32 bits PCI domains (domains > 16bits are
                          ignored by default). WARNING: This breaks the
                          library ABI, don't enable unless really needed.
```

But that is not needed as of now as far as I can tell.

PR #728, opened 2025-11-07
nvidia package updates

piyush-jena:
**Description of changes:**
* update `libnvidia-container` to v1.18.0
  * Remove the allow `clock_gettime` syscall patch because it has now been upstreamed and released.
* update `nvidia-container-toolkit` to v1.18.0
  * The `gofips` build is forced by `bottlerocket-sdk` due to `nvidia-ctk` binary's indirect dependency on `crypto/tls`. The direct dependency that indirectly forces this `crypto/tls` dependency is `github.com/urfave/cli-altsrc/v3`. Check  https://github.com/NVIDIA/nvidia-container-toolkit/blame/754cfa7480d9cef7dc956f30ef25112af44dc788/cmd/nvidia-ctk/cdi/generate/config.go#L25. 
* update `nvidia-k8s-device-plugin` to v0.18.0
  * Add a patch to disable `enable-cuda-compat` oci hook - This oci hook was the cause of a High Severity CVE (CVE-2025-23266). We add this patch because there's no way to disable this oci hook. For more details visit - https://www.wiz.io/blog/nvidia-ai-vulnerability-cve-2025-23266-nvidiascape. Enabling `features.disable-cuda-compat-lib-hook` flag in `nvidia-container-toolkit`, which is the recommended mitigation of this CVE doesn't work.

**Testing done:**
1. `nvidia-smoke-test` passes
2. Made sure that the `enable-cuda-compat` oci hook is not present.

**Terms of contribution:**


packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
```diff
@@ -95,6 +98,7 @@ install -m 0644 %{S:7} %{buildroot}%{_cross_unitdir}/
 %{_cross_attribution_file}
 %{_cross_bindir}/nvidia-container-runtime-hook
 %{_cross_bindir}/nvidia-ctk
+%{_cross_fips_bindir}/nvidia-ctk
```
arnaldo2792: This is not correct, you have to create a `fips-bin` package. You will have to restructure slightly how the binaries in the package are provided. @koooosh has done this before so he might be able to help you.

arnaldo2792 (changes requested):

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
koooosh: Here's [the commit in `amazon-ssm-agent`](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/22/commits/3073ff800939ba9e6c1d444c7e42d04e9eebdfea) where I did something similar. LMK if you have further q's @piyush-jena !

PR #729, opened 2025-11-08
ecr-credential-provider: patch to support AWS EUSC

sam-berning:
**Issue number:**

Closes #

**Description of changes:**

Patches ecr-credential-provider in order to support a new AWS partition.

**Testing done:**

Built ecr-credential-provider with the patches, and confirmed I could get credentials for a registry in the new partition:

```
echo '{"kind":"CredentialProviderRequest","apiVersion":"credentialprovider.kubelet.k8s.io/v1","image":"123456789012.dkr.ecr.eusc-de-east-1.amazonaws.eu"}' | ./ecr-credential-provider
```

Also built the core-kit and an AMI, and confirmed the same worked on the instance.

**Terms of contribution:**


vigh-m:
Please add a `Signed-off-by` tag to all your patches. [Example](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/c078d1ba8c49fdecc191acee71092f1cf84560dc/packages/systemd-257/9017-meson-set-DOPENSSL_NO_UI_CONSOLE-when-using-openssl.patch#L6)

piyush-jena (approved):

bcressey (approved):

PR #734, opened 2025-11-11
pluto: improve cluster DNS IP generation

jpculp:
**Description of changes:**

Rather than assuming IPv4, match the `ipFamily` in the EKS cluster's `kubernetesNetworkConfig`.

**Testing done:**

- Launched Bottlerocket AMI attached to IPv6 cluster.
- Launched Bottlerocket AMI attached to IPv4 cluster.

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/main.rs
+++ b/sources/api/pluto/src/main.rs
@@ -37,6 +37,7 @@ mod ec2;
 mod eks;
 
 use api::{settings_view_get, settings_view_set, SettingsViewDelta};
+use aws_sdk_eks::types::IpFamily;
 use aws_smithy_experimental::hyper_1_0::CryptoMode;
 use base64::Engine;
 use bottlerocket_modeled_types::{KubernetesClusterDnsIp, KubernetesHostnameOverrideSource};
@@ -44,7 +45,7 @@ use imdsclient::ImdsClient;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::fs::File;
 use std::io::{BufRead, BufReader, Write};
-use std::net::IpAddr;
+use std::net::{IpAddr, Ipv6Addr};
 use std::path::Path;
 use std::str::FromStr;
 use std::string::String;
@@ -220,10 +221,10 @@ async fn get_max_pods_from_file(instance_type: &str, pods_file: &'static str) ->
 
 /// Returns the cluster's DNS address.
 ///
-/// For IPv4 clusters, first it attempts to call EKS describe-cluster to find the `serviceIpv4Cidr`.
+/// First attempts to call EKS describe-cluster to find the `ipFamily` and its corresponding CIDR.
 /// If that works, it returns the expected cluster DNS IP address which is obtained by substituting
 /// `10` for the last octet. If the EKS call is not successful, it falls back to using IMDS MAC CIDR
-/// blocks to return one of two default addresses.
+/// blocks to return one of two default IPv4 addresses.
 async fn generate_cluster_dns_ip(
     client: &mut ImdsClient,
     aws_k8s_info: &mut SettingsViewDelta,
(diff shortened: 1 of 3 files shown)


Issacwww (approved):

sources/api/pluto/src/main.rs
cbgbt, line 274: The nested if let statements in the match arms make this a bit dense. Consider flattening the logic using and_then:

```suggestion
    // Derive cluster-dns-ip from service CIDR based on IP family
    match config.ip_family {
        Some(IpFamily::Ipv6) => config
            .service_ipv6_cidr
            .and_then(|cidr| get_dns_from_ipv6_cidr(&cidr).ok()),
        _ => config
            .service_ipv4_cidr
            .and_then(|cidr| get_dns_from_ipv4_cidr(&cidr).ok()),
    }
    .map(Some)
    .unwrap_or(Ok(None))
```

sources/api/pluto/src/main.rs
```diff
@@ -557,6 +587,21 @@ mod test {
         assert!(result.is_err());
     }
 
+    #[test]
```
cbgbt: Given the issue with compressed IPv6 addresses, can you add more test cases? I really like the [test-case crate](https://crates.io/crates/test_case) because it lets you write test case data in a way that still preserves pleasant output from `cargo test`. We use this dependency elsewhere in Bottlerocket.

```rust
#[test_case("fd00::/108", "fd00::a" ; "compressed ULA - common in EKS")]
#[test_case("2001:db8::/108", "2001:db8::a" ; "compressed middle")]
#[test_case("2001:db8:1234::/108", "2001:db8:1234::a" ; "partially expanded")]
#[test_case("2001:db8:1234:5678::/108", "2001:db8:1234:5678::a" ; "fully expanded")]
#[test_case("fe80::/108", "fe80::a" ; "link local")]
fn test_get_dns_from_ipv6_cidr_variants(input: &str, expected: &str) {
    let actual = get_dns_from_ipv6_cidr(input).unwrap();
    assert_eq!(expected, actual);
}

#[test_case("not-an-ipv6" ; "no slash")]
#[test_case("" ; "empty string")]
fn test_get_dns_from_ipv6_cidr_errors(input: &str) {
    let result = get_dns_from_ipv6_cidr(input);
    assert!(result.is_err());
}
```

sources/api/pluto/src/main.rs
```diff
@@ -296,6 +308,24 @@ fn get_dns_from_ipv4_cidr(cidr: &str) -> Result<String> {
     split[3] = "10";
     Ok(split.join("."))
 }
+/// Replicates [this] logic from the EKS AMI:
+///
+/// ```sh
+/// DNS_CLUSTER_IP=$(awk -F/ '{print $1}' <<< $SERVICE_IPV6_CIDR)a
+/// ```
+/// [this]: https://github.com/awslabs/amazon-eks-ami/blob/0f5c129/templates/al2/runtime/bootstrap.sh#L463
+fn get_dns_from_ipv6_cidr(cidr: &str) -> Result<String> {
```
cbgbt, line 316: This won't work with compressed IPv6 notation. AWS EKS returns compressed ULA addresses like `fd00::/108`, which would only have 3 components when split by `:`.

Looking at the [referenced EKS AMI code](https://github.com/awslabs/amazon-eks-ami/blob/0f5c129/templates/al2/runtime/bootstrap.sh#L463), the bash implementation is:

```bash
DNS_CLUSTER_IP=$(awk -F/ '{print $1}' <<< $SERVICE_IPV6_CIDR)a
```

This splits on `/`, takes the first part, and appends `a`, which will work with compressed addresses regardless of the number of segments they have.

We could use a proper IPv6 parser here, since we have the benefit of not using bash:

```rust
use std::net::Ipv6Addr;

fn get_dns_from_ipv6_cidr(cidr: &str) -> Result<String> {
    // Extract the address part before the slash
    let addr_str = cidr.split('/').next().context(error::CidrParseSnafu {
        cidr,
        reason: "missing address component",
    })?;

    let base_addr: Ipv6Addr = addr_str.parse().context(error::BadIpSnafu {
        ip: addr_str,
    })?;

    // Set the last segment to 0xa (10 in hex)
    let mut segments = base_addr.segments();
    segments[7] = 0xa;

    Ok(Ipv6Addr::from(segments).to_string())
}
```

jpculp:
Addressed feedback from @cbgbt and re-ran tests.

sources/api/pluto/src/main.rs
```diff
@@ -609,6 +655,43 @@ mod test {
...
+        }
+
+        if let Some(cluster_name) = cluster_name {
+            settings_view_set!(settings.kubernetes.cluster_name = cluster_name.try_into().unwrap());
+        }
+
+        // Since we can't easily mock the EKS service in unit tests, we test the logic paths
+        // The function will attempt to call EKS when both region and cluster_name are present,
+        // but will fail gracefully in test environment and return Ok(None)
+        let result = get_eks_network_config(&settings).await;
```
cbgbt: We shouldn't make network API calls in tests, even if we know they will fail. This limitation also makes it impossible for us to test the positive case, and only allows us to test the negative.

It's not easy, but we should either:
* Modify the logic to make mocking possible (Rust's `mockall` crate is excellent, but you will need to introduce traits)
* Find a different way to isolate the compute logic from the IO, and then only test the logic

cbgbt: Here's an example of how you could improve the unit tests: https://gist.github.com/cbgbt/770870b6f292f9e60b21e6913ebcf259

jpculp:
Moved DNS block into its own function and simplified testing based on @cbgbt's suggestion.

cbgbt (approved):

piyush-jena (approved):

bcressey (approved):

PR #748, opened 2025-11-14
Amd device plugin

mgsharm:
**Description of changes:**

This PR adds support for AMD GPU detection and device plugin functionality to Bottlerocket:

  - **libdrm package**: Added libdrm 2.4.129 as one of the device plugin dependencies. Reference:
  https://github.com/ROCm/k8s-device-plugin/blob/763445e18f3838fa72b22e31a04ec25987334bff/Dockerfile#L15
  - **rocm-k8s-device-plugin package**: Added AMD K8s device plugin v1.31.0.8 from ROCm for GPU resource management

  **Dependencies**

  - [link-to-kernel-kit-pr ](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/320)- Adds AMD GPU DKMS driver and detection services
  
  ## Testing done

  ### AMD GPU Instance (MI300X with 8 GPUs)

 PCI device detection:

  ```bash
  bash-5.1# lspci -d 1002:75a3
  51:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
  52:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
  62:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
  63:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
  73:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
  74:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
  84:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
  85:00.0 Processing accelerators: Advanced Micro Devices, Inc. [AMD/ATI] Device 75a3
```

ROCm device plugin running:

```bash
bash-5.1# systemctl list-units | grep -i rocm
  rocm-k8s-device-plugin.service                                                                                                                loaded active     
running   Start ROCm kubernetes device plugin
```

```bash
bash-5.1# systemctl status rocm-k8s-device-plugin.service
● rocm-k8s-device-plugin.service - Start ROCm kubernetes device plugin
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/rocm-k8s-device-plugin.service; enabled; preset: enabled)
             └─00-aws-config.conf
     Active: active (running) since Wed 2025-11-26 06:22:34 UTC; 2min 47s ago
 Invocation: ae2b73174a7245b5976c1b9eb7835695
    Process: 47801 ExecStartPre=/usr/bin/sleep 0.1 (code=exited, status=0/SUCCESS)
    Process: 47823 ExecStartPre=/usr/bin/test -S /var/lib/kubelet/device-plugins/kubelet.sock (code=exited, status=0/SUCCESS)
   Main PID: 47825 (rocm-device-plu)
      Tasks: 12 (limit: 629145)
     Memory: 26.9M (peak: 28M)
        CPU: 63ms
     CGroup: /system.slice/rocm-k8s-device-plugin.service
             └─47825 /usr/bin/rocm-device-plugin -logtostderr=true -stderrthreshold=INFO -v=5

Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799514   47825 amdgpu.go:261] Devices map: map[0000:51:00.0:map[card:0 computePartitionType:spx devID:10833827524210284120 memoryPartitionType:nps1 nodeId:2 numaNode:0 renderD:128] 0000:52:00.0:map[card:8 computePartitionType:spx devID:5789222631865278233 memoryPartitionType:nps1 nodeId:3 numaNode:0 renderD:136] 0000:62:00.0:map[card:16 computePartitionType:spx devID:1537282919108789315 memoryPartitionType:nps1 nodeId:4 numaNode:0 renderD:144] 0000:63:00.0:map[card:24 computePartitionType:spx devID:17650589009955793669 memoryPartitionType:nps1 nodeId:5 numaNode:0 renderD:152] 0000:73:00.0:map[card:32 computePartitionType:spx devID:17599352850648317852 memoryPartitionType:nps1 nodeId:6 numaNode:1 renderD:160] 0000:74:00.0:map[card:40 computePartitionType:spx devID:16824009056103876010 memoryPartitionType:nps1 nodeId:7 numaNode:1 renderD:168] 0000:84:00.0:map[card:48 computePartitionType:spx devID:7459436697213601231 memoryPartitionType:nps1 nodeId:8 numaNode:1 renderD:176] 0000:85:00.0:map[card:56 computePartitionType:spx devID:10369896732957786260 memoryPartitionType:nps1 nodeId:9 numaNode:1 renderD:184]]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799540   47825 amdgpu.go:278] Partition counts: map[spx_nps1:8]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799548   47825 plugin.go:254] Watching GPU with bus ID: 0000:52:00.0 NUMA Node: [0]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799558   47825 plugin.go:254] Watching GPU with bus ID: 0000:62:00.0 NUMA Node: [0]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799561   47825 plugin.go:254] Watching GPU with bus ID: 0000:63:00.0 NUMA Node: [0]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799564   47825 plugin.go:254] Watching GPU with bus ID: 0000:73:00.0 NUMA Node: [1]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799567   47825 plugin.go:254] Watching GPU with bus ID: 0000:74:00.0 NUMA Node: [1]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799570   47825 plugin.go:254] Watching GPU with bus ID: 0000:84:00.0 NUMA Node: [1]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799572   47825 plugin.go:254] Watching GPU with bus ID: 0000:85:00.0 NUMA Node: [1]
Nov 26 06:22:34 ip-172-31-60-25.us-west-2.compute.internal rocm-device-plugin[47825]: I1126 06:22:34.799575   47825 plugin.go:254] Watching GPU with bus ID: 0000:51:00.0 NUMA Node: [0]
```

GPU capacity advertised to Kubernetes:

```bash
  ➜ kubectl get nodes -o json | jq '.items[].status.capacity'
  {
    "amd.com/gpu": "8",
    "cpu": "192",
    "ephemeral-storage": "15471392Ki",
    "hugepages-1Gi": "0",
    "hugepages-2Mi": "0",
    "memory": "4206782968Ki",
    "pods": "110"
  }
```

All services and targets that `rocm-k8s-device-plugin.service` depends on:

```bash
bash-5.1#   systemctl list-dependencies rocm-k8s-device-plugin.service
rocm-k8s-device-plugin.service
● ├─kubelet.service
● ├─modprobe@amdgpu.service
● ├─system.slice
● └─sysinit.target
●   ├─dev-hugepages.mount
●   ├─dev-mqueue.mount
●   ├─kmod-static-nodes.service
●   ├─ldconfig.service
●   ├─proc-sys-fs-binfmt_misc.mount
●   ├─sys-fs-fuse-connections.mount
●   ├─sys-kernel-config.mount
●   ├─sys-kernel-debug.mount
●   ├─sys-kernel-tracing.mount
●   ├─systemd-journal-catalog-update.service
●   ├─systemd-journal-flush.service
●   ├─systemd-journald.service
●   ├─systemd-machine-id-commit.service
●   ├─systemd-modules-load.service
●   ├─systemd-network-generator.service
○   ├─systemd-pstore.service
●   ├─systemd-random-seed.service
○   ├─systemd-repart.service
●   ├─systemd-resolved.service
●   ├─systemd-sysctl.service
●   ├─systemd-sysusers.service
●   ├─systemd-tmpfiles-setup-dev-early.service
●   ├─systemd-tmpfiles-setup-dev.service
●   ├─systemd-tmpfiles-setup.service
●   ├─systemd-udev-load-credentials.service
●   ├─systemd-udev-trigger.service
●   ├─systemd-udevd.service
●   ├─local-fs.target
●   │ ├─\x2ebottlerocket.mount
●   │ ├─has-boot-ever-succeeded.service
●   │ ├─local.mount
●   │ ├─mnt.mount
●   │ ├─opt.mount
●   │ ├─prepare-local-fs.service
●   │ ├─prepare-opt.service
●   │ ├─prepare-var-lib-containerd.service
●   │ ├─prepare-var-lib-kubelet.service
●   │ ├─prepare-var.service
○   │ ├─repart-data-fallback.service
○   │ ├─repart-data-preferred.service
●   │ ├─repart-local.service
●   │ ├─selinux-policy-files.service
●   │ ├─systemd-remount-fs.service
●   │ ├─tmp.mount
●   │ ├─var.mount
●   │ └─x86_64\x2dbottlerocket\x2dlinux\x2dgnu-sys\x2droot-usr-lib-modules.mount
●   └─swap.target
```

All services that depend on modprobe@amdgpu.service :

```bash
bash-5.1#   systemctl list-dependencies --reverse modprobe@amdgpu.service
modprobe@amdgpu.service
● └─rocm-k8s-device-plugin.service
```

Terms of contribution:


packages/amd-k8s-device-plugin/amd-k8s-device-plugin.spec
```diff
@@ -0,0 +1,75 @@
+%global goproject github.com/ROCm
+%global gorepo k8s-device-plugin
+%global goimport %{goproject}/%{gorepo}
+
+%global gover 1.31.0.8
+%global rpmver %{gover}
+
+Name: %{_cross_os}amd-k8s-device-plugin
```
yeazelm: Let's keep this consistent with the upstream naming 
```suggestion
Name: %{_cross_os}rocm-k8s-device-plugin
```

packages/rocm-k8s-device-plugin/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/ROCm/k8s-device-plugin/releases"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/ROCm/k8s-device-plugin/archive/v1.31.0.8.tar.gz"
```
yeazelm, line 15: This is marked pre-release, is there a "latest" release or all of them just marked pre-release?

packages/amd-k8s-device-plugin/amd-k8s-device-plugin.service
```diff
@@ -0,0 +1,24 @@
...
+ConditionPathExists=/usr/bin/ghostdog
+
+[Service]
+# Verify AMD GPU is detected
+ExecStartPre=/usr/bin/ghostdog amd-gpu-present
+# Verify amdgpu driver is loaded
+ExecStartPre=/usr/bin/ghostdog match-driver amd rocm
+# Ensure that the kubelet device plugin socket exists before we start
+# A brief sleep is needed to avoid the `test` failing its first check
+ExecStartPre=/usr/bin/sleep 0.1
```
yeazelm: This is a sign your dependencies are incorrectly wired up. Do you actually need kubelet.service or the socket, if you need the socket, let's make that the dependency of this unit.

packages/libdrm/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://dri.freedesktop.org/libdrm/"
+
+[[package.metadata.build-package.external-files]]
+url = "https://dri.freedesktop.org/libdrm/libdrm-2.4.123.tar.xz"
```
yeazelm: Looks like latest is 2.4.128

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,61 @@
...
+
+%cross_meson "${CONFIGURE_OPTS[@]}"
+%cross_meson_build
+
+%install
+%cross_meson_install
+
+%files
+%{_cross_attribution_file}
+%{_cross_libdir}/*.so.*
```
yeazelm: Can we list these out?

packages/amd-k8s-device-plugin/amd-k8s-device-plugin.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Start AMD kubernetes device plugin
+After=kubelet.service
+Wants=kubelet.service
+ConditionPathExists=/usr/bin/ghostdog
+
+[Service]
+# Verify AMD GPU is detected
+ExecStartPre=/usr/bin/ghostdog amd-gpu-present
```
yeazelm: The way this works for the other device plugin we have is that preconfigured.target is blocked if the driver is not loaded, that way the kubelet never loads, and the device plugin doesn't either. In this case, you could have the situation where the driver fails to load, kubelet starts, becomes ready, but this fails, leaving the node in a degraded state. I'd prefer we dealt with GPU detection and driver loading as a separate unit. The assumption should be we can't make it to kubelet starting without the drivers being loaded.

sources/ghostdog/src/main.rs
```diff
@@ -412,6 +430,35 @@ fn match_neuron_driver(driver_flavor: &str) -> Result<()> {
     }
 }
 
+fn match_amd_driver(driver_flavor: &str) -> Result<()> {
```
yeazelm: I don't think we should mix semantics here: match-driver is for matching a chosen driver with the hardware present, in this case you are having it check if the driver loaded. Would it not be easier to just use a path check in systemd if we want `sys/class/kfd` to exist before proceeding? I don't think we want `ghostdog match-driver` concerning itself with an already loaded driver, its about providing a boolean answer to if that driver choice makes sense for the hardware present.

packages/rocm-k8s-device-plugin/Cargo.toml
mgsharm: All releases of the ROCm k8s-device-plugin (https://github.com/ROCm/k8s-device-plugin/releases?page=1) are marked as pre-release on their GitHub. v1.31.0.8 is the latest available release.

packages/amd-k8s-device-plugin/amd-k8s-device-plugin.service
mgsharm: Should I create a `load-amd-kernel-modules.service` in the kernel-kit (similar to how NVIDIA does it) that blocks
  `preconfigured.target` if the driver isn't loaded? And then simplify the device plugin service in core-kit to just check for the kubelet socket, since we'd be guaranteed the driver is loaded by that point?

  Also, for `ghostdog match-driver amd` - since AMD only has one driver type, should I just remove that entirely and use systemd's `ConditionPathExists` instead?

packages/amd-k8s-device-plugin/amd-k8s-device-plugin.spec
mgsharm: Makes sense. Will update.

packages/libdrm/Cargo.toml
mgsharm: Will update.

packages/amd-k8s-device-plugin/amd-k8s-device-plugin.service
mgsharm: Discussed offline with @yeazelm - here's the plan:

  Following the Neuron pattern, I'll create:
  1. `modprobe@amdgpu.service.drop-in.conf` - adds conditions to check AMD GPU presence
  2. `load-amd-modules.service` - loads AMD kernel modules, `RequiredBy=drivers.target`

These will be added to the kernel package (similar to how Neuron does it in `kernel-6.12.spec`), so driver loading happens before `drivers.target` completes. This ensures kubelet never starts if the AMD driver fails to load.

Will also simplify the device plugin service in core-kit to remove driver checks since they'll be guaranteed to be loaded by that point.

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,70 @@
...
+%{_cross_libdir}/libdrm_amdgpu.so.1.128.0
+%{_cross_libdir}/libdrm_radeon.so.1
+%{_cross_libdir}/libdrm_radeon.so.1.128.0
+%{_cross_includedir}/libsync.h
+%{_cross_datadir}/libdrm/amdgpu.ids
+
+%files devel
+%{_cross_libdir}/libdrm.so
+%{_cross_libdir}/libdrm_amdgpu.so
+%{_cross_libdir}/libdrm_radeon.so
```
arnaldo2792: All these files must be shipped in the main package.

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,70 @@
...
+%{_cross_includedir}/libdrm/
+%{_cross_includedir}/xf86drm.h
+%{_cross_includedir}/xf86drmMode.h
```
arnaldo2792: You could just `*` all of `includedir`.

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,70 @@
...
+%{_cross_pkgconfigdir}/libdrm.pc
+%{_cross_pkgconfigdir}/libdrm_amdgpu.pc
+%{_cross_pkgconfigdir}/libdrm_radeon.pc
```
arnaldo2792: The same with `pkgconfigdir` with the wildcard

Cargo.toml
```diff
@@ -66,6 +66,7 @@ members = [
   "packages/libcap",
```
arnaldo2792, line 66: For the commit message, could you please follow what was done for other commits in this repo?

packages/rocm-k8s-device-plugin/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+sha512 = "23a127b46ad15cabbdd9abe18a8b75140340dcb10f41c9efdcfd30b38db7142edabaf6a97ed2c621f0414c78c0cb87fdc81a30b5e3fb016d16cbd11210143326"
+force-upstream = true
```
arnaldo2792: You shouldn't add this:

```suggestion
```

packages/rocm-k8s-device-plugin/rocm-k8s-device-plugin.service
```diff
@@ -0,0 +1,19 @@
...
+Description=Start ROCm kubernetes device plugin
+After=kubelet.service
+Wants=kubelet.service
+
+[Service]
+# Ensure that the kubelet device plugin socket exists before we start
+# A brief sleep is needed to avoid the `test` failing its first check
+ExecStartPre=/usr/bin/sleep 0.1
+ExecStartPre=/usr/bin/test -S /var/lib/kubelet/device-plugins/kubelet.sock
+ExecStart=/usr/bin/rocm-device-plugin -logtostderr=true -stderrthreshold=INFO -v=5
```
arnaldo2792, line 11: For other reviewers, all these flags are parsed automatically by Go's flag parsing library:

https://github.com/ROCm/k8s-device-plugin/blob/763445e18f3838fa72b22e31a04ec25987334bff/vendor/github.com/golang/glog/glog_flags.go#L397

packages/libdrm/libdrm.spec
mgsharm: These .so symlinks are required in the -devel package. When I removed them the build failed with "Installed (but unpackaged) file(s) found" because libdrm's build system creates these development symlinks automatically. The symlinks belong in -devel per standard conventions - they're needed for linking during compilation.

```bash
  7.774 error: Installed (but unpackaged) file(s) found:
  7.774    /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/libdrm.so
  7.774    /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/libdrm_amdgpu.so
  7.774    /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/libdrm_radeon.so
  7.774     Installed (but unpackaged) file(s) found:
  7.774    /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/libdrm.so
  7.774    /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/libdrm_amdgpu.so
  7.774    /aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/libdrm_radeon.so
  7.774
  7.774 RPM build errors:
```

Cargo.toml
mgsharm: Updated commit message.

packages/libdrm/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+releases-url = "https://dri.freedesktop.org/libdrm/"
+
+[[package.metadata.build-package.external-files]]
+url = "https://dri.freedesktop.org/libdrm/libdrm-2.4.128.tar.xz"
+sha512 = "b80e6be1c9d0427e1c3ffd018213d7230333f037498cf98819a8a6c50d923ad3472002044e010ca9dc646ef79dbca241bd47eaa992014cb7063b31cdb84037c7"
+force-upstream = true
```
yeazelm: Remove this

packages/libdrm/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[[package.metadata.build-package.external-files]]
+url = "https://dri.freedesktop.org/libdrm/libdrm-2.4.128.tar.xz.sig"
+sha512 = "985bb55a51d04cedde1368b9593cbb67a6ed0779f67aa3f8fdefb296be046ad7de22f9715bec4f8c96bf27de592810a4d686d0de13009ef11b5d1c1b7ac103cd"
+force-upstream = true
```
yeazelm: Remove this.

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,66 @@
...
+%cross_meson "${CONFIGURE_OPTS[@]}"
+%cross_meson_build
+
+%install
+%cross_meson_install
+
+%files
+%{_cross_attribution_file}
+%{_cross_libdir}/libdrm.so.2
+%{_cross_libdir}/libdrm.so.2.128.0
```
yeazelm: We can probably * after the 2 otherwise these lines will be updated with every bump: 
```suggestion
%{_cross_libdir}/libdrm.so.2.*
```

packages/libdrm/libdrm.spec
yeazelm: I think I agree with @arnaldo2792. These non-versioned links are for helping things find them if they don't have a version dependency listed. They might be in devel as well but runtime dependencies can also use these.

sources/ghostdog/src/main.rs
```diff
@@ -73,6 +74,11 @@ struct EfaPresentArgs {}
 /// Detect if Neuron devices are attached.
 struct NeuronPresentArgs {}
 
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "amd-gpu-present")]
```
yeazelm: nit:
```suggestion
#[argh(subcommand, name = "amdgpu-present")]
``` 
Nearly everything else we refer to is all one word and the other commands all keep it to `thing`-present.

sources/ghostdog/src/main.rs
```diff
@@ -163,6 +169,9 @@ fn main() -> Result<()> {
         SubCommand::NeuronPresent(_) => {
             is_neuron_attached()?;
         }
+        SubCommand::AmdGpuPresent(_) => {
```
yeazelm: What uses this code? Without seeing the use case, its hard to review since the context isn't present.

packages/rocm-k8s-device-plugin/rocm-k8s-device-plugin.service
```diff
@@ -0,0 +1,19 @@
...
+Type=simple
+TimeoutSec=0
+RestartSec=2
+Restart=always
+StandardError=journal+console
```
bcressey: We shouldn't send most errors to the console, it's only for services that can cause boot to fail:
```suggestion
```

packages/rocm-k8s-device-plugin/rocm-k8s-device-plugin.service
```diff
@@ -0,0 +1,19 @@
...
+After=kubelet.service
+Wants=kubelet.service
+
+[Service]
+# Ensure that the kubelet device plugin socket exists before we start
+# A brief sleep is needed to avoid the `test` failing its first check
+ExecStartPre=/usr/bin/sleep 0.1
+ExecStartPre=/usr/bin/test -S /var/lib/kubelet/device-plugins/kubelet.sock
+ExecStart=/usr/bin/rocm-device-plugin -logtostderr=true -stderrthreshold=INFO -v=5
+Type=simple
```
bcressey: nit: simple is default so you can leave it unspecified
```suggestion
```

packages/rocm-k8s-device-plugin/rocm-k8s-device-plugin.spec
```diff
@@ -0,0 +1,75 @@
...
+%description fips-bin
+%{summary}.
+
+%prep
+%autosetup -n %{gorepo}-%{gover} -p1
+%cross_go_setup %{gorepo}-%{gover} %{goproject} %{goimport}
+
+%build
+%cross_go_configure %{goimport}
+%set_cross_go_flags
```
bcressey: this is redundant with `%cross_go_configure`:
```suggestion
```

packages/libdrm/libdrm.spec
mgsharm: Yes, make sense.

packages/libdrm/libdrm.spec
mgsharm: Added .so files in both the packages.

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,63 @@
...
+
+%install
+%cross_meson_install
+
+%files
+%{_cross_attribution_file}
+%{_cross_libdir}/libdrm.so.*
+%{_cross_libdir}/libdrm_amdgpu.so.*
+%{_cross_libdir}/libdrm_radeon.so.*
+%{_cross_includedir}/libsync.h
```
bcressey: Header files belong in the devel package.

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,63 @@
...
+ -Dman-pages=disabled
+ -Dvalgrind=disabled
+ -Dfreedreno=disabled
+ -Dvc4=disabled
+ -Detnaviv=disabled
+ -Dexynos=disabled
+ -Dtegra=disabled
+ -Domap=disabled
+ -Dintel=disabled
+ -Dradeon=enabled
```
bcressey: Why enable radeon?

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,63 @@
...
+
+%package devel
+Summary: Files for development using the direct rendering manager library
+Requires: %{name}
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n libdrm-%{version} -p1
```
bcressey, line 26: Please add the GPG signature check.

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,63 @@
+Name: %{_cross_os}libdrm
+Version: 2.4.129
+Release: 1%{?dist}
+Summary: Direct Rendering Manager runtime library
```
bcressey: For consistency with the devel package summary:
```suggestion
Summary: Direct rendering manager library
```

packages/libdrm/libdrm.spec
```diff
@@ -0,0 +1,63 @@
...
+%{_cross_libdir}/libdrm.so.*
+%{_cross_libdir}/libdrm_amdgpu.so.*
+%{_cross_libdir}/libdrm_radeon.so.*
+%{_cross_includedir}/libsync.h
+%{_cross_datadir}/libdrm/amdgpu.ids
+
+%files devel
+%{_cross_libdir}/libdrm.so
+%{_cross_libdir}/libdrm_amdgpu.so
+%{_cross_libdir}/libdrm_radeon.so
```
bcressey: It's fine for `*.so` symlinks to just be in the "devel" package as a general rule.

If you inspect the libraries via `readelf` - always a good idea for new library packages - you'll something like:
```
❯ readelf -a /usr/lib64/libdrm_amdgpu.so.1 | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libdrm_amdgpu.so.1]
```

You can also run `readelf` on the plugin binary and you should see something like this:
```
❯ readelf -a <binary> | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libdrm_amdgpu.so.1]
```

packages/rocm-k8s-device-plugin/rocm-k8s-device-plugin.spec
```diff
@@ -0,0 +1,75 @@
...
+install -d %{buildroot}%{_cross_unitdir}
+install -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
+%{_cross_unitdir}/rocm-k8s-device-plugin.service
+
+%files bin
+%{_cross_bindir}/amd-device-plugin
```
yeazelm: We want these to be
```suggestion
%{_cross_bindir}/rocm-device-plugin
```

packages/libdrm/libdrm.spec
mgsharm: Makes sense.

packages/libdrm/libdrm.spec
mgsharm: Disabled it.

packages/rocm-k8s-device-plugin/rocm-k8s-device-plugin.service
```diff
@@ -0,0 +1,19 @@
...
+After=kubelet.service
+Wants=kubelet.service
+
+[Service]
+# Ensure that the kubelet device plugin socket exists before we start
+# A brief sleep is needed to avoid the `test` failing its first check
+ExecStartPre=/usr/bin/sleep 0.1
+ExecStartPre=/usr/bin/test -S /var/lib/kubelet/device-plugins/kubelet.sock
+# Ensure AMD GPU drivers are loaded before starting the device plugin
+ExecStartPre=/usr/bin/test -f /etc/.amdgpu-driver-loaded
```
yeazelm: This adds a dependency on https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/320/files which we should call out. We might need to coordinate the dependency here.

mgsharm: Its already called out in the description of this PR.

packages/rocm-k8s-device-plugin/Cargo.toml
bcressey: Please fix this:
```suggestion
```

packages/rocm-k8s-device-plugin/rocm-k8s-device-plugin.service
bcressey: Let's not create "hard" cross kit dependencies like these if we can avoid it. Especially not with a brittle mechanism like marker files in `/etc`.

It'd be better for the amdgpu kmod to influence the behavior of this unit via a drop-in, or by adding another unit that runs before it does.

bcressey (approved):

mgsharm:
Make rocm-k8s-device-plugin depend on modprobe@amdgpu.service to ensure AMD GPU driver loads first.

yeazelm (approved):

PR #754, opened 2025-11-24
docs: add boot process documentation

cbgbt:
Add documentation of Bottlerocket's boot sequence, including systemd target progression, service dependencies, and synchronization mechanisms.

Include keywords for semantic search discoverability.

**Terms of contribution:**

### DIFF
--- a/docs/boot-process.md
+++ b/docs/boot-process.md
@@ -1,0 +1,382 @@
+# Bottlerocket Boot Process
+
+This document describes how Bottlerocket boots, focusing on the systemd target progression and service dependencies.
+
+**Keywords:** boot, systemd, targets, preconfigured, configured, multi-user, fipscheck, drivers, sysinit, services, dependencies, ordering, API system, bootstrap containers, settings, startup, initialization, kernel modules
+
+## Overview
+
+Bottlerocket's boot sequence progresses through six main stages, each represented by a systemd target:
+
+```
+sysinit.target
+    ↓
+fipscheck.target (FIPS mode only)
+    ↓
+drivers.target (kernel module loading)
+    ↓
+preconfigured.target (API system initialization)
+    ↓
+configured.target (bootstrap containers)
+    ↓
+multi-user.target (workload services)
+```
+
+Each stage must complete before the next begins. Services use systemd dependencies (`After=`, `Requires=`, `Wants=`) to coordinate within and between stages.
+
+## Boot Stages
+
+### Stage 0: sysinit.target
+
+Standard systemd initialization. Most services implicitly depend on this through `DefaultDependencies=yes` (the default).
+
+Early services that need to run before normal dependency chains use `DefaultDependencies=no`:
+
+- Filesystem preparation (`prepare-var.service`, `prepare-boot.service`, etc.)
+- Data store migration (`migrator.service`)
+
+### Stage 1: fipscheck.target
+
+**Purpose:** Verify cryptographic module integrity when FIPS mode is enabled.
+
+**When it runs:** Only when the kernel command line includes `fips=1`.
+
+**Key services:**
+
+- `check-kernel-integrity.service` - Verifies kernel integrity
+- `check-fips-modules.service` - Loads and tests the `tcrypt` module
+  - Creates `/etc/.fips-module-check-passed` sentinel file on success
+  - Blocks boot if FIPS checks fail
+
+**Transition:** Completes before `drivers.target` begins.
+
+### Stage 2: drivers.target
+
+**Purpose:** Load kernel modules and hardware drivers.
+
+**When it runs:** Always runs, after `basic.target` and before `preconfigured.target`.
+
+**Key services:**
+
+- `load-neuron-inf1-modules.service` - Loads AWS Neuron Inf1 kernel modules
+- `load-neuron-latest-modules.service` - Loads AWS Neuron Latest kernel modules
+
+**Dependencies:**
+
+- Runs after `basic.target`
+- Runs before `preconfigured.target`
+- Required by `preconfigured.target` and `multi-user.target`
+
+**Note:** Some driver loading services (e.g., NVIDIA GPU drivers) are required by `preconfigured.target` directly rather than using `drivers.target`.
+
+**Transition:** Completes before `preconfigured.target` begins.
+
+### Stage 3: preconfigured.target
+
+**Purpose:** Initialize the API system and apply all boot-time configuration.
+
+**What "preconfigured" means:** The system has:
+
+- A populated data store with default and user-provided settings
+- A running API server
+- All configuration files generated from settings
+
+This is the most complex boot stage. Services run in a specific order to build up the system configuration:
+
+#### 3.1 Data Store Setup
+
+**migrator** (`migrator.service`)
+
+- **When:** Runs with `DefaultDependencies=no`, before everything else
+- **What:** Updates data store schema if the OS version changed
+- **Dependencies:** Required by `apiserver.service`, `storewolf.service`, and `preconfigured.target`
+
+**storewolf** (`storewolf.service`)
+
+- **When:** After `migrator.service`
+- **What:** Creates data store directories and populates default settings
+- **Details:**
+  - Reads defaults from variant-specific `defaults.d` directories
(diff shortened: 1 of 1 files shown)


docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+**Keywords:** boot, systemd, targets, preconfigured, configured, multi-user, fipscheck, drivers, sysinit, services, dependencies, ordering, API system, bootstrap containers, settings, startup, initialization, kernel modules
+
+## Overview
+
+Bottlerocket's boot sequence progresses through six main stages, each represented by a systemd target:
+
+```
+sysinit.target
+    ↓
+fipscheck.target (FIPS mode only)
```
bcressey, line 14: This isn't the correct order, but the reason why is subtle. There's a `bootconfig-fips.conf` snippet in `release` that sets the default systemd target to `fipscheck.target`.

Since all the units that are needed by that target are `DefaultDependencies=no`, `sysinit.target` doesn't end up added to the job queue until we reach `activate-preconfigured.target`.

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+This document describes how Bottlerocket boots, focusing on the systemd target progression and service dependencies.
+
+**Keywords:** boot, systemd, targets, preconfigured, configured, multi-user, fipscheck, drivers, sysinit, services, dependencies, ordering, API system, bootstrap containers, settings, startup, initialization, kernel modules
+
+## Overview
+
+Bottlerocket's boot sequence progresses through six main stages, each represented by a systemd target:
+
+```
+sysinit.target
```
bcressey, line 12: I would probably add `local-fs.target` as one of the most important prerequisites to `sysinit.target`.

Quite a lot of Bottlerocket-specific work happens for local storage setup 😀 while the `sysinit` phase is rather vanilla.

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+Early services that need to run before normal dependency chains use `DefaultDependencies=no`:
+
+- Filesystem preparation (`prepare-var.service`, `prepare-boot.service`, etc.)
+- Data store migration (`migrator.service`)
+
+### Stage 1: fipscheck.target
+
+**Purpose:** Verify cryptographic module integrity when FIPS mode is enabled.
+
+**When it runs:** Only when the kernel command line includes `fips=1`.
```
bcressey, line 42: Not correct because of the bootconfig override, which is also where `fips=1` comes from.

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+
+**Key services:**
+
+- `check-kernel-integrity.service` - Verifies kernel integrity
+- `check-fips-modules.service` - Loads and tests the `tcrypt` module
+  - Creates `/etc/.fips-module-check-passed` sentinel file on success
+  - Blocks boot if FIPS checks fail
+
+**Transition:** Completes before `drivers.target` begins.
```
bcressey, line 51: In terms of coordinated state transitions, we have these:
1. fipscheck -> preconfigured (optional)
1. preconfigured -> configured
2. configured -> multi-user

These are our "runlevels" or discrete stages. Targets don't work like runlevels, they just activate in response to something else causing them to be enqueued. `sysinit.target` is enqueued because services in `preconfigured.target` depend on it (because of default dependencies).

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+
+### Stage 2: drivers.target
+
+**Purpose:** Load kernel modules and hardware drivers.
+
+**When it runs:** Always runs, after `basic.target` and before `preconfigured.target`.
```
bcressey, line 57: It might be better to write this in terms of what "pulls" on various targets. Many of them are pulled in parallel (sysinit.target, drivers.target, network-online.target) by the units in preconfigured.

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+**sundog** (`sundog.service`)
+
+- **When:** After `network-online.target`, `apiserver.service`, `early-boot-config.service`
+- **What:** Generates dynamic settings that can't be determined until runtime
+- **Details:**
+  - Examples: primary IP address, cluster DNS settings
+  - Runs `settings-committer` first to access user data settings
+  - PATCHes generated settings to API in _pending_ state
+- **Dependencies:** Required by `preconfigured.target`
+- **Subcomponent:** `pluto.service` generates Kubernetes-specific settings
```
bcressey, line 134: I wouldn't really call `pluto` a subcomponent any more, now that it's not invoked as a generator. It's its own thing, an additional bonus action that generates more k8s settings.

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+
+- `kubelet.service` (Kubernetes variants)
+- `ecs.service` (ECS variants)
+- `host-containers@admin.service` (admin container)
+- `host-containers@control.service` (control container)
+
+**Dependencies:**
+
+- Requires `basic.target` and `configured.target`
+- This ensures all configuration is complete before workloads start
```
bcressey, line 208: Kind of, but we never actually enqueue `multi-user.target` until these dependencies are satisfied.

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+
+- `After=` - This service starts after the specified units
+- `Before=` - This service starts before the specified units
+
+### Requirement Dependencies
+
+- `Requires=` - This service requires the specified units (hard dependency)
+- `Wants=` - This service wants the specified units (soft dependency)
+- `RequiredBy=` - Reverse of `Requires=` (specified in `[Install]` section)
+- `WantedBy=` - Reverse of `Wants=` (specified in `[Install]` section)
```
bcressey, line 222: I find the hard / soft dependency language insufficiently precise.

The "systemd as job queue" formulation would say:
1. Wants/WantedBy - causes the wanted unit to be enqueued (pretend it happens in random order), only if and exactly when the wanting unit is enqueued
2. After/Before - affects the order in which units are enqueued
3. Requires/RequiredBy - like wants, but the requiring unit won't be started if the required unit fails

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+
+### Early Boot Services
+
+Services that need to run very early use `DefaultDependencies=no` to avoid the standard dependency chain:
+
+- `migrator.service`
+- `prepare-*.service` (filesystem preparation)
+- `activate-preconfigured.service`
```
bcressey, line 230: While this is true, I don't think it's helpful - most of these services are special and have unique reasons for `DefaultDependencies=no`, either for performance or because they're essential to reach `sysinit.target`.

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+## Synchronization Mechanisms
+
+### Systemd Targets
+
+Targets serve as synchronization points. A target is "reached" when all services required by or wanted by that target have completed.
+
+Target relationships:
+
+```
+drivers.target:
```
bcressey, line 241: `drivers.target` is kind of a category error here - we can still be running some units from it while starting other units that `preconfigured.target` wants. The others are much stronger synchronization points.

In general I don't really see targets as a synchronization point. They are more of an abstraction over a bunch of units - "I need everything to bring the network online or to make the TPM2 device to also go into the queue when you put my job in."

They let you synchronize *what* you enqueue but not *when*, exactly - "when" is just "the same instant you enqueue some other job".

docs/boot-process.md
```diff
@@ -0,0 +1,382 @@
...
+
+### Sentinel Files
+
+Services use sentinel files to track state across reboots:
+
+- `/var/lib/bottlerocket/early-boot-config.ran` - Prevents `early-boot-config.service` from running after first boot
+- `/run/bootstrap-containers/<name>.ran` - Prevents bootstrap containers from re-running
+- `/etc/.fips-module-check-passed` - Marks FIPS check completion
+
+Services use `ConditionPathExists=` or `ConditionPathExists=!` to check for these files.
```
bcressey, line 265: True, though I dislike this pattern and it's more of a last resort ideally.

If we had a `sentineldog` that could create either a persistent (`/.bottlerocket`) or ephemeral  (`/etc`) marker then that might be a nicer interface for programs that find themselves doing this.

PR #756, opened 2025-11-25
docs: add settings system overview documentation

cbgbt:
Add documentation explaining how Bottlerocket's settings system works, including:

- Data store location and filesystem structure
- Key-to-path mapping and metadata storage
- Complete settings lifecycle, including defaults to running services
- Persistence model (persistent datastore, non-persistent /etc, immutable templates)
- Configuration flow with visual diagrams
- User interaction examples

**Terms of contribution:**

### DIFF
--- a/docs/settings-system-overview.md
+++ b/docs/settings-system-overview.md
@@ -1,0 +1,328 @@
+# Bottlerocket Settings System Overview
+
+This document describes how Bottlerocket's settings system works, including storage, organization, and the flow from settings to configuration files.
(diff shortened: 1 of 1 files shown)


cbgbt:
I'm not sure that the `sources` section is the right call for long-lived documentation. I used it as part of writing the document, and I like that it could make finding relevant information easier.

But it is also somewhat duplicated with the "related documents" section and could become stale. Curious to hear what folks think.

PR #760, opened 2025-12-05
host-ctr: use docker resolver

KCSesh:
Issue Related to https://github.com/aws/amazon-ecs-agent/issues/4538

Code change dependency: https://github.com/bottlerocket-os/bottlerocket/pull/4715

**Description of changes:**
Recent ecs-agent updates introduced an incompatibility when using FIPS ECR endpoints alongside use_fips_endpoint=true, requiring us to choose one approach.

We opted to let users specify FIPS ECR endpoints directly. However, amazon-ecr-containerd-resolver doesn't support FIPS endpoints without use_fips_endpoint=true, and the library has been tech debt we've wanted to migrate away from.

This change replaces amazon-ecr-containerd-resolver with containerd's Docker resolver.

**Testing done:**

#### ECS Agent Conformance Testing

Ran internal ECS conformance tests across multiple variants and architectures with use_fips_endpoint=false:

| Variant | Architecture | Region |
|---------|--------------|--------|
| aws-ecs-2 | x86_64 | commercial |
| aws-ecs-2-nvidia-fips | aarch64 | commercial |
| aws-ecs-3 | x86_64 | commercial |
| aws-ecs-2 | aarch64 | us-gov-west-1 |
| aws-ecs-2-nvidia-fips | x86_64 | us-gov-west-1  |

Additionally verified ECS task execution with both FIPS and non-FIPS containers to confirm expected behavior.

##### GovCloud (us-gov-west-1)

| Variant | Image Endpoint | FIPS | ECR Auth Endpoint | Result |
|---------|----------------|------|-------------------|--------|
| aws-ecs-2 | dkr.ecr | false | api.ecr.us-gov-west-1.amazonaws.com | ✅ |
| aws-ecs-2 | dkr.ecr-fips | true | api.ecr-fips.us-gov-west-1.amazonaws.com | ✅ |
| aws-ecs-2-fips | dkr.ecr | false | api.ecr.us-gov-west-1.amazonaws.com | ✅ |
| aws-ecs-2-fips | dkr.ecr-fips | true | api.ecr-fips.us-gov-west-1.amazonaws.com | ✅ |

##### China Region (cn-north-1)

| Variant | Image Endpoint | FIPS | Result |
|---------|----------------|------|--------|
| aws-ecs-2 | dkr.ecr | false | ✅ |
| aws-ecs-2-fips | dkr.ecr-fips | true | ❌ Expected failure: invalid FIPS region: cn-north-1 |
| aws-ecs-2-fips | dkr.ecr | false | ✅ |

##### Special Region Test For New Regions (ap-southeast-7)

Verified host container image pull works in special region:

```
level=info msg="setting up ECR client" fips=false region=ap-southeast-7
level=info msg="pulling private ECR image" ref="<account>.dkr.ecr.ap-southeast-7.amazonaws.com/test-alpine:latest"
level=info msg="pulled image successfully"
```

##### Digest test:
<details>

```
bash-5.1# journalctl -u host-containers@test
Dec 24 02:24:13 ip-172-31-24-3.us-west-2.compute.internal systemd[1]: Started Host container: test.
Dec 24 02:24:14 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:14Z" level=info msg="Image does not exist, proceeding to pull image from source." ref="111111111111.dkr.ecr.us-west-2.amazonaws.com/test-alpine@sha256:7b9b6a044d921dfcaea2a843ff19d725948590352198f93cb878fd2c19d7ba3c"
Dec 24 02:24:14 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:14Z" level=info msg="setting up ECR client" fips=false region=us-west-2
Dec 24 02:24:14 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:14Z" level=info msg="pulling private ECR image" ref="111111111111.dkr.ecr.us-west-2.amazonaws.com/test-alpine@sha256:7b9b6a044d921dfcaea2a843ff19d725948590352198f93cb878fd2c19d7ba3c" region=us-west-2
Dec 24 02:24:15 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:15Z" level=info msg="pulled image successfully" img="111111111111.dkr.ecr.us-west-2.amazonaws.com/test-alpine@sha256:7b9b6a044d921dfcaea2a843ff19d725948590352198f93cb878fd2c19d7ba3c"
Dec 24 02:24:15 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:15Z" level=info msg="unpacking image..." img="111111111111.dkr.ecr.us-west-2.amazonaws.com/test-alpine@sha256:7b9b6a044d921dfcaea2a843ff19d725948590352198f93cb878fd2c19d7ba3c"
Dec 24 02:24:15 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:15Z" level=info msg="Container does not exist, proceeding to create it" ctr-id=test
Dec 24 02:24:15 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:15Z" level=info msg="container task does not exist, proceeding to create it" container-id=test
Dec 24 02:24:15 ip-172-31-24-3.us-west-2.compute.internal host-containers@test[2236]: time="2025-12-24T02:24:15Z" level=info msg="successfully started container task"
bash-5.1#
```

</details>
**Terms of contribution:**


sources/host-ctr/cmd/host-ctr/ecr.go
```diff
@@ -0,0 +1,234 @@
...
+
+		authOpt := docker.WithAuthCreds(func(host string) (string, string, error) {
+			if host != ecrHost {
+				return "", "", fmt.Errorf("ecr-private: unexpected host %s, expected %s", host, ecrHost)
+			}
+			return username, password, nil
+		})
+		authorizer := docker.NewDockerAuthorizer(authOpt)
+		c.Resolver = docker.NewResolver(docker.ResolverOptions{
+			// TODO: Consider adding support for user-provided credentials with registryConfig as fallback,
```
KCSesh: This is the same TODO as before: 
https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/host-ctr/cmd/host-ctr/main.go#L1250

sources/host-ctr/go.mod
```diff
@@ -7,11 +7,11 @@ toolchain go1.24.6
 require (
 	github.com/aws/aws-sdk-go-v2 v1.40.0
 	github.com/aws/aws-sdk-go-v2/config v1.32.1
+	github.com/aws/aws-sdk-go-v2/service/ecr v1.54.0
 	github.com/aws/aws-sdk-go-v2/service/ecrpublic v1.38.5
-	github.com/awslabs/amazon-ecr-containerd-resolver v0.0.0-20250903194853-2883582d177b
 	github.com/containerd/containerd v1.7.29
 	github.com/containerd/containerd/api v1.10.0
-	github.com/containerd/errdefs v1.0.0
+	github.com/containerd/errdefs v0.3.0
```
ytsssun: This downgrade feels odd, is this intentional?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1199,124 +983,23 @@ func pullImage(ctx context.Context, source string, client *containerd.Client, re
 	return img, nil
 }
 
-// tagImage adds a tag to the image in containerd's metadata storage.
-//
-// Image tag logic derived from:
-//
-// https://github.com/containerd/containerd/blob/d80513ee8a6995bc7889c93e7858ddbbc51f063d/cmd/ctr/commands/images/tag.go#L67-L86
-func tagImage(ctx context.Context, imageName string, newImageName string, client *containerd.Client) error {
```
ytsssun, line 1207: What is the implication of removing this tagImage operation? Would that introduce any breaking change?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -34,44 +28,6 @@ import (
...
-	"ap-southeast-6":  "ecr.aws/arn:aws:ecr:ap-southeast-6:",
-	"us-northeast-1":  "ecr.aws/arn:aws:ecr:us-northeast-1:",
-	"eu-isoe-west-1":  "ecr.aws/arn:aws-iso-e:ecr:eu-isoe-west-1:",
-	"eusc-de-east-1":  "ecr.aws/arn:aws-eusc:ecr:eusc-de-east-1:",
-	"us-iso-east-1":   "ecr.aws/arn:aws-iso:ecr:us-iso-east-1:",
-	"us-iso-west-1":   "ecr.aws/arn:aws-iso:ecr:us-iso-west-1:",
-	"us-isob-east-1":  "ecr.aws/arn:aws-iso-b:ecr:us-isob-east-1:",
-	"us-isob-west-1":  "ecr.aws/arn:aws-iso-b:ecr:us-isob-west-1:",
-	"us-isof-south-1": "ecr.aws/arn:aws-iso-f:ecr:us-isof-south-1:",
-	"us-isof-east-1":  "ecr.aws/arn:aws-iso-f:ecr:us-isof-east-1:",
```
ytsssun, line 62: Did you confirm that all these regions are now present in the go sdk?

sources/host-ctr/cmd/host-ctr/ecr.go
```diff
@@ -0,0 +1,266 @@
...
+
+// A set of the currently supported FIPS regions for ECR: https://docs.aws.amazon.com/general/latest/gr/ecr.html
+// FIPS-supported ECR regions: https://docs.aws.amazon.com/general/latest/gr/ecr.html
+var fipsSupportedEcrRegionSet = map[string]bool{
+	"us-east-1":     true,
+	"us-east-2":     true,
+	"us-west-1":     true,
+	"us-west-2":     true,
+	"us-gov-east-1": true,
+	"us-gov-west-1": true,
```
ytsssun, line 62: Looking at the linked regions, there seems to be more regions than what's listed here. Should we include them?

sources/host-ctr/go.mod
KCSesh: I think it is a result from a clean-tidy run. 

```
| Package | Requires |
|---------|----------|
| host-ctr (direct) | v1.0.0 |
| containerd@v1.7.29 | v0.3.0 |
| hcsshim@v0.13.0 | v0.3.0 |
| containernetworking/plugins@v1.9.0 | v0.3.0 |
| errdefs/pkg@v0.3.0 | v0.3.0 |
```
Non-functional change: 
https://github.com/containerd/errdefs/compare/v1.0.0...v0.3.0

However, I see no reason for it to change, so I'll make sure the direct dependency remains and keep it on `v1.0.0`

sources/host-ctr/cmd/host-ctr/main.go
KCSesh: The `tagImage` function was only needed because [amazon-ecr-containerd-resolver](https://github.com/awslabs/amazon-ecr-containerd-resolver?tab=readme-ov-file#ref) required images to be referenced in its canonical format (ecr.aws/arn:aws:ecr:<region>:<account>:repository/<name>:<tag>). 

When we pulled an image, it was stored in containerd under that canonical name, so
`tagImage` added the original Docker-style URI as an alias to allow lookups by either name. 

Note that this was only used for private ECR - ECR Public already used the Docker resolver directly and never needed tagImage.

With the new Docker resolver approach, images are stored directly under the original URI (111111111111.dkr.ecr.us-west-2.amazonaws.com/repo:tag) - no conversion happens, so no aliasing is needed.

sources/host-ctr/cmd/host-ctr/main.go
KCSesh: Yes, the unit test: `TestECRRegionHandling` ensures they load correctly from the sdk. 

If you think of other test suggestions for this, let me know!

sources/host-ctr/cmd/host-ctr/ecr.go
KCSesh: The [URL](https://docs.aws.amazon.com/general/latest/gr/ecr.html) is a link to which shows all/(most?) of ECR endpoints. 

Of that these 6 have fips endpoints. We currently capture all fips regions

sources/host-ctr/cmd/host-ctr/ecr.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// parsedECR contains the parsed components of an ECR image URI.
+type parsedECR struct {
+	Region   string
+	Account  string
+	RepoPath string
+	Fips     bool
+}
```
sam-berning, line 69: nit: `Account` and `RepoPath` seem unused (outside of tests) now that we don't need to construct the `ecr.aws/arn...` ref. we can potentially remove them and remove the account capture group from the regex

sam-berning (approved):

sources/host-ctr/cmd/host-ctr/ecr.go
KCSesh: Yeah, in fear of making an incompatible regex change - I oped on not doing it. Let me know if you think otherwise though.

sam-berning: I'm fine with not changing the regex, but I think it's probably worth removing the dead code for parsing out/setting the repoPath and account.

packages/ecs-agent/Cargo.toml
```diff
@@ -13,8 +13,9 @@ path = "../packages.rs"
 releases-url = "https://github.com/aws/amazon-ecs-agent/releases"
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/aws/amazon-ecs-agent/archive/v1.91.2/amazon-ecs-agent-1.91.2.tar.gz"
-sha512 = "c079dc22ee60ff0701d9a66f59add26fcab02baae36c72f98e8397ea6747a1858c4df2cada9ed3e2af3657d65920d2495b0b94c88dfbd573a6485ce2a4d6a816"
+# Verify the Git submodule commit of amazon-vpc-cni-plugins matches what is shipped
```
cbgbt: Good comment but admittedly it took me a moment to realize you meant shipped seperately in core-kit. Maybe revise to "... shipped in ../amazon-vpc-cni-plugins"

sources/host-ctr/cmd/host-ctr/ecr.go
```diff
@@ -0,0 +1,266 @@
...
+// ecrRegex matches ECR image names of the form:
+//
+// Example 1: 777777777777.dkr.ecr.us-west-2.amazonaws.com/my_image:latest
+// Example 2: 777777777777.dkr.ecr.cn-north-1.amazonaws.com.cn/my_image:latest
+// Example 3: 777777777777.dkr.ecr.eu-isoe-west-1.cloud.adc-e.uk/my_image:latest
+// Example 4: 777777777777.dkr.ecr-fips.us-west-2.amazonaws.com/my_image:latest
+//
+// Capture groups: [1] = account ID, [2] = "-fips" or empty, [3] = region
+//
+// ECR hostname pattern also used in the ecr-credential-provider:
```
cbgbt, line 30: We've since deviated from this in order to support .eu domain suffix.

I think this regex predates your PR. It's a gnarly one to read. It would be great if we could reign it in or do away with it somehow. I know in Python regexes have "verbose" mode where you can add inline comments explaining parts of the regex.

I want to say I remember interacting with this one in the past, so there's a chance I tried to do battle with it and failed. Might be a dead end.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -339,22 +295,14 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 	}
 	defer client.Close()
 
-	// Check if the image source is an ECR image. If it is, then we need to handle it with the ECR resolver.
-	isECRImage := ecrRegex.MatchString(source)
+	// Use unified image fetching for all registries
```
cbgbt: nit: Feel a bit mixed about this comment, as it is more a comment about the history of the code than the current state.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -580,19 +528,11 @@ func pullImageOnly(containerdSocket string, namespace string, source string, reg
...
-		if err != nil {
-			return err
-		}
-	} else {
-		_, err = fetchImage(ctx, source, client, registryConfigPath, useCachedImage, labels)
-		if err != nil {
-			log.G(ctx).WithField("ref", source).Error(err)
-			return err
-		}
+	// Use unified image fetching for all registries
```
cbgbt: same nit here

sources/host-ctr/cmd/host-ctr/ecr.go
```diff
@@ -0,0 +1,266 @@
...
+// https://github.com/kubernetes/cloud-provider-aws/blob/212135d0d7b448cd34e2e11e5e81f59e3e6c2d7a/cmd/ecr-credential-provider/main.go#L45
+var ecrRegex = regexp.MustCompile(`^(\d{12})\.dkr[\.\-]ecr(-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.(amazonaws(\.com(?:\.cn)?|\.eu)|on\.(?:aws|amazonwebservices\.com\.cn)|sc2s\.sgov\.gov|c2s\.ic\.gov|cloud\.adc-e\.uk|csp\.hci\.ic\.gov).*$`)
+
+const ecrPublicHost = "public.ecr.aws"
+const ecrPublicRegion = "us-east-1"
+
+// A set of the currently supported FIPS regions for ECR: https://docs.aws.amazon.com/general/latest/gr/ecr.html
+// FIPS-supported ECR regions: https://docs.aws.amazon.com/general/latest/gr/ecr.html
+var fipsSupportedEcrRegionSet = map[string]bool{
```
cbgbt: nit: dupe comment and the official list from the link appears to be larger now.

not nit: Is this something we can lean on the SDK for now? It seems like in the old code we needed to understand if we were doing FIPS to avoid hitting an error condition in the resolver.

Now we only use it to raise an error - but the SDK might take care of that for us.

ytsssun (approved):

sources/host-ctr/cmd/host-ctr/ecr.go
KCSesh: > nit: dupe comment 

Fixed.

> Re: Not Not: 

I did dig into this before the PR... and checked the SDK. U

Unfortunately it won't catch this for us. The SDK has a default FIPS hostname template `(ecr-fips.{region}.amazonaws.com)` that it falls back to for any region, so requesting FIPS for eu-west-1 (a non-fips region) would just construct `ecr-fips.eu-west-1.amazonaws.com` without error. 

The failure would only happen later as a DNS/connection error, which would be confusing for users.
Keeping the validation gives a clear "invalid FIPS region" error upfront. 

Also ECR doesn't appear to support all FIPS endpoints, which is another variable here:
https://aws.amazon.com/compliance/fips/ - see `ca-west-1`

>the official list from the link appears to be larger now.

This is not quite true, I have updated the comment to be more clear. We current support the list of fips end points that ECR supports. There is not new FIPS regions for ECR.

sources/host-ctr/cmd/host-ctr/ecr.go
KCSesh: I decided to better match:
https://github.com/kubernetes/cloud-provider-aws/blob/d1c7c02d2da22e87175802ec94c73bd8871691bc/cmd/ecr-credential-provider/main.go#L46

which includes `eu` and also I added in-line comments. 

Let me know what you think!

sources/host-ctr/cmd/host-ctr/main.go
KCSesh: Fixed, nice catch!

sources/host-ctr/cmd/host-ctr/ecr.go
KCSesh: Based off this and the discussion below with @cbgbt, I updated the regex and struct: 
https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/760#discussion_r2677309765

cbgbt (approved):
Nice!

PR #764, opened 2025-12-08
containerd-2.1: disable concurrent layer fetch by default

KCSesh:
The ConcurrentLayerFetchBuffer feature was previously set to 8MiB, enabling parallel layer downloads. Rolling out this feature revealed compatibility challenges with certain registry implementations and race conditions in the concurrent fetch logic.

Setting the buffer to 0 disables layer chunking, changing to non-chunked layer downloads. This provides more reliable behavior across different registry implementations while the underlying issues are addressed.

**Issue number:**

Related issues #https://github.com/bottlerocket-os/bottlerocket/issues/4709

**Description of changes:**

Disabling the 8MiB default. Users can still set it via:

```
apiclient set settings.container-runtime.concurrent-download-chunk-size="8mib"
```

**Testing done:**

Launched AMI and confirmed settings are updated. 

**Terms of contribution:**

### DIFF
--- a/packages/containerd-2.1/containerd-config-toml_k8s_nvidia_containerd_sock
+++ b/packages/containerd-2.1/containerd-config-toml_k8s_nvidia_containerd_sock
@@ -26,7 +26,7 @@ address = "/run/containerd/containerd.sock"
 {{#if settings.container-runtime.max-concurrent-downloads}}
 max_concurrent_downloads = {{settings.container-runtime.max-concurrent-downloads}}
 {{/if}}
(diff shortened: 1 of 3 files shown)


packages/containerd-2.1/containerd-config-toml_k8s_nvidia_containerd_sock
bcressey, line 53: The default here should also be changed.

bcressey (approved):

ytsssun (approved):

koooosh (approved):

PR #813, opened 2026-01-28
re-packaging the Go packages

sky1122:
**Issue number:**

Closes #814 

**Description of changes:**
re-packaging the Go packages to remove the fips binary and `gofips` build

**Testing done:**
- ecs/eks boot test and ecs ec2 running task.
- ecs-2-fips: internal ECS testing passes
- k8s-1.29-fips: Conformance test pass
- k8s-1.29-fips test
- build all the k8s fips/non-fips variants successful
- build `k8s-1.35-nvidia-fips` smoke test pass
```
╰$ ls
x86_64-aws-ecs-2-fips         x86_64-aws-k8s-1.30       x86_64-aws-k8s-1.31-fips  x86_64-aws-k8s-1.33       x86_64-aws-k8s-1.34-fips
x86_64-aws-ecs-2-nvidia-fips  x86_64-aws-k8s-1.30-fips  x86_64-aws-k8s-1.32       x86_64-aws-k8s-1.33-fips  x86_64-aws-k8s-1.35
x86_64-aws-k8s-1.29-fips      x86_64-aws-k8s-1.31       x86_64-aws-k8s-1.32-fips  x86_64-aws-k8s-1.34       x86_64-aws-k8s-1.35-fips
```

```
│127.0.0.1 - - [02/Feb/2026:23:08:47 +0000] "HEAD /v2/test/manifests/latest HTTP/1.1" 400 0 "-" "containerd/1.7.30+bottlerocket"                                                     ││://localhost:5001"]}]}}'                                                                                                                                                            │
│2026/02/02 23:09:59 [info] 30#30: *6 SSL_do_handshake() failed (SSL: error:0A0000C1:SSL routines::no shared cipher) while SSL handshaking, client: 127.0.0.1, server: 0.0.0.0:5001  ││                                                                                                                                                                                    │
│2026/02/02 23:10:10 [info] 30#30: *7 SSL_do_handshake() failed (SSL: error:0A0000C1:SSL routines::no shared cipher) while SSL handshaking, client: 127.0.0.1, server: 0.0.0.0:5001  ││                                                                                                                                                                                    │
│2026/02/02 23:10:36 [info] 30#30: *8 SSL_do_handshake() failed (SSL: error:0A0000C1:SSL routines::no shared cipher) while SSL handshaking, client: 127.0.0.1, server: 0.0.0.0:5001  ││Then recreate the nonfips test pod.                                                                                                                                                 │
[...]

PR #818, opened 2026-02-03
add additional tools

bcressey:
**Issue number:**
N/A

**Description of changes:**
I've carried these packages in my local tree for a while; they all seem like useful additions.

1. `erofs-utils` - I'd like to switch over to the erofs snapshotter for containerd at some point, or make that a configurable option.
2. `cri-tools` - `crictl` is useful to exercise containerd's CRI interface specifically. Can be updated when we add a new `kubelet` package.
3. `perf` - `perf` is essential for profiling performance. Can be updated when we add a new kernel.

**Testing done:**
<details>

<summary>perf top works</summary>

```
   PerfTop:      24 irqs/sec  kernel:83.3%  exact:  0.0% lost: 0/0 drop: 0/0 [4000Hz cycles:P],  (all, 1 CPU)
----------------------------------------------------------------------------------------------------------------------------------------------------------------------

    19.65%  [kernel]       [k] finish_task_switch.isra.0
    13.40%  [kernel]       [k] default_idle_call
    12.33%  [kernel]       [k] uart_write
     4.15%  [kernel]       [k] arch_local_irq_restore
     2.08%  [kernel]       [k] tick_nohz_idle_exit
     1.51%  [kernel]       [k] el0_svc
```

</details>

<details>

<summary>crictl can pull and inspect images</summary>

```
bash-5.2# crictl pull amazonlinux:2023
WARN[0000] Config "/etc/crictl.yaml" does not exist, trying next: "/aarch64-bottlerocket-linux-gnu/sys-root/usr/bin/crictl.yaml"
WARN[0000] Image connect using default endpoints: [unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock]. As the default settings are now deprecated, you should set the endpoint instead.
Image is up to date for sha256:e131830c9f16032602c25a25c31cc9e5b658ff910c44fc9b6010be94d08101c5

bash-5.2# crictl inspecti localhost/kubernetes/pause:0.1.0
WARN[0000] Config "/etc/crictl.yaml" does not exist, trying next: "/aarch64-bottlerocket-linux-gnu/sys-root/usr/bin/crictl.yaml"
WARN[0000] Image connect using default endpoints: [unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock]. As the default settings are now deprecated, you should set the endpoint instead.
{
...
  "status": {
    "id": "sha256:160fa372abdd58cb0ef24e8505a7982e2ff35d13f19eda2ffb383ad8f44e5773",
    "pinned": true,
    "repoDigests": [],
    "repoTags": [
      "localhost/kubernetes/pause:0.1.0"
    ],
    "size": "164632",
    "username": ""
  }
}
```
</details>

<details>

<summary>mkfs.erofs works when invoked by the containerd erofs snapshotter</summary>

```
bash-5.2# findmnt /var/lib/containerd/io.containerd.snapshotter.v1.erofs/snapshots/1/fs -o target,source,fstype
TARGET                                                                SOURCE                                                                         FSTYPE
/var/lib/containerd/io.containerd.snapshotter.v1.erofs/snapshots/1/fs /var/lib/containerd/io.containerd.snapshotter.v1.erofs/snapshots/1/layer.erofs erofs
```

</details>

**Terms of contribution:**


yeazelm (approved):
LGTM!

packages/cri-tools/cri-tools.spec
```diff
@@ -0,0 +1,73 @@
...
+
+%prep
+%setup -n %{gorepo}-%{gover} -q
+%setup -T -D -n %{gorepo}-%{version} -b 1
+
+%build
+%set_cross_go_flags
+export GO_MAJOR="1.25"
+go build -ldflags="${GOLDFLAGS}" -o crictl ./cmd/crictl
+gofips build -ldflags="${GOLDFLAGS}" -o fips/crictl ./cmd/crictl
```
arnaldo2792: Should we avoid the second compilation in prep of:

https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/813

packages/perf/perf.spec
```diff
@@ -0,0 +1,84 @@
...
+  NO_LIBDW_DWARF_UNWIND=1 \
+  NO_LIBLLVM=1 \
+  NO_LIBNUMA=1 \
+  NO_LIBPERL=1 \
+  NO_LIBPFM4=1 \
+  NO_LIBPYTHON=1 \
+  NO_LIBTRACEEVENT=1 \
+  NO_LIBUNWIND=1 \
+  NO_LIBZSTD=1 \
+  NO_LZMA=1 \
```
arnaldo2792, line 57: I'm curious, why not enabling the compression support for `perf` but it was enabled for`erofs-utils`? Is it to match the EROFS compression enabled in the kernel?

packages/perf/perf.spec
```diff
@@ -0,0 +1,84 @@
...
+
+%global pmake %{shrink: \
+make -s \
+  ARCH="%{_cross_karch}" \
+  CROSS_COMPILE="%{_cross_target}-" \
+  -C tools/perf \
+  prefix="%{_cross_prefix}" \
+  DESTDIR="%{buildroot}" \
+  EXTRA_CFLAGS="%{_cross_cflags} -fno-lto" \
+  EXTRA_CXXFLAGS="%{_cross_cxxflags} -fno-lto" \
```
arnaldo2792, line 30: Out of curiosity, why not enabling `lto` for `perf`?

arnaldo2792 (approved):

packages/cri-tools/cri-tools.spec
bcressey: > Should we avoid the second compilation in prep of:

Depends on who merges first 😀

packages/perf/perf.spec
bcressey: > I'm curious, why not enabling the compression support for perf but it was enabled for erofs-utils? Is it to match the EROFS compression enabled in the kernel?

`perf` can use libzstd to compress traces, which sounds useful although I've never used it.

LZMA is for xz-compressed kernel modules, which we don't use.

packages/perf/perf.spec
bcressey: > Out of curiosity, why not enabling lto for perf?

The `perf` package started life as sub-packages of `kernel-6.1` and `kernel-6.12`, though ultimately I needed the `libelf` dependency to get traces with symbol names, and that ended up being too much userspace to package in the kernel kit.

During that time, I kept running into LTO-related build issues. Sometimes they'd be fixed upstream but not backported, or affect one kernel but not the other. Disabling LTO made the pain go away.

With the move into the core kit, the package now tracks current kernel releases, so in theory there are no more LTO problems. But I haven't gone back to test that, and I have a lingering concern that the minor performance win might not be worth the upkeep effort.

packages/perf/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+build = "../build.rs"
+
+[package.metadata.build-package]
+package-name = "perf"
+
+[lib]
+path = "../packages.rs"
+
+[[package.metadata.build-package.external-files]]
+url = "https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.18.8.tar.xz"
```
KCSesh: nit: linux-6.18.9 exists seems it has 1 [perf](https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.18.9) related commit callout.

KCSesh (approved):

bcressey:
⬆️ force push to rebase for Go package changes

bcressey:
⬆️ force push:
* adjust packaging for `cri-tools` to eliminate fips/non-fips distinction
* update `perf` to 6.18.10

KCSesh (approved):

PR #721, opened 2025-11-07
add support for encrypted storage

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket/issues/4680

**Description of changes:**
Most of the code to support this is in #717; what's left is a lot of minor integration chores.

High level summary of changes:
1. Improve error handling for local storage setup.
2. Make existing local storage units customizable via drop-ins.
3. Encrypt BOTTLEROCKET-DATA (`/local`) if encrypted storage is enabled.
4. Encrypt BOTTLEROCKET-PRIVATE (`/.bottlerocket/datastore`) if encrypted storage is enabled.
5. Measure boot phases to protect sealed encryption keys.
6. Measure kernel command line and OS settings.

See documentation for additional notes.

**Testing done:**
<details>
<summary>Storage is encrypted when the feature is enabled</summary>

```
bash-5.2# findmnt /local
TARGET SOURCE                        FSTYPE OPTIONS
/local /dev/mapper/BOTTLEROCKET-DATA xfs    rw,nosuid,nodev,noatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota

bash-5.2# rw check directory /.bottlerocket/datastore encrypted
directory '/.bottlerocket/datastore' is encrypted

bash-5.2# apiclient ephemeral-storage init
bash-5.2# apiclient ephemeral-storage bind
bash-5.2# findmnt /mnt/.ephemeral/
TARGET          SOURCE                     FSTYPE OPTIONS
/mnt/.ephemeral /dev/mapper/EPHEMERAL-DATA xfs    rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=256k,sunit=512,swidth=1024,noquota
```
</details>

<details>
<summary>Storage is NOT encrypted when the feature is disabled</summary>

```
bash-5.2# findmnt /local
TARGET SOURCE    FSTYPE OPTIONS
/local /dev/vdd1 xfs    rw,nosuid,nodev,noatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota

bash-5.2# rw check directory /.bottlerocket/datastore encrypted
bash: rw: command not found

bash-5.2# apiclient ephemeral-storage init
bash-5.2# apiclient ephemeral-storage bind
bash-5.2# findmnt /mnt/.ephemeral/
TARGET          SOURCE     FSTYPE OPTIONS
[...]

PR #822, opened 2026-02-04
fixes for ephemeral storage

bcressey:
**Issue number:**
N/A

**Description of changes:**
Refactor the implementation for the ephemeral storage API and fix four problems.

`/mnt/ephemeral` was previously mounted without the "private" option, leading to mount amplification as every directory under `/var/lib/kubelet` would appear twice. Parsing mount information can be a bottleneck for `systemd` on busy systems. 

Mounting `/mnt/ephemeral` without the "private" option leads to another issue, which is that paths like `/mnt/ephemeral/._var_lib_kubelet` would show some of the same content as `/var/lib/kubelet`, but not all, because pod mounts will no longer propagate back to the parent. Rather than accept the inconsistency, we mount over the parent to "mask" it with a read-only mount and prevent it from being used as another way to view the content.

See also the notes in https://github.com/bottlerocket-os/bottlerocket/pull/2240 from when the same change was made for `/local` and `/var`.

#395 included a fix to make the API idempotent, but #721 introduced a regression when initializing encrypted storage. Fix that in the same way, by removing the symlink if it already exists.

Bind mounts might not happen in the correct order - parent directories before children. Sort bind mounts first to ensure this.

**Testing done:**
<details>

<summary>Mounts no longer propagate back to /mnt/ephemeral</summary>

Before:
```
bash-5.2# findmnt -R /mnt -o TARGET,PROPAGATION,OPTIONS
TARGET                                                                                                          PROPAGATION OPTIONS
/mnt                                                                                                            shared      rw,nosuid,nodev,noatime,seclabel,attr2,inod
`-/mnt/.ephemeral                                                                                               shared      rw,relatime,seclabel,attr2,inode64,logbufs=
  |-/mnt/.ephemeral/._var_lib_kubelet/pods/5bf2ab9e-e6c6-4e37-b90f-874cafaafd4c/volumes/kubernetes.io~projected/kube-api-access-46q8s
  |                                                                                                             shared      rw,relatime,seclabel,size=409600k,noswap
  |-/mnt/.ephemeral/._var_lib_kubelet/pods/82eae8b0-5a7a-4b7c-9437-b5961c5cde14/volumes/kubernetes.io~projected/kube-api-access-gtnh9
  |                                                                                                             shared      rw,relatime,seclabel,size=6933428k,noswap
  |-/mnt/.ephemeral/._var_lib_kubelet/pods/0a20da32-a3aa-4617-9f98-bd2057450f51/volumes/kubernetes.io~projected/kube-api-access-j29rg
```

After:
```
bash-5.2#  findmnt -R /mnt -o TARGET,PROPAGATION,OPTIONS
TARGET                                PROPAGATION OPTIONS                                                                                                              /mnt                                  shared      rw,nosuid,nodev,noatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,sunit=8,swidth=8,noquota
`-/mnt/.ephemeral                     private     rw,nosuid,nodev,noatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
  |-/mnt/.ephemeral/._var_lib_kubelet private     ro,nosuid,nodev,noexec,relatime,seclabel,user_xattr,acl,cache_strategy=readaround                                      `-/mnt/.ephemeral/._var_log_pods    private     ro,nosuid,nodev,noexec,relatime,seclabel,user_xattr,acl,cache_strategy=readaround
```

</details>

<details>

<summary>Mounts have expected options</summary>

Before:
```
bash-5.2# findmnt /var/lib/kubelet/ -o TARGET,OPTIONS
TARGET           OPTIONS
/var/lib/kubelet rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
```

After:
```
bash-5.2# findmnt /var/lib/kubelet/ -o TARGET,OPTIONS
TARGET           OPTIONS
/var/lib/kubelet rw,nosuid,nodev,noatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
```

</details>

<details>

<summary>Init API is idempotent</summary>

Before:
```
bash-5.2# apiclient ephemeral-storage init
Failed to initialize ephemeral storage: Failed POST request to '/actions/ephemeral-storage/init': Status 500 when POSTing /actions/ephemeral-storage/init: Unable to initialize ephemeral storage: Failed to create disk symlink File exists (os error 17)
```

After (later boot phase, expected failure):
```
bash-5.2# apiclient ephemeral-storage init
Failed to initialize ephemeral storage: Failed POST request to '/actions/ephemeral-storage/init': Status 500 when POSTing /actions/ephemeral-storage/init: Unable to initialize ephemeral storage: Failed to run 'rottweiler' with args 'attach block-device /dev/disk/EPHEMERAL-DATA ephemeral-storage' on device '/dev/disk/EPHEMERAL-DATA': stdout: , stderr: Error: /usr/bin/systemd-creds failed with exit code 1: Failed to unseal secret using TPM2: Operation not permitted
```

After (same boot phase, success):
```
bash-5.3# apiclient ephemeral-storage init
bash-5.3# apiclient ephemeral-storage init
```

</details>

<details>

<summary>Parent directories mounted before children</summary>

Before:
```
bash-5.3# apiclient ephemeral-storage bind --dirs /mnt/foo/bar /mnt/foo
Failed to initialize ephemeral storage: Failed POST request to '/actions/ephemeral-storage/bind': Status 500 when POSTing /actions/ephemeral-storage/bind: Unable to bind ephemeral storage: Failed to bind directory '/mnt/.ephemeral/._mnt_foo_bar' to '/mnt/foo/bar': mount: /mnt/foo/bar: mount point does not exist.
```

After:
```
bash-5.3# apiclient ephemeral-storage bind --dirs /mnt/foo/bar/baz /mnt/foo /mnt/bar
bash-5.3# findmnt -R /mnt -o TARGET,SOURCE
TARGET                                SOURCE
/mnt                                  /dev/mapper/BOTTLEROCKET-DATA[/mnt]
├─/mnt/.ephemeral                     /dev/mapper/EPHEMERAL-DATA
│ ├─/mnt/.ephemeral/._mnt_foo         /dev/dm-0[/srv]
│ ├─/mnt/.ephemeral/._mnt_bar         /dev/dm-0[/srv]
│ └─/mnt/.ephemeral/._mnt_foo_bar_baz /dev/dm-0[/srv]
├─/mnt/bar                            /dev/mapper/EPHEMERAL-DATA[/._mnt_bar]
└─/mnt/foo                            /dev/mapper/EPHEMERAL-DATA[/._mnt_foo]
  └─/mnt/foo/bar/baz                  /dev/mapper/EPHEMERAL-DATA[/._mnt_foo_bar_baz]
```

</details>

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/ephemeral_storage.rs
+++ b/sources/api/apiserver/src/server/ephemeral_storage.rs
@@ -17,6 +17,7 @@ static MKFSEXT4: &str = "/usr/sbin/mkfs.ext4";
 static FINDMNT: &str = "/usr/bin/findmnt";
 
 static EPHEMERAL_MNT: &str = "/mnt/.ephemeral";
+static SRV_DIR: &str = "/srv";
 
 /// Name of the device and its path from the MD driver
 static RAID_DEVICE_DIR: &str = "/dev/md/";
@@ -170,7 +171,7 @@ pub fn initialize(
     Ok(())
 }
 
-/// binds the specified directories to the pre-configured array, creating those directories if
+/// Binds the specified directories to the pre-configured array, creating those directories if
 /// they do not exist.
 pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
     let device_name = EPHEMERAL_STORAGE_LINK;
@@ -243,8 +244,9 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
 
     let dirs: Vec<OsString> = dirs.iter().map(OsString::from).collect();
     let mut dirs_to_bind = HashSet::new();
+    let mut dirs_to_mask = HashSet::new();
 
-    // Check which directories need binding
+    // Check which directories need binding and/or masking
     for target_dir in &dirs {
         // Transform the directory path to a unique name
         let source_subdir = transform_dir_name(target_dir);
@@ -256,9 +258,19 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
         })?;
         std::fs::create_dir_all(target_dir).context(error::MkdirSnafu { dir: target_dir })?;
 
+        let is_source_masked = is_masked(&source_dir)?;
         let is_target_mounted = is_mounted(target_dir)?;
-        if !is_target_mounted {
-            dirs_to_bind.insert((source_dir.clone(), target_dir.clone()));
+
+        match (is_target_mounted, is_source_masked) {
+            (true, true) => continue,
+            (true, false) => {
+                dirs_to_mask.insert(source_dir.into());
+            }
+            (false, false) => {
+                dirs_to_bind.insert((source_dir.clone(), target_dir.clone()));
+                dirs_to_mask.insert(source_dir.into_os_string());
+            }
+            (false, true) => error::DirectoryAlreadyMaskedSnafu { dir: source_dir }.fail()?,
         }
     }
 
@@ -284,6 +296,29 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
             error::BindDirectoryFailureSnafu {
                 source_dir,
                 target_dir,
+                output,
+            }
+        );
+    }
+
+    // Mask source directories that need it
+    for source_dir in &dirs_to_mask {
+        info!("masking {}", source_dir.display());
+        let output = Command::new(MOUNT)
+            .args([
+                OsStr::new("--bind"),
+                OsStr::new("--options"),
+                OsStr::new("nosuid,nodev,noexec,private"),
+                OsStr::new(SRV_DIR),
+                source_dir,
+            ])
+            .output()
+            .context(error::ExecutionFailureSnafu { command: MOUNT })?;
+
+        ensure!(
+            output.status.success(),
+            error::MaskDirectorySnafu {
+                dir: source_dir,
                 output,
             }
         );
@@ -323,6 +358,24 @@ fn is_mounted(path: impl AsRef<OsStr>) -> Result<bool> {
         .status()
         .context(error::FindMntFailureSnafu {})?;
     Ok(status.success())
+}
+
+/// is_masked returns true if the specified path is already masked
+fn is_masked(path: impl AsRef<OsStr>) -> Result<bool> {
+    let output = Command::new(FINDMNT)
+        .args([OsStr::new("-no"), OsStr::new("SOURCE")])
+        .arg(&path)
+        .output()
+        .context(error::CheckMaskSnafu { dir: &path })?;
+
+    // Check if the command was successful and if the output contains "[/srv]"
+    if output.status.success() {
+        let source = String::from_utf8_lossy(&output.stdout);
+        Ok(source.trim().ends_with(&format!("[{SRV_DIR}]")))
+    } else {
+        // If the command failed, the path is not mounted at all
+        Ok(false)
+    }
 }
 
 /// creates the array with the given name from the specified disks
@@ -625,6 +678,20 @@ pub mod error {
             args: String,
             output: std::process::Output,
         },
+        #[snafu(display("Failed to check if directory '{}' is masked: {}", dir.display(), source))]
+        CheckMask {
+            dir: PathBuf,
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Failed to mask directory '{}': {}", dir.display(), String::from_utf8_lossy(output.stderr.as_slice())))]
+        MaskDirectory {
+            dir: PathBuf,
+            output: std::process::Output,
+        },
+
+        #[snafu(display("Cannot bind directory '{}': directory is masked", dir.display()))]
+        DirectoryAlreadyMasked { dir: PathBuf },
     }
 }
 
--- a/sources/api/apiserver/src/server/ephemeral_storage.rs
+++ b/sources/api/apiserver/src/server/ephemeral_storage.rs
@@ -220,7 +220,12 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
         std::fs::create_dir_all(EPHEMERAL_MNT).context(error::MkdirSnafu { dir: EPHEMERAL_MNT })?;
         info!("mounting {device_name} as {EPHEMERAL_MNT}");
         let output = Command::new(MOUNT)
-            .args([OsString::from(device_name), OsString::from(EPHEMERAL_MNT)])
+            .args([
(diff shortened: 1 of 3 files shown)


sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -459,6 +533,11 @@ fn should_encrypt() -> Result<bool> {
 fn encrypt_ephemeral_device(device: &str) -> Result<String> {
     info!("encrypting ephemeral device {device:?}");
 
+    // Clear previous link if it exists
+    if std::fs::exists(EPHEMERAL_DATA_LINK).is_ok_and(|x| x) {
+        std::fs::remove_file(EPHEMERAL_DATA_LINK).context(error::DiskUnlinkFailureSnafu {})?;
+    }
```
ytsssun, line 546: Q - for the idempotency problem, I wonder if we can introduce any testing to avoid future regression. Unit test might be hard as we might need to introduce a series of mocking to be able to test functions like `encrypt_ephemeral_device`.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -214,79 +217,125 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
...
+        match (is_target_mounted, is_source_masked) {
+            (true, true) => continue,
+            (true, false) => {
+                dirs_to_mask.insert(source_dir.into());
+            }
+            (false, false) => {
+                dirs_to_bind.insert((source_dir.clone(), target_dir.clone()));
+                dirs_to_mask.insert(source_dir.into_os_string());
+            }
+            (false, true) => error::DirectoryAlreadyMaskedSnafu { dir: source_dir }.fail()?,
```
ytsssun, line 273: Can we wrap this logic in a modular function and unit test them? Right now they are not quite testable if included as part of bind().

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -214,79 +217,125 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
...
-    let mount_point = Path::new(&mount_point);
-    for dir in &dirs {
-        // construct a directory name (E.g. /var/lib/kubelet => ._var_lib_kubelet) that will be
-        // unique between the binding targets
-        let mut directory_name = dir.replace('/', "_");
-        directory_name.insert(0, '.');
-        let mount_destination = mount_point.join(&directory_name);
+    let dirs: Vec<OsString> = dirs.iter().map(OsString::from).collect();
+    let mut dirs_to_bind = HashSet::new();
+    let mut dirs_to_mask = HashSet::new();
```
ytsssun: Q - does the order matter here for the dirs to be binded/masked? If yes, we should probably use Vec<> instead.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -168,7 +171,7 @@ pub fn initialize(
     Ok(())
 }
 
-/// binds the specified directories to the pre-configured array, creating those directories if
+/// Binds the specified directories to the pre-configured array, creating those directories if
 /// they do not exist.
 pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
```
ytsssun, line 177: nit: With more logic adding to bind() it has become a rather large function. To maintain readability, should we consider modularize some of the new code added here?

sources/api/apiserver/src/server/ephemeral_storage.rs
piyush-jena: Also ensures minimum unique set. Vec may have duplicates.

KCSesh: The only problem I can see with this is if mount points are children of another. 

Which makes me wonder about adding additional input sanity on overlapping paths and if we want to protect against that. 

`/mnt/test` and `/mnt/test/sub` - technically with a hashset mount order is non-deterministic, which with overlapping paths **could** make data inaccessible.

We could catch that as invalid 🤔 

<details>

```
for (i, dir_a) in dirs.iter().enumerate() {
    let path_a = Path::new(dir_a);
    for dir_b in dirs.iter().skip(i + 1) {
        let path_b = Path::new(dir_b);
        if path_a.starts_with(path_b) || path_b.starts_with(path_a) {
            return error::InvalidParameterSnafu {
                parameter: format!("{}, {}", dir_a, dir_b),
                reason: "overlapping bind directories not allowed",
            }
            .fail();
        }
    }
}

```
</details>

This problem exists today though - but across multiple calls: 

```
apiclient ephemeral-storage bind --targets /mnt/a/sub
# later...
apiclient ephemeral-storage bind --targets /mnt/a

```
I could see this happening across multiple bootstrap containers that run and set different binds (where we don't guarantee launch order)

bcressey: Ordering matters for bind mounts, but not for the masking part since masking happens on `._path_to_dir`.

Should be easy to fix the mount ordering, e.g. by sorting, I can add that.

sources/api/apiserver/src/server/ephemeral_storage.rs
bcressey: > Can we wrap this logic in a modular function and unit test them? Right now they are not quite testable if included as part of bind().

Making this unit testable is a good idea, but it's a larger refactor than I want to tackle here. Are you OK with a backlog issue?

sources/api/apiserver/src/server/ephemeral_storage.rs
bcressey: > I could see this happening across multiple bootstrap containers that run and set different binds (where we don't guarantee launch order)

Fixing this - or at least detecting it - across multiple calls is a reasonable idea, but will have to wait for a future change 😀 

I can fix it for the single call case though.

sources/api/apiserver/src/server/ephemeral_storage.rs
ytsssun: sure, this is not a blocking comment!

KCSesh (approved):

ytsssun (approved):

sources/api/apiserver/src/server/ephemeral_storage.rs
bcressey: Opened https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/840 for this.

Issue #540, opened 2025-06-06
CIS Benchmark Level 2 benchmark failure on EKS


justingood:
**Package I'm using:**
bloodhound/kubelet

**What I expected to happen:**
After following the instructions in [this blog post](https://aws.amazon.com/blogs/containers/validating-amazon-eks-optimized-bottlerocket-ami-against-the-cis-benchmark/), I expected to be able to run `apiclient report cis -l 2` and pass all the tests.

**Configuration**
Bottlerocket: v1.40.0 (x86_64)
EKS: v1.31

<details>

<summary>Here is our bottlerocket configuration</summary>

We're using Terraform to template the file, so there are some variables and a base64 encoding function in there

```toml
[settings.kubernetes]
api-server = "${api_server}"
cluster-certificate = "${cluster_certificate}"
cluster-name = "${cluster_name}"
node-labels = { pool-name = "${pool_name}" }
cluster-dns-ip = "${dns_cluster_ip}"

[settings.kernel]
lockdown = "integrity"

[settings.kernel.sysctl]
"net.ipv4.conf.all.send_redirects" = "0"
"net.ipv4.conf.default.send_redirects" = "0"
"net.ipv4.conf.all.accept_redirects" = "0"
"net.ipv4.conf.default.accept_redirects" = "0"
"net.ipv6.conf.all.accept_redirects" = "0"
"net.ipv6.conf.default.accept_redirects" = "0"
"net.ipv4.conf.all.secure_redirects" = "0"
"net.ipv4.conf.default.secure_redirects" = "0"
"net.ipv4.conf.all.log_martians" = "1"
"net.ipv4.conf.default.log_martians" = "1"

[settings.kernel.modules.udf]
allowed = false

[settings.kernel.modules.sctp]
allowed = false

[settings.bootstrap-containers.iptables]
source = "public.ecr.aws/bottlerocket/bottlerocket-bootstrap:v0.2.0"
mode = "always"
essential = true
user-data = "${base64encode(<<EOT
#!/usr/bin/env bash
set -xeuo pipefail

dnf install -y iptables-legacy

iptables -F
iptables -P INPUT DROP
iptables -P OUTPUT DROP
iptables -P FORWARD DROP
iptables -I INPUT -p tcp -m tcp --dport 10250 -j ACCEPT
iptables -A INPUT -i lo -j ACCEPT
iptables -A OUTPUT -o lo -j ACCEPT
[...]

PR #833, opened 2026-02-13
remove kubelet ipv4 ACCEPT override

rpkelly:
**Issue number:** #540

**Description of changes:**
Removing `FORWARD ACCEPT` override of iptables for kubernetes variants.

**Testing done:**

Built k8s 1.34 and ran k8s e2e tests on ipv4 cluster

**Terms of contribution:**


bcressey:
Please squash this together into one commit that explains (at length) why the change is being made and why it's safe now (with citations) when presumably it wasn't in the past.

It's not obvious to me that it's going to be compatible with deployed (i.e., not necessarily latest) versions of the various CNI solutions: aws-vpc, Cilium, Calico.

PR #836, opened 2026-02-17
source: add ecr-credential-helper-shim

sky1122:
**Issue number:**

Closes #

**Description of changes:**
This is part of the changes to enable `notation` in `fips140=only` mode.
Introduce ecr-credential-helper-shim, a Rust binary that downgrades GODEBUG from fips140=only to fips140=on before exec'ing the Go-based ecr-credential-helper, since the Go ECR credential helper does not support fips140=only mode.

**Testing done:**
- build ecs-3-fips and ecs-3 variants imaging signing test pass.
```
== when pulling public.ecr.aws/o7o0w6s5/public-alpine-clone:latest ==
Digest: sha256:85f2b723e106c34644cd5851d7e81ee87da98ac54672b29947c052a45d31dc2f
Status: Downloaded newer image for public.ecr.aws/o7o0w6s5/public-alpine-clone:latest
public.ecr.aws/o7o0w6s5/public-alpine-clone:latest
=======================

Waiter CommandExecuted failed: Waiter encountered a terminal failure state: For expression "Status" we matched expected path: "Failed"
== when pulling hello-world ==
Using default tag: latest
Error response from daemon: image verifier bindir blocked pull of docker.io/library/hello-world:latest with digest sha256:ef54e839ef541993b4e87f25e752f7cf4238fa55f017957c2eb44077083d7a6a for reason: verifier notation-image-verifier rejected image (exit code 1): verifying image: docker.io/library/hello-world@sha256:ef54e839ef541993b4e87f25e752f7cf4238fa55f017957c2eb44077083d7a6a
image verification failed: Error: signature verification failed: no signature is associated with "docker.io/library/hello-world@sha256:ef54e839ef541993b4e87f25e752f7cf4238fa55f017957c2eb44077083d7a6a", make sure the artifact was signed successfully
=======================

Waiter CommandExecuted failed: Waiter encountered a terminal failure state: For expression "Status" we matched expected path: "Failed"
== when pulling public.ecr.aws/o7o0w6s5/public-nginx-clone:latest ==
Error response from daemon: image verifier bindir blocked pull of public.ecr.aws/o7o0w6s5/public-nginx-clone:latest with digest sha256:bd1578eec775d0b28fd7f664b182b7e1fb75f1dd09f92d865dababe8525dfe8b for reason: verifier notation-image-verifier rejected image (exit code 1): verifying image: public.ecr.aws/o7o0w6s5/public-nginx-clone@sha256:bd1578eec775d0b28fd7f664b182b7e1fb75f1dd09f92d865dababe8525dfe8b
image verification failed: Error: signature verification failed for all the signatures associated with public.ecr.aws/o7o0w6s5/public-nginx-clone@sha256:bd1578eec775d0b28fd7f664b182b7e1fb75f1dd09f92d865dababe8525dfe8b
```

**Terms of contribution:**

### DIFF
--- a/packages/ecr-credential-helper-shim/Cargo.toml
+++ b/packages/ecr-credential-helper-shim/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "ecr-credential-helper-shim"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[package.metadata.build-package]
+source-groups = [ "ecr-credential-helper-shim" ]
+
+[lib]
+path = "../packages.rs"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
--- a/sources/ecr-credential-helper-shim/src/main.rs
+++ b/sources/ecr-credential-helper-shim/src/main.rs
@@ -1,0 +1,21 @@
+use std::env;
+use std::process::{self, Command};
+
+fn main() {
+    let godebug = env::var("GODEBUG")
+        .unwrap_or_default()
+        .replace("fips140=only", "fips140=on");
+
+    let status = Command::new("/usr/libexec/docker-credential-ecr-login")
+        .args(env::args().skip(1))
+        .env("GODEBUG", &godebug)
+        .status();
+
+    match status {
+        Ok(s) => process::exit(s.code().unwrap_or(1)),
+        Err(err) => {
+            eprintln!("Failed to exec /usr/libexec/docker-credential-ecr-login: {err}");
+            process::exit(1);
+        }
+    }
+}
--- a/sources/notation-image-verifier/main.go
+++ b/sources/notation-image-verifier/main.go
@@ -64,6 +64,7 @@ func main() {
 	// Bottlerocket does not have a $HOME set by default and notation expect to find
 	// credentials from the ecr-credential-helper here.
 	os.Setenv("HOME", "/root")
+	os.Setenv("GODEBUG", "fips140=on")
 
 	cmd := exec.Command("notation", "verify", imageRef)
 	output, err := cmd.CombinedOutput()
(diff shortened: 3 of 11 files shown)


sources/ecr-credential-helper-shim/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+
+fn main() {
+    let godebug = env::var("GODEBUG")
+        .unwrap_or_default()
+        .replace("fips140=only", "fips140=on");
+
+    let status = Command::new("/usr/libexec/docker-credential-ecr-login")
+        .args(env::args().skip(1))
+        .env("GODEBUG", &godebug)
+        .status();
```
arnaldo2792, line 12: Rather than executing the call of the helper in a separate process (fork), you should `exec` the call instead:

https://www.man7.org/linux/man-pages/man3/exec.3.html

So that you don't have to capture the status, and you don't have to worry about handling STDOUT/STDERR in case they require special handling by whatever is calling your shim.

sources/notation-image-verifier/main.go
```diff
@@ -64,6 +64,7 @@ func main() {
 	// Bottlerocket does not have a $HOME set by default and notation expect to find
 	// credentials from the ecr-credential-helper here.
 	os.Setenv("HOME", "/root")
+	os.Setenv("GODEBUG", "fips140=on")
```
arnaldo2792, line 67: Will notation  break if `fips140=only`? Or did you do this to guarantee that the `notation` process that calls your shim also enforces `fips140=on`? If the only thing that breaks is the ECR helper, this shouldn't be needed.

packages/ecr-credential-helper-shim/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
```
arnaldo2792, line 1: Can you clarify why a new package rather than using the OS package?

sky1122:
Thanks for the review!!! That answer a lot of my questions and pros and cons. 
Close this PR in favor of https://github.com/awslabs/amazon-ecr-credential-helper/pull/1064

PR #837, opened 2026-02-17
Update Cargo.lock to bump the dependencies

ytsssun:
**Issue number:**
N/A

**Description of changes:**
Bump the Cargo.lock in sources to update the dependencies.

**Testing done:**
- AMI built.
- Launched a `aws-k8s-1.35` that joins the cluster. 
- Verified `pluto` and `cfsignal` components works:
```
bash-5.2# journalctl -u pluto
Feb 17 23:22:31 localhost systemd[1]: Starting Generate additional settings for Kubernetes...
Feb 17 23:22:31 localhost settings-committer[1568]: 23:22:31 [INFO] Committing settings.
Feb 17 23:22:31 localhost systemd[1]: Finished Generate additional settings for Kubernetes.
bash-5.2# journalctl -u cfsignal
Feb 17 23:22:32 localhost systemd[1]: Started Send signal to CloudFormation Stack.
Feb 17 23:22:35 ip-192-168-16-235.us-west-2.compute.internal cfsignal[1630]: 23:22:35 [INFO] System status is: running [0]
Feb 17 23:22:35 ip-192-168-16-235.us-west-2.compute.internal systemd[1]: cfsignal.service: Deactivated successfully.
bash-5.2#
```

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -59,7 +59,7 @@ dependencies = [
  "actix-macros",
  "actix-rt",
  "actix_derive",
- "bitflags 2.9.4",
+ "bitflags 2.11.0",
  "bytes",
  "crossbeam-channel",
  "futures-core",
@@ -81,7 +81,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
- "bitflags 2.9.4",
+ "bitflags 2.11.0",
  "bytes",
  "futures-core",
  "futures-sink",
@@ -94,16 +94,16 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.11.1"
+version = "3.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "44cceded2fb55f3c4b67068fa64962e2ca59614edc5b03167de9ff82ae803da0"
+checksum = "7926860314cbe2fb5d1f13731e387ab43bd32bca224e82e6e2db85de0a3dba49"
 dependencies = [
  "actix-codec",
  "actix-rt",
  "actix-service",
  "actix-utils",
- "base64 0.22.1",
- "bitflags 2.9.4",
+ "base64",
+ "bitflags 2.11.0",
  "bytes",
  "bytestring",
  "derive_more",
@@ -134,7 +134,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.106",
+ "syn 2.0.116",
 ]
 
 [[package]]
@@ -264,7 +264,7 @@ checksum = "b6ac1e58cded18cb28ddc17143c4dea5345b3ad575e14f32f66e4054a56eb271"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.106",
+ "syn 2.0.116",
 ]
 
 [[package]]
@@ -288,7 +288,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.16",
+ "getrandom 0.2.17",
  "once_cell",
  "version_check",
 ]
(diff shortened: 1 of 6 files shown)


ytsssun:
Force pushed changes to address clippy warning and license check failure.

sources/Cargo.lock
```diff
@@ -6792,11 +6810,13 @@ checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
 
 [[package]]
 name = "uuid"
-version = "1.18.1"
+version = "1.21.0"
```
jpculp: It looks like if you hold off on 1.21.0 and bump to 1.20.0 instead you can avoid pulling in a another version of `getrandom` and some other dependencies.

sources/api/pluto/src/aws.rs
```diff
@@ -25,7 +25,7 @@ pub(crate) async fn sdk_config(region: &str) -> SdkConfig {
         .imds_client(sdk_imds_client())
         .build()
         .await;
-    aws_config::defaults(BehaviorVersion::v2025_08_07())
+    aws_config::defaults(BehaviorVersion::v2026_01_12())
```
cbgbt, line 28: As far as I can tell, the big change here is that retries are enabled by default.

Do we need to re-consider any of our own outer retry loops when taking this?

In other wrods, I'm curious if this is actually meaningfully changing the semantic meaning of some of our API calls in here.

ytsssun:
Pinned `uuid` to 1.20.0 per @jpculp . This would avoid the duplicate version of `getrandom`. It also would avoid introduction of a few other crates.

sources/api/pluto/src/aws.rs
ytsssun: Good question, I think for `pluto` we have explicitly configured the retry behavior, so the default behavior will not have effect to it - https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/sources/api/pluto/src/aws.rs#L19-L21

For `cfsignal` it would be interesting, because we right now do not have any explicit retry config for the client - https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/16f54fff16eed78e4ac45491ee174441a0fb7a59/sources/cfsignal/src/cloudformation.rs#L25-L28. 

`apiclient` is even more interesting - we are always using the Latest() https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/16f54fff16eed78e4ac45491ee174441a0fb7a59/sources/api/apiclient/src/tls_resolvers.rs#L41-L45, so with any SDK bump, it just pick the newest one. 

This change is to address a warning from clippy for deprecated `BehaviorVersion`. To be safe, we can also suppress that with `#[allow(deprecated)]`. But then we might want to update `apiclient` with a specific version instead of using `Latest()`.

ytsssun:
Follow up on @cbgbt 's comment - I can confirm that we do not have any custom retry logic in `cfsignal` and `apiclient` where we make `aws` API calls via the rust sdk. This means the default retries introduced in `BehaviorVersion::v2026_01_12()` is purely beneficial and has no bad semantic meaning change of our API calls.

And `pluto` is not affected as we have configured retry_config() explicitly.

However, there is a finding out of this - we were using `BehaviorVersion::latest()` in `apiclient`, which we probably shouldn't. So I update the PR to pin the version there to `BehaviorVersion::v2026_01_12()`.

jpculp (approved):

sources/api/pluto/src/aws.rs
cbgbt: I think we should pin to a specific behavior version that isn't latest, that way we can reason about the new behavior whenever we make the switch in this code.

If it's possible, we should try to get clippy to deny the use of `latest()`. We do this for a few things in twoliter: https://github.com/bottlerocket-os/twoliter/blob/develop/clippy.toml

ytsssun: Yes, this PR would enforce that - https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/839

ytsssun:
> Follow up on @cbgbt 's comment - I can confirm that we do not have any custom retry logic in `cfsignal` and `apiclient` where we make `aws` API calls via the rust sdk. This means the default retries introduced in `BehaviorVersion::v2026_01_12()` is purely beneficial and has no bad semantic meaning change of our API calls.
> 
> And `pluto` is not affected as we have configured retry_config() explicitly.
> 
> However, there is a finding out of this - we were using `BehaviorVersion::latest()` in `apiclient`, which we probably shouldn't. So I update the PR to pin the version there to `BehaviorVersion::v2026_01_12()`.

And I added separate PR(https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/839) to forbid the use of `BehaviorVersion::latest()` as well per @cbgbt .

sources/api/pluto/src/aws.rs
cbgbt: Nice!

cbgbt (approved):

KCSesh (approved):

PR #851, opened 2026-03-04
fixes for the latest SDK

bcressey:
**Issue number:**
Related: bottlerocket-os/bottlerocket-sdk/pull/334, https://github.com/bottlerocket-os/bottlerocket-sdk/pull/332

**Description of changes:**
Enable minimal debuginfo for Rust crates and glibc.

~~Upgrade the minimum kernel version for glibc to 6.1, to match the new SDK requirement.~~

With symbols included, more binaries run afoul of the FIPS check. Remove the TLS feature from `shibaken`, which doesn't need to make TLS calls. Add a separate build job for crates that use `aws-lc-rs`, to avoid feature unification that would cause it to be used for more binaries.

**Testing done:**
Verified that the build succeeds with the newer SDK.

**Terms of contribution:**

### DIFF
--- a/packages/glibc/glibc.spec
+++ b/packages/glibc/glibc.spec
@@ -71,7 +71,7 @@ Requires: %{name}
 %autosetup -Sgit -n glibc-%{version} -p1
 
 %global glibc_configure %{shrink: \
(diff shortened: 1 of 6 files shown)


bcressey:
⬆️ force push drops the kernel version bump for glibc for now, to make it possible to merge this in advance of the new SDK.

vigh-m (approved):

arnaldo2792 (approved):

PR #863, opened 2026-03-13
Prevent race condition between system services and `/tmp`

arnaldo2792:
**Issue number:**

Related to: https://github.com/bottlerocket-os/bottlerocket/issues/4788

**Description of changes:**

System services with sandboxing options need systemd to create directories under /tmp. If tmp.mount isn't ready when these services start, systemd attempts to create them on the immutable root filesystem, causing startup failures.

Add a global service drop-in that orders all system services after tmp.mount via Wants= and After=, preventing the race.

**Testing done:**

I launched 1000 instances with a custom AMI. Prior this fix, 60% of my launches resulted in failures that hit the race condition with this error:

```
(whippet)[829]: whippet.service: Failed to set up mount namespacing: /run/systemd/unit-root/dev: Read-only file system
```

After this fix, I no longer hit the error.

**Terms of contribution:**

### DIFF
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -111,6 +111,7 @@ Source1109: modprobe-no-exit.conf
 Source1110: tmp-mount-noexec.conf
 Source1111: network-pre-target-dbus-dep.conf
 Source1112: fips-go.conf
+Source1113: requires-tmp.conf
 
 # network link rules
 Source1200: 80-release.link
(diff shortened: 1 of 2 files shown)


JacobHenner:
What build process did you follow to introduce this change into a custom AMI?

arnaldo2792:
The process is slight involved, but here it is:

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/BUILDING.md

That covers how you can build the core kit with my change (you will have pull down my change) and use it in your own variant build.

JacobHenner:
> The process is slight involved, but here it is:
> 
> https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/BUILDING.md
> 
> That covers how you can build the core kit with my change (you will have pull down my change) and use it in your own variant build.

Thanks - do you have a rough sense for when this fix will make its way to the distributed EKS AMIs?

bcressey (approved):

vigh-m (approved):

PR #868, opened 2026-03-17
selinux-policy: limit interactions between container runtimes and containers

bcressey:
**Issue number:**
N/A

**Description of changes:**
The overall goal is to limit interactions between container runtimes and containers to just the pipes used for logging output, thereby reducing the exposed surface area.

In policy terms, that's accomplished by this change:
```
 ; Allow containers to communicate with runtimes via pipes.
-(allow container_s runtime_t (files (mutate)))
+(allow container_s runtime_fifo_t (files (mutate)))
```

For that to work, it must be possible to give pipes a different label than that of the parent process, which in turn requires other changes.

The first set of changes enables the use of `glblub` ("greatest lower bound, lowest upper bound") as the default range for file types, and adjusts other range transitions so that everything continues to get the intended range.
```
selinux-policy: use single level for containers
host-ctr: use single SELinux level for processes
selinux-policy: adjust level for container runtimes
selinux-policy: update to policy version 32
selinux-policy: use "glblub" for the default range
```

The second logical change is enabling transition rules for `pipefs`, rather than using task-labeling, so that pipes can receive a different label. We also separate out the ability to "use" or receive file descriptors from the broader "load" permission. Permissions for ioctls are split out as well to prepare for future restrictions.
```
selinux-policy: use transition SIDs for pipefs
selinux-policy: split out perms for ioctls and fds
```

Finally, pipes are given a newly created label that both container runtimes and containers can read and write to. Since containers no longer need to read and write to the container runtime type, such access can then be explicitly blocked.
```
selinux-policy: relabel pipes for container runtimes
selinux-policy: restrict reads for container runtimes
```

**Testing done:**
Ran the SELinux test suite to check for process and file labels. The only change there was the expected one; privileged containers now run as `system_u:system_r:control_t:s0:c0.c1023` instead of `system_u:system_r:control_t:s0-s0:c0.c1023`.

Pipe labels before:
```
# find /proc/*/fd -lname 'pipe:*' -exec ls -1LZ {} + 2>/dev/null | grep -oP '^\s*\S+' | tr -d ' ' | sort -u
system_u:system_r:container_t:s0:c560,c574
system_u:system_r:control_t:s0-s0:c0.c1023
system_u:system_r:runtime_t:s0
system_u:system_r:super_t:s0-s0:c0.c1023
system_u:system_r:super_t:s0:c493,c600
```

Pipe labels after:
```
# find /proc/*/fd -lname 'pipe:*' -exec ls -1LZ {} + 2>/dev/null | grep -oP '^\s*\S+' | tr -d ' ' | sort -u
system_u:object_r:container_t:s0:c134,c708
system_u:object_r:control_t:s0:c0.c1023
system_u:object_r:runtime_fifo_t:s0
system_u:object_r:super_t:s0:c0.c1023
system_u:object_r:super_t:s0:c13,c1000
```

FIFO labels before and after:
```
# find / -type p -exec ls -1LZ {} + 2>/dev/null | grep -oP '^\s*\S+' | sort -u
system_u:object_r:any_t:s0
```

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/base.cil
+++ b/packages/selinux-policy/base.cil
@@ -9,11 +9,21 @@
 (sensitivityorder (s0))
 (sensitivitycategory s0 (range c0 c1023))
 
+; A "range" of one: one sensitivity, no categories. Used for contexts
+; where categories don't matter.
 (level s0 (s0))
 (levelrange s0 (s0 s0))
 
+; A range from one sensitivity with no categories, through one
+; sensitivity with all categories. Used for statements that need to
+; cover the entire range of possibilities.
 (level s0-s0 (s0 (range c0 c1023)))
 (levelrange s0-s0 (s0 s0-s0))
+
+; A "range" of one: one sensitivity with all categories. Used for
+; contexts where categories matter, and so the same sensitivity with
+; no categories cannot be included in the range.
+(levelrange s0-c0-c1023 (s0-s0 s0-s0))
 
 ; The system_r role is used for processes.
 (role system_r)
--- a/packages/selinux-policy/base.cil
+++ b/packages/selinux-policy/base.cil
@@ -38,12 +38,26 @@
 (userlevel system_u s0-s0)
 (userrange system_u s0-s0)
 
-; Take the context from the target file rather than the source
-; process when computing the level for new file objects. We can
-; expect the directory where files are created to have the right
-; range of categories applied, but the process creating the file
-; may be privileged and have the full range or no range at all.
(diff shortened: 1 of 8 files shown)


bcressey:
⬆️ force push to fix merge conflict via rebase on develop

arnaldo2792 (approved):

KCSesh (approved):

mgsharm (approved):

PR #883, opened 2026-03-26
rottweiler: add directory protection

bcressey:
**Issue number:**

**Description of changes:**
Extend `rottweiler` with the ability to lock down directories by remounting the underlying filesystem read-only, and installing a BPF LSM hook to prevent subsequent remounts, unmounts, or overmounts.

Label the directory on `bpffs` such that only API processes can modify the map used by the BPF LSM hook.

Update the lockdown API to also lock `/etc` and `/.bottlerocket/datastore`.

**Testing done:**
```
# fails for non-API user
bash-5.3# rw protect directory /.bottlerocket/datastore/
libbpf: failed to pin map: -EACCES
libbpf: map 'protected_devices': failed to auto-pin at '/sys/fs/bpf/rottweiler/protected_devices': -EACCES
libbpf: map 'protected_devices': failed to create: -EACCES
libbpf: failed to load BPF skeleton 'protect_dirs_bpf': -EACCES
Error: failed to load BPF program

Caused by this error:
  1: Permission denied (os error 13)

# succeeds via API
bash-5.3# apiclient lockdown
22:58:32 [INFO] Lockdown completed

# map can't be read by non-API user
bash-5.3# rw check directory /.bottlerocket/datastore protected
Error: failed to open pinned protected_mounts map

Caused by this error:
  1: Permission denied (os error 13)

# capture messages logged by BPF LSM hook
bash-5.3# cat /sys/kernel/debug/tracing/trace_pipe &
[1] 1920

bash-5.3# umount /.bottlerocket
          umount-1923    [000] ....1   599.620999: bpf_trace_printk: Blocked umount on protected device 265289767
umount: /.bottlerocket: must be superuser to unmount.

bash-5.3# mount -oremount,rw /.bottlerocket/
           mount-1924    [000] ....1   609.010153: bpf_trace_printk: Blocked remount on protected device 265289767
mount: /.bottlerocket: permission denied.
       dmesg(1) may have more information after failed mount system call.

bash-5.3# mount -t tmpfs none /.bottlerocket/
           mount-1925    [000] ....1   621.933652: bpf_trace_printk: Blocked mount over protected device 265289767
mount: /.bottlerocket: permission denied.
       dmesg(1) may have more information after failed mount system call.
```

**Terms of contribution:**

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -3,7 +3,7 @@ description: "Performs setup for caching and other common needs."
 runs:
   using: "composite"
   steps:
-    - run: sudo apt -y install build-essential openssl libssl-dev pkg-config liblz4-tool clang
+    - run: sudo apt -y install build-essential openssl libbpf-dev libelf-dev libssl-dev pkg-config liblz4-tool clang
       shell: bash
     - uses: actions/cache/restore@v4
       # Restore most recent cache if available.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1082,6 +1082,7 @@ name = "rottweiler"
 version = "0.1.0"
 dependencies = [
  "glibc",
+ "libbpf",
 ]
 
 [[package]]
--- a/packages/rottweiler/Cargo.toml
+++ b/packages/rottweiler/Cargo.toml
@@ -16,3 +16,4 @@ source-groups = [
 
 [build-dependencies]
 glibc = { path = "../glibc" }
+libbpf = { path = "../libbpf" }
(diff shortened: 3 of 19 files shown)


bcressey:
Keeping this in draft as there's a significant blocker: the system fails to boot if `apiclient lockdown` is applied via bootstrap command, since then `/etc` becomes read-only and systemd cannot update its default targets.

I'm also unsure whether the recursive remount under `/etc` is truly desirable; for example, it means that `/etc/cni/net.d` could not be populated by a CNI plugin later.

Issue #806, opened 2026-01-20
Upgrade to containerd 2.2


vrdn-23:
**What I'd like:**

Curious to know what the timeline/ETA for the Containerd 2.2 upgrade looks like 

Particularly interested in the speed-ups available through https://github.com/containerd/containerd/pull/12332

From https://github.com/containerd/containerd/issues/8881
>If user can improve disk performance, the image pull latency by this proposal can be reduced significantly. For example, Tao was able to achieve 3X faster image pull (120 seconds -> 40 seconds) for pulling a popular container: [gcr.io/deeplearning-platform-release/base-cu113:m106](http://gcr.io/deeplearning-platform-release/base-cu113:m106) (5.4 GB) with a common setup of a deep learning node (2500 GB PD-SSD, 32 vCPUs).

**Any alternatives you've considered:**

vrdn-23:
cc @piyush-jena @ytsssun

vrdn-23:
Any update on this @piyush-jena @ytsssun @KCSesh

ytsssun:
Hi @vrdn-23, thanks for your interests here! we are actively working on adding containerd 2.2. We will use this issue to track the progress.

PR #886, opened 2026-03-30
Containerd 2.2: support setting max-concurrent-unpacks

ytsssun:
**Issue number:**

Related: https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/806

**Description of changes:**

Expose containerd 2.2's `max_concurrent_unpacks` transfer service setting via `settings.container-runtime.max-concurrent-unpacks`, and add warning infrastructure for variants using older containerd versions.

**Containerd 2.2 templates:**
- Added `{{#if settings.container-runtime.max-concurrent-unpacks}}` block to `containerd-config-toml_k8s_containerd_sock` and `containerd-config-toml_k8s_nvidia_containerd_sock`, rendering `max_concurrent_unpacks` under `[plugins."io.containerd.transfer.v1.local"]`

**Unsupported setting warning (containerd 1.7 and 2.1):**
- New `unsupported-setting-warning@.service` and `unsupported-setting-warning@.timer` systemd template units in the release package (same pattern as `deprecation-warning@`)
- Warning template `max-concurrent-unpacks-unsupported` added to containerd-1.7 and containerd-2.1 packages — writes an env file to `/etc/unsupported-settings/` when the setting is configured
- Timer fires on startup and every 6 hours, logging: `The setting "container-runtime.max-concurrent-unpacks" is not supported by the containerd version in this variant and has no effect.`

Depends on https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/119 for the settings model.

**Testing done:**

Built aws-k8s-1.35 variant with both containerd 2.2 and containerd 2.1, using the setting introduced in https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/119. Launched EC2 instances with `settings.container-runtime.max-concurrent-unpacks = 5` in user data.

**Containerd 2.2:**
- `/etc/containerd/config.toml` renders `max_concurrent_unpacks = 5` under `[plugins."io.containerd.transfer.v1.local"]` ✅
- No warning in journal ✅

**Containerd 2.1:**
- `max_concurrent_unpacks` not rendered in containerd config ✅
- Timer active, journal warning: `The setting container-runtime.max-concurrent-unpacks is not supported by the containerd version in this variant and has no effect.` ✅

**Terms of contribution:**

### DIFF
--- a/packages/containerd-2.2/containerd-config-toml_k8s_containerd_sock
+++ b/packages/containerd-2.2/containerd-config-toml_k8s_containerd_sock
@@ -46,6 +46,9 @@ enable_unprivileged_icmp = {{settings.container-runtime.enable-unprivileged-icmp
 {{#if settings.container-runtime.max-concurrent-downloads}}
 max_concurrent_downloads = {{settings.container-runtime.max-concurrent-downloads}}
 {{/if}}
+{{#if settings.container-runtime.max-concurrent-unpacks}}
+max_concurrent_unpacks = {{settings.container-runtime.max-concurrent-unpacks}}
+{{/if}}
 concurrent_layer_fetch_buffer = {{default 0 settings.container-runtime.concurrent-download-chunk-size}}
 
 [[plugins."io.containerd.transfer.v1.local".unpack_config]]
--- a/packages/release/unsupported-setting-warning@.service
+++ b/packages/release/unsupported-setting-warning@.service
@@ -1,0 +1,8 @@
+[Unit]
+Description=Unsupported setting warning for %i
+
+[Service]
+Type=simple
+EnvironmentFile=/etc/unsupported-settings/%i
+ExecStart=/usr/bin/logger -p4 The setting "${UNSUPPORTED_SETTING}" is not supported and has no effect. ${UNSUPPORTED_REASON}
+StandardError=journal+console
--- a/packages/containerd-1.7/containerd-1.7.spec
+++ b/packages/containerd-1.7/containerd-1.7.spec
@@ -24,6 +24,7 @@ Source4: containerd-config-toml_k8s_nvidia_containerd_sock
 Source5: containerd-tmpfiles.conf
 Source6: containerd-cri-base-json
 Source7: snapshotter-toml
+Source8: max-concurrent-unpacks-unsupported
 
 # Mount for writing containerd configuration
 Source100: etc-containerd.mount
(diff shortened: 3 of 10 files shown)


ytsssun:
Putting to draft as the depending https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/119 is not yet merged.

vigh-m:
I'm wondering if there is a way to piggyback on what already exists for `deprecation-warning@` instead of creating new timers and services.
We can potentially put the `unsupported-setting@.service` in the same path as the `deprecation-warning` and reuse the time?

ytsssun:
> I'm wondering if there is a way to piggyback on what already exists for `deprecation-warning@` instead of creating new timers and services. We can potentially put the `unsupported-setting@.service` in the same path as the `deprecation-warning` and reuse the time?

Yeah, the implementation is basically mimicking what we did for `deprecation-warning`, but you can probably tell that semantically, it has different meaning here - the setting `max-concurrent-unpacks` is not a deprecated setting, instead it is a new setting that the older containerd versions does not support.

And due to the difference in the semantic of these two, the warning message will be different, in which case a separate service makes sense to me. We can arguably reuse the timer, but `unsupported-setting-warning@.timer` seems more consistent here?

vigh-m (approved):

vigh-m:
That's fair. The consistency of naming over code reuse is an acceptable tradeoff. :)

packages/release/unsupported-setting-warning@.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Unsupported setting warning for %i
+
+[Service]
+Type=simple
+EnvironmentFile=/etc/unsupported-settings/%i
+ExecStart=/usr/bin/logger -p4 The setting "${UNSUPPORTED_SETTING}" is not supported by the containerd version in this variant and has no effect.
```
arnaldo2792: This is not correct, the purpose of a template is to be agnostic and not specific to containerd. You will have to control the message through the environment variable from the `EnvironmentFile`.

arnaldo2792 (changes requested):

packages/containerd-2.2/containerd-config-toml_k8s_containerd_sock
```diff
@@ -46,6 +46,9 @@ enable_unprivileged_icmp = {{settings.container-runtime.enable-unprivileged-icmp
 {{#if settings.container-runtime.max-concurrent-downloads}}
```
arnaldo2792, line 46: The commit messages in the first two commits don't follow the convention followed by the project:

```
<subsystem>: description
```

For this commit, it should be something like:

```
api: expose max-concurrent-unpacks for containerd 2.2
```

piyush-jena:
The commits are a bit messed up on this one. The settings-sdk bump has schnauzer fixes as well. You may rebase not as my PR with the settings-sdk is merged.

packages/release/unsupported-setting-warning@.service
ytsssun: Updated. Nice catch!

ytsssun:
> The commits are a bit messed up on this one. The settings-sdk bump has schnauzer fixes as well. You may rebase not as my PR with the settings-sdk is merged.

Yes, just dropped the commit in my PR for the settings sdk bump.

packages/containerd-2.2/containerd-config-toml_k8s_containerd_sock
ytsssun: Updated

ytsssun:
Force pushed drop the commits to bump settings sdk as that is merged along with https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/901. 

Also addressed comments per @arnaldo2792

piyush-jena (approved):
nice

arnaldo2792 (approved):

PR #890, opened 2026-04-03
feat(containerd): plumb cgroup_writable setting into config templates

bilby91:
## Summary
- Adds `cgroup_writable` to containerd 2.1 and 2.2 k8s config templates (standard and NVIDIA)
- The setting is placed under each runtime handler section (`runtimes.runc`, `runtimes.nvidia`, etc.) matching containerd's config structure
- Only rendered when explicitly set by the user — no default value

Relates to: https://github.com/bottlerocket-os/bottlerocket/issues/4666
Companion PR: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/128
See also: https://github.com/containerd/containerd/pull/11131

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/containerd-2.1/containerd-config-toml_k8s_containerd_sock
+++ b/packages/containerd-2.1/containerd-config-toml_k8s_containerd_sock
@@ -62,6 +62,9 @@ platform = "linux/amd64"
 [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc]
 runtime_type = "io.containerd.runc.v2"
 base_runtime_spec = "/etc/containerd/cri-base.json"
+{{#if settings.container-runtime.cgroup-writable}}
+cgroup_writable = {{settings.container-runtime.cgroup-writable}}
+{{/if}}
 
 [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc.options]
 SystemdCgroup = true
--- a/packages/containerd-2.1/containerd-config-toml_k8s_containerd_sock
+++ b/packages/containerd-2.1/containerd-config-toml_k8s_containerd_sock
@@ -2,7 +2,7 @@
 container-registry = "v1"
 container-runtime = "v1"
 kubernetes = "v1"
-std = { version = "v1", helpers = ["join_array", "default"]}
+std = { version = "v1", helpers = ["join_array", "default", "is_cgroup_v2"]}
 +++
 version = 3
 root = "/var/lib/containerd"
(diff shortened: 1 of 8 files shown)


packages/containerd-2.1/containerd-config-toml_k8s_containerd_sock
```diff
@@ -62,6 +62,9 @@ platform = "linux/amd64"
 [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc]
 runtime_type = "io.containerd.runc.v2"
 base_runtime_spec = "/etc/containerd/cri-base.json"
+{{#if settings.container-runtime.cgroup-writable}}
+cgroup_writable = {{settings.container-runtime.cgroup-writable}}
+{{/if}}
```
bcressey, line 68: At a minimum, this needs to be scoped to nodes where cgroup v2 is used. That's the default on Bottlerocket but it's still possible to switch back to v1. For that, I'd recommend adding a guard helper to `schnauzer`.

I'm also skeptical that it makes sense to enable this system-wide; if a container isn't prepared to lock down the delegated hierarchy, it could be exposed to additional risks.

Better would be a per-pod annotation so that individual pods can opt-in.

bilby91: @bcressey  Thanks for the review!

For concern 1, I agree — I'll add an is_cgroup_v2 guard helper in schnauzer (following the existing pattern for system-detection helpers like fips_enabled) so this only takes effect on nodes running cgroup v2.

Regarding concerns 2 and 3, I've been looking into this but I'm still getting familiar with the full architecture. From what I can tell, a per-pod approach could work via Kubernetes RuntimeClasses — containerd already supports per-runtime cgroup_writable values and maps them from the pod's runtimeClassName through GetSandboxRuntime(). So we could define a second runtime (e.g. runc-cgroup-writable) in the containerd config template with the flag enabled, keeping it off by default and letting pods opt in. This wouldn't require upstream changes.

I also noticed that an NRI plugin could potentially achieve this by intercepting CreateContainer events and replacing the cgroup mount options based on pod annotations, though Bottlerocket doesn't ship any NRI plugins today so that would be a bigger lift.

I'm not deeply familiar with how Bottlerocket typically handles this kind of per-pod configuration, so I'd really appreciate your guidance on which direction makes sense here — or if there's another approach I'm not seeing.

bilby91: @bcressey Any path we can explore to continue this conversation ?

Thanks!

PR #889, opened 2026-04-02
Third party packages update

maherthomsi:
**Description of changes:**
Updates the following package:
* aws-iam-authenticator to 0.7.11

**Testing done:**
Ensured I was able to build and launch the core-kit with these changes

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #896, opened 2026-04-08
sdk: update following Go version bump

vigh-m:
**Description of changes:**
- Depends on https://github.com/bottlerocket-os/bottlerocket-sdk/pull/345
- Updates `host-ctr` to use Go v1.26
- Add the rc build of `kubernetes-1.36`

**Testing done:**
- Tested building both arches and launching AMIs
- Conformance testing of AMIs
- Tested the `k8s-1.36` rc by building an AMI and verifying that it launches

**ToDo:**
- [x] Bump the SDK version after release

**Terms of contribution:**


vigh-m:
From reading the [Go 1.26 release notes](https://go.dev/doc/go1.26), I think the below changes are worth noting:

---

### New garbage collector

The Green Tea garbage collector, previously available as an experiment in Go 1.25, is now enabled by default after incorporating feedback.

This garbage collector’s design improves the performance of marking and scanning small objects through better locality and CPU scalability. Benchmark results vary, but we expect somewhere between a 10–40% reduction in garbage collection overhead in real-world programs that heavily use the garbage collector. Further improvements, on the order of 10% in garbage collection overhead, are expected when running on newer amd64-based CPU platforms (Intel Ice Lake or AMD Zen 4 and newer), as the garbage collector now leverages vector instructions for scanning small objects when possible.

The new garbage collector may be disabled by setting GOEXPERIMENT=nogreenteagc at build time. This opt-out setting is expected to be removed in Go 1.27. If you disable the new garbage collector for any reason related to its performance or behavior, please file an issue.

### Compiler

The compiler can now allocate the backing store for slices on the stack in more situations, which improves performance. If this change is causing trouble, the bisect tool can be used to find the allocation causing trouble using the -compile=variablemake flag. All such new stack allocations can also be turned off using -gcflags=all=-d=variablemakehash=n. If you encounter issues with this optimization, please file an issue.

piyush-jena (approved):

ytsssun (approved):
LGTM, assuming CI passes

vigh-m:
Pushed the changes from https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/889 to test the SDK

vigh-m:
⬆️ Updated SDK

piyush-jena (approved):

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -33,7 +33,6 @@ Summary: A statically-linked agent to enable remote management of EC2 instances
 %set_cross_go_flags
```
arnaldo2792, line 34: The commit message should read:

```
packages: remove hardcoded GO_MAJOR

With the SDK <>, the hardcoded GO_MAJOR isn't needed.
```

packages/ecr-credential-provider-1.36/0001-ecr-credential-provider-hardcode-ECR-endpoint-for-eu.patch
```diff
@@ -0,0 +1,31 @@
...
+@@ -75,7 +75,11 @@ func defaultECRProvider(ctx context.Context, region string) (ECR, error) {
+ 		return nil, err
+ 	}
+ 
+-	return ecr.NewFromConfig(cfg), nil
++	return ecr.NewFromConfig(cfg, func(o *ecr.Options) {
++		if region == "eusc-de-east-1" {
++			o.BaseEndpoint = aws.String("https://api.ecr.eusc-de-east-1.amazonaws.eu")
++		}
++	}), nil
```
arnaldo2792, line 25: Do you know if there are any GH issues to ask them to fix this? Or what's the path forward out of the patch.

packages/aws-iam-authenticator/aws-iam-authenticator.spec
```diff
@@ -2,7 +2,7 @@
 %global gorepo aws-iam-authenticator
 %global goimport %{goproject}/%{gorepo}
```
arnaldo2792, line 3: Is this update needed for your bump? If this is just a bump in version, keep it in a separate PR.

packages/binutils/binutils.spec
```diff
@@ -1,5 +1,5 @@
 Name: %{_cross_os}binutils
-Version: 2.43.1
```
arnaldo2792, line 2: I understand this change is needed. You don't have to keep track where did you get this change from (the cherry-pick). Just keep authorship.

advisories/staging/BRSA-ljpndu5ez903.toml
```diff
@@ -0,0 +1,17 @@
+[advisory]
```
arnaldo2792, line 1: Same as elsewhere, just keep the authorship.

packages/aws-iam-authenticator/aws-iam-authenticator.spec
vigh-m: No. These are all the changes I pulled from https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/889. I can drop them as they are not needed

packages/ecr-credential-provider-1.36/0001-ecr-credential-provider-hardcode-ECR-endpoint-for-eu.patch
vigh-m: Dropped the change. We can investigate this outside of this PR

packages/aws-iam-authenticator/aws-iam-authenticator.spec
vigh-m: Dropped this change

packages/binutils/binutils.spec
vigh-m: Gotcha. Dropped the tracking commit message

vigh-m:
⬆️ Updated based on comments

vigh-m:
⬆️ Fixed typo in commit message

vigh-m:
⬆️ Fixed error in the rebase

vigh-m:
⬆️ Removed the `nvidia-container-toolkit` update

arnaldo2792 (approved):

PR #908, opened 2026-04-14
Make FIPS activation runtime-conditional on fips=1 kernel param

sparrc:
The overall purpose of this change is to include most FIPS logic in all BR variants, and to standardize on all of the systemd services only being activated with `ConditionKernelCommandLine=fips=1`, as some already are: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/748058e55349b1b1ea32d4d4cd304c07bcdcf9cd/packages/release/check-kernel-integrity.service#L9

This gets closer to the "standard" BR AMI being able to be used in FIPS environments, only needing a runtime switch to activate FIPS mode like the following userdata:

```
[settings.boot] 
reboot-to-reconcile = true 
[settings.boot.kernel-parameters] 
fips = ["1"]
```

**Description of changes:**

Move FIPS-related systemd units and drop-ins from the release-fips subpackage into the base release package so they ship in all variants. These units already use ConditionKernelCommandLine=fips=1 or depend on fipscheck.target, so they are no-ops on non-FIPS boots.

Replace the unconditional Environment=GODEBUG=fips140=on drop-in with an EnvironmentFile=-/run/fips-go.env approach: a new service copies the env file into /run only when fips=1 is set. Similarly, replace the tmpfiles-based /etc/system-fips creation with a conditional service.

The bootconfig snippet and FIPS binary overlay mounts remain in the fips subpackage since they require build-time binary selection.

Assisted-by: Cline:anthropic.claude-opus-4-6-v1

**Testing done:**

WIP

**Terms of contribution:**

### DIFF
--- a/packages/release/create-fips-marker.service
+++ b/packages/release/create-fips-marker.service
@@ -1,0 +1,17 @@
+# This file creates the /etc/system-fips marker file. This file can
+# be mounted into containers for certain distros (RHEL, Amazon Linux)
+# to trigger the use of FIPS certified crypto.
+
+[Unit]
+Description=Create FIPS system marker file
+DefaultDependencies=no
+ConditionKernelCommandLine=fips=1
+Before=sysinit.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/touch /etc/system-fips
+RemainAfterExit=true
+
+[Install]
+WantedBy=sysinit.target
--- a/packages/release/fips-go.conf
+++ b/packages/release/fips-go.conf
@@ -1,6 +1,6 @@
 [Service]
-# Enable Go FIPS 140-3 fips140=on mode. When enabled, Go's crypto packages use the
-# NIST DRBG for randomness, crypto/tls only negotiates FIPS-approved
-# algorithms, and mandatory self-tests run on initialization.
-# See: https://go.dev/blog/fips140
-Environment=GODEBUG=fips140=on
+# When FIPS mode is enabled at runtime via kernel parameter fips=1,
+# the generate-fips-env.service creates /run/fips-go.env with
+# GODEBUG=fips140=on. The "-" prefix means this is optional and
+# will not cause failures if the file does not exist (non-FIPS boot).
+EnvironmentFile=-/run/fips-go.env
--- a/packages/release/release-fips-tmpfiles.conf
+++ b/packages/release/release-fips-tmpfiles.conf
@@ -1,2 +1,2 @@
-f+ /etc/system-fips 0644 root root -
+# Clean up FIPS module check state files from fips-modprobe@.service.
 r /etc/.fips-modprobe.*
--- a/packages/release/release-fips-tmpfiles.conf
+++ b/packages/release/release-fips-tmpfiles.conf
@@ -1,2 +1,1 @@
-# Clean up FIPS module check state files from fips-modprobe@.service.
 r /etc/.fips-modprobe.*
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -143,6 +143,12 @@ Source1650: prepare-local-fs-encrypted.conf
 Source1651: local-mount-encrypted.conf
 Source1652: repart-local-encrypted.conf
 
+# Runtime FIPS activation sources.
+# These are always installed but only activate when fips=1 is on the kernel command line.
+Source1700: generate-fips-env.service
+Source1701: create-fips-marker.service
+Source1702: fips-go.env
+
 Requires: %{_cross_os}audit
 Requires: %{_cross_os}auditd
 Requires: %{_cross_os}chrony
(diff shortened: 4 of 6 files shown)


packages/release/create-fips-marker.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
sky1122, line 5: can you add a comment on why do we need this `/etc/system-fips` marker file?

like the one include in [here](https://github.com/bottlerocket-os/bottlerocket-core-kit/commit/f3309917d8839dc6146c9f95a0e4507f3abd98f6)

sparrc: pushed new rev with comment in this file

packages/release/release.spec
```diff
@@ -473,20 +485,25 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
...
+# These are always present but are no-ops on non-FIPS boots.
 %{_cross_unitdir}/service.d/00-fips-go.conf
-%{_cross_unitdir}/*-bin.mount
-%{_cross_unitdir}/*-libexec.mount
+%{_cross_unitdir}/generate-fips-env.service
+%{_cross_unitdir}/create-fips-marker.service
+%{_cross_datadir}/bottlerocket/fips-go.env
+%{_cross_tmpfilesdir}/release-fips.conf
 %{_cross_unitdir}/fipscheck.target
 %{_cross_unitdir}/activate-preconfigured.service
```
ginglis13, line 494: These also need

```
ConditionKernelCommandLine=fips=1
```

which @sky1122 has in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/918

We should merge #918 first and rebase this PR

sparrc: Ack 👍

packages/release/release.spec
```diff
@@ -473,10 +485,13 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 %{_cross_datadir}/logdog.d/logdog.common.conf
 %{_cross_unitdir}/configure-snapshotter.service
 %{_cross_templatedir}/selected-snapshotter
-%{_cross_tmpfilesdir}/release-fips.conf
+# Runtime FIPS activation — conditional on fips=1 kernel command line parameter.
+# These are always present but are no-ops on non-FIPS boots.
```
arnaldo2792: Extra comment here not really needed, we have the context from the previous comments

packages/release/fips-go.conf
```diff
@@ -1,6 +1,6 @@
...
-# Enable Go FIPS 140-3 fips140=on mode. When enabled, Go's crypto packages use the
-# NIST DRBG for randomness, crypto/tls only negotiates FIPS-approved
-# algorithms, and mandatory self-tests run on initialization.
-# See: https://go.dev/blog/fips140
-Environment=GODEBUG=fips140=on
+# When FIPS mode is enabled at runtime via kernel parameter fips=1,
+# the generate-fips-env.service creates /run/fips-go.env with
+# GODEBUG=fips140=on. The "-" prefix means this is optional and
+# will not cause failures if the file does not exist (non-FIPS boot).
+EnvironmentFile=-/run/fips-go.env
```
arnaldo2792, line 6: Just for posterity:

I thought that it will be simpler to keep the original file and copy it with a systemd service to `/etc/systemd/system/service.d`. But, it doesn't feel right that a systemd service copies a drop-in for other systemd services (and I don't want to establish the precedent that we should do that).

That said, I like your approach in the PR better because it copies an env file for all systemd services instead of a drop-in for all services.

packages/release/release.spec
```diff
@@ -486,6 +501,8 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 
 %files fips
 %{_cross_bootconfigdir}/10-fips.conf
+%{_cross_unitdir}/*-bin.mount
+%{_cross_unitdir}/*-libexec.mount
```
arnaldo2792: We should keep these in the main package as now they are conditional on the kernel parameter.

packages/release/release-fips-tmpfiles.conf
```diff
@@ -1,2 +1,2 @@
-f+ /etc/system-fips 0644 root root -
+# Clean up FIPS module check state files from fips-modprobe@.service.
```
arnaldo2792: No need for the extra comment.

sparrc: (removed)

packages/release/release.spec
sparrc: done in latest push

packages/release/release.spec
sparrc: (removed)

arnaldo2792 (approved):

sky1122 (approved):

PR #909, opened 2026-04-15
fix: correct advisory version for microcode advisories

ginglis13:
**Description of changes:**

The microcode advisories from 2.5.0 core-kit release incorrectly include
a patched-version. However, version information for this package is
intentionally made "0.0"

See https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/6965cb0d797bd7cab9f7c9b8f597291668aa4626/packages/microcode/microcode.spec#L4-L10

**Terms of contribution:**

### DIFF
--- a/advisories/2.5.0/BRSA-7esrdcy2k3ug.toml
+++ b/advisories/2.5.0/BRSA-7esrdcy2k3ug.toml
@@ -7,7 +7,7 @@ description = "A protection mechanism failure in some 3rd, 4th, and 5th Generati
 
 [[advisory.products]]
 package-name = "microcode"
-patched-version = "20240813"
+patched-version = "0"
 patched-release = "0"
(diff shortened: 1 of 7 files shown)


vigh-m (approved):

advisories/2.5.0/BRSA-7esrdcy2k3ug.toml
```diff
@@ -7,7 +7,7 @@ description = "A protection mechanism failure in some 3rd, 4th, and 5th Generati
```
arnaldo2792, line 7: Commit message should read as:

```
<subsystem>: <description>
```

So for this, you should use something along the lines of:

```
advisories: correct advisory version for microcode advisories
```

piyush-jena (approved):

PR #911, opened 2026-04-16
Build changes to bash and systemd

piyush-jena:
**Description of changes:**
- Bash follows C17 standard which we have hardcoded here. Newer versions of GCC default to newer standards which break compilation.
- Hardcoding mount and umount path in systemd meson build parameters - util-linux package in bottlerocket installs mount and umount to /usr/bin (`%{_cross_bindir}`) and this change makes sure we honour that instead of defaulting to the path used by host OS.

**Testing done:**
1. core-kit builds ✅
2. instance boots ✅ 
3. bash commands work ✅  

**Terms of contribution:**

### DIFF
--- a/packages/systemd-252/systemd-252.spec
+++ b/packages/systemd-252/systemd-252.spec
@@ -23,8 +23,8 @@ Source103: org.freedesktop.systemd1.toml
 # Reference issue: github.com/systemd/systemd/issues/25441
 Patch1001: 1001-sd-netlink-make-calc_elapse-return-USEC_INFINITY-whe.patch
 Patch1002: 1002-sd-netlink-make-the-default-timeout-configurable-by-.patch
-# Backport of upstream patch to change `Failed to execute <filepath> No such file 
-# or directory` error logs to debug 
+# Backport of upstream patch to change `Failed to execute <filepath> No such file
+# or directory` error logs to debug
 Patch1003: 1003-exec-util-make-missing-agents-a-gracefull-handled-issues.patch
 
 # Local patch to work around the fact that /var is a bind mount from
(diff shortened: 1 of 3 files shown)


packages/systemd-252/systemd-252.spec
```diff
@@ -340,6 +340,9 @@ CONFIGURE_OPTS=(
  -Danalyze=true
 
  -Dbpf-framework=false
+
+ -Dmount-path='%{_cross_bindir}/mount'
+ -Dumount-path='%{_cross_bindir}/umount'
```
vigh-m: I think you can bundle these with the other `-path` configs at line 212.

Similar for v257

packages/systemd-252/systemd-252.spec
```diff
@@ -23,8 +23,8 @@ Source103: org.freedesktop.systemd1.toml
 # Reference issue: github.com/systemd/systemd/issues/25441
 Patch1001: 1001-sd-netlink-make-calc_elapse-return-USEC_INFINITY-whe.patch
 Patch1002: 1002-sd-netlink-make-the-default-timeout-configurable-by-.patch
-# Backport of upstream patch to change `Failed to execute <filepath> No such file 
-# or directory` error logs to debug 
+# Backport of upstream patch to change `Failed to execute <filepath> No such file
+# or directory` error logs to debug
```
arnaldo2792, line 27: nit: thanks for the cleanup! but I would have done that in a different commit, not blocking tho.

piyush-jena: updated

arnaldo2792 (approved):

vigh-m (approved):

PR #919, opened 2026-05-05
Normalize NVIDIA library paths to /usr/lib/

maherthomsi:
Merge with https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/425

**Description of changes:**

Normalize NVIDIA library paths in containers so libraries appear at `/usr/lib/` instead of the Bottlerocket cross-compilation sysroot path (`/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/`).

- nvidia-container-toolkit: Add `--additional-symlinks` flag patch to `nvidia-ctk cdi generate` that creates backwards-compatibility symlinks in a specified directory pointing to each discovered library. Configure `generate-cdi-specs.service` with `--additional-symlinks /usr/lib/nvidia/tesla`.
- nvidia-k8s-device-plugin: Set `containerDriverRoot` to the Bottlerocket sysroot path so the device plugin discovers libraries correctly and generates CDI specs with normalized `/usr/lib/` container paths. Enable `CreateLibSymlinksHook` via patch 1002 so the device plugin's CDI spec includes backwards-compat symlinks. Add `--cdi-enabled-hooks create-lib-symlinks` to the exec-start template to pass the hook enablement through the systemd drop-in. Fix `containerDriverRoot` macro expansion from `%{_cross_sysroot}` (undefined) to `/%{_target_cpu}-bottlerocket-linux-gnu/sys-root`.

Result: containers see libraries at `/usr/lib/libcuda.so.580.159.03` with backwards-compat symlinks at `/usr/lib/nvidia/tesla/libcuda.so.580.159.03` pointing to `/usr/lib/libcuda.so.580.159.03`.

Note: Backwards-compat symlinks are supported in `cdi-cri` mode (the default device-list-strategy). The legacy `volume-mounts` mode does not use CDI hooks and therefore does not create symlinks in containers.

**In future variants we will be removing this symlink change.  This will only be available in existing variants**

**Testing done:**

- Built core-kit for x86_64, published, and built aws-k8s-1.35-nvidia variant AMI
- Launched g6.xlarge node on EKS cluster in ap-northeast-1
- Verified `containerDriverRoot` resolves to `/x86_64-bottlerocket-linux-gnu/sys-root`
- Verified device plugin starts and registers GPUs with kubelet
- Verified `nvidia-smi` works in container
- Verified libraries at `/usr/lib/` inside container
- Verified backwards-compat symlinks at `/usr/lib/nvidia/tesla/` point to `/usr/lib/`
- Verified `generate-cdi-specs.service` passes on boot
- Verified CDI spec at `/var/run/cdi/` includes symlink hooks
- Verified `ldconfig` resolves `libcuda.so.1` on host
- Verified no `ld.so.conf.d` entry needed (libs in standard path)
- Verified `nvidia-smi` works on host (dlopen via both paths)
- Tested negative case: without `--cdi-enabled-hooks`, symlinks do not appear in containers
- Ran nvidia smoke tests - all passed

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-container-toolkit/0002-add-additional-symlinks-flag.patch
+++ b/packages/nvidia-container-toolkit/0002-add-additional-symlinks-flag.patch
@@ -1,0 +1,190 @@
+From 75fb6b8355f68b1559818573736a7aa8c955065c Mon Sep 17 00:00:00 2001
+From: Maher Homsi <maherhom@amazon.com>
+Date: Wed, 17 Jun 2026 00:16:32 +0000
+Subject: [PATCH] Add --additional-symlinks flag for library path compatibility
+
+Add a --additional-symlinks flag to  that takes
+a directory path. When specified, for each discovered nvidia library
+mounted into the container, a symlink is created from
+<additional-symlinks-dir>/<lib> pointing to the actual library path.
+
+This enables backwards compatibility for workloads that expect nvidia
+libraries at /usr/lib/nvidia/tesla/ after libraries move to /usr/lib/.
+
+Usage:
+  nvidia-ctk cdi generate --additional-symlinks /usr/lib/nvidia/tesla
+
+Signed-off-by: Maher Homsi <maherhom@amazon.com>
+---
+ cmd/nvidia-ctk/cdi/generate/generate.go |  8 ++++
+ internal/discover/lib_path_symlinks.go  | 54 +++++++++++++++++++++++++
+ pkg/nvcdi/driver-nvml.go                | 10 +++++
+ pkg/nvcdi/lib.go                        |  2 +
+ pkg/nvcdi/options.go                    |  9 +++++
+ 5 files changed, 83 insertions(+)
+ create mode 100644 internal/discover/lib_path_symlinks.go
+
+diff --git a/cmd/nvidia-ctk/cdi/generate/generate.go b/cmd/nvidia-ctk/cdi/generate/generate.go
+index 2fff6ea..097eb07 100644
+--- a/cmd/nvidia-ctk/cdi/generate/generate.go
++++ b/cmd/nvidia-ctk/cdi/generate/generate.go
+@@ -68,6 +68,7 @@ type options struct {
+ 	enabledHooks       []string
+ 
+ 	featureFlags []string
++	libDirSymlinksDir string
+ 
+ 	csv struct {
+ 		files               []string
+@@ -243,6 +244,12 @@ func (m command) build() *cli.Command {
+ 				Destination: &opts.featureFlags,
+ 				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_FEATURE_FLAGS"),
+ 			},
++			&cli.StringFlag{
++				Name:        "additional-symlinks",
++				Destination: &opts.libDirSymlinksDir,
++				Usage:       "Create symlinks in the specified directory pointing to each nvidia library. This enables backwards compatibility for workloads expecting libraries at a legacy path.",
++				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_ADDITIONAL_SYMLINKS"),
++			},
+ 			&cli.BoolFlag{
+ 				Name:        "no-all-device",
+ 				Usage:       "Don't generate an `all` device for the resultant spec",
+@@ -397,6 +404,7 @@ func (m command) generateSpecs(opts *options) ([]generatedSpecs, error) {
+ 		nvcdi.WithDisabledHooks(opts.disabledHooks...),
+ 		nvcdi.WithEnabledHooks(opts.enabledHooks...),
+ 		nvcdi.WithFeatureFlags(opts.featureFlags...),
++		nvcdi.WithLibDirSymlinksDir(opts.libDirSymlinksDir),
+ 		// We set the following to allow for dependency injection:
+ 		nvcdi.WithNvmlLib(opts.nvmllib),
+ 	}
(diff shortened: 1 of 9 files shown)


maherthomsi:
Added `Signed off by` to commits

packages/nvidia-container-toolkit/0002-add-additional-symlinks-flag.patch
```diff
@@ -0,0 +1,192 @@
...
+ 		nvcdi.WithNvmlLib(opts.nvmllib),
+ 	}
+diff --git a/internal/discover/lib_path_symlinks.go b/internal/discover/lib_path_symlinks.go
+new file mode 100644
+index 0000000..a1b2c3d
+--- /dev/null
++++ b/internal/discover/lib_path_symlinks.go
+@@ -0,0 +1,72 @@
++/**
++# Copyright 2024 NVIDIA CORPORATION
```
piyush-jena: Not sure if this is appropriate but I would want others opinion as well.

packages/nvidia-container-toolkit/generate-cdi-specs.service
```diff
@@ -25,14 +25,10 @@ RefuseManualStop=true
 
 [Service]
 Type=oneshot
-# Explanation of the options:
-# --format json: to be consistent across Bottlerocket's variants
-# --mode nvml: the default mode ("auto") resolves to this already, make it explicit
```
piyush-jena, line 30: this comment section doesn't need to be removed. It needs to be extended.

packages/nvidia-k8s-device-plugin/1003-vendor-add-CreateLibSymlinksHook.patch
```diff
@@ -0,0 +1,179 @@
...
+ 		var transformedArgs []string
+ 		for _, arg := range args {
+ 			transformedArgs = append(transformedArgs, "--link", arg)
+diff --git a/vendor/github.com/NVIDIA/nvidia-container-toolkit/internal/discover/lib_path_symlinks.go b/vendor/github.com/NVIDIA/nvidia-container-toolkit/internal/discover/lib_path_symlinks.go
+new file mode 100644
+index 0000000..7cbf052
+--- /dev/null
++++ b/vendor/github.com/NVIDIA/nvidia-container-toolkit/internal/discover/lib_path_symlinks.go
+@@ -0,0 +1,73 @@
++/**
```
piyush-jena: Same license comment as above. I don't think we need it for a patch.

packages/nvidia-container-toolkit/0002-add-additional-symlinks-flag.patch
maherthomsi: Removed this section

packages/nvidia-k8s-device-plugin/1003-vendor-add-CreateLibSymlinksHook.patch
maherthomsi: Removed this section

packages/nvidia-container-toolkit/generate-cdi-specs.service
maherthomsi: Added it back and extended with the added options

piyush-jena (approved):

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-conf
```diff
@@ -33,7 +33,7 @@ flags:
     deviceListStrategy: "volume-mounts"
     {{/if}}
     deviceIDStrategy: {{default "index" settings.kubelet-device-plugins.nvidia.device-id-strategy}}
-    containerDriverRoot: "/"
+    containerDriverRoot: "/x86_64-bottlerocket-linux-gnu/sys-root"
```
arnaldo2792: This is not correct, you are hardcoding the path for both arches. If this is the way to go from now on, you will have to make this file an "input" templated file. Similar to what was done in the files you removed from the kernel kit.

mgsharm (approved):

vigh-m (approved):

maherthomsi:
Force-push: Added --cdi-enabled-hooks create-lib-symlinks to exec-start template

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
```diff
@@ -53,7 +55,8 @@ install -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}
 install -p -m 0644 %{S:5} %{buildroot}%{_cross_unitdir}
 install -d %{buildroot}%{_cross_unitdir}/nvidia-k8s-device-plugin.service.d
 install -d %{buildroot}%{_cross_unitdir}/nvidia-mps-control-daemon.service.d
-install -D -m 0644 %{S:2} %{buildroot}%{_cross_templatedir}/nvidia-k8s-device-plugin-conf
+sed -e 's|__PREFIX__|/%{_target_cpu}-bottlerocket-linux-gnu/sys-root|g' %{S:2} > nvidia-k8s-device-plugin-conf
```
arnaldo2792: `%{__target_cpu}` is not the correct macro, you should use `%{_cross_arch}`

KCSesh:
Nit: Please call out in the future variant releases, we will remove this symlink. This is a change for existing variants 👍

arnaldo2792 (approved):

maherthomsi:
Rebasing off latest develop

piyush-jena (approved):

vigh-m (approved):

PR #935, opened 2026-05-27
sources: add bottlerocket-crypto-provider crate

sky1122:
**Description of changes:**
Add a centralized CryptoProvider crate that provides runtime FIPS detection and TLS algorithm selection for Bottlerocket Rust binaries.

When the kernel FIPS flag is enabled (/proc/sys/crypto/fips_enabled = 1), the provider restricts TLS to FIPS-approved algorithms only (AES-GCM cipher suites, P-256/P-384 key exchange). On non-FIPS systems, the full algorithm set is available.

This crate exposes three public functions:
* fips_enabled() - detect kernel FIPS mode
* provider() - return the appropriate CryptoProvider based on runtime detection
* install_provider() - install the provider as the rustls global default

**Testing done:**
Testing with [this PR](https://github.com/awslabs/tough/pull/940) and [these two code chunks](https://gist.github.com/sky1122/cf0482d9328c0683e0b2d8c9719cbaf3) running on both fips/non-fips fedora host and test pass on both

```
 cargo test --features http -p tough --test http test_fips_crypto_provider_e2e             
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running tests/http.rs (target/debug/deps/http-e659a739e99ce22d)

running 1 test
test http_custom_provider::test_fips_crypto_provider_e2e ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.00s
```
**RSA key-length regression check**

  Regenerated the Squid leaf with a 1024-bit RSA key (CA remained 2048) to probe whether the migration or the always-on aws-lc-rs/fips build introduced FIPS-strict rejection in non-FIPS callers.

Result: no behavior change. 
rustls-with-aws_lc_rs already rejects RSA < 2048 unconditionally. it SUPPORTED_SIG_ALGS registers only RSA_PKCS1_2048_8192_* / RSA_PSS_2048_8192_* algorithm constants, and this list is not gated on the fips feature (rustls-0.23.42/src/crypto/aws_lc_rs/mod.rs:158). The runtime enforcement is inside aws-lc-rs at verify_rsa_signature (aws-lc-rs-1.17.1/src/rsa/signature.rs:281), which returns Err before any crypto if the key size is outside the algorithm's declared range present in both the FIPS and non-FIPS builds

**Terms of contribution:**

### DIFF
--- a/sources/bottlerocket-crypto-provider/src/lib.rs
+++ b/sources/bottlerocket-crypto-provider/src/lib.rs
@@ -1,0 +1,344 @@
+//! Centralized CryptoProvider for Bottlerocket Rust binaries.
+//!
+//! Provides runtime FIPS detection and TLS algorithm selection.
+//! When the kernel FIPS flag is enabled (`/proc/sys/crypto/fips_enabled` = 1),
+//! the provider restricts TLS to FIPS-approved algorithms only.
+//!
+//! The FIPS filtering uses rustls's `.fips()` trait method on each cipher suite
+//! and key exchange group, which queries the underlying aws-lc-rs library to
+//! determine FIPS approval. This means we automatically stay in sync with
+//! upstream's FIPS classifications without maintaining a manual allowlist.
+//!
+//! # Usage
+//!
+//! Call [`install_provider`] at the very start of `main()`, before any TLS
+//! connections are made. This installs the appropriate `CryptoProvider` as the
+//! process-wide global default for all rustls usage, including downstream
+//! libraries like `reqwest`, `hyper-rustls`, and `aws-smithy-http-client`.
+//!
+//! ```rust,ignore
+//! fn main() {
+//!     bottlerocket_crypto_provider::install_provider()
+//!         .expect("failed to install crypto provider");
+//!
+//!     // All subsequent TLS connections automatically use the installed
+//!     // provider — no per-client configuration needed.
+//! }
+//! ```
+//!
+//! If you need a [`CryptoProvider`] instance directly (e.g. to pass to
+//! `ClientConfig::builder_with_provider()`), use [`provider`]:
+//!
+//! ```rust,ignore
+//! use std::sync::Arc;
+//!
+//! let crypto = bottlerocket_crypto_provider::provider()
+//!     .expect("failed to detect FIPS mode");
+//! let config = rustls::ClientConfig::builder_with_provider(Arc::new(crypto))
+//!     .with_safe_default_protocol_versions()
+//!     .unwrap()
+//!     .with_root_certificates(root_store)
+//!     .with_no_client_auth();
+//! ```
+//!
+//! # How it works
+//!
+//! 1. Reads `/proc/sys/crypto/fips_enabled` to detect if the kernel has FIPS
+//!    mode enabled.
+//! 2. If FIPS is enabled, returns a provider filtered to only FIPS-approved
+//!    algorithms (AES-GCM cipher suites, P-256/P-384/X25519MLKEM768 key
+//!    exchange groups).
+//! 3. If FIPS is not enabled, returns the full default provider with all
+//!    algorithms available (including ChaCha20 and X25519).
+//!
+//! # Important
+//!
+//! - [`install_provider`] MUST be called before any other code installs a
+//!   `CryptoProvider`. The global default uses `OnceLock` and cannot be
+//!   overridden once set. If another provider is already installed, it returns
+//!   [`Error::ProviderAlreadyInstalled`].
+//! - The `aws-lc-rs` dependency MUST be compiled with `features = ["fips"]` so
+//!   that the underlying C library reports `FIPS_mode() == 1`. Without this,
+//!   `.fips()` returns `false` for all algorithms, and `fips_provider()` would
+//!   filter out everything — resulting in an empty provider with no cipher
+//!   suites or kx groups, and all TLS connections would fail.
+
+use log::info;
+use rustls::crypto::{aws_lc_rs, CryptoProvider};
+use snafu::{ResultExt, Snafu};
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Failed to read /proc/sys/crypto/fips_enabled"))]
+    ReadFipsEnabled { source: std::io::Error },
+
+    #[snafu(display(
+        "CryptoProvider already installed before install_provider() was called. \
+         install_provider() must be the first thing called in main()."
+    ))]
+    ProviderAlreadyInstalled,
+}
+
+/// Detect whether the system is running in FIPS mode by reading the kernel flag.
+pub fn fips_enabled() -> Result<bool, Error> {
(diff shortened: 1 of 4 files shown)


sky1122:
forced pushed to rebase to the latest and regenerated the Cargo.lock file

sources/bottlerocket-crypto-provider/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// This should be called once at the start of `main()` before any TLS connections are made.
+/// If a provider is already installed (by another component), this is a no-op.
+pub fn install_provider() {
+    if CryptoProvider::get_default().is_some() {
+        return;
+    }
+    let mode = if fips_enabled() { "FIPS" } else { "default" };
+    info!("Installing {} CryptoProvider", mode);
+    let _ = provider().install_default();
```
ginglis13: nit: `fips_enabled` is called twice here - once just for the log message, then again inside the `provider()` method.

sky1122:
force pushed to address feedback

sky1122:
forced pushed to turn on fips build and correct branch.

sources/bottlerocket-crypto-provider/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    #[test]
+    fn default_provider_includes_x25519() {
+        let p = default_provider();
+        assert!(p.kx_groups.iter().any(|g| g.name() == NamedGroup::X25519));
+    }
+
+    #[test]
+    fn fips_enabled_defaults_to_false() {
```
ginglis13: this test will fail consistently if I'm building the core-kit on a fips-enabled host

sources/bottlerocket-crypto-provider/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+/// FIPS-approved key exchange groups.
+const FIPS_KX_GROUPS: &[NamedGroup] = &[
+    NamedGroup::secp256r1,
+    NamedGroup::secp384r1,
+    NamedGroup::X25519MLKEM768,
+];
+
+/// Detect whether the system is running in FIPS mode by reading the kernel flag.
+pub fn fips_enabled() -> bool {
+    std::fs::read_to_string("/proc/sys/crypto/fips_enabled")
```
ginglis13: edge case, but if we fail to read this file on a fips variant, we'll silently enable non-fips ciphers. We could be stricter in this method:

```rust
pub fn fips_enabled() -> std::io::Result<bool> {
    match std::fs::read_to_string("/proc/sys/crypto/fips_enabled") {
        Ok(c) => Ok(c.trim() == "1"),
        Err(e) if e.kind() == ErrorKind::NotFound => Ok(false),
        Err(e) => Err(e),
    }
}
```

sky1122:
forced pushed to address feedback change `fips_enabled `. For `pub fn provider() -> CryptoProvider`, I decide to keep  return with no result  and default handle the edge case to return `true` because I think  if something unexpected happens, use the more restrictive mode. Breaking the caller (panicking or returning error) would prevent the binary from starting, which is worse than falling back to FIPS ciphers.

ginglis13 (approved):

sources/bottlerocket-crypto-provider/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Detect whether the system is running in FIPS mode by reading the kernel flag.
+///
+/// Returns `Ok(true)` if FIPS is enabled, `Ok(false)` if the file is missing
+/// (non-FIPS system), or `Err` if the file exists but cannot be read.
+pub fn fips_enabled() -> std::io::Result<bool> {
+    match std::fs::read_to_string("/proc/sys/crypto/fips_enabled") {
+        Ok(c) => Ok(c.trim() == "1"),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(false),
+        Err(e) => Err(e),
+    }
```
arnaldo2792, line 88: Two things:

- Lets fail if the read fails for whatever reason, we don't want to be in a situation where we use the wrong default (e.g. FIPS in non-FIPS setup).

- You could provide context of the error with `snafu`, like:

```rust
let fips_enabled_content = std::fs::read_to_string("/proc/sys/crypto/fips_enabled").context(error::ReadFipsEnabled)?;

fips_enabled_content == "1"
```

sources/bottlerocket-crypto-provider/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+    });
+    info!(
+        "Using {} CryptoProvider",
+        if fips { "FIPS" } else { "default" }
+    );
+    if fips {
+        fips_provider()
+    } else {
+        default_provider()
+    }
```
arnaldo2792: If `fips_enabled` returns an error with context, you can do this instead:

```rust

let (crypto_mode, provider) = if fips_enabled()? {
    ("FIPS", fips_provider())
} else {
    ("default",  default_provider())
};

info!("Using {} CrytpProvider", crypto_mode);
provider
```

sources/bottlerocket-crypto-provider/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+            .any(|s| s.suite() == CipherSuite::TLS13_CHACHA20_POLY1305_SHA256));
+    }
+
+    #[test]
+    fn default_provider_includes_x25519() {
+        let p = default_provider();
+        assert!(p.kx_groups.iter().any(|g| g.name() == NamedGroup::X25519));
+    }
+
+    #[test]
```
arnaldo2792, line 243: We should have a test that validates that a client in FIPS mode can't talk to a server that doesn't support any of the FIPS ciphers when `fips_provider` is used.

sky1122:
forced pushed to address feedback

sky1122:
git forced push to rebase

sky1122:
forced pushed(s) to rebase and Cargo.lock rebase issue. drop the rustls update commit as develop branch already bump the version

sources/bottlerocket-crypto-provider/src/lib.rs
sky1122: I double check again, sorry, I was wrong this file show or not depended on [kernel configuration](https://kernel.org/doc/html/latest/admin-guide/sysctl/crypto.html). In case new kernel doesn't have fips module ready yet but we already have new kernel in kernel kit, so I will still keep on this style for that case.

sky1122:
forced pushed to run `cargo fmt`

sky1122:
forced pushed to
- Replaced TCP client/server test with in-memory buffer test — avoids multi-threading, TCP sockets, and network access; uses rustls's write_tls/read_tls to exchange bytes through a Vec<u8> since TLS cipher negotiation is protocol-level and transport-independent
- add comment on how to use the libary
- install_provider() now fails if provider already set

sky1122:
forced push to change test comments and some variable order

arnaldo2792 (approved):

PR #943, opened 2026-06-08
feat(release): add no-data-partitions and no-private-partition image features

jmt-lab:
Introduce two opt-in image features that omit the BOTTLEROCKET-DATA-{A,B} and BOTTLEROCKET-PRIVATE filesystems from the image. Both default to off; behavior is unchanged when neither is enabled.

- Extend `bottlerocket-image-features` with `no_data_partitions` and `no_private_partition` fields and tests.
- Make `apiserver`'s `should_encrypt()` return false when either partition-omitting flag is set, with unit-test coverage via a pure helper.
- Declare RPM `Conflicts:` between `release-crypt` and the new image features so encrypted-storage builds cannot drop the required partitions.
- Document the design and operator-attached storage workflow in PARTITION_FEATURES.md and cross-link from ENCRYPTED_STORAGE.md.

**Terms of contribution:**

### DIFF
--- a/docs/ENCRYPTED_STORAGE.md
+++ b/docs/ENCRYPTED_STORAGE.md
@@ -14,6 +14,20 @@ Bottlerocket's encrypted storage feature provides:
 
 All encryption keys are sealed to TPM2 PCRs, ensuring data can only be decrypted when the system boots in a trusted state.
 
+## Compatibility with partition-omitting image features
+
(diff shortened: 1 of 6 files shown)


jmt-lab:
This PR is no longer needed, its changes will be folded into @sky1122 work, and will be using the unmanaged image feature.

PR #947, opened 2026-06-16
fix: conditional wipefs before systemd-makefs for false-positive signature detection

### LINKED ISSUE #4855: prepare-local-fs.service fails intermittently on encrypted EBS volumes due to mkfs.xfs false-positive signature detection
**Image I'm using:**
EKS optimized Bottlerocket AMI for EKS 1.34 and 1.35 with Bottlerocket v1.61.0

EKS worker nodes properly boot up and join EKS cluster.
In case of the existence of random partition signatures these should be wiped before proceeding with mkfs.

**What actually happened:**
Karpenter `NodePool` defines `blockDeviceMappings` as:
```
spec:
  amiFamily: Bottlerocket
  amiSelectorTerms:
  - alias: bottlerocket@latest
  blockDeviceMappings:
  - deviceName: /dev/xvda
    ebs:
      deleteOnTermination: true
      encrypted: true
      iops: 3000
      throughput: 125
      volumeSize: 20Gi
      volumeType: gp3
  - deviceName: /dev/xvdb
    ebs:
      deleteOnTermination: true
      encrypted: true
      iops: 10000
      throughput: 1500
      volumeSize: 1000Gi
      volumeType: gp3
```

On approximately 1 in 1,800 instance launches, the node fails to boot because `prepare-local-fs.service` exits with an error:
```
systemd-makefs[2158]: mkfs.xfs: /dev/disk/by-partlabel/BOTTLEROCKET-DATA appears to contain a partition table (atari).
systemd-makefs[2158]: mkfs.xfs: Use the -f option to force overwrite.
```
This cascades to prevent all of /local, /var, /opt, /mnt from mounting — containerd and kubelet never start, the node never registers with the EKS cluster.

Karpenter (node scaler) removes the node after not being registered for 15min.

Happens intermittently

**Additional information:**
mkfs.xfs performs a signature scan before formatting. The Atari partition table signature is only 2 bytes at a specific offset. With pseudorandom data filling the partition (see Amazon EBS [Data protection in Amazon EC2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/data-protection.html)), there is a non-trivial probability of a false-positive match.

This is a well-documented issue in the Linux ecosystem:

[util-linux/util-linux#1116](https://github.com/util-linux/util-linux/issues/1116) — "wipefs atari partition table false positives"
[rook/rook#7940 ](https://github.com/rook/rook/issues/7940)— "unexpected partition on disks >= 1TB (atari partitions)"

Evidence:
EC2 console output from a failed instance:
```
[  OK  ] Found device Amazon Elastic Block Store BOTTLEROCKET-DATA.
         Starting Prepare Local Filesystem (/local)...
[    1.514487] systemd-makefs[2158]: mkfs.xfs: /dev/disk/by-partlabel/BOTTLEROCKET-DATA appears to contain a partition table (atari).
[FAILED] Failed to start Prepare Local Filesystem (/local).
[    1.517224] systemd-makefs[2158]: mkfs.xfs: Use the -f option to force overwrite.
[DEPEND] Dependency failed for Local Directory (/local).
[DEPEND] Dependency failed for Resize Data Partition.
[DEPEND] Dependency failed for Local File Systems.
... (39 services/targets fail in cascade)
```
A successful instance for comparison:
```
[  OK  ] Found device /dev/disk/by-partlabel/BOTTLEROCKET-DATA.
         Starting Prepare Local Filesystem (/local)...
[  OK  ] Finished Prepare Local Filesystem (/local).
         Mounting Local Directory (/local)...
[  OK  ] Mounted Local Directory (/local).
```

youwalther65:
## Summary

Fixes https://github.com/bottlerocket-os/bottlerocket/issues/4855

On encrypted EBS volumes, pseudorandom data on the `BOTTLEROCKET-DATA` partition can match an Atari partition table signature (only 2 bytes at a specific offset), causing `mkfs.xfs` to refuse formatting (~1/1800 instance launches). This cascades to prevent all of `/local`, `/var`, `/opt`, `/mnt` from mounting — containerd and kubelet never start, the node never joins the EKS cluster.

This adds a conditional `wipefs -a` before `systemd-makefs` in `prepare-local-fs.service` that only runs when no valid filesystem is detected.

## Implementation Details

- `wipefs -a` erases all magic strings/signatures from a device (partition tables, filesystem superblocks, etc.)
- `blkid` exits 0 when it finds a filesystem, 2 when it does not — the `!` inverts this
- `systemd-makefs` internally checks for existing filesystems and is a no-op if one exists
- The `-` prefix on the ExecStart line makes the entire wipefs step non-fatal
- The encrypted variant (`prepare-local-fs-encrypted.conf`) overrides `DATA_PARTITION_BLOCK_DEVICE` to `/dev/mapper/BOTTLEROCKET-DATA` — the same fix applies to both the plain and encrypted paths since the decrypted mapper device also contains random-looking data before first format

## Test plan

- [ ] Boot a Bottlerocket instance with encrypted EBS data volume and verify `prepare-local-fs.service` succeeds
- [ ] Verify on an instance with an already-formatted data partition that `wipefs` is skipped (blkid returns 0, short-circuits)
- [ ] Simulate the false positive by writing Atari signature bytes to a test block device and confirm wipefs clears them before mkfs runs

### DIFF
--- a/packages/release/prepare-local-fs.service
+++ b/packages/release/prepare-local-fs.service
@@ -13,6 +13,13 @@ Type=oneshot
 EnvironmentFile=/usr/share/bottlerocket/image-features.env
 Environment=DATA_PARTITION_BLOCK_DEVICE=/dev/disk/by-partlabel/BOTTLEROCKET-DATA
 
+# Wipe stale or false-positive partition/filesystem signatures before
+# formatting. Only runs when no valid filesystem is detected, so existing
+# data partitions are left untouched. Prevents mkfs.xfs from refusing to
+# format due to pseudorandom data matching signature patterns (e.g. Atari
+# partition table on encrypted EBS volumes).
+ExecStart=-/bin/sh -c '! /usr/sbin/blkid ${DATA_PARTITION_BLOCK_DEVICE} && /usr/sbin/wipefs -a ${DATA_PARTITION_BLOCK_DEVICE}'
+
 # Create the filesystem on the partition, if it doesn't exist.
 ExecStart=/usr/lib/systemd/systemd-makefs ${DATA_PARTITION_FILESYSTEM} ${DATA_PARTITION_BLOCK_DEVICE}
 


youwalther65:
Did a test in an admin container  `sudo sheltie` environment and it failed:
```bash
[root@admin]# sudo sheltie

bash-5.2# ls -l /dev/disk/by-partlabel/BOTTLEROCKET-DATA
lrwxrwxrwx. 1 root root 15 Jun 16 17:09 /dev/disk/by-partlabel/BOTTLEROCKET-DATA -> ../../nvme1n1p1
bash-5.2# ls -l /usr/sbin/blkid
-rwxr-xr-x. 1 root root 64976 May  1 04:20 /usr/sbin/blkid

bash-5.2# ls -l /usr/sbin/wipefs
-rwxr-xr-x. 1 root root 41440 May  1 04:20 /usr/sbin/wipefs

bash-5.2# /bin/sh -c '! /usr/sbin/blkid /dev/disk/by-partlabel/BOTTLEROCKET-DATA && /usr/sbin/wipefs -a /dev/disk/by-partlabel/BOTTLEROCKET-DATA'
expected first arg 'sh', not '/bin/sh'

bash-5.2# sh -c '! /usr/sbin/blkid /dev/disk/by-partlabel/BOTTLEROCKET-DATA && /usr/sbin/wipefs -a /dev/disk/by-partlabel/BOTTLEROCKET-DATA'     
'!' could not be found

Caused by these errors (recent errors listed first):
  1: failed to locate '!' in PATH
  2: cannot find binary path
```

So unfortunately this simple approach does not work because of the usage of underlying `brush` instead of a real bash like shell.

piyush-jena:
Thank you for your contribution. We have fixed the issue differently as you might be aware. I am closing this PR.

PR #963, opened 2026-07-02
Add create-device subcommand to ghostdog

arnaldo2792:
**Description of changes:**

Some out-of-tree kernel modules (e.g. those with licensing restrictions) do not create their own device nodes.  Add a create-device subcommand to ghostdog to provide a helper for creating device nodes on behalf of such modules.

**Testing done:**

In combination with https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/462, I confirmed that the device `/dev/gdrdrv` was created after the kernel module was loaded:

```bash
bash-5.2# ls /dev/gdrdrv
/dev/gdrdrv
bash-5.2#
```

**Terms of contribution:**

### DIFF
--- a/sources/ghostdog/src/create_device.rs
+++ b/sources/ghostdog/src/create_device.rs
@@ -1,0 +1,223 @@
+//! Create device nodes using mknod based on /proc/devices lookup.
+
+use crate::error::{self, Result};
+use crate::CreateDeviceArgs;
+use snafu::{ensure, ResultExt};
+use std::collections::HashMap;
+use std::path::Path;
+use std::process::Command;
+use std::{fs, str};
+
+const MKNOD_PATH: &str = "/usr/bin/mknod";
+const PROC_DEVICES_PATH: &str = "/proc/devices";
+
+/// Type of device (character or block)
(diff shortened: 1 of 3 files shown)


sources/ghostdog/src/create_device.rs
```diff
@@ -0,0 +1,223 @@
...
+                .fail();
+            }
+        }
+    }
+
+    Ok(ProcDevices {
+        char_devices,
+        block_devices,
+    })
+}
```
piyush-jena, line 107: parse_proc_devices LGTM

piyush-jena (approved):
LGTM!

sources/ghostdog/src/create_device.rs
```diff
@@ -0,0 +1,223 @@
...
+
+/// Look up a device name in the parsed /proc/devices and return (major, type)
+fn lookup_device(devices: &ProcDevices, name: &str) -> Result<(u32, DeviceType)> {
```
ginglis13, line 110: why not impl on `ProcDevices` ?

ginglis13 (approved):

PR #966, opened 2026-07-02
rottweiler: dynamically exclude settings from settings.measurement.excluded-settings

ginglis13:
**Issue number:**

Related: https://github.com/bottlerocket-os/bottlerocket/issues/4872

**Description of changes:**

When measuring OS settings into PCR 8, rottweiler now reads settings.measurement.excluded-settings and appends them as additional --exclude arguments to apiclient.

**Testing done:**

In combination with https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/140

## Test Matrix

| Test Case | User Data Config | Purpose |
|---|---|---|
| Baseline  | No `settings.measurement` | Control |
| Excluded  | `excluded-settings = ["host-containers"]` | Verify exclusion works |
| Disallowed  | `excluded-settings = ["settings", "*", "settings.measurement", "host-containers"]` | Verify disallowed patterns are rejected |

### PCR 8 Values

| Test Case | PCR 8 (sha256) |
|---|---|
| Baseline | `B1ED3892838081122EB1ECBACE7C8225041B9D5CE9F24A1D4556713285B3B752` |
| Excluded | `A94FE2DA15EFA2DB5C120B2A9DBCF4632E1B92446B858CD7CD1A30DC59643FB0` |
| Disallowed | <failure to boot, bad settings provided> |

**Key observation:** All three PCR 8 values are different, confirming that the exclusion configuration affects the measurement.

### Test Case 1: Baseline (No Exclusions)

**User data:**
```toml
[settings.host-containers.admin]
enabled = true
source = "328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.11.16"
```

**Settings keys present:** `autoscaling`, `aws`, `boot`, `cloudformation`, `ecs`, `host-containers`, `kernel`, `metrics`, `motd`, `network`, `ntp`, `oci-defaults`, `oci-hooks`, `updates`

**Observations:**
- `host-containers` is present in settings (containing admin container config)
- `measurement` is not present (setting was not configured)
- PCR 8 was measured from ALL settings (minus hardcoded exclusions: `updates.seed`, `network.hostname`)

✅ Baseline works as expected with no exclusions.

### Test Case 2: With Exclusion (`host-containers`)

**User data:**
```toml
[settings.host-containers.admin]
enabled = true
source = "328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.11.16"

[settings.measurement]
excluded-settings = ["host-containers"]
```

**Settings keys present:** `autoscaling`, `aws`, `boot`, `cloudformation`, `ecs`, `host-containers`, `kernel`, `measurement`, `metrics`, `motd`, `network`, `ntp`, `oci-defaults`, `oci-hooks`, `updates`

**Measurement settings value:**
```json
{
  "excluded-settings": ["host-containers"]
}
```

**Observations:**
- PCR 8 (`A94F...`) differs from baseline (`B1ED...`): exclusion changed the measurement
- `host-containers` is still present in settings (the setting exists, just excluded from measurement)
- `measurement` is now present in settings and is measured (not self-excluded)
- The difference in PCR 8 between baseline and this case is due to:
  1. `host-containers` being excluded from the measured payload
  2. `settings.measurement` being ADDED to the measured payload (it didn't exist in baseline)

✅ Exclusion of `host-containers` from PCR 8 measurement is working correctly.

### Test Case 3: Disallowed Values

**User data:**
```toml
[settings.host-containers.admin]
enabled = true
source = "328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.11.16"

[settings.measurement]
excluded-settings = ["settings", "*", "settings.measurement", "host-containers"]
```

Instance boot fails.

| Check | Expected | Actual | Status |
|---|---|---|---|
| Baseline PCR 8 != Excluded PCR 8 | Different values | `B1ED...` != `A94F...` | **PASS** |
| Baseline PCR 8 != Disallowed PCR 8 | Different values | `B1ED...` != `0DCC...` | **PASS** |
| Excluded PCR 8 != Disallowed PCR 8 | Different (different measurement content) | `A94F...` != `0DCC...` | **PASS** |
| host-containers present in API on all instances | True | True on all 3 | **PASS** |
| settings.measurement present when configured | True | True on excluded + disallowed | **PASS** |
| Disallowed values passed to API | Failure to boot | Failure to boot | **PASS** |
| settings.measurement not self-excluded | Contributes to PCR | Different PCR when content differs | **PASS** 

**Terms of contribution:**


sources/rottweiler/src/system.rs
```diff
@@ -161,21 +161,109 @@ fn get_tpm2_pcrs() -> Result<String> {
...
+                    "WARNING: ignoring measurement exclusion pattern: {:?}",
+                    v.trim()
+                );
+                return None;
+            }
+            if normalized.starts_with(SELF_PREFIX) {
+                eprintln!(
+                    "WARNING: ignoring self-referencing measurement exclusion: {:?}",
+                    v.trim()
+                );
```
arnaldo2792: I would prefer if we use `log` as the logging library.

sources/rottweiler/src/system.rs
```diff
@@ -161,21 +161,109 @@ fn get_tpm2_pcrs() -> Result<String> {
...
+
+    // Parse the JSON response. Expected to be a JSON array of strings,
+    // or a JSON object like {"settings.measurement.excluded-settings": [...]}
+    let text = String::from_utf8_lossy(&output);
+    let text = text.trim();
+
+    // Try parsing as a direct array first
+    if let Ok(values) = serde_json::from_str::<Vec<String>>(text) {
+        return filter_excluded_settings(&values);
+    }
```
arnaldo2792: The API server wouldn't return just an array of values. It will actually return the verbose JSON so it should be fine to drop this check.

sources/rottweiler/src/system.rs
```diff
@@ -161,21 +161,109 @@ fn get_tpm2_pcrs() -> Result<String> {
...
+            .and_then(|s| s.get("measurement"))
+            .and_then(|m| m.get("excluded-settings"))
+            .and_then(|v| v.as_array())
+    {
+        let strings: Vec<String> = values
+            .iter()
+            .filter_map(|v| v.as_str().map(String::from))
+            .collect();
+        return filter_excluded_settings(&strings);
+    }
```
arnaldo2792: In here, the function is silently failing if the API server returned a spurious output. If the function leverages on serde's deserialization, it could return an error with context around the failure. The caveat is that it needs 3 structs to properly deserialize, but I think the benefit outweighs the verbosity of the structs:
- Idiomatic, strong types
- A good error message

sources/rottweiler/src/system.rs
```diff
@@ -161,21 +161,109 @@ fn get_tpm2_pcrs() -> Result<String> {
     })
 }
 
-/// Get canonicalized settings from apiclient, excluding seed and hostname
-pub fn apiclient_get_settings() -> Result<Zeroizing<Vec<u8>>> {
-    execute(
+/// Patterns that must not be allowed as measurement exclusions.
+const DISALLOWED_PATTERNS: &[&str] = &["settings", "", "*"];
```
arnaldo2792: `"settings."` will cause an error right? API server won't return anything if the customer sets that by mistake. Just asking.

ginglis13: At set-time, yes, https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/140 's validate() rejects it. Added this at settings get for defense in depth

piyush-jena: can't we just use the model in settings-sdk instead of repeated code? This kind of repeated code also adds additional burden on keeping it in sync in both places.

piyush-jena: and I believe we changed this part in settings-sdk (so its out of sync now)

ginglis13: Good point, yea. I'll work on deduplicating this instead and lean on the modeled type from the settings-sdk now that we have that released

ginglis13: Updated to model responses from apiserver and settings model shape rather than have this redundant parsing and validation.

piyush-jena:
Based on yesterday's changes, TC3 should error out, right? Because of wildcards and "settings"?

ginglis13:
> Based on yesterday's changes, TC3 should error out, right? Because of wildcards and "settings"?

Correct, will update the PR description

ginglis13:
^ force push rebase and address merge conflicts

PR #976, opened 2026-07-10
Bug fixes for thar-be-registries port inference

vigh-m:
**Issue number:**

`thar-be-registries` has two interrelated bugs in registry URL parsing that cause containerd to fail resolving certain registry configurations:

1. Default ports stripped from directory names: Rust's `Url::port()` strips default ports (80, 443, 21), but containerd preserves them. When a user configures a registry like `registry.example.com:443`, the generated directory path omits the port, creating a mismatch with containerd's expected lookup path under `certs.d/`.
```bash
# Output of: find /etc/containerd/certs.d
/etc/containerd/certs.d
/etc/containerd/certs.d/196.18.89.98
/etc/containerd/certs.d/196.18.89.98/credentials.toml
/etc/containerd/certs.d/public.ecr.aws
/etc/containerd/certs.d/public.ecr.aws/credentials.toml
/etc/containerd/certs.d/public.ecr.aws/hosts.toml
```

2. Incorrect scheme for port-80 registries: `parse_registry()` defaults to `https` for bare hostnames. This produces `server = "https://registry:80"` in `hosts.toml`, causing containerd to attempt a TLS handshake against a plaintext HTTP server.

3. Incorrect path validation for hosts: When using a registry without a scheme, the `hosts.toml` would omit the `override_path` flag leading to incorrect configurations:
```bash
bash-5.2# cat /etc/containerd/certs.d/docker.io/hosts.toml
server = "https://registry-1.docker.io"

[host."196.18.8.18:443/v2/eks-a-test"]
capabilities = ["pull", "resolve"]
bash-5.2# cat /etc/containerd/certs.d/public.ecr.aws/hosts.toml
server = "https://public.ecr.aws"

[host."196.18.8.18:443/v2/eks-a-test"]
capabilities = ["pull", "resolve"]
```

**Description of changes:**
- Adds a `parsed_port()` helper that checks the original input string for an explicit `:80` or `:443` suffix when `Url::port()` returns `None
- Also When no explicit scheme is provided, the scheme is now inferred from the port:
  - Port 80 → http
  - All other ports / no port → https
Explicit schemes (http:// or https://) are always preserved regardless of port.

- Update `Endpoint::has_path_component()` with an explicit retry call to `Url::Parse` with a scheme when provided a url without a scheme

- All unit tests updated.

**Testing done:**
- Validate that `/etc/containerd/certs.d/` folders are generated correctly for user-data using port 443
```bash
bash-5.2# find /etc/containerd/certs.d/
/etc/containerd/certs.d/
/etc/containerd/certs.d/192.168.176.179_443_
/etc/containerd/certs.d/192.168.176.179_443_/credentials.toml
/etc/containerd/certs.d/public.ecr.aws
/etc/containerd/certs.d/public.ecr.aws/hosts.toml
```
- Validate `hosts.toml` are generated with the correct `override_path` flag
```bash
bash-5.2# cat /etc/containerd/certs.d/docker.io/hosts.toml
server = "https://registry-1.docker.io"

[host."196.18.8.18:443/v2/eks-a-test"]
capabilities = ["pull", "resolve"]
override_path = true
bash-5.2# cat /etc/containerd/certs.d/public.ecr.aws/hosts.toml
server = "https://public.ecr.aws"

[host."196.18.8.18:443/v2/eks-a-test"]
capabilities = ["pull", "resolve"]
override_path = true
```

**Terms of contribution:**

### DIFF
--- a/sources/api/thar-be-registries/src/registry.rs
+++ b/sources/api/thar-be-registries/src/registry.rs
@@ -135,6 +135,20 @@ mod tests {
     // [fe80::443] has no port — the "443" is part of the address, not a port.
     #[test_case("[fe80::443]", "[fe80::443]", "https"; "ipv6 addr containing 443 no port")]
     #[test_case("https://[fe80::443]", "[fe80::443]", "https"; "ipv6 addr containing 443 with scheme no port")]
+    // URLs with IP, port, and path — parse_registry extracts only host:port and scheme.
+    // With scheme
+    #[test_case("https://10.0.0.1:443/path/to/resource", "10.0.0.1:443", "https"; "https ip with 443 and path")]
+    #[test_case("http://172.16.0.5:80/path/to/resource", "172.16.0.5:80", "http"; "http ip with 80 and path")]
+    #[test_case("https://10.0.0.1:8443/path/to/resource", "10.0.0.1:8443", "https"; "https ip with custom port and path")]
+    #[test_case("https://registry.example.com/path/to/resource", "registry.example.com", "https"; "https hostname no port with path")]
+    #[test_case("https://registry.example.com:5000/path/to/resource", "registry.example.com:5000", "https"; "https hostname custom port with path")]
+    #[test_case("http://registry.example.com:80/path/to/resource", "registry.example.com:80", "http"; "http hostname port 80 with path")]
+    // Schemeless with path
+    #[test_case("196.18.8.18:443/path/to/resource", "196.18.8.18:443", "https"; "schemeless ip port 443 with path")]
+    #[test_case("192.168.1.1:5000/path/to/resource", "192.168.1.1:5000", "https"; "schemeless ip custom port with path")]
+    #[test_case("192.168.1.1:80/path/to/resource", "192.168.1.1:80", "http"; "schemeless ip port 80 with path")]
+    #[test_case("registry.example.com:5000/path/to/resource", "registry.example.com:5000", "https"; "schemeless hostname custom port with path")]
+    #[test_case("registry.example.com:443/path/to/resource", "registry.example.com:443", "https"; "schemeless hostname port 443 with path")]
     fn test_parse_registry(input: &str, expected_host: &str, expected_scheme: &str) {
         let (host, scheme) = parse_registry(input).unwrap();
         assert_eq!(host, expected_host);
--- a/sources/api/thar-be-registries/src/registry.rs
+++ b/sources/api/thar-be-registries/src/registry.rs
@@ -7,9 +7,18 @@ use crate::error::{ParseRegistrySnafu, Result};
 pub(crate) const DOCKER_HUB_HOST: &str = "docker.io";
 pub(crate) const DOCKER_HUB_REGISTRY: &str = "registry-1.docker.io";
 
-/// Parse registry string, extracting host:port and optional scheme.
-/// Works with full URLs (https://docker.io) or bare hostnames (docker.io, registry:5000).
-/// Defaults to https scheme when none is provided.
+/// Parse a registry string into its `(host, scheme)` components.
+///
+/// Returns:
+/// - `host`: The registry host and optional port (e.g. `docker.io`, `registry.example.com:5000`).
+///   Used to construct the directory name under `certs.d/` and the `server` URL in `hosts.toml`.
+/// - `scheme`: Either `http` or `https`. Written into the `server` field of `hosts.toml`
+///   (e.g. `server = "https://registry.example.com:5000"`).
+///
+/// Scheme is determined as follows:
+/// - Explicit scheme in input (`http://` or `https://`) → preserved as-is.
+/// - No scheme, port 80 → `http`.
+/// - No scheme, any other port or no port → `https`.
 pub(crate) fn parse_registry(registry: &str) -> Result<(String, String)> {
     // Try parsing as-is first (handles URLs with scheme like https://docker.io)
     // Only accept if it has a host (registry:5000 parses as scheme with no host)
@@ -24,7 +33,8 @@ pub(crate) fn parse_registry(registry: &str) -> Result<(String, String)> {
     if let Ok(url) = Url::parse(&format!("https://{}", registry)) {
         if let Some(host) = url.host_str() {
             let port = parsed_port(&url, registry);
-            return Ok((format_host_port(host, port), "https".to_string()));
+            let scheme = if port == Some(80) { "http" } else { "https" };
+            return Ok((format_host_port(host, port), scheme.to_string()));
         }
     }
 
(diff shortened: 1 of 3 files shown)


sources/api/thar-be-registries/src/registry.rs
```diff
@@ -7,28 +7,57 @@ use crate::error::{ParseRegistrySnafu, Result};
...
 
+/// Returns the port from a parsed URL, preserving explicit default ports (e.g. :443, :80)
+/// that `Url::port()` would strip. This matches containerd's Go-based directory lookup
+/// which uses `net/url` and preserves explicit ports.
+fn parsed_port(url: &Url, original_url: &str) -> Option<u16> {
+    if let Some(port) = url.port() {
+        return Some(port);
+    }
+    if let Some(default_port) = url.port_or_known_default() {
+        if original_url.contains(&format!(":{}", default_port)) {
```
michaelb990: nit: thoughts on adding `url.host_str().unwrap_or_default()` to the matching fragment to ensure we aren't picking up a `:<port>` somewhere else in the URL such as an IPv6 address or in the path?

vigh-m: Yup! Makes sense. Added that and some unit tests for it. Thanks!

sam-berning (approved):

michaelb990 (approved):

sources/api/thar-be-registries/src/registry.rs
```diff
@@ -15,20 +15,41 @@ pub(crate) fn parse_registry(registry: &str) -> Result<(String, String)> {
...
+        return Some(port);
+    }
+    if let Some(default_port) = url.port_or_known_default() {
+        if original_url.contains(&format!(
+            "{}:{}",
+            url.host_str().unwrap_or_default(),
+            default_port
+        )) {
+            return Some(default_port);
+        }
```
arnaldo2792, line 57: Is this how Go's `net/url` does it? It is strange that the function checks whether the url contains the re-created URL? Why is the check needed?

sources/api/thar-be-registries/src/registry.rs
```diff
@@ -7,9 +7,20 @@ use crate::error::{ParseRegistrySnafu, Result};
...
-/// Defaults to https scheme when none is provided.
+/// Parse a registry string into its `(host, scheme)` components.
+///
+/// Returns:
+/// - `host`: The registry host and optional port (e.g. `docker.io`, `registry.example.com:5000`).
+///   Used to construct the directory name under `certs.d/` and the `server` URL in `hosts.toml`.
+/// - `scheme`: Either `http` or `https`. Written into the `server` field of `hosts.toml`
+///   (e.g. `server = "https://registry.example.com:5000"`). containerd uses this to decide
+///   whether to open a TLS or plaintext connection — see `parseHostConfig` in
+///   `core/remotes/docker/config/hosts.go`.
```
arnaldo2792: Rather than a reference to the path, a permalink will be best.

sources/api/thar-be-registries/src/registry.rs
```diff
@@ -24,7 +35,8 @@ pub(crate) fn parse_registry(registry: &str) -> Result<(String, String)> {
     if let Ok(url) = Url::parse(&format!("https://{}", registry)) {
         if let Some(host) = url.host_str() {
             let port = parsed_port(&url, registry);
-            return Ok((format_host_port(host, port), "https".to_string()));
+            let scheme = if port == Some(80) { "http" } else { "https" };
```
arnaldo2792, line 36: Is this logic also taken from upstream containerd?

sources/api/thar-be-registries/src/registry.rs
vigh-m: To clarify, Go does not do this, which is why we need to add this logic to explicitly return the port when a well-known port like `443` or `21` is used.

The check is to avoid matching against a url like `https://example.com/:443` being detected as having port 443 when actually `/:443` is part of the url

sources/api/thar-be-registries/src/registry.rs
vigh-m: Removed that line. It wasn't adding any extra context

sources/api/thar-be-registries/src/registry.rs
vigh-m: No. Like above, this is added to bring compatibility between `thar-be-registry` parsing and what containerd expects. This specific change specifically parses a url without a scheme like `example.com:443` to use `http://` as the scheme instead of defaulting to `https://` which was the previous default

vigh-m:
⬆️ Updated commit messages and addressed comments

sources/api/thar-be-registries/src/registry.rs
```diff
@@ -73,6 +111,30 @@ mod tests {
...
+    // Default port preservation: explicit :443 and :80 must be retained
+    #[test_case("192.168.1.1:443", "192.168.1.1:443", "https"; "bare host with explicit 443")]
+    #[test_case("192.168.1.1:80", "192.168.1.1:80", "http"; "bare host with explicit 80")]
+    #[test_case("registry.example.com:443", "registry.example.com:443", "https"; "bare hostname with explicit 443")]
+    #[test_case("registry.example.com:80", "registry.example.com:80", "http"; "bare hostname with explicit 80")]
+    #[test_case("https://192.168.1.1:443", "192.168.1.1:443", "https"; "https url with explicit 443")]
+    #[test_case("http://192.168.1.1:80", "192.168.1.1:80", "http"; "http url with explicit 80")]
+    #[test_case("https://registry.example.com:443", "registry.example.com:443", "https"; "https hostname with explicit 443")]
+    #[test_case("http://registry.example.com:80", "registry.example.com:80", "http"; "http hostname with explicit 80")]
+    // Explicit scheme always wins over port-based inference.
```
piyush-jena, line 123: nice! good coverage!

sources/api/thar-be-registries/src/registry.rs
```diff
@@ -73,6 +111,30 @@ mod tests {
...
+    #[test_case("https://registry.example.com:80", "registry.example.com:80", "https"; "explicit https on port 80")]
+    #[test_case("http://registry.example.com:443", "registry.example.com:443", "http"; "explicit http on port 443")]
+    // Bare IP with no port defaults to https.
+    #[test_case("192.168.1.1", "192.168.1.1", "https"; "bare ip no port")]
```
piyush-jena, line 127: is there something that recommends this or are we choosing to do this?

piyush-jena: or is it for security reasons?

sources/api/thar-be-registries/src/registry.rs
piyush-jena: `https://example.com/:443` is this a valid url?

piyush-jena: in this case we are explicitly checking for this corner case. Has this been encountered somewhere?

piyush-jena: can you try a TC for this? I have a feeling parsing into Url struct might fail.

maherthomsi (approved):

sources/api/thar-be-registries/src/registry.rs
vigh-m: No. This is an assumption we are making that all urls without an explicit scheme are `https://` unless they explicitly have port `:80`

sources/api/thar-be-registries/src/registry.rs
vigh-m: > `https://example.com/:443` is this a valid url?

Yeah, it's unconventional but it is a valid url where `/:443` is the [url path and not the port](https://www.rfc-editor.org/info/rfc3986/#section-3.3). The port with a url-path is represented as `https://example.com:443/path/to/resource`
I don't know of this being used in the wild but we're also avoiding the case for ipv6 IPs like `[fe80::443]` where `:443` is the IP address and not the port.

I have test cases like below to cover that scenario which I think should suffice.

```rust
#[test_case("[fe80::443]", "[fe80::443]", "https"; "ipv6 addr containing 443 no port")]
#[test_case("https://[fe80::443]", "[fe80::443]", "https"; "ipv6 addr containing 443 with scheme no port")]
```

vigh-m:
⬆️ Fix missing `override_path` flag for urls without an explicit scheme
- Also add more test cases to validate urls with port and url-path

sam-berning (approved):

arnaldo2792 (approved):

PR #999, opened 2026-08-06
feat: add support for ephemeral-encryption-keys

piyush-jena:
**Description of changes:**
- Adds support for ephemeral encryption keys. The encryption keys are placed in `tmpfs` (`/run`) instead of the keystore and deleted after unlock.
- As part of this feature, we encrypt the private partition with `LUKS2` cipher. Because the encryption key is temporary, we format the partition every boot (with the same parameters) and re-encrypt every boot.
- The datastore inside of private partition is further encrypted with `fscrypt`. This requires an additional fs feature to be enabled (`encrypt`) which we pass in the previous step (or using `tune2fs` in case of data partition). 
- Data Partition is formatted with the default parameters which is why we use `systemd-makefs`. We force the reformat by running `wipefs` first. This deletes all fs related signatures tricking `systemd-makefs` into reformatting and avoiding the additional overhead of writing 0s to the entire partition.
- Additionally we use image features environment variables in various places to choose keystore path or trick `early-boot-config` into populating the datastore every boot (because we format the private partition every boot).

**Testing done:**
- The tmpfs keystore has SELinux label `private_t`
```
bash-5.2# ls -lZ /run/rottweiler
total 0
bash-5.2# ls -lZ /run | grep rottweiler
drwx------.  2 root   root   system_u:object_r:private_t:s0     40 Aug 12 16:22 rottweiler
```
- Verified the new keystore doesn't contain the keys.
```
[root@admin]# sheltie
bash-5.2# ls -lia /run/rottweiler/
total 0
918 drwx------.  2 root root  40 Aug 10 23:13 .
  1 drwxr-xr-x. 21 root root 520 Aug 10 23:13 ..
```
- Verified the filesystems and the directories are encrypted.
```
bash-5.2# lsblk -o NAME,TYPE,FSTYPE,MOUNTPOINT
NAME                     TYPE  FSTYPE         MOUNTPOINT
nvme1n1                  disk
`-nvme1n1p1              part  crypto_LUKS
  `-BOTTLEROCKET-DATA    crypt xfs            /local
nvme0n1                  disk
|-nvme0n1p1              part
|-nvme0n1p2              part  vfat
|-nvme0n1p3              part  vfat           /boot
|-nvme0n1p4              part  erofs
|-nvme0n1p5              part  DM_verity_hash
|-nvme0n1p6              part
|-nvme0n1p7              part  crypto_LUKS
| `-BOTTLEROCKET-PRIVATE crypt ext4           /var/lib/bottlerocket
`-nvme0n1p8              part
bash-5.2# rottweiler check directory /.bottlerocket/datastore encrypted
directory '/.bottlerocket/datastore' is encrypted
```
- Reboot (with user-data) works - used bootstrap commands as user data to change `settings.motd` and in every boot its the updated value
user-data
```
[settings.bootstrap-commands.mybootstrap]
commands = [["apiclient", "set", "settings.motd=hello"]]
essential = true
mode = "always"
```
output:
```
bash-5.2# apiclient get settings.motd
{
  "settings": {
    "motd": "hello"
  }
}
```
A second consequence of reformatting data partition is that the journal for previous boot is lost which was also checked.
- Moving the EBS volume to a different instance makes it unreadable
- lockdown mode works
```
bash-5.2# apiclient get settings.motd
{
  "settings": {
    "motd": "hello"
  }
}
bash-5.2# apiclient lockdown
01:21:33 [INFO] Lockdown completed
bash-5.2# apiclient get settings.motd
Failed to get settings: Failed GET request to '/?prefix=settings.motd': Status 500 when GETing /?prefix=settings.motd: Data store error during get_prefix 'settings.motd' for Live: Data store integrity violation at /var/lib/bottlerocket/datastore/current/live: Live datastore missing
```
- (A)I went through journal and it didn't find any errors during all this experiments
**Terms of contribution:**

### DIFF
--- a/packages/release/encrypt-datastore-private-luks.conf
+++ b/packages/release/encrypt-datastore-private-luks.conf
@@ -1,0 +1,7 @@
+[Service]
+# In case of ephemeral-encryption-keys, we format the BOTTLEROCKET-PRIVATE with
+# `encrypt` feature flag. So, we drop the tune2fs call, which is used to enable
+# the flag and generate the key and encrypt the datastore.
+ExecStart=
+ExecStart=/usr/bin/rottweiler generate-key datastore
+ExecStart=/usr/bin/rottweiler encrypt directory ${DATASTORE_DIR} datastore
--- a/packages/release/encrypt-private-fs.service
+++ b/packages/release/encrypt-private-fs.service
@@ -1,0 +1,27 @@
+[Unit]
+Description=Encrypt Private Filesystem (/.bottlerocket)
+DefaultDependencies=no
+Conflicts=umount.target
+Wants=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device
+After=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device tpm2.target cryptsetup-pre.target
+Before=unlock-private-fs.service
+ConditionSecurity=tpm2
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+Environment=BOTTLEROCKET_PRIVATE=/dev/disk/by-partlabel/BOTTLEROCKET-PRIVATE
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
+
+# Generate the per-boot key into the /run keystore and LUKS2-format PRIVATE.
+ExecStart=/usr/bin/rottweiler generate-key bottlerocket-private
+ExecStart=/usr/bin/rottweiler encrypt block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private
+
+UMask=0077
+RemainAfterExit=true
+StandardOutput=journal+console
+StandardError=inherit
+
+[Install]
+RequiredBy=unlock-private-fs.service
--- a/packages/release/prepare-private-fs.service
+++ b/packages/release/prepare-private-fs.service
@@ -1,0 +1,27 @@
+[Unit]
+Description=Prepare Private Filesystem (/.bottlerocket)
+DefaultDependencies=no
+Conflicts=umount.target
+Wants=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device
+After=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device unlock-private-fs.service
+BindsTo=unlock-private-fs.service
+Before=\x2ebottlerocket.mount
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+Environment=PRIVATE_PARTITION_BLOCK_DEVICE=/dev/mapper/BOTTLEROCKET-PRIVATE
+# Relative to the ext4 defaults, we:
+# - adjust the inode ratio since we expect lots of small files
+# - retain the inode size to allow most settings to be stored inline
+# - retain the block size to handle worse-case alignment for hardware
+# - enable the ext4's native per-file encryption capability
+ExecStart=/usr/sbin/mkfs.ext4 -b 4096 -i 4096 -I 256 -F -O encrypt ${PRIVATE_PARTITION_BLOCK_DEVICE}
+
+RemainAfterExit=true
+StandardOutput=journal
+StandardError=journal+console
+
+[Install]
+RequiredBy=\x2ebottlerocket.mount
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -64,6 +64,7 @@ Source1027: root-.aws.mount
 Source1028: opt-cni.mount
 Source1029: opt-csi.mount
 Source1030: opt-civ.mount
+Source1031: run-rottweiler.mount
 
 # Mounts that require helper programs.
 Source1040: prepare-boot.service
@@ -145,6 +146,19 @@ Source1650: prepare-local-fs-encrypted.conf
 Source1651: local-mount-encrypted.conf
 Source1652: repart-local-encrypted.conf
 
+# Ephemeral (per-boot) encryption key drop-ins.
+Source1661: encrypt-datastore-ephemeral.conf
+Source1662: unlock-datastore-ephemeral.conf
+Source1663: encrypt-local-fs-ephemeral.conf
+Source1664: repart-local-ephemeral.conf
+
+# Full private-partition LUKS encryption units (ephemeral-encryption-keys).
+Source1670: encrypt-private-fs.service
+Source1671: unlock-private-fs.service
+Source1672: prepare-private-fs.service
+Source1673: bottlerocket-mount-encrypted.conf
+Source1674: encrypt-datastore-private-luks.conf
+
 # Runtime FIPS activation sources.
 # These are always installed but only activate when fips=1 is on the kernel command line.
 Source1700: generate-fips-env.service
@@ -191,6 +205,8 @@ Requires: %{_cross_os}xfsprogs
 Requires: %{_cross_os}libkcapi
 Requires: (%{name}-fips if %{_cross_os}image-feature(fips))
 Requires: (%{name}-crypt if %{_cross_os}image-feature(encrypted-storage))
+Requires: (%{name}-ephemeral-crypt if %{_cross_os}image-feature(ephemeral-encryption-keys))
+
 
 %description
 %{summary}.
@@ -210,6 +226,13 @@ Requires: (%{_cross_os}image-feature(encrypted-storage) and %{name})
 Requires: %{_cross_os}rottweiler
 
 %description crypt
+%{summary}.
+
+%package ephemeral-crypt
+Summary: Bottlerocket release, with ephemeral encryption keys
+Requires: (%{_cross_os}image-feature(ephemeral-encryption-keys) and %{name}-crypt)
+
+%description ephemeral-crypt
 %{summary}.
 
 %package swap
@@ -275,7 +298,7 @@ install -p -m 0644 \
   %{S:1006} %{S:1007} %{S:1008} %{S:1009} %{S:1010} \
   %{S:1020} %{S:1021} %{S:1022} %{S:1023} %{S:1024} \
   %{S:1025} %{S:1026} %{S:1027} %{S:1028} %{S:1029} \
-  %{S:1030} \
+  %{S:1030} %{S:1031} \
   %{S:1040} %{S:1041} %{S:1042} %{S:1043} %{S:1044} \
   %{S:1045} %{S:1046} %{S:1047} %{S:1048} %{S:1049} \
   %{S:1050} \
@@ -396,6 +419,27 @@ install -p -m 0644 %{S:1651} %{buildroot}%{_cross_unitdir}/local.mount.d/10-encr
 install -d %{buildroot}%{_cross_unitdir}/repart-local.service.d
 install -p -m 0644 %{S:1652} %{buildroot}%{_cross_unitdir}/repart-local.service.d/10-encrypted.conf
 
+install -d %{buildroot}%{_cross_unitdir}/encrypt-datastore.service.d
+install -p -m 0644 %{S:1661} %{buildroot}%{_cross_unitdir}/encrypt-datastore.service.d/20-ephemeral.conf
+
+install -d %{buildroot}%{_cross_unitdir}/unlock-datastore.service.d
+install -p -m 0644 %{S:1662} %{buildroot}%{_cross_unitdir}/unlock-datastore.service.d/20-ephemeral.conf
+
+install -d %{buildroot}%{_cross_unitdir}/encrypt-local-fs.service.d
+install -p -m 0644 %{S:1663} %{buildroot}%{_cross_unitdir}/encrypt-local-fs.service.d/20-ephemeral.conf
+
+install -d %{buildroot}%{_cross_unitdir}/repart-local.service.d
+install -p -m 0644 %{S:1664} %{buildroot}%{_cross_unitdir}/repart-local.service.d/20-ephemeral.conf
+
+# Full private-partition LUKS encryption units (ephemeral-encryption-keys).
+install -p -m 0644 %{S:1670} %{S:1671} %{S:1672} %{buildroot}%{_cross_unitdir}
+
+BOTTLEROCKET_PATH=$(systemd-escape --path /.bottlerocket)
+install -d %{buildroot}%{_cross_unitdir}/${BOTTLEROCKET_PATH}.mount.d
+install -p -m 0644 %{S:1673} %{buildroot}%{_cross_unitdir}/${BOTTLEROCKET_PATH}.mount.d/10-encrypted.conf
+
+install -p -m 0644 %{S:1674} %{buildroot}%{_cross_unitdir}/encrypt-datastore.service.d/30-private-luks.conf
+
 ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 
 %files
@@ -428,6 +472,7 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 %{_cross_unitdir}/prepare-var.service
 %{_cross_unitdir}/repart-local.service
 %{_cross_unitdir}/\x2ebottlerocket.mount
+%dir %{_cross_unitdir}/\x2ebottlerocket.mount.d
 %{_cross_unitdir}/var.mount
 %{_cross_unitdir}/opt.mount
 %{_cross_unitdir}/mnt.mount
@@ -510,7 +555,9 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 
 %files crypt
 %{_cross_unitdir}/encrypt-datastore.service
+%dir %{_cross_unitdir}/encrypt-datastore.service.d
 %{_cross_unitdir}/encrypt-local-fs.service
+%dir %{_cross_unitdir}/encrypt-local-fs.service.d
 %{_cross_unitdir}/measure-cmdline.service
 %{_cross_unitdir}/measure-settings.service
 %{_cross_unitdir}/systemd-pcrphase-configured.service
@@ -518,10 +565,23 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 %{_cross_unitdir}/systemd-pcrphase-preconfigured.service
 %{_cross_unitdir}/systemd-pcrphase-sysinit.service
 %{_cross_unitdir}/unlock-datastore.service
+%dir %{_cross_unitdir}/unlock-datastore.service.d
 %{_cross_unitdir}/unlock-local-fs.service
 %{_cross_unitdir}/local.mount.d/10-encrypted.conf
 %{_cross_unitdir}/prepare-local-fs.service.d/10-encrypted.conf
 %{_cross_unitdir}/repart-local.service.d/10-encrypted.conf
+
+%files ephemeral-crypt
+%{_cross_unitdir}/encrypt-datastore.service.d/20-ephemeral.conf
+%{_cross_unitdir}/encrypt-datastore.service.d/30-private-luks.conf
+%{_cross_unitdir}/unlock-datastore.service.d/20-ephemeral.conf
+%{_cross_unitdir}/encrypt-local-fs.service.d/20-ephemeral.conf
+%{_cross_unitdir}/repart-local.service.d/20-ephemeral.conf
+%{_cross_unitdir}/encrypt-private-fs.service
+%{_cross_unitdir}/unlock-private-fs.service
+%{_cross_unitdir}/prepare-private-fs.service
+%{_cross_unitdir}/\x2ebottlerocket.mount.d/10-encrypted.conf
+%{_cross_unitdir}/run-rottweiler.mount
 
 %files swap
 %{_cross_sysctldir}/81-release-swap.conf
--- a/packages/release/repart-local-ephemeral.conf
+++ b/packages/release/repart-local-ephemeral.conf
@@ -1,0 +1,5 @@
+[Service]
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
+# Delete the sealed data-partition encryption key after it has been used for the
+# time in the boot, forcing format and fresh key generation in the next boot.
+ExecStartPost=/usr/bin/rottweiler delete-key bottlerocket-data
--- a/packages/release/run-rottweiler.mount
+++ b/packages/release/run-rottweiler.mount
@@ -1,0 +1,16 @@
+[Unit]
+Description=Ephemeral keystore directory
+DefaultDependencies=no
+Conflicts=umount.target
+After=selinux-policy-files.service
+Wants=selinux-policy-files.service
+Before=encrypt-datastore.service encrypt-local-fs.service encrypt-private-fs.service
+
+[Mount]
+What=tmpfs
+Where=/run/rottweiler
+Type=tmpfs
+Options=nosuid,nodev,noexec,noatime,noswap,context=system_u:object_r:private_t:s0,mode=0700
+
+[Install]
+RequiredBy=local-fs.target
--- a/packages/release/unlock-datastore-ephemeral.conf
+++ b/packages/release/unlock-datastore-ephemeral.conf
@@ -1,0 +1,5 @@
+[Service]
+# Delete the key used to encrypt the datastore after it has been used for the final
+# time in the boot, forcing new key generation and a complete rebuild of the datastore
+# in the subsequent boot.
+ExecStartPost=/usr/bin/rottweiler delete-key datastore
--- a/packages/release/unlock-private-fs.service
+++ b/packages/release/unlock-private-fs.service
@@ -1,0 +1,29 @@
+[Unit]
+Description=Unlock Private Filesystem (/.bottlerocket)
+DefaultDependencies=no
+Conflicts=umount.target
+IgnoreOnIsolate=true
+After=cryptsetup-pre.target systemd-udevd-kernel.socket dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device
+Wants=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device blockdev@dev-mapper-BOTTLEROCKET\x2dPRIVATE.target
+Before=cryptsetup.target blockdev@dev-mapper-BOTTLEROCKET\x2dPRIVATE.target
+ConditionSecurity=tpm2
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+Environment=BOTTLEROCKET_PRIVATE=/dev/disk/by-partlabel/BOTTLEROCKET-PRIVATE
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
+
+ExecStart=/usr/bin/rottweiler attach block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private
+# Delete the private-partition encryption key from the tmpfs keystore after it has
+# been used to unlock the volume, so a fresh key is generated on the next boot.
+ExecStartPost=/usr/bin/rottweiler delete-key bottlerocket-private
+ExecStop=/usr/bin/rottweiler detach block-device ${BOTTLEROCKET_PRIVATE}
+
+RemainAfterExit=yes
+StandardOutput=journal
+StandardError=journal+console
+
+[Install]
+RequiredBy=prepare-private-fs.service
--- a/sources/bottlerocket-image-features/src/lib.rs
+++ b/sources/bottlerocket-image-features/src/lib.rs
@@ -15,6 +15,7 @@ Currently supported feature flags:
 
 - `IN_PLACE_UPDATES` - Controls whether in-place updates are enabled (default: true)
 - `ENCRYPTED_STORAGE` - Controls whether encrypted storage is enabled (default: false)
+- `EPHEMERAL_ENCRYPTION_KEYS` - Controls whether encryption keys are regenerated every boot (default: false)
 
 ## Usage
 
@@ -56,6 +57,8 @@ pub struct ImageFeatures {
     pub in_place_updates: bool,
     #[serde(default)]
     pub encrypted_storage: bool,
+    #[serde(default)]
+    pub ephemeral_encryption_keys: bool,
 }
 
 fn default_true() -> bool {
@@ -68,6 +71,7 @@ pub fn parse_image_features() -> Result<ImageFeatures> {
         return Ok(ImageFeatures {
             in_place_updates: true,
             encrypted_storage: false,
+            ephemeral_encryption_keys: false,
         });
     }
 
@@ -148,5 +152,26 @@ IN_PLACE_UPDATES="false"
         let content = "";
         let features = parse_image_features_from_str(content).unwrap();
         assert_eq!(features.in_place_updates, true);
+        assert_eq!(features.ephemeral_encryption_keys, false);
+    }
+
+    #[test]
+    fn test_parse_env_ephemeral_encryption_keys_false() {
+        let content = r#"EPHEMERAL_ENCRYPTION_KEYS="false""#;
+        let features = parse_image_features_from_str(content).unwrap();
+        assert_eq!(features.ephemeral_encryption_keys, false);
+    }
+
+    #[test]
+    fn test_parse_env_ephemeral_encryption_keys_true() {
+        let content = r#"EPHEMERAL_ENCRYPTION_KEYS="true""#;
+        let features = parse_image_features_from_str(content).unwrap();
+        assert_eq!(features.ephemeral_encryption_keys, true);
+    }
+
+    #[test]
+    fn test_parse_env_ephemeral_encryption_keys_invalid() {
+        let content = r#"EPHEMERAL_ENCRYPTION_KEYS="yes""#;
+        assert!(parse_image_features_from_str(content).is_err());
     }
 }
--- a/sources/rottweiler/src/key.rs
+++ b/sources/rottweiler/src/key.rs
@@ -1,6 +1,6 @@
 use snafu::prelude::*;
 use std::fs;
-use std::io::Read;
+use std::io::{self, Read};
 use std::path::PathBuf;
 use zeroize::Zeroizing;
 
@@ -12,15 +12,21 @@ const DEV_RANDOM: &str = "/dev/random";
 const KEYSTORE: &str = "/.bottlerocket/keystore";
 const KEY_SIZE: usize = 64;
 
+fn validate_key_id(key_id: &str) -> Result<()> {
+    snafu::ensure_whatever!(
+        !key_id.is_empty()
+            && key_id
+                .chars()
+                .all(|c| c.is_alphanumeric() || c == '-' || c == '_'),
+        "key_id must be non-empty and contain only alphanumerics, dashes, and underscores"
+    );
+
+    Ok(())
+}
+
 /// Generate a random encryption key and encrypt it with TPM2 PCRs 7+14
 pub fn generate(key_id: String) -> Result<()> {
-    if !key_id
-        .chars()
-        .all(|c| c.is_alphanumeric() || c == '-' || c == '_')
-    {
-        whatever!("key_id must contain only alphanumerics, dashes, and underscores");
-    }
-
+    validate_key_id(&key_id)?;
     let key_path = PathBuf::from(KEYSTORE).join(&key_id);
 
     // Skip generation if key already exists
@@ -48,8 +54,27 @@ pub fn generate(key_id: String) -> Result<()> {
     Ok(())
 }
 
+/// Delete a sealed key from the keystore.
+pub fn delete(key_id: String) -> Result<()> {
+    validate_key_id(&key_id)?;
+    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
+
+    // Skip deletion if key doesn't exists
+    if !key_path.exists() {
+        return Ok(());
+    }
+
+    match fs::remove_file(&key_path) {
+        Ok(()) => Ok(()),
+        Err(e) if e.kind() == io::ErrorKind::NotFound => Ok(()),
+        Err(e) => Err(e)
+            .with_whatever_context(|_| format!("failed to delete key '{}'", key_path.display())),
+    }
+}
+
 /// Load and decrypt a TPM2-encrypted key from the keystore
 pub fn load(key_id: String) -> Result<Zeroizing<Vec<u8>>> {
+    validate_key_id(&key_id)?;
     let key_path = PathBuf::from(KEYSTORE).join(&key_id);
 
     let encrypted = fs::read(&key_path)
--- a/sources/rottweiler/src/key.rs
+++ b/sources/rottweiler/src/key.rs
@@ -9,8 +9,26 @@ use crate::system;
 type Result<T> = std::result::Result<T, snafu::Whatever>;
 
 const DEV_RANDOM: &str = "/dev/random";
-const KEYSTORE: &str = "/.bottlerocket/keystore";
+const KEYSTORE_PERMANENT: &str = "/.bottlerocket/keystore";
+const KEYSTORE_EPHEMERAL: &str = "/run/rottweiler";
 const KEY_SIZE: usize = 64;
+
+/// Checks if ephemeral encryption keys feature is enabled via image features
+fn ephemeral_encryption_keys_enabled() -> Result<bool> {
+    let features = bottlerocket_image_features::parse_image_features()
+        .with_whatever_context(|_| "failed to load image features")?;
+    Ok(features.ephemeral_encryption_keys)
+}
+
+fn keystore_dir() -> PathBuf {
+    let is_ephemeral_encryption_keys_enabled = ephemeral_encryption_keys_enabled().unwrap_or(false);
+    // If ephemeral encryption keys are enabled, we store the keys in tmpfs rather than the disk
+    if is_ephemeral_encryption_keys_enabled {
+        PathBuf::from(KEYSTORE_EPHEMERAL)
+    } else {
+        PathBuf::from(KEYSTORE_PERMANENT)
+    }
+}
 
 fn validate_key_id(key_id: &str) -> Result<()> {
     snafu::ensure_whatever!(
@@ -27,7 +45,8 @@ fn validate_key_id(key_id: &str) -> Result<()> {
 /// Generate a random encryption key and encrypt it with TPM2 PCRs 7+14
 pub fn generate(key_id: String) -> Result<()> {
     validate_key_id(&key_id)?;
-    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
+    let keystore = keystore_dir();
+    let key_path = keystore.join(&key_id);
 
     // Skip generation if key already exists
     if key_path.exists() {
@@ -45,8 +64,12 @@ pub fn generate(key_id: String) -> Result<()> {
 
     let encrypted = system::systemd_creds_encrypt(&key_id, &random_bytes)?;
 
-    fs::create_dir_all(KEYSTORE)
-        .with_whatever_context(|_| format!("failed to create keystore directory '{}'", KEYSTORE))?;
+    fs::create_dir_all(&keystore).with_whatever_context(|_| {
+        format!(
+            "failed to create keystore directory '{}'",
+            keystore.display()
+        )
+    })?;
 
     fs::write(&key_path, encrypted)
         .with_whatever_context(|_| format!("failed to write key to '{}'", key_path.display()))?;
@@ -57,7 +80,7 @@ pub fn generate(key_id: String) -> Result<()> {
 /// Delete a sealed key from the keystore.
 pub fn delete(key_id: String) -> Result<()> {
     validate_key_id(&key_id)?;
-    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
+    let key_path = keystore_dir().join(&key_id);
 
     // Skip deletion if key doesn't exists
     if !key_path.exists() {
@@ -75,7 +98,7 @@ pub fn delete(key_id: String) -> Result<()> {
 /// Load and decrypt a TPM2-encrypted key from the keystore
 pub fn load(key_id: String) -> Result<Zeroizing<Vec<u8>>> {
     validate_key_id(&key_id)?;
-    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
+    let key_path = keystore_dir().join(&key_id);
 
     let encrypted = fs::read(&key_path)
         .with_whatever_context(|_| format!("failed to read key from '{}'", key_path.display()))?;
--- a/sources/rottweiler/src/main.rs
+++ b/sources/rottweiler/src/main.rs
@@ -12,6 +12,7 @@ interface for encrypting and managing encrypted storage resources including:
 
 ### Key Management
 - `generate-key <key-id>` - Generate an encryption key
+- `delete-key <key-id>` - Delete an encryption key from the keystore
 
 ### Block Device Operations
 - `encrypt block-device <path> <key-id>` - Encrypt a block device using LUKS
@@ -92,6 +93,7 @@ fn main() -> Result<()> {
 
     match args.command {
         Command::GenerateKey(cmd) => key::generate(cmd.key_id),
+        Command::DeleteKey(cmd) => key::delete(cmd.key_id),
         Command::Encrypt(cmd) => match cmd.resource {
             EncryptResource::BlockDevice(cmd) => block_device::encrypt(cmd.path, cmd.key_id),
             EncryptResource::Directory(cmd) => directory::encrypt(cmd.path, cmd.key_id),
@@ -193,6 +195,7 @@ struct Args {
 #[argh(subcommand)]
 enum Command {
     GenerateKey(GenerateKeyCmd),
+    DeleteKey(DeleteKeyCmd),
     Encrypt(EncryptCmd),
     Attach(AttachCmd),
     Detach(DetachCmd),
@@ -207,6 +210,14 @@ enum Command {
 #[argh(subcommand, name = "generate-key")]
 /// Generate an encryption key
 struct GenerateKeyCmd {
+    #[argh(positional)]
+    key_id: String,
+}
+
+#[derive(FromArgs)]
+#[argh(subcommand, name = "delete-key")]
+/// Delete an encryption key from the keystore
+struct DeleteKeyCmd {
     #[argh(positional)]
     key_id: String,
 }
--- a/packages/release/bottlerocket-mount-encrypted.conf
+++ b/packages/release/bottlerocket-mount-encrypted.conf
@@ -1,0 +1,8 @@
+[Unit]
+# mount the LUKS mapper instead of the raw partition and only after the
+# partition has been encrypted, unlocked and formatted.
+After=unlock-private-fs.service prepare-private-fs.service
+Requires=prepare-private-fs.service
+
+[Mount]
+What=/dev/mapper/BOTTLEROCKET-PRIVATE
--- a/packages/release/encrypt-datastore-ephemeral.conf
+++ b/packages/release/encrypt-datastore-ephemeral.conf
@@ -1,0 +1,4 @@
+[Service]
+# Drop the "only if unencrypted" guard so the datastore directory is
+# re-encrypted with a freshly generated key on every boot.
+ExecCondition=
--- a/packages/release/encrypt-datastore.service
+++ b/packages/release/encrypt-datastore.service
@@ -11,6 +11,7 @@ RefuseManualStop=true
 [Service]
 Type=oneshot
 Environment=DATASTORE_DIR=/.bottlerocket/datastore
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
 
 # Only run if the directory is not yet encrypted.
 ExecCondition=/usr/bin/rottweiler check directory ${DATASTORE_DIR} unencrypted
--- a/packages/release/encrypt-local-fs-ephemeral.conf
+++ b/packages/release/encrypt-local-fs-ephemeral.conf
@@ -1,0 +1,4 @@
+[Service]
+# Drop the "only if unencrypted" guard so the Data partition is
+# re-encrypted with a freshly generated key on every boot.
+ExecCondition=
--- a/packages/release/encrypt-local-fs.service
+++ b/packages/release/encrypt-local-fs.service
@@ -12,6 +12,7 @@ RefuseManualStop=true
 [Service]
 Type=oneshot
 Environment=BOTTLEROCKET_DATA=/dev/disk/by-partlabel/BOTTLEROCKET-DATA
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
 
 # Only format if not already encrypted.
 ExecCondition=/usr/bin/rottweiler check block-device ${BOTTLEROCKET_DATA} unencrypted
--- a/packages/release/unlock-datastore.service
+++ b/packages/release/unlock-datastore.service
@@ -9,6 +9,7 @@ RefuseManualStop=true
 [Service]
 Type=oneshot
 Environment=DATASTORE_DIR=/.bottlerocket/datastore
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
 
 ExecStart=/usr/bin/rottweiler unlock directory ${DATASTORE_DIR} datastore
 
--- a/packages/release/unlock-local-fs.service
+++ b/packages/release/unlock-local-fs.service
@@ -13,6 +13,7 @@ RefuseManualStop=true
 [Service]
 Type=oneshot
 Environment=BOTTLEROCKET_DATA=/dev/disk/by-partlabel/BOTTLEROCKET-DATA
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
 
 ExecStart=/usr/bin/rottweiler attach block-device ${BOTTLEROCKET_DATA} bottlerocket-data
 ExecStop=/usr/bin/rottweiler detach block-device ${BOTTLEROCKET_DATA}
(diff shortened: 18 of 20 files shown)


sources/rottweiler/src/key.rs
```diff
@@ -48,6 +48,30 @@ pub fn generate(key_id: String) -> Result<()> {
...
+    if !key_path.exists() {
+        return Ok(());
+    }
+
+    match fs::remove_file(&key_path) {
+        Ok(()) => Ok(()),
+        Err(e) => Err(e).with_whatever_context(|_| {
+            format!("failed to delete key '{}'", key_path.display())
+        }),
+    }
```
jpculp:
```suggestion
    fs::remove_file(&key_path)
        .with_whatever_context(|_| format!("failed to delete key '{}'", key_path.display()))
```
Would this be more consistent with `load()`?

packages/release/early-boot-config-ephemeral.conf
```diff
@@ -0,0 +1,8 @@
+[Unit]
+# ephemeral-encryption-keys feature wipes and re-encrypts the datastore on
+# every boot, which drops all settings that early-boot-config applied on first
+# boot. In case of persistent datastore, early-boot-config uses a marker file
+# /var/lib/bottlerocket/early-boot-config.ran to run on the first boot only.
+# Resetting this allows early-boot-config to run every boot and re-apply
+# user data after each wipe.
+ConditionPathExists=
```
piyush-jena: instead of this just don't create the marker file

packages/early-boot-config/early-boot-config.service
```diff
@@ -16,6 +16,7 @@ RefuseManualStop=true
 
 [Service]
 Type=oneshot
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
```
jmt-lab: **[P1] EnvironmentFile is mandatory: unit fails if `image-features.env` is missing**

`early-boot-config.service` is a pre-existing unit used by *every* Bottlerocket variant. Adding an unconditional `EnvironmentFile=/usr/share/bottlerocket/image-features.env` means any image that does not ship that file (older variants, or a broken build) will fail the unit with `Failed to load environment files: No such file or directory`, blocking `preconfigured.target` and therefore first-boot configuration. This is a regression path for images that don't opt into ephemeral-encryption-keys at all.

Per `systemd.exec(5)`: "If the specified file does not exist... this is considered an error unless the filename is prefixed with `-`".

**Fix:** prefix with `-`:

```ini
EnvironmentFile=-/usr/share/bottlerocket/image-features.env
```

Consider the same defensive prefix on the other new units (`encrypt-private-fs.service`, `unlock-private-fs.service`, `repart-local-ephemeral.conf`, etc.) unless the RPM dependency graph guarantees the file is present in every variant that installs them.

packages/release/unlock-datastore-ephemeral.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+# Delete the key used to encrypt the datastore after it has been used for the final
+# time in the boot, forcing new key generation and a complete rebuild of the datastore
+# in the subsequent boot.
+ExecStartPost=-/usr/bin/rottweiler delete-key datastore
```
jmt-lab: **[P1] Wrong key deleted — datastore fscrypt key is NOT scrubbed after unlock**

This drop-in is for `unlock-datastore.service` (fscrypt-encrypted `/.bottlerocket/datastore` directory, key name `datastore`), not the block-device data partition (key name `bottlerocket-data`).

`repart-local-ephemeral.conf` correctly deletes `bottlerocket-data`. This drop-in should delete `datastore` to match `encrypt-datastore.service` (`generate-key datastore`).

**Consequence:** After `unlock-datastore.service` runs, `/run/rottweiler/datastore` remains in tmpfs for the rest of the boot because we deleted the wrong key. `rottweiler delete-key bottlerocket-data` becomes a no-op (already deleted upstream). This silently defeats the ephemeral property for the datastore fscrypt key — the very thing users are opting into with this feature.

**Fix:**

```ini
[Service]
# Delete the datastore fscrypt encryption key after it has been used for the
# final time in the boot, forcing fresh key generation on the next boot.
ExecStartPost=-/usr/bin/rottweiler delete-key datastore
```

packages/release/unlock-private-fs.service
```diff
@@ -0,0 +1,30 @@
...
+[Service]
+Type=oneshot
+Environment=BOTTLEROCKET_PRIVATE=/dev/disk/by-partlabel/BOTTLEROCKET-PRIVATE
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
+
+ExecStart=/usr/bin/rottweiler attach block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private
+# Delete the sealed data-partition encryption key after it has been used for the
+# time in the boot, forcing format and fresh key generation in the next boot.
+ExecStartPost=-/usr/bin/rottweiler delete-key bottlerocket-private
+ExecStop=/usr/bin/rottweiler detach block-device ${BOTTLEROCKET_PRIVATE}
```
jmt-lab, line 22: **[P1] Key not scrubbed on unlock failure — ephemeral property violated**

`systemd.service(5)`: `ExecStartPost=` runs only when the main `ExecStart=` succeeds. If `rottweiler attach block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private` fails (corrupted LUKS header, TPM2 unseal mismatch, udev race), the plaintext key stays in `/run/rottweiler/bottlerocket-private` for the remainder of the boot. On an emergency shell (or any process running as root) it is readable, violating the design invariant that "the key exists only briefly during boot."

The PR description notes that the next boot reformats, so on-disk data is not at risk — but the *key* exposure window during the current boot is larger than intended.

The same concern applies to `repart-local-ephemeral.conf:5` and `unlock-datastore-ephemeral.conf:5`.

**Fix (any one of):**
- Add `ExecStopPost=-/usr/bin/rottweiler delete-key bottlerocket-private` in addition to the current post-success cleanup; `ExecStopPost=` runs on both success and failure paths.
- Wrap into a single subcommand (`rottweiler unlock-and-forget`) that scrubs the key in a Rust `finally`-style block.
- Alternatively, since the key already exists in tmpfs and the unit is oneshot, have `rottweiler attach` itself unlink the key file after (successful or failed) LUKS open.

```ini
ExecStart=/usr/bin/rottweiler attach block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private
ExecStartPost=-/usr/bin/rottweiler delete-key bottlerocket-private
# Ensure cleanup even on failure paths
ExecStopPost=-/usr/bin/rottweiler delete-key bottlerocket-private
ExecStop=/usr/bin/rottweiler detach block-device ${BOTTLEROCKET_PRIVATE}
```

Also verify `rottweiler detach` doesn't itself need the key — by ExecStop time the key file will already be gone.

packages/release/tmpfs-keystore.conf
```diff
@@ -0,0 +1,2 @@
+d /run/rottweiler 0700 root root -
+t /run/rottweiler - - - - security.selinux=system_u:object_r:private_t:s0
```
jmt-lab: **[P1] tmpfs is swappable — key material can hit persistent storage**

`/run/rottweiler` lives in `/run`, which is `tmpfs`. On Linux, tmpfs pages *are* swappable unless swap is disabled or the pages are `mlock`ed. The design uses this location to hold plaintext LUKS/fscrypt keys between generate/attach/delete.

`release-swap` subpackage (`release.spec:584`, `prepare-swap.service`) enables zram-based swap on Bottlerocket. If an ephemeral-encryption-keys image also ships swap (nothing in `ephemeral-crypt` prevents it), a memory-pressure event during the boot window between `generate-key` and `delete-key` can page the key out. For zram this is in-RAM (probably fine), but any variant configuring disk-backed swap or hibernation makes this a real key-persistence bug — undermining the "ephemeral" property.

**Fix options (pick one):**
1. Back the key with a kernel keyring entry (`keyctl`, already pulled in via `Requires: keyutils` at `release.spec:187`) or a `memfd_create(..., MFD_CLOEXEC)` handle, plus `mlock()` in the `rottweiler` process. This is the most robust option.
2. Mount `/run/rottweiler` as its own `ramfs` (never swappable) via a mount unit rather than reusing `/run`.
3. Add `Conflicts: %{name}-swap` on `%package ephemeral-crypt` (`release.spec:227`) or declare a hard incompatibility with disk-backed swap in the image-feature docs.

packages/release/prepare-private-fs.service
```diff
@@ -0,0 +1,27 @@
...
+Type=oneshot
+Environment=PRIVATE_PARTITION_BLOCK_DEVICE=/dev/mapper/BOTTLEROCKET-PRIVATE
+
+# Since we are re-encrypting wthe private partition, we run wipefs to reliably erase stale signatures.
+ExecStartPre=-/usr/sbin/wipefs --all --force ${PRIVATE_PARTITION_BLOCK_DEVICE}
+# Relative to the ext4 defaults, we:
+  # - adjust the inode ratio since we expect lots of small files
+  # - retain the inode size to allow most settings to be stored inline
+  # - retain the block size to handle worse-case alignment for hardware
+ExecStart=/usr/sbin/mkfs.ext4 -b 4096 -i 4096 -I 256 -F -O encrypt ${PRIVATE_PARTITION_BLOCK_DEVICE}
```
jmt-lab: **[P1] Reformat-every-boot chain has weak observability**

`ExecStartPre=-/usr/sbin/wipefs --all --force ...` swallows any wipefs failure, and `mkfs.ext4 -F` forces mkfs regardless. In the happy path this is fine — the mapper was just created by `cryptsetup luksFormat` in `encrypt-private-fs.service` and is expected to be blank. But if `encrypt-private-fs.service` was skipped or partially completed, or the mapper is somehow busy, there is no signal that anything went wrong: `mkfs -F` overwrites, and failure shows up only as a downstream mount failure.

**Fix:**
- Drop the `-` on `ExecStartPre=` — a genuine wipefs failure on a freshly-created mapper is unexpected and worth failing on:
  ```ini
  ExecStartPre=/usr/sbin/wipefs --all --force ${PRIVATE_PARTITION_BLOCK_DEVICE}
  ```
- Alternatively, remove `wipefs` entirely since a fresh LUKS mapper has no filesystem signatures to erase.
- Add `StandardOutput=journal+console` so `mkfs` / `wipefs` progress is captured (see the separate P2 finding on this file).

packages/release/prepare-private-fs.service
```diff
@@ -0,0 +1,27 @@
...
+
+RemainAfterExit=true
+StandardError=journal+console
```
jmt-lab, line 24: **[P2] Inconsistent journaling: `StandardOutput` unset**

This unit sets `StandardError=journal+console` but omits `StandardOutput=`, so `mkfs.ext4` progress and `wipefs` output go to `/dev/null` (systemd default). Siblings `encrypt-private-fs.service:22-23` and `unlock-private-fs.service:26-27` set `StandardOutput=journal+console` (with `StandardError=inherit`) — this unit is the odd one out.

**Fix:**

```ini
RemainAfterExit=true
StandardOutput=journal+console
StandardError=journal+console
```

or match the sibling pattern with `StandardOutput=journal+console` + `StandardError=inherit`.

packages/release/encrypt-private-fs.service
```diff
@@ -0,0 +1,26 @@
+[Unit]
+Description=Encrypt Private Filesystem (/.bottlerocket)
+DefaultDependencies=no
+Wants=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device
+After=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device tpm2.target
```
jmt-lab: **[P2] Missing `After=cryptsetup-pre.target` — potential udev race on luksFormat**

`unlock-private-fs.service:6` and `unlock-local-fs.service:6` both order themselves `After=cryptsetup-pre.target systemd-udevd-kernel.socket`. This unit orders only `After=... .device tpm2.target`.

`rottweiler encrypt block-device` invokes `cryptsetup luksFormat` under the hood. cryptsetup expects the device-mapper subsystem and udev to be initialized (which is exactly what `cryptsetup-pre.target` gates). On a slow-udev boot the format can race with "Cannot use device ..., name is invalid or still in use" or similar. This is why upstream `cryptsetup.target` uses `cryptsetup-pre.target`.

**Fix:**

```ini
After=cryptsetup-pre.target systemd-udevd-kernel.socket dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dPRIVATE.device tpm2.target
Before=cryptsetup.target unlock-private-fs.service
```

packages/release/tmpfs-keystore.conf
jmt-lab: **[P2] SELinux labeling via tmpfiles `t` line is fragile — prefer `file_contexts`**

The `t` type in `tmpfiles.d(5)` sets extended attributes; the 6-field syntax is correct. But the `d` line above it creates the directory *first*, at which point SELinux auto-labels it based on policy `file_contexts` (likely `var_run_t` or an unlabeled fallback). The `t` directive then overrides. That works, but the correct fix lives in the SELinux policy, not in tmpfiles:

```
/run/rottweiler(/.*)?    system_u:object_r:private_t:s0
```

That way `restorecon` and initial creation both DTRT, and any future `systemd-tmpfiles --remove` + `--create` cycle doesn't leave a race window where the label is wrong.

**Also worth checking:** `private_t` is a very generic type name. If it's the same type as the encrypted private partition, using it for a tmpfs keystore may be too coarse for policy auditing. Consider a dedicated `bottlerocket_keystore_t`.

**Fix:** Move the label rule into the Bottlerocket SELinux policy package's `file_contexts` and drop line 2 (or keep it but add a comment cross-referencing the policy source so the two stay in sync).

packages/release/encrypt-private-fs.service
```diff
@@ -0,0 +1,26 @@
...
+ExecStart=/usr/bin/rottweiler generate-key bottlerocket-private
+ExecStart=/usr/bin/rottweiler encrypt block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private
+
+UMask=0077
+RemainAfterExit=true
+StandardOutput=journal+console
+StandardError=inherit
+
+[Install]
+RequiredBy=unlock-private-fs.service
```
jmt-lab, line 27: **[P2] `[Install]` sections need a preset to take effect on `RefuseManualStart=true` units**

The dependency graph is a valid DAG:
- `encrypt-private-fs.service` → `unlock-private-fs.service` → `prepare-private-fs.service` → `\x2ebottlerocket.mount`
- `bottlerocket-mount-encrypted.conf` reinforces the tail via `Requires=prepare-private-fs.service`.

No cycles. Good.

However, `[Install] RequiredBy=` only works if `systemctl enable` (or a systemd preset) creates the symlink. For `RefuseManualStart=true` units, only a preset is realistic. Bottlerocket's `%{_cross_os}systemd` package typically ships a preset, and `encrypt-datastore.service:31` already uses this pattern — so presumably it works. Worth explicitly confirming for the three *new* units (`encrypt-private-fs.service`, `unlock-private-fs.service`, `prepare-private-fs.service`).

**Fix / suggestion:** Add these three unit names to the preset list, or lean on `bottlerocket-mount-encrypted.conf`-style drop-in `Requires=` (which doesn't depend on `[Install]`). The mount drop-in already covers the tail; consider also propagating `Requires=` up the chain in each unit's `[Unit]` section, so the graph is preset-independent.

packages/release/encrypt-datastore-private-luks.conf
```diff
@@ -0,0 +1,7 @@
+[Service]
+# In case of ephemeral-encryption-keys, we format the BOTTLEROCKET-PRIVATE with
+# `encrypt` feature flag. So, we drop the tune2fs call, which is used to enable
+# the flag and generate the key and encrypt the datastore.
+ExecStart=
+ExecStart=/usr/bin/rottweiler generate-key datastore
+ExecStart=/usr/bin/rottweiler encrypt directory ${DATASTORE_DIR} datastore
```
jmt-lab, line 7: **[P2] Drop-in assumes BOTTLEROCKET-PRIVATE was formatted with `-O encrypt`, but ephemeral-crypt has no TPM dependency**

The comment describes this override as specifically for the case where BOTTLEROCKET-PRIVATE was formatted with `-O encrypt` in `prepare-private-fs.service` — i.e. when ephemeral-encryption-keys is active on a TPM system.

`encrypt-datastore.service:5` has `ConditionSecurity=tpm2`, so on a non-TPM system the whole unit (and hence this drop-in) is a no-op. That mitigates the concern *at unit level*, but it means the entire ephemeral-crypt subpackage is silently inert on non-TPM systems — yet `%package ephemeral-crypt` (release.spec:227-229) does not declare a TPM dependency.

**Fix (suggestion):** Either
- add a TPM-capable-image dependency to `%package ephemeral-crypt`, or
- document the TPM assumption in `bottlerocket-image-features` docs and add a header comment on this drop-in noting that it's contingent on the TPM-gated parent unit.

Also confirm drop-in ordering: `20-ephemeral.conf` clears `ExecCondition=` (correct), then `30-private-luks.conf` clears `ExecStart=` and appends fresh commands (correct). systemd applies drop-ins lexically, so `20 < 30` — good — but a comment stating the intended ordering would help future maintainers.

packages/release/unlock-private-fs.service
```diff
@@ -0,0 +1,30 @@
...
+
+[Service]
+Type=oneshot
+Environment=BOTTLEROCKET_PRIVATE=/dev/disk/by-partlabel/BOTTLEROCKET-PRIVATE
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
+
+ExecStart=/usr/bin/rottweiler attach block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private
+# Delete the sealed data-partition encryption key after it has been used for the
+# time in the boot, forcing format and fresh key generation in the next boot.
+ExecStartPost=-/usr/bin/rottweiler delete-key bottlerocket-private
```
jmt-lab: **[P2] Wrong-partition comment (copy-paste from repart-local-ephemeral.conf)**

Line 20-21 reads "Delete the sealed data-partition encryption key" but the unit deletes `bottlerocket-private` (line 22) — the private-partition key, not the data-partition key.

These comments are the primary in-repo documentation of the ephemeral-key flow. Wrong ones lead to exactly the kind of copy-paste key-name mismatch already visible in `unlock-datastore-ephemeral.conf` (see the P1 finding on that file).

**Fix:**

```
# Delete the private-partition encryption key from the tmpfs keystore after
# it has been used to unlock the volume, so a fresh key is generated on the
# next boot.
```

Also fixes the awkward "for the time in the boot" phrasing.

packages/release/release.spec
```diff
@@ -515,6 +564,23 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
...
+%{_cross_unitdir}/encrypt-datastore.service.d/20-ephemeral.conf
+%{_cross_unitdir}/encrypt-datastore.service.d/30-private-luks.conf
+%{_cross_unitdir}/unlock-datastore.service.d/20-ephemeral.conf
+%{_cross_unitdir}/encrypt-local-fs.service.d/20-ephemeral.conf
+%{_cross_unitdir}/repart-local.service.d/20-ephemeral.conf
+%{_cross_unitdir}/prepare-local-fs.service.d/20-ephemeral.conf
+%{_cross_unitdir}/encrypt-private-fs.service
+%{_cross_unitdir}/unlock-private-fs.service
+%{_cross_unitdir}/prepare-private-fs.service
+%{_cross_unitdir}/\x2ebottlerocket.mount.d/10-encrypted.conf
```
jmt-lab, line 583: **[P3] RPM `%dir` treatment is asymmetric across subpackages**

`%files crypt` (lines 552–565) packages `prepare-local-fs.service.d/10-encrypted.conf` and `repart-local.service.d/10-encrypted.conf` with **no** `%dir` declarations for those directories. `%files ephemeral-crypt` (lines 567–582) packages sibling `20-ephemeral.conf` files with **no** `%dir` for `prepare-local-fs.service.d` / `repart-local.service.d`, but **does** declare `%dir` for `encrypt-datastore.service.d`, `unlock-datastore.service.d`, `encrypt-local-fs.service.d`, and `\x2ebottlerocket.mount.d`.

RPM auto-owns parent directories via each contained file, so installation works today. But this is fragile: if `crypt`'s `10-encrypted.conf` is ever moved elsewhere, the shared directory becomes orphaned.

**Fix (for consistency):** Since `ephemeral-crypt` already `Requires: %{name}-crypt` (line 229), let `crypt` own all shared drop-in directories via `%dir` and remove the `%dir` lines from `ephemeral-crypt`. Directories will always exist when ephemeral files are installed.

packages/release/prepare-private-fs.service
```diff
@@ -0,0 +1,27 @@
...
+Requires=unlock-private-fs.service
+Before=\x2ebottlerocket.mount
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+Environment=PRIVATE_PARTITION_BLOCK_DEVICE=/dev/mapper/BOTTLEROCKET-PRIVATE
+
+# Since we are re-encrypting wthe private partition, we run wipefs to reliably erase stale signatures.
```
jmt-lab: **[P3] Typo + slightly-inaccurate description**

Line 15 reads:

```
# Since we are re-encrypting wthe private partition, we run wipefs to reliably erase stale signatures.
```

- `wthe` → `the`.
- This unit doesn't re-encrypt — it re-formats the just-created LUKS mapper with a fresh ext4.

**Fix:**

```
# The LUKS mapper is freshly created on every boot with a new key. Wipe any
# stale signatures before mkfs to be safe.
```

packages/release/unlock-datastore-ephemeral.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+# Delete the key used to encrypt the datastore after it has been used for the final
+# time in the boot, forcing new key generation and a complete rebuild of the datastore
+# in the subsequent boot.
```
jmt-lab, line 4: **[P3] Awkward phrasing: missing "final"**

`repart-local-ephemeral.conf:3-4` has the correct phrasing "used for the **final** time in the boot". Two other comments dropped the word:

- `unlock-private-fs.service:20-21`
- `unlock-datastore-ephemeral.conf:3-4`

**Fix:** insert "final" so each reads "after it has been used for the final time in the boot".

sources/rottweiler/src/key.rs
```diff
@@ -9,9 +10,25 @@ use crate::system;
 type Result<T> = std::result::Result<T, snafu::Whatever>;
 
 const DEV_RANDOM: &str = "/dev/random";
-const KEYSTORE: &str = "/.bottlerocket/keystore";
+const KEYSTORE_PERMANENT: &str = "/.bottlerocket/keystore";
+const KEYSTORE_EPHEMERAL: &str = "/run/rottweiler";
+const EPHEMERAL_ENCRYPTION_KEYS_ENVVAR: &str = "EPHEMERAL_ENCRYPTION_KEYS";
```
jmt-lab: **[P1] Three parsers, three grammars, silent fallback to `false`**

The same feature flag is parsed by three different code paths, two of which silently swallow parse errors:

1. `bottlerocket-image-features/src/lib.rs` uses `envy` → serde bool → **strict** `true`/`false`, returns an error otherwise.
2. `rottweiler/src/key.rs::keystore_dir()` (this file, lines 13–24) calls `env::var(...).ok().and_then(|v| v.parse().ok()).unwrap_or(false)`.
3. `early-boot-config/src/main.rs::run()` uses the same lossy pattern.

`str::parse::<bool>` only accepts exact lowercase `"true"` / `"false"`. Anything else — `"yes"`, `"1"`, `"True"`, `" true"` (leading whitespace), empty string — parses as `Err`, is dropped by `.ok()`, and becomes `false`.

**Data-loss scenario:** Operator writes `EPHEMERAL_ENCRYPTION_KEYS="yes"` (or `"1"`, `"True"`) into `image-features.env` intending to enable the feature. `rottweiler generate-key` silently writes the key to `/.bottlerocket/keystore/` (permanent, on disk) — the very failure mode the flag exists to prevent.

**Fix:** route both binaries through the shared crate, or extract a single helper that logs a warning on unparseable values:

```rust
pub fn ephemeral_encryption_keys_enabled() -> bool {
    match env::var("EPHEMERAL_ENCRYPTION_KEYS") {
        Ok(v) => match v.trim().parse::<bool>() {
            Ok(b) => b,
            Err(_) => {
                warn!("EPHEMERAL_ENCRYPTION_KEYS has non-boolean value {:?}; defaulting to false", v);
                false
            }
        }
        Err(env::VarError::NotPresent) => false,
        Err(e) => {
            warn!("EPHEMERAL_ENCRYPTION_KEYS unreadable: {}; defaulting to false", e);
            false
        }
    }
}
```

sources/rottweiler/src/key.rs
```diff
@@ -9,9 +10,25 @@ use crate::system;
...
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(false);
+
+    // If ephemeral encryption keys are enabled, we store the keys in tmpfs rather than the disk
+    if ephemeral_encryption_keys_enabled {
+        PathBuf::from(KEYSTORE_EPHEMERAL)
+    } else {
+        PathBuf::from(KEYSTORE_PERMANENT)
+    }
+}
```
jmt-lab: **[P1] `keystore_dir()` re-reads env on every call — contract lives in every systemd unit**

`keystore_dir()` is called from `generate()`, `load()`, and `delete()`. Each call re-reads `EPHEMERAL_ENCRYPTION_KEYS` from the process environment. Since `rottweiler` is invoked as a short-lived oneshot per unit, the "environment" is whatever the invoking unit sets via `EnvironmentFile=`.

**Failure scenario:** A future contributor writes a new unit or a debugging `systemctl` invocation that forgets `EnvironmentFile=/usr/share/bottlerocket/image-features.env`. `keystore_dir()` silently returns `/.bottlerocket/keystore`. `load()` reads a nonexistent path and returns a generic "failed to read key" error; nothing points to the real cause. Worst case: `encrypt-local-fs.service` was supposed to run ephemeral, but a mis-propagated env var causes the key to be written to the persistent keystore — the disk is now LUKS-encrypted with an at-rest-persistent key on the very partition that was supposed to have zero at-rest persistence.

**Fix (defense in depth):**

- Cache the value in `std::sync::OnceLock` so all calls in one invocation are guaranteed consistent.
- Log the resolved mode at startup: `info!("rottweiler keystore = {} (ephemeral={})", ..., ...)`.
- Better yet: promote the mode to an explicit `--ephemeral` CLI flag so the systemd units make intent visible at each call site.

```rust
use std::sync::OnceLock;

fn keystore_dir() -> &'static Path {
    static KEYSTORE: OnceLock<PathBuf> = OnceLock::new();
    KEYSTORE.get_or_init(|| {
        let ephemeral = env::var(EPHEMERAL_ENCRYPTION_KEYS_ENVVAR)
            .ok()
            .and_then(|v| v.trim().parse().ok())
            .unwrap_or(false);
        let path = if ephemeral { KEYSTORE_EPHEMERAL } else { KEYSTORE_PERMANENT };
        info!("rottweiler: using keystore {} (ephemeral={})", path, ephemeral);
        PathBuf::from(path)
    })
}
```

sources/rottweiler/src/key.rs
```diff
@@ -39,18 +57,44 @@ pub fn generate(key_id: String) -> Result<()> {
...
+    fs::remove_file(&key_path)
+        .with_whatever_context(|_| format!("failed to delete key '{}'", key_path.display()))?;
+
+    Ok(())
+}
+
 /// Load and decrypt a TPM2-encrypted key from the keystore
 pub fn load(key_id: String) -> Result<Zeroizing<Vec<u8>>> {
-    let key_path = PathBuf::from(KEYSTORE).join(&key_id);
+    let key_path = keystore_dir().join(&key_id);
```
jmt-lab, line 101: **[P1] Dropped invariant: `load()` no longer sanitizes `key_id`**

`generate()` (lines 34–39) and the new `delete()` (lines 75–80) both enforce that `key_id` matches `[A-Za-z0-9_-]+`. `load()` (line 73 in the new file) does not — and this PR changed `load()` from `PathBuf::from(KEYSTORE).join(...)` to `keystore_dir().join(...)`, so the path traversal surface is worth re-examining.

**Scenario:** any future caller passing untrusted input as `key_id` (e.g. a user-data field) can request `key_id = "../../etc/machine-id"`, `"foo/../bar"`, or absolute `"/etc/passwd"` (with an absolute rhs, `Path::join` *replaces* the base). `fs::read` on the traversed path succeeds; the contents are then fed to `systemd-creds decrypt --name <untrusted>`. Decryption fails safely (wrong AAD), so this is not an exploitable data leak *today* — all in-repo callers pass literals — but the whole point of `key.rs` is to be a safe API for the storage stack.

**Fix:** factor validation into a private helper and call it from all three functions, or introduce a `KeyId` newtype with `TryFrom<String>`:

```rust
fn validate_key_id(key_id: &str) -> Result<()> {
    snafu::ensure_whatever!(
        !key_id.is_empty()
            && key_id.chars().all(|c| c.is_alphanumeric() || c == '-' || c == '_'),
        "key_id must be non-empty and contain only alphanumerics, dashes, and underscores"
    );
    Ok(())
}

pub fn load(key_id: String) -> Result<Zeroizing<Vec<u8>>> {
    validate_key_id(&key_id)?;
    let key_path = keystore_dir().join(&key_id);
    // ...
}
```

sources/rottweiler/src/key.rs
```diff
@@ -39,18 +57,44 @@ pub fn generate(key_id: String) -> Result<()> {
...
+
+    let key_path = keystore_dir().join(&key_id);
+
+    // Skip deletion if key doesn't exists
+    if !key_path.exists() {
+        return Ok(());
+    }
+
+    fs::remove_file(&key_path)
+        .with_whatever_context(|_| format!("failed to delete key '{}'", key_path.display()))?;
```
jmt-lab: **[P2] `delete()` is not idempotent under concurrency (TOCTOU)**

```rust
if !key_path.exists() {
    return Ok(());
}
fs::remove_file(&key_path)
    .with_whatever_context(|_| format!("failed to delete key '{}'", key_path.display()))?;
```

`unlock-private-fs.service` (`ExecStartPost=`) and `repart-local-ephemeral.conf` (`ExecStartPost=`) both call `rottweiler delete-key`. During shutdown/restart, or in a debug scenario, two invocations can race: both see `exists()` → true, both try `remove_file`, the loser gets `ENOENT` and the unit reports failure — even though the desired end state (key gone) is achieved. The `-` prefix on the systemd side hides this, but it's still an unnecessary error path.

Also: two syscalls where one suffices.

**Fix:**

```rust
use std::io;

match fs::remove_file(&key_path) {
    Ok(()) => Ok(()),
    Err(e) if e.kind() == io::ErrorKind::NotFound => Ok(()),
    Err(e) => Err(e).with_whatever_context(|_| {
        format!("failed to delete key '{}'", key_path.display())
    }),
}
```

Atomic and idempotent.

sources/rottweiler/src/key.rs
```diff
@@ -39,18 +57,44 @@ pub fn generate(key_id: String) -> Result<()> {
...
+
+    Ok(())
+}
+
```
jmt-lab, line 97: **[P2] `unlink()` doesn't zero storage — residue may linger on the permanent keystore**

`fs::remove_file` is a plain `unlink(2)`. For the tmpfs ephemeral keystore this is genuinely secure — pages are freed to the kernel and never touch disk. For the permanent keystore (`/.bottlerocket/keystore`, if `delete-key` is ever invoked there due to env-var mis-propagation), the encrypted blob's on-disk sectors are simply unreferenced.

**Why it's reachable:** `main.rs::DeleteKey` unconditionally routes through `keystore_dir()`. If a `delete-key` invocation is missing its `EnvironmentFile=`, it operates on the persistent path (see the sibling `keystore_dir()` P1 finding).

**Fix (pick one):**

- **Preferred (explicit invariant):**
  ```rust
  let keystore = keystore_dir();
  snafu::ensure_whatever!(
      keystore == Path::new(KEYSTORE_EPHEMERAL),
      "delete() is only supported on the ephemeral keystore",
  );
  ```
- Or overwrite the file with zeros + `fsync` before `unlink` (note: journaled/CoW filesystems may not honor this).
- At minimum, document the intended keystore in the function's docstring.

sources/early-boot-config/early-boot-config/src/main.rs
```diff
@@ -234,9 +242,11 @@ async fn run() -> Result<()> {
...
     }
 
-    fs::write(MARKER_FILE, "").unwrap_or_else(|e| {
-        warn!("Failed to create marker file {MARKER_FILE}, may unexpectedly run again: {e}")
-    });
+    if !ephemeral_encryption_keys_enabled {
+        fs::write(MARKER_FILE, "").unwrap_or_else(|e| {
+            warn!("Failed to create marker file {MARKER_FILE}, may unexpectedly run again: {e}")
+        });
+    }
```
jmt-lab: **[P2] Silent parser default combined with marker-file logic → stateful failure on mode switch**

Same lossy parse pattern as `rottweiler`:

```rust
let ephemeral_encryption_keys_enabled = env::var(EPHEMERAL_ENCRYPTION_KEYS_ENVVAR)
    .ok()
    .and_then(|v| v.parse().ok())
    .unwrap_or(false);
```

Combined with the `if !ephemeral_encryption_keys_enabled { fs::write(MARKER_FILE, ...) }` block below.

**Concrete failure mode:** operator ships `EPHEMERAL_ENCRYPTION_KEYS="yes"`. Both binaries treat it as `false`. `early-boot-config` writes the marker → subsequent boots skip user-data application via `ConditionPathExists=/var/lib/bottlerocket/early-boot-config.ran`. Later, the operator fixes the value to `"true"` and upgrades. If the marker survives the upgrade path (e.g. `/var/lib/bottlerocket` was not reformatted on this particular mode switch), the system boots with an empty, freshly-reformatted datastore but `early-boot-config` never runs — no user data applied, services depending on API settings misbehave.

**Fix:**

1. Route through the shared `bottlerocket-image-features` crate (see the sibling P1 finding on parser divergence).
2. When the value is present but unparseable, `error!` with the value and either abort or explicitly default with a loud log.
3. Consider actively *removing* the marker file at the start of `run()` when ephemeral mode is enabled — that cleans up stale marker state on a mode switch.

sources/bottlerocket-image-features/src/lib.rs
```diff
@@ -68,6 +71,7 @@ pub fn parse_image_features() -> Result<ImageFeatures> {
         return Ok(ImageFeatures {
             in_place_updates: true,
             encrypted_storage: false,
+            ephemeral_encryption_keys: false,
```
jmt-lab, line 74: **[P2] No unit tests for `ephemeral_encryption_keys` — grammar is only implicitly pinned**

The existing test module tests `IN_PLACE_UPDATES` four different ways but never touches `ENCRYPTED_STORAGE` or the new `ephemeral_encryption_keys`. Since this crate is the *canonical* parser and there is now a real divergence risk with the ad-hoc parsers in `rottweiler` and `early-boot-config`, the following tests should exist:

1. `EPHEMERAL_ENCRYPTION_KEYS="true"` → `features.ephemeral_encryption_keys == true`.
2. `EPHEMERAL_ENCRYPTION_KEYS="false"` → `false`.
3. Missing key → default `false`.
4. **Contract-pinning test**: `"yes"`, `"1"`, `"True"` — assert the current behavior (parser errors) so a future `envy` / `serde` bump doesn't silently change the accepted grammar.
5. Round-trip test that the same string in `image-features.env` is parsed the same way by both `parse_image_features_from_str` and the ad-hoc `env::var(...).parse::<bool>()` path used in `rottweiler` and `early-boot-config` — a test that *surfaces* the divergence would document intent.

**Doc gap:** module-level docstring says "Boolean values (\"true\" or \"false\")" but does not state case sensitivity, whitespace handling, or quote handling. Systemd's `EnvironmentFile=` strips outer quotes and does not further trim. Pinning this in prose and tests is worthwhile.

sources/rottweiler/src/main.rs
```diff
@@ -211,6 +214,14 @@ struct GenerateKeyCmd {
...
 }
 
+#[derive(FromArgs)]
+#[argh(subcommand, name = "delete-key")]
+/// Delete an encryption key from the keystore
+struct DeleteKeyCmd {
+    #[argh(positional)]
+    key_id: String,
+}
+
```
jmt-lab, line 224: **[P3] Silent destructive op + missing idempotency documentation**

Two small improvements:

1. **Silent-on-success.** `delete-key` prints nothing on success (and silently succeeds when the key doesn't exist). If a human ever runs it interactively (e.g. debugging a stuck boot) with a wrong `key-id`, there's no signal until the next unlock fails. A one-liner like `println!("deleted key '{key_id}' from {}", keystore_dir().display())` on success (and a distinct message on "already gone") would help.

2. **Docstring gap.** The module docstring says `delete-key <key-id> - Delete an encryption key from the keystore`. It should note:
   - This is a no-op if the key does not exist (idempotent contract).
   - In ephemeral-keys mode this is the mechanism used to force re-generation on next boot (see the systemd unit comments in `repart-local-ephemeral.conf`).

Neither blocks the PR.

jmt-lab (changes requested):

packages/release/unlock-datastore-ephemeral.conf
piyush-jena: hehe! Guess it misread something.

packages/release/tmpfs-keystore.conf
piyush-jena: The suggestion didn't work. Ended up using a tmpfs mount.

sources/rottweiler/src/key.rs
piyush-jena: opted using `bottlerocket-image-features` library to handle this.

sources/early-boot-config/early-boot-config/src/main.rs
piyush-jena: decided to use the bottlerocket-image-features library to parse.

packages/release/release.spec
piyush-jena: No these dropins are specific to ephemeral encruption keys. It doesn't make sense to include them in crypt.

packages/release/unlock-private-fs.service
piyush-jena: A failure here would mean a boot failure. The ephemeral key in itself doesn't hold value.

packages/early-boot-config/early-boot-config.service
piyush-jena: There's precedent - https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/0ef38ed222e2436989f134bcfa11eb1aa07719b1/packages/release/prepare-local-fs.service#L13. It should be present.

piyush-jena: twoliter adds this file to the image in rpm2img.

packages/release/tmpfs-keystore.conf
arnaldo2792: Keys are encrypted, not stored in plain-text, the next version wil include a mount with noswap.

packages/release/encrypt-datastore-private-luks.conf
arnaldo2792: This is a nit. Is that what P2 means?

sources/rottweiler/src/key.rs
arnaldo2792:
```
Failure scenario: A future contributor writes a new unit or a debugging systemctl invocation that forgets EnvironmentFile=/usr/share/bottlerocket/image-features.env. keystore_dir() silently returns /.bottlerocket/keystore. load() reads a nonexistent path and returns a generic "failed to read key" error; nothing points to the real cause. Worst case: encrypt-local-fs.service was supposed to run ephemeral, but a mis-propagated env var causes the key to be written to the persistent keystore — the disk is now LUKS-encrypted with an at-rest-persistent key on the very partition that was supposed to have zero at-rest persistence.
```

This is a test case we should cover in our release.

```
Better yet: promote the mode to an explicit --ephemeral CLI flag so the systemd units make intent visible at each call site.
```

The intention is to have one standalone service that knows how to behave based of its environment. Adding a flag will require us to have two services, and still conditionally run one or the other, or include one or the other. So to actually prevent the failure mode, we need an integ test that prevents the failure mode.

packages/release/encrypt-datastore-private-luks.conf
piyush-jena: dependency enforced through encrypted storage.

arnaldo2792 (approved):

packages/release/run-rottweiler.mount
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Ephemeral keystore directory
+DefaultDependencies=no
+Conflicts=umount.target
+After=selinux-policy-files.service
+Wants=selinux-policy-files.service
+Before=encrypt-datastore.service encrypt-localfs.service encrypt-privatefs.service
```
ginglis13:
```suggestion
Before=encrypt-datastore.service encrypt-local-fs.service encrypt-private-fs.service
```

sources/early-boot-config/early-boot-config/src/main.rs
```diff
@@ -234,9 +245,13 @@ async fn run() -> Result<()> {
...
-    fs::write(MARKER_FILE, "").unwrap_or_else(|e| {
-        warn!("Failed to create marker file {MARKER_FILE}, may unexpectedly run again: {e}")
-    });
+    // We do not write the marker file in case ephemeral encryption keys feature is enabled. This
+    // makes sure that post a reboot we repopulate the user-data in the datastore.
+    if is_ephemeral_encryption_keys_enabled {
+        fs::write(MARKER_FILE, "").unwrap_or_else(|e| {
+            warn!("Failed to create marker file {MARKER_FILE}, may unexpectedly run again: {e}")
+        });
+    }
```
ginglis13: The comment doesn't match the logic here - 

> We do not write the marker file in case ephemeral encryption keys feature is enabled.

> if is_ephemeral_encryption_keys_enabled -> write the file

packages/release/tmpfs-keystore.conf
ginglis13, line 1: This file doesn't appear to be used

packages/release/run-rottweiler.mount
vigh-m: Similar error in the `[Install]` section below

jmt-lab (approved):

vigh-m:
⬆️ Updated following the comment and retested:
```bash
[root@admin]# sheltie
bash-5.2# ls -lZ /run/rottweiler
total 0
bash-5.2# ls -lZ /run | grep rottweiler
drwx------.  2 root   root   system_u:object_r:private_t:s0     40 Aug 15 04:17 rottweiler
bash-5.2# ls -lia /run/rottweiler/
total 0
1 drwx------.  2 root root  40 Aug 15 04:17 .
1 drwxr-xr-x. 21 root root 500 Aug 15 04:17 ..
bash-5.2# lsblk -o NAME,TYPE,FSTYPE,MOUNTPOINT
NAME                     TYPE  FSTYPE         MOUNTPOINT
nvme1n1                  disk
`-nvme1n1p1              part  crypto_LUKS
  `-BOTTLEROCKET-DATA    crypt xfs            /local
nvme0n1                  disk
|-nvme0n1p1              part
|-nvme0n1p2              part  vfat
|-nvme0n1p3              part  vfat           /boot
|-nvme0n1p4              part  erofs
|-nvme0n1p5              part  DM_verity_hash
|-nvme0n1p6              part
|-nvme0n1p7              part  crypto_LUKS
| `-BOTTLEROCKET-PRIVATE crypt ext4           /var/lib/bottlerocket
`-nvme0n1p8              part
bash-5.2# apiclient lockdown
04:19:26 [INFO] Lockdown completed
bash-5.2# apiclient get settings.motd
Failed to get settings: Failed GET request to '/?prefix=settings.motd': Status 500 when GETing /?prefix=settings.motd: Data store error during get_prefix 'settings.motd' for Live: Data store integrity violation at /var/lib/bottlerocket/datastore/current/live: Live datastore missing
bash-5.2# systemctl --failed
  UNIT LOAD ACTIVE SUB DESCRIPTION

0 loaded units listed.
```

piyush-jena:
Removed `early-boot-config` change in the force push because private partition is recreated. WIth the latest twoliter and kernel-kit (built from develop), fips works as well

```
bash-5.2# ls -lZ /run/rottweiler
total 0
bash-5.2# ls -lZ /run | grep rottweiler
drwx------.  2 root   root   system_u:object_r:private_t:s0     40 Aug 17 20:05 rottweiler
bash-5.2# lsblk -o NAME,TYPE,FSTYPE,MOUNTPOINT
NAME                     TYPE  FSTYPE         MOUNTPOINT
nvme0n1                  disk
|-nvme0n1p1              part
|-nvme0n1p2              part  vfat
|-nvme0n1p3              part  vfat           /boot
|-nvme0n1p4              part  erofs
|-nvme0n1p5              part  DM_verity_hash
|-nvme0n1p6              part
|-nvme0n1p7              part  crypto_LUKS
| `-BOTTLEROCKET-PRIVATE crypt ext4           /var/lib/bottlerocket
`-nvme0n1p8              part
nvme1n1                  disk
`-nvme1n1p1              part  crypto_LUKS
  `-BOTTLEROCKET-DATA    crypt xfs            /local
bash-5.2# rottweiler check directory /.bottlerocket/datastore encrypted
directory '/.bottlerocket/datastore' is encrypted
bash-5.2# apiclient get settings.motd
{
  "settings": {
    "motd": "hello"
  }
}
bash-5.2# apiclient report fips
Benchmark name:  FIPS Security Policy
Version:         v1.0.0
Reference:       https://csrc.nist.gov/
Benchmark level: 1
Start time:      2026-08-17T20:07:55.010321100Z

[PASS] 1.0       FIPS mode is enabled. (Automatic)
[PASS] 1.1       FIPS module is Amazon Linux 2023 Kernel Cryptographic API. (Automatic)
[PASS] 1.2       FIPS self-tests passed. (Automatic)

Passed:          3
Failed:          0
Skipped:         0
Total checks:    3

Compliance check result: PASS
bash-5.2# systemctl status encrypt-local-fs.service
● encrypt-local-fs.service - Encrypt Local Filesystem (/local)
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/encrypt-local-fs.service; enabled; preset: enabled)
             └─00-aws-config.conf, 00-fips-go.conf, 10-requires-tmp.conf
             /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/encrypt-local-fs.service.d
             └─20-ephemeral.conf
     Active: active (exited) since Mon 2026-08-17 20:05:54 UTC; 2min 24s ago
 Invocation: b86902673df749078bd77b1e3bf01b75
   Main PID: 1650 (code=exited, status=0/SUCCESS)
   Mem peak: 2.5M
        CPU: 443ms

Aug 17 20:05:53 localhost systemd[1]: Starting Encrypt Local Filesystem (/local)...
Aug 17 20:05:54 localhost systemd[1]: Finished Encrypt Local Filesystem (/local).
bash-5.2# journalctl | head -5
Aug 17 20:05:51 localhost kernel: Linux version 6.18.39 (builder@buildkitsandbox) (x86_64-bottlerocket-linux-gnu-gcc (Buildroot 2025.05.1) 13.4.0, GNU ld (GNU Binutils) 2.44) #2 SMP PREEMPT_DYNAMIC Mon Aug 17 18:20:33 UTC 2026
Aug 17 20:05:51 localhost kernel: Command line: console=tty0 console=ttyS0,115200n8 net.ifnames=0 netdog.default-interface=eth0:dhcp4,dhcp6? quiet fips=1 module_blacklist=i8042 SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1 SYSTEMD_DEFAULT_MOUNT_RATE_LIMIT_BURST=25 module_blacklist=i8042 fips=1 init.systemd.unit=fipscheck.target root=/dev/dm-0 rootwait ro raid=noautodetect random.trust_cpu=on selinux=1 enforcing=1dm-mod.create="root,,,ro,0 661264 verity 1 PARTUUID=D52978BD-EF33-445C-9444-CFFD7748FC2D/PARTNROFF=1 PARTUUID=D52978BD-EF33-445C-9444-CFFD7748FC2D/PARTNROFF=2 4096 4096 82658 1 sha256 577af503ec391ccdcc1738294cbd37287392152a0b57e356ebfb73ee4cfd12fa 7c42a7fe3dcb190463029c6998e50628f0bda2c81f03dadf895daf6fe51c74fd 2 restart_on_corruption ignore_zero_blocks" -- systemd.log_target=journal-or-kmsg systemd.log_color=0 systemd.show_status=true
Aug 17 20:05:51 localhost kernel: KASLR enabled
Aug 17 20:05:51 localhost kernel: BIOS-provided physical RAM map:
Aug 17 20:05:51 localhost kernel: BIOS-e820: [mem 0x0000000000000000-0x000000000009ffff] usable
bash-5.2# journalctl -u bootstrap-commands
Aug 17 20:05:56 ip-172-31-28-220.us-west-2.compute.internal systemd[1]: Starting Bootstrap Commands...
Aug 17 20:05:56 ip-172-31-28-220.us-west-2.compute.internal bootstrap-commands[2004]: 20:05:56 [INFO] Processing bootstrap command 'mybootstrap' ...
Aug 17 20:05:56 ip-172-31-28-220.us-west-2.compute.internal bootstrap-commands[2004]: 20:05:56 [INFO] Successfully ran bootstrap command 'mybootstrap'
Aug 17 20:05:56 ip-172-31-28-220.us-west-2.compute.internal systemd[1]: Finished Bootstrap Commands.
```
In the first boot we get the following journal logs. This shows that the bootstrap-commands run every time.
```
bash-5.2# journalctl | head -5
Aug 17 19:59:18 localhost kernel: Linux version 6.18.39 (builder@buildkitsandbox) (x86_64-bottlerocket-linux-gnu-gcc (Buildroot 2025.05.1) 13.4.0, GNU ld (GNU Binutils) 2.44) #2 SMP PREEMPT_DYNAMIC Mon Aug 17 18:20:33 UTC 2026
Aug 17 19:59:18 localhost kernel: Command line: console=tty0 console=ttyS0,115200n8 net.ifnames=0 netdog.default-interface=eth0:dhcp4,dhcp6? quiet fips=1 module_blacklist=i8042 SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1 SYSTEMD_DEFAULT_MOUNT_RATE_LIMIT_BURST=25 module_blacklist=i8042 fips=1 init.systemd.unit=fipscheck.target root=/dev/dm-0 rootwait ro raid=noautodetect random.trust_cpu=on selinux=1 enforcing=1dm-mod.create="root,,,ro,0 661264 verity 1 PARTUUID=D52978BD-EF33-445C-9444-CFFD7748FC2D/PARTNROFF=1 PARTUUID=D52978BD-EF33-445C-9444-CFFD7748FC2D/PARTNROFF=2 4096 4096 82658 1 sha256 577af503ec391ccdcc1738294cbd37287392152a0b57e356ebfb73ee4cfd12fa 7c42a7fe3dcb190463029c6998e50628f0bda2c81f03dadf895daf6fe51c74fd 2 restart_on_corruption ignore_zero_blocks" -- systemd.log_target=journal-or-kmsg systemd.log_color=0 systemd.show_status=true
Aug 17 19:59:18 localhost kernel: KASLR enabled
Aug 17 19:59:18 localhost kernel: BIOS-provided physical RAM map:
Aug 17 19:59:18 localhost kernel: BIOS-e820: [mem 0x0000000000000000-0x000000000009ffff] usable
```

piyush-jena:
In `aws-ecs-4` we see that `once` mode bootstrap-commands don't apply again
```
bash-5.2# journalctl -u bootstrap-commands
Aug 17 20:13:44 ip-172-31-42-194.us-west-2.compute.internal systemd[1]: Starting Bootstrap Commands...
Aug 17 20:13:44 ip-172-31-42-194.us-west-2.compute.internal bootstrap-commands[1794]: 20:13:44 [INFO] Processing bootstrap command 'mybootstrap' ...
Aug 17 20:13:44 ip-172-31-42-194.us-west-2.compute.internal bootstrap-commands[1794]: 20:13:44 [INFO] Turning off bootstrap command 'mybootstrap'
Aug 17 20:13:44 ip-172-31-42-194.us-west-2.compute.internal bootstrap-commands[1794]: 20:13:44 [INFO] Successfully ran bootstrap command 'mybootstrap'
Aug 17 20:13:44 ip-172-31-42-194.us-west-2.compute.internal systemd[1]: Finished Bootstrap Commands.
Aug 17 20:15:30 ip-172-31-42-194.us-west-2.compute.internal systemd[1]: bootstrap-commands.service: Deactivated successfully.
Aug 17 20:15:30 ip-172-31-42-194.us-west-2.compute.internal systemd[1]: Stopped Bootstrap Commands.
-- Boot 4b2fdec6fdd14cbc9d83708c8573e8ea --
Aug 17 20:15:38 ip-172-31-42-194.us-west-2.compute.internal systemd[1]: Starting Bootstrap Commands...
Aug 17 20:15:38 ip-172-31-42-194.us-west-2.compute.internal bootstrap-commands[1695]: 20:15:38 [INFO] Bootstrap command mode for 'mybootstrap' is 'off'
Aug 17 20:15:38 ip-172-31-42-194.us-west-2.compute.internal systemd[1]: Finished Bootstrap Commands.
```

piyush-jena:
Tried changing the settings in ecs-5-fips and post reboot I found that only settings through user-data persisted.
In case of ecs-4, they behaved as expected. Changes in settings persisted after reboots.

packages/release/prepare-private-fs.service
```diff
@@ -0,0 +1,26 @@
...
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+Environment=PRIVATE_PARTITION_BLOCK_DEVICE=/dev/mapper/BOTTLEROCKET-PRIVATE
+# Relative to the ext4 defaults, we:
+# - adjust the inode ratio since we expect lots of small files
+# - retain the inode size to allow most settings to be stored inline
+# - retain the block size to handle worse-case alignment for hardware
+ExecStart=/usr/sbin/mkfs.ext4 -b 4096 -i 4096 -I 256 -F -O encrypt ${PRIVATE_PARTITION_BLOCK_DEVICE}
```
arnaldo2792, line 20: nit: doesn't explain `encrypt`

packages/release/prepare-private-fs.service
```diff
@@ -0,0 +1,26 @@
...
+
+RemainAfterExit=true
+StandardOutput=journal+console
```
arnaldo2792: For STDOUT, the convention is to send it to the journal only.

packages/release/release.spec
```diff
@@ -212,6 +228,13 @@ Requires: %{_cross_os}rottweiler
 %description crypt
 %{summary}.
 
+%package ephemeral-crypt
+Summary: Bottlerocket release, with ephemeral (per-boot) encryption keys
```
arnaldo2792: No need to put `(per-boot)`:


```suggestion
Bottlerocket release, ephemeral encryption keys edition
```

packages/release/release.spec
```diff
@@ -523,6 +567,22 @@ ln -s preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 %{_cross_unitdir}/prepare-local-fs.service.d/10-encrypted.conf
 %{_cross_unitdir}/repart-local.service.d/10-encrypted.conf
 
+%files ephemeral-crypt
+%dir %{_cross_unitdir}/encrypt-datastore.service.d
+%dir %{_cross_unitdir}/unlock-datastore.service.d
+%dir %{_cross_unitdir}/encrypt-local-fs.service.d
+%dir %{_cross_unitdir}/\x2ebottlerocket.mount.d
```
arnaldo2792: these technically are owned by `crypt`

packages/release/repart-local-ephemeral.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
+# Delete the sealed data-partition encryption key after it has been used for the
+# time in the boot, forcing format and fresh key generation in the next boot.
+ExecStartPost=-/usr/bin/rottweiler delete-key bottlerocket-data
```
arnaldo2792: Actually, if we failed to delete, we shouldn't silently continue - we don't want to leave keys behind.

packages/release/run-rottweiler.mount
```diff
@@ -0,0 +1,16 @@
...
+Before=encrypt-datastore.service encrypt-local-fs.service encrypt-private-fs.service
+
+[Mount]
+What=tmpfs
+Where=/run/rottweiler
+Type=tmpfs
+Options=nosuid,nodev,noexec,noatime,noswap,context=system_u:object_r:private_t:s0,mode=0700
+
+[Install]
+RequiredBy=encrypt-datastore.service encrypt-local-fs.service encrypt-private-fs.service
```
arnaldo2792: Can you make it `RequiredBy=local-fs.target`. It should run early enough, way before all these services.

packages/release/unlock-datastore-ephemeral.conf
arnaldo2792: Same as elsewhere, hard-fail if we failed to delete the key.

packages/release/unlock-private-fs.service
```diff
@@ -0,0 +1,29 @@
...
+
+ExecStart=/usr/bin/rottweiler attach block-device ${BOTTLEROCKET_PRIVATE} bottlerocket-private
+# Delete the private-partition encryption key from the tmpfs keystore after it has
+# been used to unlock the volume, so a fresh key is generated on the next boot.
+ExecStartPost=-/usr/bin/rottweiler delete-key bottlerocket-private
+ExecStop=/usr/bin/rottweiler detach block-device ${BOTTLEROCKET_PRIVATE}
+
+RemainAfterExit=yes
+StandardOutput=journal+console
+StandardError=inherit
```
arnaldo2792: STDOUT, as elsewhere, to the journal only. STDERR, journal+console.

piyush-jena:
force push addresses the above comments.

arnaldo2792 (approved):

PR #1001, opened 2026-08-06
Set ephemeral storage allowlist via packaging

### LINKED ISSUE #1000: Make ephemeral storage allowlist variant agnostic
**What I'd like:**

The ephemeral storage allowlist for directories that can be bind-mounted is gated by variant name, e.g.:

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/ba49cc484a4894ec98b5abf894d084f278f20d1b/sources/api/apiserver/src/server/ephemeral_storage.rs#L456-L460

Downstream builders of custom variants cannot bind for example `/var/lib/soci-snapshotter` on their variant if its name does not include `"k8s"`.

Instead, package configuration for paths that can be bind-mounted, and have apiserver read these configuration files. This way, a path is allowlisted if the package that corresponds to that path is installed.

Keep the status quo

ginglis13:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/1000 (1000!)

**Description of changes:**

Package configuration files alongside packages that provide paths allowlisted by the ephemeral storage utility. This way, the allowlist can be built via package installation and is no longer gated by variant name.

For ephemeral storage logic, replace the variant-name-sniffing `allowed_bind_dirs(variant)` with a reader that sources exact-match paths from package-contributed fragments in `/usr/lib/bottlerocket/ephemeral-storage.d/`. Each package that owns a bind-target directory ships its own fragment, so the allowlist is determined by variant package composition rather than substring matching on the variant name.

Prefix and disallowed rules remain hardcoded apiserver policy.

The drop-ins are in /usr/lib, which is part of the readonly fs, so they can't be modified, nor can a new file be written to `/usr/lib/bottlerocket/ephemeral-storage.d/`

**Testing done:**

Unit tests and manual testing below:

Built custom `aws-k8s-1.33` and `aws-ecs-2` images from these changes containing the drop-ins. 

**`aws-k8s-1.33`**

```
$ ls /usr/lib/bottlerocket/ephemeral-storage.d/
containerd.conf
host-containerd.conf
kubelet.conf
soci-snapshotter.conf
```

Absent (as expected): `docker.conf`, `ecs-agent.conf`

Drop-in contents:

| File | Contents |
|------|----------|
| `kubelet.conf` | `/var/lib/kubelet`, `/var/log/pods` |
| `soci-snapshotter.conf` | `/var/lib/soci-snapshotter` |
| `containerd.conf` | `/var/lib/containerd` |
| `host-containerd.conf` | `/var/lib/host-containerd` |

```
$ apiclient ephemeral-storage list-dirs
/var/lib/kubelet
/var/log/pods
/var/lib/soci-snapshotter
/var/lib/containerd
/var/lib/host-containerd
```

**Correctly absent:** `/var/lib/docker`, `/var/log/ecs`

**`aws-ecs-2`**

```
$ ls /usr/lib/bottlerocket/ephemeral-storage.d/
containerd.conf
docker.conf
ecs-agent.conf
host-containerd.conf
```

Absent (as expected): `kubelet.conf`, `soci-snapshotter.conf`

Drop-in file contents

| File | Contents |
|------|----------|
| `ecs-agent.conf` | `/var/log/ecs` |
| `docker.conf` | `/var/lib/docker` |
| `containerd.conf` | `/var/lib/containerd` |
| `host-containerd.conf` | `/var/lib/host-containerd` |

```
$ apiclient ephemeral-storage list-dirs output
/var/lib/docker
/var/log/ecs
/var/lib/containerd
/var/lib/host-containerd
```

**Correctly absent:** `/var/lib/kubelet`, `/var/log/pods`, `/var/lib/soci-snapshotter`

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/ephemeral_storage.rs
+++ b/sources/api/apiserver/src/server/ephemeral_storage.rs
@@ -38,9 +38,10 @@ static EPHEMERAL_PATH: &str = "/dev/disk/ephemeral";
 static EPHEMERAL_EBS_PATH: &str = "/dev/disk/ephemeral-ebs";
 /// Key ID to use for ephemeral storage encryption
 static EPHEMERAL_STORAGE_KEY_ID: &str = "ephemeral-storage";
+static BIND_DIRS_DROPIN_DIR: &str = "/usr/lib/bottlerocket/ephemeral-storage.d";
 
 pub struct BindDirs {
-    pub allowed_exact: HashSet<&'static str>,
+    pub allowed_exact: HashSet<String>,
     pub allowed_prefixes: HashSet<&'static str>,
     pub disallowed_contains: HashSet<&'static str>,
 }
@@ -174,7 +175,7 @@ pub fn initialize(
 
 /// Binds the specified directories to the pre-configured array, creating those directories if
 /// they do not exist.
-pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
+pub fn bind(dirs: Vec<String>) -> Result<()> {
     let device_name = EPHEMERAL_STORAGE_LINK;
     if !std::fs::exists(device_name).is_ok_and(|x| x) {
         info!("ephemeral storage not initialized, skipping binding");
@@ -182,12 +183,8 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
     }
 
     let dirs = if dirs.is_empty() {
-        let allowed_dirs = allowed_bind_dirs(variant);
-        allowed_dirs
-            .allowed_exact
-            .into_iter()
-            .map(|s| s.to_string())
-            .collect()
+        let allowed_dirs = allowed_bind_dirs()?;
+        allowed_dirs.allowed_exact.into_iter().collect()
     } else {
         dirs
     };
@@ -198,9 +195,9 @@ pub fn bind(variant: &str, dirs: Vec<String>) -> Result<()> {
         .map(|dir| dir.trim_end_matches("/").to_string())
         .collect();
 
-    let allowed_dirs = allowed_bind_dirs(variant);
+    let allowed_dirs = allowed_bind_dirs()?;
     for dir in &dirs {
-        let exact_match = allowed_dirs.allowed_exact.contains(dir.as_str());
+        let exact_match = allowed_dirs.allowed_exact.contains(dir);
         let prefix_match = allowed_dirs
             .allowed_prefixes
             .iter()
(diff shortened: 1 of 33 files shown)


ginglis13:
force push force some fmt issues ^ but seems actions didn't triggered because of ongoing GH incident: https://www.githubstatus.com/incidents/qcvjkzcs7j74

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -38,9 +38,10 @@ static EPHEMERAL_PATH: &str = "/dev/disk/ephemeral";
 static EPHEMERAL_EBS_PATH: &str = "/dev/disk/ephemeral-ebs";
 /// Key ID to use for ephemeral storage encryption
 static EPHEMERAL_STORAGE_KEY_ID: &str = "ephemeral-storage";
+static BIND_DIRS_DROPIN_DIR: &str = "/usr/share/bottlerocket/ephemeral-storage.d";
```
arnaldo2792: If we follow other `.d` drop-in directories used by other system services, then `%{_cross_libdir}/bottlerocket/ephemeral-storage.d` seems like a better fit.

e.g. `%{_cross_tmpfilesdir}` maps to `/usr/lib/tmpfiles.d`

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -448,20 +445,74 @@ fn discover_disks(path: &str) -> Result<Vec<String>> {
...
+    };
+
+    for entry in entries {
+        let entry = match entry {
+            Ok(e) => e,
+            Err(e) => {
+                warn!("Failed to read entry in {}: {}", dir, e);
+                continue;
+            }
+        };
```
arnaldo2792: I would prefer if we hard-fail. Any misconfiguration will result in warnings, but if we (or downstreams) don't look at the journal warnings, this will be a silent failure that will be easy to miss.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -448,20 +445,74 @@ fn discover_disks(path: &str) -> Result<Vec<String>> {
...
+
+        let path = entry.path();
+        if !path.is_file() {
+            continue;
+        }
```
arnaldo2792: Is the contract we are setting that we don't expect directories withing the drop-ins directory? If so, I'll prefer a hard failure rather than silently ignoring the path and not traversing.

sources/api/apiserver/src/server/ephemeral_storage.rs
```diff
@@ -448,20 +445,74 @@ fn discover_disks(path: &str) -> Result<Vec<String>> {
...
+            if !line.starts_with('/') {
+                warn!("Skipping non-absolute path in {:?}: {}", path, line);
+                continue;
+            }
+
+            // Reject paths containing ".."
+            if line.contains("..") {
+                warn!("Skipping path containing '..' in {:?}: {}", path, line);
+                continue;
+            }
```
arnaldo2792: Same as elsewhere, hard failure.

ZephirFXEC:
lgtm :)

sources/api/apiserver/src/server/ephemeral_storage.rs
ginglis13: I'll move the `installs` to  `%{_cross_libdir}/bottlerocket/ephemeral-storage.d` for all the path configuration files

ginglis13: A counter might be https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/ba49cc484a4894ec98b5abf894d084f278f20d1b/packages/early-boot-config/early-boot-config.spec#L81 where we configure under `/usr/libexec/early-boot-config/data-providers.d`, but I'm going with your suggestion.

ginglis13:
^ address feedback to fail hard instead of warning on failures. Add unreleased changelog section



retested with latest force-pushed changes, this time on ecs-4 instead of ecs-2:

<details>

### `aws-k8s-1.33` test



```bash
# /usr/lib/bottlerocket/ephemeral-storage.d/
containerd.conf         #  /var/lib/containerd
host-containerd.conf    # /var/lib/host-containerd
kubelet.conf            # /var/lib/kubelet, /var/log/pods
soci-snapshotter.conf   # /var/lib/soci-snapshotter
```

Absent as expected: `docker.conf`, `ecs-agent.conf`

```bash
$ apiclient ephemeral-storage list-dirs
/var/lib/kubelet
/var/log/pods
/var/lib/soci-snapshotter
/var/lib/containerd
/var/lib/host-containerd
```

Correctly absent: `/var/lib/docker`, `/var/log/ecs`

---

### `aws-ecs-4` test

```bash
# /usr/lib/bottlerocket/ephemeral-storage.d/
containerd.conf         # /var/lib/containerd
docker.conf             # /var/lib/docker
ecs-agent.conf          # /var/log/ecs
host-containerd.conf    # /var/lib/host-containerd
```

Absent as expected: `kubelet.conf`, `soci-snapshotter.conf`



```
$ apiclient ephemeral-storage list-dirs
/var/lib/docker
/var/log/ecs
/var/lib/containerd
/var/lib/host-containerd
```

Correctly absent: `/var/lib/kubelet`, `/var/log/pods`, `/var/lib/soci-snapshotter`

</details>

vigh-m (approved):

maherthomsi (approved):

ginglis13:
^force push rebase to resolve changelog conflict https://github.com/bottlerocket-os/bottlerocket-core-kit/compare/22be4981250da92692fcc66f7fb9bbd6bf0427b4..6945f580e52287af33125d2fc29e9d2864a551cf

maherthomsi (approved):

PR #1007, opened 2026-08-11
Prepare to support UKIs

arnaldo2792:
**Description of changes:**

This series prepares the core kit to support Unified Kernel Images.

Build `systemd-boot` and `systemd-stub` for systemd-257. A patch for systemd was needed to prevent it from building these binaries for the wrong CPU architectures (32 bits).

The release package was updated to conditionally include `grub`, since now there is another package that provides a bootloader.

Given that in-place upgrades aren't supported for the time being, `signpost` is skipped entirely as a binary. However, the library component was updated to find the correct FAT partition that includes the UKI to mount it at `/boot`.  This is needed for other parts of the system, like the FIPS check.

User-space PCR9 measurements are skipped for UKI-based images, since PCR9 is measured at the kernel level when systemd-boot is used. 

**Testing done:**

As part of: https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/523

- Booted UKI
- Confirmed cmdline measurements are skipped in UKI images and ran in GRUB based images
- /boot partition mounted for both formats
- Build fails when `in-place-upgrades` + `uki` are used

**Terms of contribution:**


arnaldo2792:
(Force push to fix commit messages)

arnaldo2792:
(Forced push to fix commit messages, for real)

arnaldo2792:
(Forced push includes missing systemd service)

packages/systemd-257/loader.conf
```diff
@@ -0,0 +1,31 @@
+# There is no interactive user at the the console, so the defaults below
+# aim for "boot straight through, expose nothing extra, keep exactly
+# one local recovery path."
+#
+# Note that this file is measured into TPM PCR 5 before parsing: editing
+# it changes PCR 5 and invalidates anything sealed against it.
```
ginglis13, line 6: I'd suggest this being the first line of the file, seems like a significant callout. Maybe a "DO NOT EDIT" line, or "WARNING" line too?

packages/systemd-257/systemd-257.spec
```diff
@@ -3,6 +3,14 @@
 
 %global package_priority_epoch 0
 
+# Install the systemd-boot bootloader at the same location as the GRUB and
+# shim EFI binaries, so image builds find all ESP loaders in one place.
+%global efidir /boot/efi/EFI/BOOT
+
+# systemd-boot reads its own configuration from \loader\loader.conf on the
```
ginglis13: did you mean /loader/loader.conf?

https://www.freedesktop.org/software/systemd/man/latest/loader.conf.html
```suggestion
# systemd-boot reads its own configuration from /loader/loader.conf on the
```

packages/systemd-257/loader.conf
```diff
@@ -0,0 +1,31 @@
...
+auto-firmware no
+
+# No "Power Off"/"Reboot The System" menu entries.
+auto-poweroff no
+auto-reboot no
+
+# No PC-speaker beeping during the countdown.
+beep no
+
+# Never scan \loader\keys or enroll Secure Boot keys from the ESP.
```
ginglis13:
```suggestion
# Never scan /loader/keys or enroll Secure Boot keys from the ESP.
```
https://www.freedesktop.org/software/systemd/man/latest/loader.conf.html

packages/systemd-257/systemd-257.spec
```diff
@@ -372,6 +401,20 @@ install -d %{buildroot}%{_cross_bootconfigdir}
...
+# Relocate the systemd-boot EFI binary from the meson install location to
+# %%{efidir}, matching the GRUB and shim convention. The UKI stub
+# (linux%%{_cross_efi_arch}.efi.stub) stays in %%{_cross_libdir}/systemd/boot/efi
+# since it is a build-time input for ukify, not an ESP loader.
+install -d %{buildroot}%{efidir}
+mv %{buildroot}%{_cross_libdir}/systemd/boot/efi/systemd-boot%{_cross_efi_arch}.efi \
+  %{buildroot}%{efidir}/systemd-boot%{_cross_efi_arch}.efi
+
+# Ship sensible systemd-boot defaults. This has to land on the ESP rather than
+# the boot partition, since systemd-boot only reads \loader\loader.conf from
```
ginglis13:
```suggestion
# the boot partition, since systemd-boot only reads /loader/loader.conf from
```

sources/api/prairiedog/src/main.rs
```diff
@@ -186,29 +207,114 @@ fn capture_dump() -> Result<()> {
...
+///
+/// The partition table is the primary signal, so `uki_image` is only consulted when the disk has no
+/// XBOOTLDR partition - either because it has none or because the scan for one failed.
+fn boot_action(xbootldr: Option<PathBuf>, uki_image: impl FnOnce() -> bool) -> BootAction {
+    match xbootldr {
+        Some(boot_partition_path) => BootAction::MountXbootldr(boot_partition_path),
+        // A UKI image boots from the XBOOTLDR partition and has no A/B partition sets to fall back
+        // on, so a missing partition means /boot can never be populated. Fail loudly instead of
+        // leaving it unmounted, which would silently break everything that reads /boot.
+        None if uki_image() => BootAction::MissingXbootldr,
```
ginglis13: can't we just


```suggestion
        None if uki_image_from_env() => BootAction::MissingXbootldr,
```

and bring this function's signature to just

```
fn boot_action(xbootldr: Option<PathBuf>) -> BootAction
```

? The callsite in `prepare_boot` is the only instance of using `uki_image_from_env()`, which is just wired through to this method.

sources/api/prairiedog/src/main.rs
```diff
@@ -186,29 +207,114 @@ fn capture_dump() -> Result<()> {
...
+            // rather than claiming the partition is absent.
+            warn!("Unable to scan the OS disk for an XBOOTLDR partition: {e}");
+            (None, Some(e))
+        }
+    };
+
+    match boot_action(xbootldr, uki_image_from_env) {
+        BootAction::MountXbootldr(boot_partition_path) => {
+            info!(
+                "Found XBOOTLDR partition {} on the OS disk, so this is a UKI image",
```
ginglis13:
```suggestion
                "UKI image detected (XBOOTLDR partition '{}')",
```

sources/updater/signpost/src/state.rs
```diff
@@ -400,6 +407,98 @@ impl State {
...
+                    g[14],
+                    g[15],
+                )
+            },
+            partitions,
+        }
+    );
+
+    Ok(partitions.pop())
+}
```
jmt-lab: **[P2] Deduplicate disk/GPT lookup shared by `xbootldr_partition` and `State::load`**

`xbootldr_partition()` (lines ~418–461) is a near line-for-line copy of the first ~60 lines of `State::load()` (lines 71–131): the `BlockDevice::from_device_path("/")` walk, the `lower_devices()` selection, the `disk()` unwrap, the `GPT::find_from` open, and the `device_from_part_num` closure are all repeated verbatim.

**Why it matters.** `prepare_boot` in prairiedog calls both `signpost::xbootldr_partition()` and `signpost::State::load()` back-to-back in the GRUB path (`main.rs:243` and `main.rs:274`), so the same disk is stat-walked and its GPT parsed twice on every boot. More importantly, any future change to how the OS disk is discovered has to be applied in two places, and drift here is silent.

**Failure scenario.** A follow-up change that teaches `State::load` to unwrap an extra dm layer (say, "boot from LUKS") applied only to `State::load` leaves `xbootldr_partition` looking at the wrong disk. It returns `None`, prairiedog silently takes the `MountActiveSet` branch on what is really a UKI image, and boot tries to mount an ext4 boot partition on a disk that no longer has one.

**Fix.** Extract a private helper like `fn parse_os_disk_table() -> Result<(Disk, GPT), Error>` and have both entry points call it. Alternatively, expose the XBOOTLDR lookup as a method on `State` so a single scan feeds both.

sources/updater/signpost/src/state.rs
```diff
@@ -400,6 +407,98 @@ impl State {
...
+            .context(error::PartitionNotFoundOnDeviceSnafu {
+                num,
+                device: os_disk.path(),
+            })?
+            .path())
+    };
+
+    let mut partitions = table
+        .iter()
+        .filter(|(_, p)| p.is_used() && p.partition_type_guid == XBOOTLDR)
```
jmt-lab: **[P2] `xbootldr_partition` filters on `is_used()` but `State::load` does not**

`xbootldr_partition` filters entries with `p.is_used() && p.partition_type_guid == XBOOTLDR` (this line). `State::load`'s `nth_guid` closure (lines 104–110) filters only on `p.partition_type_guid == guid` — no `is_used()` check.

In the current schema this cannot matter, because unused entries have an all-zero type GUID. But the asymmetry is exactly the kind of subtle drift the "read the GPT once" comment in `main.rs:239–242` is trying to avoid: if a future change ever ties partition identity to something other than the type GUID (a partition-name convention, an "in transition" tombstone), the two functions will disagree.

Also: the test `ignores_unused_and_other_types` (line 583) exercises `partition_nums_with_type` — a test helper — not `State::load`'s actual `nth_guid` filter.

**Fix.** Add `p.is_used()` to `State::load`'s `nth_guid` filter for consistency, or extract a shared helper.

sources/updater/signpost/src/state.rs
```diff
@@ -400,6 +407,98 @@ impl State {
...
+                    g[11],
+                    g[12],
+                    g[13],
+                    g[14],
+                    g[15],
+                )
+            },
+            partitions,
+        }
+    );
```
jmt-lab: **[P2] `MultiplePartitionsOfType` builds the GUID string at runtime with a 16-arg `format!`**

The XBOOTLDR type GUID is a compile-time constant — its canonical form is already in a comment three lines above the array literal (line 23) and in the `MissingXbootldrPartition` / `ScanForXbootldrPartition` display messages in prairiedog (`error.rs:29`, `error.rs:47`).

Three separate places now hand-serialize the same 128-bit constant, and one of them does it with a 16-argument positional `format!` that will silently produce the wrong string if the byte-swap indices are ever transcribed incorrectly. There is a byte-order test (`xbootldr_guid_byte_order`, line 597), but no test for the *formatting*.

**Fix.** Since `XBOOTLDR` is the only type GUID this call currently reports, just pass the canonical string literal: `partition_type: "bc13c2ff-59e6-4262-a352-b275fd6f7172".to_string()`. If more type GUIDs need reporting later, introduce a `fn guid_to_string(g: [u8; 16]) -> String` helper with its own test.

sources/api/prairiedog/src/main.rs
```diff
@@ -186,29 +207,114 @@ fn capture_dump() -> Result<()> {
...
+        },
+
+        BootAction::MountActiveSet => {
+            info!("No XBOOTLDR partition found on the OS disk; using the boot partition of the active partition set");
+            // Get the current partitions state
+            let state = signpost::State::load().context(error::LoadStateSnafu)?;
+
+            mount_boot_partition(&state.active_set().boot, BOOT_FS_EXT4, None)
+        }
+    }
```
jmt-lab: **[P2] Multiple XBOOTLDR partitions on a "GRUB" image silently fall back to mounting the A-set ext4 boot partition**

`xbootldr_partition()` returns `Err(MultiplePartitionsOfType)` when more than one XBOOTLDR partition is on the OS disk (`state.rs:469–497`). Here in `prepare_boot`, that error is stashed as `scan_error` and downgraded to a `warn!` log (lines 250–252). Then `boot_action(None, uki_image_from_env)` is called: if `UKI_IMAGE` is unset or `"false"`, control flows to `BootAction::MountActiveSet` and prairiedog mounts the A-set ext4 boot partition (lines 271–277).

**Why it matters.** "Two XBOOTLDR partitions on disk" is a structural corruption of the partition table, not a legitimate GRUB layout — a plain GRUB image has zero of them. Suppressing this to a warning on the GRUB path lets a broken disk boot as if nothing is wrong; if that disk was meant to be a UKI image but `image-format.env` got lost, the operator sees a successful `/boot` mount that contains none of the UKI's kernels. The failure only surfaces later, when systemd-boot or a kernel update tool trips over it.

**Fix.** In the `Err` arm at line 245, distinguish "no XBOOTLDR to report" (log and continue) from "the scan found a real problem" (return the error unconditionally). Only demote truly ambiguous scan errors (I/O failures on the raw disk) and propagate `MultiplePartitionsOfType` as a hard failure. Alternatively, keep the scan-error stash but check its variant when choosing the fallback.

sources/api/prairiedog/src/main.rs
```diff
@@ -186,29 +207,114 @@ fn capture_dump() -> Result<()> {
...
+            value.trim() == "true"
+        }
+        Err(_) => {
+            info!(
+                "{UKI_IMAGE_ENV} is not set in the environment; assuming this is not a UKI image"
+            );
+            false
+        }
+    }
+}
```
jmt-lab, line 259: **[P2] `uki_image_from_env` treats non-UTF-8 and misspelled values as "not a UKI image" without warning**

`uki_image_from_env` uses `env::var(UKI_IMAGE_ENV)`:
- `Ok(value)` where `value.trim() != "true"` (including `"TRUE"`, `"1"`, `"yes"`, or an accidental leading `#`) silently returns `false` after logging the value at `info`.
- `Err(NotUnicode(_))` (garbage bytes in the env var) is collapsed into the same "not set" branch as `Err(NotPresent)`, so the operator sees `"UKI_IMAGE is not set in the environment"` even though it *was* set — just unreadably.

**Why it matters.** This env variable is the sole fallback used to decide whether missing partitions are fatal (line 232). A UKI image whose `image-format.env` was corrupted or transcoded (e.g. an editor added a BOM, or a build emitted `UKI_IMAGE=True`) will be misclassified as a GRUB image. On a UKI image with a transient GPT read error at boot, the operator loses the useful `ScanForXbootldrPartition` error message and gets a misleading `State::load` failure instead.

**Fix.** Either
1. Case-insensitive parse (`value.trim().eq_ignore_ascii_case("true")`) and explicitly reject unrecognized values with a warning, or
2. Return `Result<bool, Error>` and propagate parse/`NotUnicode` errors so `prepare_boot` can decide.

At minimum, log `warn!` (not `info!`) when the value is set but not exactly `"true"` — right now that case is indistinguishable from a legitimate `UKI_IMAGE=false` in the journal.

sources/api/prairiedog/src/main.rs
```diff
@@ -355,3 +461,31 @@ fn main() {
...
+    #[test]
+    fn uki_image_without_an_xbootldr_partition_fails() {
+        assert_eq!(boot_action(None, || true), BootAction::MissingXbootldr);
+    }
+
+    #[test]
+    fn grub_image_without_an_xbootldr_partition_uses_the_active_partition_set() {
+        assert_eq!(boot_action(None, || false), BootAction::MountActiveSet);
+    }
+}
```
jmt-lab: **[P2] Tests never exercise `prepare_boot`'s scan-error handoff to `boot_action`**

The new `tests` module covers three cases of `boot_action` in isolation, all of which pass `xbootldr` as either `Some(_)` or `None` — never a value that came from a failed scan. The interesting logic in `prepare_boot` is the *pairing* between `scan_error` (lines 243–253) and the `BootAction::MissingXbootldr` branch (lines 264–269): scan error is propagated only when we've concluded the image is UKI. That coupling is untested.

**Why it matters.** The whole reason `scan_error` is stashed and threaded through is so a UKI image can report the underlying cause via `ScanForXbootldrPartitionSnafu`. If someone refactors `prepare_boot` and drops the `Some(source) => Err(source).context(...)` arm, the tests still pass but UKI images with a failing GPT read now report `MissingXbootldrPartition` — an outright wrong root cause.

**Fix.** Refactor `prepare_boot` so the scan-error handoff is testable (e.g., extract `fn resolve_boot_action(scan: Result<Option<PathBuf>, Error>, uki: bool) -> Result<BootAction, Error>`). Then add tests for: (a) scan errored + uki=true → `ScanForXbootldrPartition`; (b) scan errored + uki=false → `MountActiveSet` (or intended fallback); (c) scan `Ok(None)` + uki=true → `MissingXbootldrPartition`.

sources/updater/signpost/src/state.rs
```diff
@@ -400,6 +407,98 @@ impl State {
...
+
+    // Parse the partition table on the disk.
+    let table = GPT::find_from(&mut File::open(os_disk.path()).context(error::OpenSnafu {
+        path: os_disk.path(),
+        what: "reading",
+    })?)
+    .map_err(error::GPTError)
+    .context(error::GPTFindSnafu {
+        device: os_disk.path(),
+    })?;
```
jmt-lab: **[P2] `xbootldr_partition` re-opens the disk and re-parses the GPT that `State::load` will parse again**

`xbootldr_partition()` opens the OS disk with `File::open` and parses its GPT here, and `State::load()` does the same open+parse a few lines later (lines 94–101). In the GRUB path, `prepare_boot` calls both in sequence (`main.rs:243`, `main.rs:274`), so the raw disk is opened and its GPT parsed twice per boot.

Beyond the wasted work, the two reads are not guaranteed to be consistent: if some other tool is mid-write to the partition table (signpost itself calls `write_into` in `State::write`), the two callers can observe different GPTs.

**Fix.** Fold the XBOOTLDR probe into `State` (e.g., `State::xbootldr(&self) -> Option<&Path>`) so a single GPT parse covers both. Downstream callers then just do `let state = State::load()?; match state.xbootldr() { ... }`. This also solves the code-duplication finding elsewhere in this file.

packages/release/release.spec
```diff
@@ -165,7 +165,10 @@ Requires: %{_cross_os}filesystem
 Requires: %{_cross_os}findutils
 Requires: %{_cross_os}glibc
 Requires: %{_cross_os}grep
-Requires: %{_cross_os}grub
+# For newer versions of twoliter that support UKIs, explicitly request the bootloader(efi) capability
+Requires: (%{_cross_os}bootloader(efi) if (%{_cross_os}image-feature(uki-image) or %{_cross_os}image-feature(no-uki-image)))
+# Older versions of twoliter that don't support UKIs always get GRUB
+Requires: (%{_cross_os}grub or %{_cross_os}image-feature(uki-image) or %{_cross_os}image-feature(no-uki-image))
```
jmt-lab, line 173: **[P1] `no-uki-image` variants will fail dependency resolution — no package provides `bootloader(efi)`**

The new dependency logic here requires `bootloader(efi)` whenever *either* `image-feature(uki-image)` *or* `image-feature(no-uki-image)` is set:

```
Requires: (%{_cross_os}bootloader(efi) if (%{_cross_os}image-feature(uki-image) or %{_cross_os}image-feature(no-uki-image)))
Requires: (%{_cross_os}grub or %{_cross_os}image-feature(uki-image) or %{_cross_os}image-feature(no-uki-image))
```

But `bootloader(efi)` is provided **only** by `%{_cross_os}systemd-257-bootloader` (`systemd-257.spec:193`), which itself has:
```
Requires:  %{_cross_os}image-feature(uki-image)          (line 192)
Conflicts: %{_cross_os}image-feature(no-uki-image)       (line 194)
```

**Trigger.** A variant built with newer twoliter that sets `image-feature(no-uki-image)`:
- Line 169 → demands `bootloader(efi)`.
- Line 171 → satisfied by `image-feature(no-uki-image)`, so `grub` is **not** pulled in.
- The only `bootloader(efi)` provider is forbidden by its own `Conflicts` line.

Result: unresolvable transaction, or an image with no bootloader.

**Fix options.**
- Have `grub` (or a companion subpackage) `Provides: %{_cross_os}bootloader(efi)`, or
- Change line 169 to require `bootloader(efi)` only when `image-feature(uki-image)` is set, and drop the `no-uki-image` disjunct from line 171 so it falls back to `grub`:
  ```
  Requires: (%{_cross_os}bootloader(efi) if %{_cross_os}image-feature(uki-image))
  Requires: (%{_cross_os}grub or %{_cross_os}image-feature(uki-image))
  ```

packages/systemd-257/systemd-257.spec
```diff
@@ -175,6 +186,18 @@ Conflicts: %{_cross_os}auditd
...
 %{summary}.
 
+%package bootloader
+Summary: EFI boot loader and stub binaries
+Requires: %{name}
+Requires: %{_cross_os}image-feature(uki-image)
+Provides: %{_cross_os}bootloader(efi)
+Conflicts: %{_cross_os}image-feature(no-uki-image)
+Conflicts: %{_cross_os}image-feature(standalone-image)
+Conflicts: %{_cross_os}image-feature(in-place-updates)
```
jmt-lab: **[P1] `Conflicts: image-feature(in-place-updates)` / `image-feature(standalone-image)` reference feature names that do not exist**

The image-feature naming convention throughout the repository is the *negative* form. A grep for `image-feature(` across the tree shows:
- `image-feature(no-in-place-updates)` — used across `os.spec` as the "off" flag.
- `image-feature(in-place-updates)`     — **not defined or used anywhere else**.
- `image-feature(standalone-image)`     — **not defined or used anywhere else**.

**Why it matters.** These are dead conflict clauses. The intended safety guard — "the systemd-boot bootloader package must never be installed alongside the signpost/GRUB update flow" — is silently inert. If a future variant is misconfigured to include both signpost/GRUB *and* the UKI bootloader package, rpm/dnf won't catch it.

**Fix.** Either invert the polarity to match the codebase convention, or drop these lines since `Requires: image-feature(uki-image)` on line 192 combined with `Conflicts: image-feature(no-uki-image)` on line 194 already prevents installation on any non-UKI image. If two *new* positive feature names are being introduced by this PR series, they need to actually be defined on the image side and referenced consistently.

packages/release/measure-cmdline.service
```diff
@@ -10,6 +10,12 @@ RefuseManualStop=true
 
 [Service]
 Type=oneshot
+EnvironmentFile=-/usr/share/bottlerocket/image-format.env
+
+# Skip on UKI images; the kernel command line is measured as part of the
```
jmt-lab, line 15: **[P2] `ExecCondition=[ "${UKI_IMAGE}" != "true" ]` relies on the `[` binary + systemd `$PATH` search, not a shell**

systemd does *not* wrap `ExecCondition=` in a shell. systemd tokenizes the line and execs the first token — which here is the literal string `[`. This works only because:
1. Since ~systemd v247, `ExecStart=`/`ExecCondition=` will do a `$PATH` lookup on non-absolute command names.
2. `coreutils` installs `%{_cross_bindir}/[`.
3. `[ "" != "true" ]` (when `UKI_IMAGE` is unset) evaluates true, so it's the safe default.

**Why it's still worth flagging.**
- systemd docs recommend absolute paths in unit files. On other minimal distributions this would silently break.
- If someone ever slims down the base image by dropping the `[` binary from coreutils (a rarely-needed standalone), every `ExecCondition=[ … ]` in the repo becomes `status=203/EXEC`, and this service would then *always* fail its condition and skip PCR9 measurement.
- The comment above ("Skip on UKI images; ... measured as part of the UKI PE binary itself") makes the correctness of this condition security-relevant: if the `ExecCondition` ever fails-open in the *non-UKI* direction, kernel command line measurement is silently skipped and TPM-based attestation loses coverage of PCR 9.

**Fix (recommended).** Use an absolute path:
```
ExecCondition=/usr/bin/test "${UKI_IMAGE}" != "true"
```
or invoke a small helper (e.g. `prairiedog is-not-uki-image`) that mirrors the existing `uki_image_from_env()` in `sources/api/prairiedog/src/main.rs:282`, so the guardrail is independent of coreutils' `[`/`test`.

packages/os/os.spec
```diff
@@ -239,6 +240,7 @@ Requires: %{_cross_os}nvme-cli
 
 %package -n %{_cross_os}signpost
 Summary: Bottlerocket GPT priority querier/switcher
+Conflicts: %{_cross_os}image-feature(uki-image)
 %description -n %{_cross_os}signpost
 %{summary}.
```
jmt-lab, line 245: **[P2] `signpost` `Conflicts: image-feature(uki-image)` may block installs, and UKI images seem to lose boot-success signalling**

The pairing here is internally consistent — on UKI images, line 122's `Requires: (signpost or image-feature(uki-image))` is satisfied by the image feature and `signpost` isn't pulled in.

Two things worth verifying before merging:

1. **Transitive `Requires: signpost`.** If any other package (in this repo, or in downstream code kits) has `Requires: %{_cross_os}signpost` without an `or image-feature(uki-image)` escape hatch, the transaction becomes unresolvable on `uki-image` variants because signpost is Conflicts-blocked but still required. This deserves an audit across the org's kits.

2. **Boot-success signalling on UKI.** `%files -n %{_cross_os}signpost` still packages `mark-successful-boot.service` and `has-boot-ever-succeeded.service`. On UKI variants these disappear. systemd-boot has `systemd-bless-boot`, but the new `systemd-257-bootloader` subpackage explicitly excludes `systemd-bless-boot*` at `systemd-257.spec:501, 511-512`. So **UKI images built from this PR appear to have no boot-success signalling at all**. Please confirm this is handled elsewhere.

packages/systemd-257/9018-build-correct-sd-boot-EFI-arch-on-the-Bottlerocket-S.patch
```diff
@@ -0,0 +1,94 @@
...
++
++# Single corrected arch key for all EFI flag lookups, so no flag site can
++# silently regress to the -m32 (ia32) path on this SDK.
++if efi_arch == 'x64'
++    efi_cpu_family = 'x86_64'
++elif efi_arch == 'ia32'
++    efi_cpu_family = 'x86'
++else
++    efi_cpu_family = host_machine.cpu_family()
++endif
```
jmt-lab, line 53: **[P2] Overriding `efi_arch` in systemd's `meson.build` masks a real SDK bug and is fragile against upstream refactors**

The patch works around a Bottlerocket-SDK quirk where `host_machine.cpu_family()` returns `'x86'` on x86_64 and `'arm'` on aarch64, and post-processes the value here.

1. **Root cause is elsewhere.** `host_machine.cpu_family()` being wrong is a meson cross-file / SDK configuration bug (`[host_machine] cpu_family = 'x86_64'` / `'aarch64'` are the correct settings). Patching every consumer of `cpu_family` in systemd is whack-a-mole — the SDK cross-file should be fixed so *every* build sees the right value. If the SDK fix ever lands, this patch becomes actively misleading.

2. **Incomplete coverage.** The patch only replaces `host_machine.cpu_family()` at two call sites in `src/boot/meson.build`. Other consumers inside systemd 257 (e.g., arch-specific seccomp lists, syscall filter tables, ukify machinery) will still see the wrong value. The commit message says "no flag site can silently regress" but only sd-boot's `efi_archspecs` was fixed.

3. **`efi_arch_alt` sentinel semantics.** The patch removes the `if have and efi_arch == 'x64' and cc.links(…)` block but leaves the `efi_arch_alt = ''` initializer. Please confirm that `efi_archspecs` guards the `_alt` entries with `if efi_arch_alt != ''` — otherwise meson may still try to build a nameless alt binary.

4. **Aarch64 asserts an unenforced invariant.** The comment says "efi_cpu_family is deliberately left resolving to 'arm'" for aarch64, relying on `efi_arch_c_args.get('arm', []) == efi_arch_c_args.get('aarch64', [])`. Any future systemd change that diverges the two entries will silently be lost.

**Fix.**
- Preferred: fix the SDK meson cross-file and drop this patch.
- If the patch must stay: set `efi_cpu_family = 'aarch64'` explicitly when `efi_arch == 'aa64'`, and add a `message()`/`error()` guard that asserts `efi_arch_c_args.get('arm') == efi_arch_c_args.get('aarch64')` at configure time.

packages/release/measure-cmdline.service
```diff
@@ -10,6 +10,12 @@ RefuseManualStop=true
 
 [Service]
 Type=oneshot
+EnvironmentFile=-/usr/share/bottlerocket/image-format.env
```
jmt-lab, line 13: **[P2] `measure-cmdline.service` is only shipped in the `crypt` subpackage — verify UKI behavior for its siblings**

This service is placed under `%files crypt` in `release.spec` (around lines 506–514), which is gated behind `image-feature(encrypted-storage)`. So the `ExecCondition` guard here only runs on encrypted-storage images.

The service itself is used with the encrypted-storage TPM sealing path, so its UKI-skip scope is consistent. But two things worth verifying explicitly:

1. **Siblings without the same guard.** `measure-settings.service` and the four `systemd-pcrphase-*.service` units in the same `%files crypt` block do **not** get an equivalent `ExecCondition`. Are those safe to run on UKI (e.g., because they measure different PCRs that sd-stub doesn't already cover)? If not, they'll re-extend PCRs that sd-stub has already measured, corrupting attestation.

2. **`image-format.env` provisioning ordering.** The `EnvironmentFile=-/usr/share/bottlerocket/image-format.env` (also used by `prepare-boot.service:9`) uses `-`, so a missing file won't fail the unit — but the file's *contents* determine behavior silently. Please confirm that whichever kit/image-build step writes `/usr/share/bottlerocket/image-format.env` sets `UKI_IMAGE=true` **before** the initrd/UKI image is composed (in the sysroot image), not at first boot — otherwise `measure-cmdline.service` could race with it on first boot after upgrade.

**Fix.** Add an `ExecCondition` to `measure-settings.service` and the `systemd-pcrphase-*` services if they should also be skipped on UKI, or add a comment explaining why they don't need one. And document (here or in a nearby README) where `image-format.env` is produced.

jmt-lab (changes requested):

packages/release/release.spec
arnaldo2792: This isn't accurate, the kernel kit will provide `bootloader(efi)`

packages/systemd-257/systemd-257.spec
arnaldo2792: Not accurate again, the image features exist in Twoliter.

packages/release/measure-cmdline.service
arnaldo2792: there are other places in bottlerocket where we do this. If this is a concern, I would rather fix them all together in one PR as a cleanup.

packages/os/os.spec
arnaldo2792: This is exactly what we want since we don't support in-place upgrades yet for UKIs

packages/systemd-257/9018-build-correct-sd-boot-EFI-arch-on-the-Bottlerocket-S.patch
arnaldo2792: The SDK fix, we can do in a subsequent release. The patch must stay so I'll fix the suggestion.

packages/release/measure-cmdline.service
arnaldo2792: 1. siblings don't need the same guard, we don't want this service running in UKIs because the PCR9 measurements will occur in the kernel
2. This is to prevent the service failing in variants built with an older twoliter that doesn't provide the file.

packages/systemd-257/systemd-257.spec
```diff
@@ -175,6 +193,18 @@ Conflicts: %{_cross_os}auditd
 %description journald-audit
 %{summary}.
 
+%package bootloader
+Summary: EFI boot loader and stub binaries
+Requires: %{name}
+Requires: %{_cross_os}image-feature(uki-image)
+Provides: %{_cross_os}bootloader(efi)
+Conflicts: %{_cross_os}image-feature(no-uki-image)
+Conflicts: %{_cross_os}image-feature(standalone-image)
```
ginglis13: standalone-image feature should be compatible with systemd-bootloader

```suggestion
```

packages/systemd-257/systemd-257.spec
```diff
@@ -81,6 +90,11 @@ Patch9016: 9016-bootctl-disable-secure-boot-autoenroll.patch
 
 # Patch meson to set OPENSSL_NO_UI_CONSOLE CFLAGS for the build
 Patch9017: 9017-meson-set-DOPENSSL_NO_UI_CONSOLE-when-using-openssl.patch
+# Patch meson to skip unsupported architecture builds 
+Patch9018: 9018-do-not-build-32-bit-ia32-sd-boot.patch
+# Remove undesired SMBIOS functionality
+Patch9019: 9019-boot-remove-SMBIOS-Type-11-kernel-cmdline-extra-mech.patch
+Patch9020: 9020-boot-vmspawn-finish-removing-SMBIOS-cmdline-extra-bi.patch
```
vigh-m, line 97: Do these need to be 2 separate patches? Looks like they're modifying the same file.

packages/systemd-257/9018-build-correct-sd-boot-EFI-arch-on-the-Bottlerocket-S.patch
```diff
@@ -0,0 +1,94 @@
+From 7a35885f3294c0cde362edcb5021556b8906be80 Mon Sep 17 00:00:00 2001
+From: Arnaldo Garcia Rincon <agarrcia@amazon.com>
+Date: Tue, 4 Aug 2026 03:50:03 +0000
+Subject: [PATCH] build: correct sd-boot EFI arch on the Bottlerocket SDK
+
+The SDK cross sysroot mis-reports host_machine.cpu_family(), so systemd's
+efi_arch lookup resolves to the wrong EFI target: 'ia32' on x86_64 and 'arm' on
+aarch64. Correct both from host_machine.cpu(), and drop the IA-32 mixed-mode
+alternate build, which cannot link in this sysroot.
```
sam-berning, line 9: This patch (as well as 9019 and 9020) is missing a `Signed-off-by:` line

piyush-jena (approved):

piyush-jena:
Tested jointly with #999 . Stress tested and no issues found. Assuming that the sign offs in the patches will be added.

packages/systemd-257/9018-build-correct-sd-boot-EFI-arch-on-the-Bottlerocket-S.patch
arnaldo2792: Actually, changing the meson's SDK to `x86_64` will be a large blast radius change. The "whack-a-mole" approach here should be fine as this is a problem with how systemd's build machinery chooses the "wrong" default. Yes this is self inflicting pain (the x86 hardcoded default in the SDK predates my time), but we have to evaluate if changing that default doesn't break other parts of the system.

ginglis13 (approved):

jmt-lab (approved):

packages/systemd-257/systemd-257.spec
```diff
@@ -887,3 +951,11 @@ install -p -m 0644 %{S:2} %{buildroot}%{_cross_bootconfigdir}/21-cgroup-enable-l
 
 %files journald-audit
 %{_cross_unitdir}/systemd-journald-audit.socket
+
+%files bootloader
+%dir %{efidir}
+%{efidir}/systemd-boot%{_cross_efi_arch}.efi
+%dir %{loaderdir}
+%{loaderdir}/loader.conf
```
vigh-m, line 959: Let's drop this and just keep systemd-stub

PR #1024, opened 2026-08-27
updates to ephemeral encryption keys

piyush-jena:
**Description of changes:**
- Adds documentation for this preview feature
- use plain mode device encryption (which is headerless) when ephemeral encryption keys are used
- encrypt and unlock in the same service to avoid storing keys in any storage (except datastore encryption)

**Testing done:**
Containers run with containerd:
1. Tested `nvidia-smoke-test` - both `nvidia-smi` and CUDA Workload
2. Tested `dotnet-test` workload.

FIPS test:
```
[ssm-user@control]$ apiclient report fips
Benchmark name:  FIPS Security Policy
Version:         v1.0.0
Reference:       https://csrc.nist.gov/
Benchmark level: 1
Start time:      2026-08-27T06:49:34.743345588Z

[PASS] 1.0       FIPS mode is enabled. (Automatic)
[PASS] 1.1       FIPS module is Amazon Linux 2023 Kernel Cryptographic API. (Automatic)
[PASS] 1.2       FIPS self-tests passed. (Automatic)

Passed:          3
Failed:          0
Skipped:         0
Total checks:    3

Compliance check result: PASS
```

lsblk output
```
bash-5.2# lsblk
NAME                     MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS
nvme1n1                  259:0    0    20G  0 disk
`-nvme1n1p1              259:12   0    20G  0 part
  `-BOTTLEROCKET-DATA    252:2    0    20G  0 crypt /var
                                                    /opt
                                                    /mnt
                                                    /local
nvme0n1                  259:1    0     2G  0 disk
|-nvme0n1p1              259:4    0     4M  0 part
|-nvme0n1p2              259:5    0    10M  0 part
|-nvme0n1p3              259:6    0    80M  0 part  /boot
|-nvme0n1p4              259:7    0   1.8G  0 part
|-nvme0n1p5              259:8    0    20M  0 part
|-nvme0n1p6              259:9    0    50M  0 part
|-nvme0n1p7              259:10   0    41M  0 part
| `-BOTTLEROCKET-PRIVATE 252:1    0    41M  0 crypt /var/lib/bottlerocket
|                                                   /.bottlerocket
`-nvme0n1p8              259:11   0     1M  0 part
nvme2n1                  259:2    0 109.9G  0 disk
```

keystore for datastore encryption is correctly labeled and the key is ephemeral
```
bash-5.2# ls -lZ /run/ | grep rottweiler
drwx------.  2 root   root   system_u:object_r:private_t:s0     40 Aug 27 06:31 rottweiler
bash-5.2# ls -lZ /run/rottweiler
total 0
```

systemctl status output
```
bash-5.2# systemctl status
● ip-172-31-16-76.us-west-2.compute.internal
    State: running
    Units: 417 loaded (incl. loaded aliases)
     Jobs: 0 queued
   Failed: 0 units
    Since: Thu 2026-08-27 06:29:39 UTC; 22min ago
  systemd: 257.13
```

older services were removed
```
bash-5.2# find . -name unlock-datastore.service
bash-5.2# find . -name encrypt-unlock-datastore.service
./etc/systemd/system/storewolf.service.requires/encrypt-unlock-datastore.service
./etc/systemd/system/migrator.service.requires/encrypt-unlock-datastore.service
./x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/encrypt-unlock-datastore.service
bash-5.2# find . -name unlock-datastore.service
bash-5.2# find . -name encrypt-datastore.service
bash-5.2# find . -name encrypt-local-fs.service
bash-5.2# find . -name unlock-local-fs.service
bash-5.2# find . -name encrypt-unlock-local-fs.service
./etc/systemd/system/local-fs.target.requires/encrypt-unlock-local-fs.service
./x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/encrypt-unlock-local-fs.service
bash-5.2# find . -name encrypt-unlock-private-fs.service
./etc/systemd/system/prepare-private-fs.service.requires/encrypt-unlock-private-fs.service
./x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/encrypt-unlock-private-fs.service
bash-5.2# find . -name encrypt-private-fs.service
bash-5.2# find . -name unlock-private-fs.service
```
Stress test:
Ran reboot loop on 3 nvidia instances with bootstrap command setting up ephemeral storage with a large data partition. Didn't get any bricked device across 300 boots.

**Terms of contribution:**


packages/release/encrypt-unlock-datastore.service
```diff
@@ -0,0 +1,26 @@
+[Unit]
+Description=Encrypt and Unlock Datastore (/.bottlerocket/datastore)
+DefaultDependencies=no
+Before=migrator.service storewolf.service
+RequiresMountsFor=/.bottlerocket
+RefuseManualStart=true
+RefuseManualStop=true
```
piyush-jena, line 10: `RequiresMountsFor=/run/rottweiler`

considering adding this

packages/release/release.spec
```diff
@@ -229,6 +230,15 @@ Requires: %{_cross_os}rottweiler
 %description crypt
 %{summary}.
 
+%package crypt-luks
+Summary: Bottlerocket release, with LUKS block-device encryption
+Requires: (%{_cross_os}image-feature(encrypted-storage) and %{name}-crypt)
+Conflicts: %{_cross_os}image-feature(ephemeral-encryption-keys)
```
vigh-m, line 249: Should this conflict with `ephemeral-crypt`?

piyush-jena:
Force push removes the requirement of a `udev settle` by reordering the commands in `encrypt-attach-local-fs.service`. This is achieved by doing the cheaper operations first - `encrypt`, `unlock`, `systemd-makefs` (on the smaller filesystem) followed by `systemd-repart` and `rottweiler resize`. We run `systemd-growfs` in a separate service because this takes the most time and is heavily dependent on the size of the block device and requires the data partition to be mounted as local.

docs/ENCRYPTED_STORAGE.md
```diff
@@ -208,6 +210,139 @@ This ensures encrypted data can only be accessed when booting a trusted configur
...
+
+Directory encryption for `/.bottlerocket/datastore` still uses fscrypt with a TPM2-sealed key, and it runs as a single service (`encrypt-unlock-datastore.service`) instead of the `encrypt-datastore.service` / `unlock-datastore.service` pair, with the sealed key living in the `/run/rottweiler` tmpfs keystore rather than on `/.bottlerocket`.
+
+### Keys
+
+No key survives the service that uses it. Each key is 64 bytes read from `/dev/random`, TPM2-sealed with `systemd-creds`, and written to the `/run/rottweiler` tmpfs keystore only for the duration of the service that opens the device:
+
+- The `bottlerocket-data`, `bottlerocket-private`, and ephemeral-storage keys are generated, used, and deleted within a single service (an `ExecStartPost=rottweiler delete-key` removes each one)
+- Keys live in the `/run/rottweiler` tmpfs keystore, never on `/.bottlerocket` and never on a persistent disk
+- Data does not survive a reboot: the next boot generates a new key, so prior contents are unreadable and the filesystem is recreated
```
arnaldo2792: The above is enough, you can remove this.

docs/ENCRYPTED_STORAGE.md
```diff
@@ -208,6 +210,139 @@ This ensures encrypted data can only be accessed when booting a trusted configur
...
+### Boot Flow (Every Boot)
+
+Nothing is conditional on prior state, so every boot follows the same path.
+
+```
+1. encrypt-unlock-local-fs.service
+   - Generates a per-boot key, opens BOTTLEROCKET-DATA as /dev/mapper/BOTTLEROCKET-DATA with it,
+     creates the filesystem, then grows the partition and resizes the mapper, and finally deletes
+     the key. Opening the mapper before growing the partition avoids waiting on the by-partlabel
+     device node to be recreated.
```
arnaldo2792:
```suggestion
     the key.
```

docs/ENCRYPTED_STORAGE.md
```diff
@@ -208,6 +210,139 @@ This ensures encrypted data can only be accessed when booting a trusted configur
...
+#### .bottlerocket.mount (modified)
+
+**Drop-in:** `bottlerocket-mount-ephemeral.conf`, installed as
+`.bottlerocket.mount.d/20-ephemeral.conf`
+
+**Changes:**
+- Mounts `/dev/mapper/BOTTLEROCKET-PRIVATE` instead of the raw partition
+- Requires `prepare-private-fs.service`
+
+`repart-local.service` gets a `10-ephemeral.conf` drop-in in this mode that masks the base unit's `ExecStart` (the `systemd-repart` grow), because the partition grow and mapper resize are performed inside `encrypt-unlock-local-fs.service` so the key can be generated and deleted within that single service. The base unit's `systemd-growfs /local` still runs to grow the filesystem to fill the resized mapper.
```
arnaldo2792: Why not following the convention already established for all other services?

sources/rottweiler/src/system.rs
```diff
@@ -88,6 +88,64 @@ pub fn cryptsetup_luks_format(device: &str, key_data: &[u8]) -> Result<()> {
...
+        "--key-size",
+        PLAIN_KEY_SIZE_BITS,
+        "--hash",
+        "plain",
+        // Without `--key-file=-`, cryptsetup treats stdin as an interactive *passphrase* and stops
+        // reading at the first newline; `--keyfile-size` is what makes it read exactly the key even
+        // from stdin. Together they are the only reason the mapper gets all of the key bytes: a
+        // 0x0A anywhere in a random 64-byte key would otherwise silently truncate it (and a leading
+        // 0x0A would yield an all-zero key, accepted as-is on a non-FIPS kernel and rejected by
+        // xts_verify_key() as -EINVAL on a FIPS one).
```
arnaldo2792: Document this in the function, not here. And less verbose: `prefer key-file over stdin for reasons...`

sources/rottweiler/src/system.rs
```diff
@@ -250,4 +308,52 @@ mod tests {
...
+
+    /// Regression test for a truncated plain-mode key.
+    ///
+    /// The key is 64 raw bytes from /dev/random handed to cryptsetup on stdin. Without
+    /// `--key-file=-` cryptsetup reads stdin as an interactive passphrase and stops at the first
+    /// newline, and `--hash plain` then zero-pads whatever it got to `--key-size`. Observed live on
+    /// aws-mantle-1{,-fips} x86_64: a key whose first byte was 0x0A produced an all-zero dm-crypt
+    /// key, silently accepted on the non-FIPS kernel and rejected on the FIPS kernel as
+    /// `crypt: Error decoding and setting key (-EINVAL)`, which failed
+    /// `encrypt-unlock-local-fs.service` and took `preconfigured.target` with it.
```
arnaldo2792:
```suggestion
    /// The key is 64 raw bytes from /dev/random handed to cryptsetup on stdin. Without
    /// `--key-file=-` cryptsetup reads stdin as an interactive passphrase and stops at the first
    /// newline, and `--hash plain` then zero-pads whatever it got to `--key-size`. 
```

piyush-jena:
Force push shortens comments and adds a fix to race in rottweiler's `cryptsetup_resize` function. It occurs in case of plain mode only. We pass the key through STDIN. In case of LUKS2, the child process (`cryptsetup-resize`), reads the STDIN which contains the key and then resizes. In case of plain mode, the child process actually picks up the key from kernel keyring and runs the resize and exits. The race is that the child sometimes does this faster than the parent (rottweiler) can write to STDIN which leads to broken pipe error.

sources/rottweiler/src/key.rs
```diff
@@ -13,15 +13,9 @@ const KEYSTORE_PERMANENT: &str = "/.bottlerocket/keystore";
...
-fn ephemeral_encryption_keys_enabled() -> Result<bool> {
-    let features = bottlerocket_image_features::parse_image_features()
-        .with_whatever_context(|_| "failed to load image features")?;
-    Ok(features.ephemeral_encryption_keys)
-}
-
 fn keystore_dir() -> PathBuf {
-    let is_ephemeral_encryption_keys_enabled = ephemeral_encryption_keys_enabled().unwrap_or(false);
+    let is_ephemeral_encryption_keys_enabled =
+        system::ephemeral_encryption_keys_enabled().unwrap_or(false);
```
maherthomsi: This portion of code would be potentially swallowing an error from `ephemeral_encryption_keys_enabled()`.

It might be better to refactor the calls that use `.unwrap_or(false)` and instead have the logic looped into `ephemeral_encryption_keys_enabled()` to either display the error message, hard fail, or proceed with false if the feature is disabled

piyush-jena: updated. makes sense.

maherthomsi (approved):

packages/release/release.spec
piyush-jena: The conflict logic is made on the basis of image-feature. `crypt-luks` conflicts with `image-feature(ephemeral-encryption-keys)` and `ephemeral-crypt` requires `image-feature(ephemeral-encryption-keys)`. So the additional conflict with `ephemeral-crypt` would just be a redundant check.

piyush-jena:
Force push above rebases and fixes comments to hard fail if image-feature function fails instead of assuming default.

sources/rottweiler/src/key.rs
```diff
@@ -42,26 +34,30 @@ fn validate_key_id(key_id: &str) -> Result<()> {
     Ok(())
 }
 
+pub fn random_bytes() -> Result<Zeroizing<Vec<u8>>> {
```
arnaldo2792: nit: why is this public and not documented? I couldn't see where it was called outside this module so does it really have to be public.

sources/rottweiler/src/main.rs
```diff
@@ -262,6 +270,31 @@ struct EncryptDirectoryCmd {
     key_id: String,
 }
 
+#[derive(FromArgs)]
+#[argh(subcommand, name = "encrypt-and-attach")]
+/// Encrypt (plain mode) and attach a resource
```
arnaldo2792:
```suggestion
/// Encrypt using dm-crypt in plain mode and attach a resource
````

Same for the other places

File: sources/rottweiler/src/system.rs [hunk 11]
```diff
@@ -11,6 +11,9 @@ const SYSTEMD_CRYPTSETUP: &str = "/usr/lib/systemd/systemd-cryptsetup";
 const CRYPTSETUP: &str = "/usr/sbin/cryptsetup";
 const APICLIENT: &str = "/usr/bin/apiclient";
 const TPM2_PCREXTEND: &str = "/usr/bin/tpm2_pcrextend";
+// Size in bytes of the plain-mode dm-crypt key, and the same value in bits
+const PLAIN_KEY_BYTES: usize = 64;
+const PLAIN_KEY_SIZE_BITS: &str = "512";
```
arnaldo2792:
```suggestion
const PLAIN_KEY_BYTES: usize = 64;
const PLAIN_KEY_SIZE_BITS: usize= 64 * 8;;
```

And then you do `PLAIN_KEY_SIZE_BITS.to_string()`, just for consistency in the types.

sources/rottweiler/src/system.rs
```diff
@@ -88,16 +91,60 @@ pub fn cryptsetup_luks_format(device: &str, key_data: &[u8]) -> Result<()> {
...
 }
 
+/// Resize a block device to match the underlying device size.
+pub fn cryptsetup_resize(volume_name: &str, key_data: &[u8]) -> Result<()> {
+    if ephemeral_encryption_keys_enabled()? {
+        // In case of plain-mode dm-crypt the cryptsetup-resize command opts to read
+        // the key from kernel keyring if available. Because of this, the child
+        // process running cryptsetup-resize may finish before the parent process
+        // can put the key_data in the STDIN buffer causing the parent process
+        // rottweiler to fail with errno EPIPE. We opt to skip providing the keydata.
```
arnaldo2792: Comment is a bit unclear, are you saying that `--key-file` is omitted here to fix what you are describing?

packages/release/run-rottweiler.mount
```diff
@@ -4,12 +4,13 @@ DefaultDependencies=no
...
 After=selinux-policy-files.service
 Wants=selinux-policy-files.service
-Before=encrypt-datastore.service encrypt-local-fs.service encrypt-private-fs.service
+Before=encrypt-unlock-datastore.service encrypt-unlock-private-fs.service encrypt-unlock-local-fs.service
 
 [Mount]
 What=tmpfs
 Where=/run/rottweiler
 Type=tmpfs
+# noswap prevents encryption key from reaching persistent swap
```
arnaldo2792:
```suggestion
# noswap prevents the keystore from reaching persistent swap
```

docs/ENCRYPTED_STORAGE.md
```diff
@@ -208,6 +210,125 @@ This ensures encrypted data can only be accessed when booting a trusted configur
 - Before: `migrator.service`, `storewolf.service`
 - Required by: `migrator.service`, `storewolf.service`
 
+## Ephemeral Encryption Keys
+
+Variants with the `ephemeral-encryption-keys` image feature (in addition to `encrypted-storage`) replace the LUKS2 flow with plain-mode dm-crypt for `BOTTLEROCKET-DATA`, `BOTTLEROCKET-PRIVATE`, and the `EPHEMERAL-DATA` device:
+
+- Each key is 64 bytes from `/dev/random`, TPM2-sealed into the `/run/rottweiler` tmpfs keystore, and deleted (`ExecStartPost=rottweiler delete-key`) within the service that opened the device
```
arnaldo2792:
```suggestion
- Each key is 64 bytes from `/dev/random`, TPM2-sealed into the `/run/rottweiler` tmpfs keystore to prevent writing plain-text keys, and deleted (`ExecStartPost=rottweiler delete-key`) within the service that opened the device
```

docs/ENCRYPTED_STORAGE.md
```diff
@@ -299,6 +420,22 @@ cryptsetup luksFormat \
 
 This matches systemd's behavior and avoids unnecessary key stretching.
 
+### Plain Mode Formatting
+
+With `ephemeral-encryption-keys`, block devices are opened in plain mode. No header is written,
+so encryption and attach are a single operation, and `--hash plain` uses the key bytes verbatim:
```
arnaldo2792, line 431: Consistency in line-breaks

sources/rottweiler/src/key.rs
piyush-jena: we moved away from plain-text key, when this was required. Now it doesn't need to be pub. I fixed it.

piyush-jena:
Force push adds details to commit messages.

arnaldo2792 (approved):

PR #1035, opened 2026-09-03
packages: add nvidia-dra-driver-gpu

maherthomsi:
**Description of changes:**
Adds the nvidia-dra-driver-gpu package, which builds the NVIDIA GPU DRA driver (gpu-kubelet-plugin) and installs it as a host systemd service. The package is added to the bottlerocket-core-kit workspace and included in the kit manifest so it is available to consuming variants.

Bump settings sdk for this change to include https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/148

The DRA driver (`nvidia-dra-driver-gpu.service`) and the kubernetes device plugin (`nvidia-k8s-device-plugin.service`) both allocate the same GPUs, so only one runs at a time. This is enforced without a systemd `Conflicts=`: the DRA unit gates on `ConditionFileNotEmpty=/etc/nvidia-dra-driver-gpu/enabled`, and the marker is rendered `run` (non-empty) **only when DRA is enabled and the device plugin is disabled**. DRA is therefore opt-in and off by default.

  ### Both settings present

  | `kubelet-dra-drivers.nvidia.enabled` | `kubelet-device-plugins.nvidia.enabled` | DRA service | device-plugin service | GPU publisher |
  |:---:|:---:|:---:|:---:|:---:|
  | enabled  | disabled | **running**   | off         | DRA (ResourceSlice) |
  | enabled  | enabled  | off (defers)  | **running** | device plugin (`nvidia.com/gpu`) |
  | disabled | enabled  | off           | **running** | device plugin |
  | disabled | disabled | off           | off         | none |

  ### Settings extension not present (failure to adopt the DRA settings API)

  | Condition | DRA marker | DRA service | device-plugin service | GPU publisher |
  |:---|:---:|:---:|:---:|:---:|
  | `kubelet-dra-drivers` extension absent | not rendered | off | **running** (its own default) | device plugin |
   | `kubelet-dra-drivers` extension absent | not rendered | off | **not running** (disabled via API) | DRA |

  ### What "unset" means per setting

  | Setting | If unset | Effect |
  |:---|:---:|:---|
  | `kubelet-dra-drivers.nvidia.enabled` | treated as **disabled** (`default "false"`; shipped default is `false`) | DRA off |
  | `kubelet-device-plugins.nvidia.enabled` | treated as **enabled** (`default "true"`; classic behavior) | device plugin runs |

  **Out-of-the-box default** (both unset): DRA **off**, device plugin **running**.

  **Guarantees:** exactly one GPU publisher at any time; the device plugin wins when both are enabled (deterministic, no boot-time race); with no settings API the DRA driver stays off

**Testing done:**
- [x] Built the kits for x86_64 and aarch64
- [x] Launched a custom ami with the driver added to the variant and ensured that the DRA driver was loaded.
- - [x] Tested on aarch64
- - [x] Tested on x86_64 

**Terms of contribution:**


packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu.spec
```diff
@@ -0,0 +1,64 @@
...
+# Name of the upstream repo, used for the archive directory / source filename.
+%global gorepo dra-driver-nvidia-gpu
+
+%global gover 25.8.1
+%global rpmver %{gover}
+
+Name: %{_cross_os}nvidia-dra-driver-gpu
+Version: %{rpmver}
+Release: 1%{?dist}
+Epoch: 1
```
arnaldo2792: Should-fix: drop Epoch: 1. It isn't needed for a brand-new package with no prior release to sort against; Epoch is typically added retroactively only to fix version-ordering issues. Please remove it unless there's a specific version-ordering reason to keep it.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu.spec
```diff
@@ -0,0 +1,64 @@
...
+Summary: NVIDIA DRA driver (GPU kubelet plugin) for Kubernetes Dynamic Resource Allocation
```
arnaldo2792: Should-fix: tighten the Summary: to match house style. Convention is a short noun phrase: first word capitalized, no trailing period, no parenthetical, roughly 30-60 characters (for example Summary: Kubernetes device plugin for NVIDIA GPUs or Summary: An industry-standard container runtime). This one runs 84 characters and includes a parenthetical aside; since %description is %{summary}. by convention, the summary doubles as the description, so tightening it matters twice over.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu.spec
```diff
@@ -0,0 +1,64 @@
...
+License: Apache-2.0
+URL: https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu
+Source0: https://github.com/kubernetes-sigs/%{gorepo}/archive/refs/tags/v%{gover}.tar.gz#/%{gorepo}-%{gover}.tar.gz
```
arnaldo2792: Nit: refs/tags/ isn't the dominant house form for Source0. Most specs use archive/v%{gover}/%{gorepo}-%{gover}.tar.gz (for example nvidia-container-toolkit, containerd-2.3, cri-tools). Consider matching that pattern for consistency, though both forms work on GitHub.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu.spec
```diff
@@ -0,0 +1,64 @@
...
+Source2: nvidia-dra-driver-gpu-exec-start-conf
+
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}. This package ships the gpu-kubelet-plugin binary from the NVIDIA
+DRA driver and runs it as a host systemd service (rather than a Helm-installed
+DaemonSet). The plugin registers with the local kubelet, discovers GPUs via
+NVML on the host, generates CDI specifications, and publishes ResourceSlices to
+the Kubernetes API server using the node's own kubeconfig identity.
```
arnaldo2792: Should-fix: %description should be the bare %{summary}. macro. All 24 existing %description stanzas in the kit (mains and subpackages) use exactly this form, with no prose added. This package's hybrid %{summary}. <prose> has no precedent.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu.spec
```diff
@@ -0,0 +1,64 @@
...
+
+install -d %{buildroot}%{_cross_unitdir}
+install -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_unitdir}/nvidia-dra-driver-gpu.service.d
+
+install -D -m 0644 %{S:2} %{buildroot}%{_cross_templatedir}/nvidia-dra-driver-gpu-exec-start-conf
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
```
arnaldo2792: Two licensing/attribution gaps. 1. The archive ships a NOTICE file alongside LICENSE, but it isn't installed; consider %license LICENSE NOTICE. 2. %install doesn't call %cross_scan_attribution and %files doesn't ship %{_cross_attribution_vendor_dir}, even though upstream vendors its dependencies at this tag and the default GOFLAGS includes -mod=vendor. Please confirm whether vendor attribution scanning is expected for this package.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu.service
```diff
@@ -0,0 +1,23 @@
+[Unit]
+Description=NVIDIA DRA driver GPU kubelet plugin
+After=kubelet.service
+Wants=kubelet.service
+# Only meaningful on a node that actually has an NVIDIA GPU present.
+ConditionPathExists=/dev/nvidiactl
```
arnaldo2792: Not needed.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
```diff
@@ -0,0 +1,32 @@
...
+# to the API server using the node's own kubeconfig identity.
+ExecStart=/usr/bin/gpu-kubelet-plugin \
+    --node-name {{settings.kubernetes.hostname-override}} \
+    --namespace kube-system \
+    --nvidia-driver-root / \
+    --container-driver-root / \
+    --cdi-root /etc/cdi \
+    --nvidia-cdi-hook-path /usr/bin/nvidia-cdi-hook \
+    --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
+    --image-name nvcr.io/nvidia/k8s-dra-driver-gpu:v25.8.1 \
```
arnaldo2792: Set --image-name to "". MPS support in the DRA driver is alpha and disabled by default, and an empty string still satisfies the Required: true flag on this argument, so there's no need to hardcode v25.8.1 here and risk it drifting out of sync with %global gover in the spec and the url/sha512 in Cargo.toml. Once MPS support lands, the image should be provided from the node itself, similar to how the pause container is handled, rather than pulled externally. That needs more research since the binary appears to call chroot, so the image may need to be supplied via mounts instead. We should decide how we want to provide this image before enabling MPS.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
```diff
@@ -0,0 +1,32 @@
...
+    --healthcheck-port -1
```
arnaldo2792, line 25: Add a comment explaining why -1 is used here. Per cmd/gpu-kubelet-plugin/health.go:50-53, any negative value short-circuits startHealthcheck before it creates a listener, so -1 disables the gRPC healthcheck service. An inline comment documenting that intent would help future readers, since -1 is also the flag's default and it's not obvious this line is deliberate.

maherthomsi:
Added settings-sdk changes to this PR

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
```diff
@@ -0,0 +1,37 @@
...
+# --image-name is only used to render MPS control-daemon pods; MPS is alpha and
+# disabled by default, and an empty string satisfies the required flag without
+# hardcoding an externally-pulled image tag.
+# --healthcheck-port -1 disables the gRPC healthcheck service: any negative
+# value short-circuits startHealthcheck before it creates a listener.
+RemainAfterExit=false
+{{else}}
+Restart=no
+ExecStart=/usr/bin/true
+Type=oneshot
```
vigh-m, line 47: I think the `if true` case here should also have a `Type=oneshot`. Once the DRA plugin is started by your `ExecStart` command, the service can exit successfully and does not need to be held open by systemd.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
```diff
@@ -0,0 +1,37 @@
...
+kubernetes = "v1"
+kubelet-dra-drivers = "v1"
+std = { version = "v1", helpers = ["if_not_null", "default"] }
++++
+[Service]
+ExecStartPre=
+ExecStart=
+{{#if (eq (default "false" settings.kubelet-dra-drivers.nvidia.enabled) "true")}}
+ExecStartPre=/usr/bin/sleep 0.1
+ExecStartPre=/usr/bin/test -d /var/lib/kubelet/plugins_registry
```
vigh-m, line 13: Any reason that these are repeated in the base service and in the rendered drop-in as well?

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
maherthomsi: We need this in order to keep the kubelet DRA plugin alive and supervised.  This is a live service the kubelet keeps talking to in order to answer GPU pod scheduling, keeping the GPU's ResourceSlices published for the scheduler, and holds the socket the kubelet connects through.  We need this running so that the socket does not disappear.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
maherthomsi: Good catch, theres no reason to keep them in both.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
```diff
@@ -0,0 +1,35 @@
...
+    --container-driver-root / \
+    --cdi-root /etc/cdi \
+    --nvidia-cdi-hook-path /usr/bin/nvidia-cdi-hook \
+    --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
+    --image-name "" \
+    --healthcheck-port -1
+# --image-name is only used to render MPS control-daemon pods; MPS is alpha and
+# disabled by default, and an empty string satisfies the required flag.
+# --healthcheck-port -1 disables the gRPC healthcheck service: any negative
+# value short-circuits startHealthcheck before it creates a listener.
```
arnaldo2792: nit: they should go on top of the flags they reference.

maherthomsi:
Addresses review feedback on the nvidia-dra-driver-gpu package:

- spec: add `Requires: %{_cross_os}nvidia-container-toolkit` - the rendered unit invokes `/usr/bin/nvidia-cdi-hook`, which that package ships.
- service (base unit): replace the `ExecStart=/usr/bin/false + Type=simple + Restart=always placeholder` with an inert no-op (`Type=oneshot`, `RemainAfterExit=true`, `ExecStart=/usr/bin/true`). It no longer retries forever / logs to console before the drop-in renders, and the ExecStartPre checks are no longer duplicated between the base unit and the drop-in.
- exec-start-conf (drop-in): reference the optional `settings.kubernetes.hostname-override` only inside `{{#if_not_null}}`, so an unset value no longer fails the strict render. Move the daemon's `Type=simple/Restart=always/RestartSec/RemainAfterExit=false` into the enabled branch (previously inherited from the base) and drop the redundant `{{else}}` branch since the inert base now covers the disabled/unset case.

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
```
arnaldo2792, line 1: You should mimic what the kubelet-device-plugin does:

- Provide a unit that just works - with the drop-in for the host name
- Provide an override that runs `true` when disabled
- And in the same template, provide the "enabled" case and run the unit

packages/nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-exec-start-conf
maherthomsi: We disable the healthcheck-port here because the DRA driver lives as a systemd service and not a pod. Nothing probes it and the listener would serve no purpose.  Liveness is handled by systemd managing restarts via Restart=always.

vigh-m (approved):

piyush-jena (dismissed):

piyush-jena:
Looks good to me!

I had this design in mind which I think should work:
`nvidia-k8s-device-plugin.service` creates a marker file (`nvidia-k8s-device-plugin.ran`) after running. `nvidia-dra-driver-plugin.service` runs after/requires `nvidia-k8s-device-plugin.service`, if it sees the marker file then doesn't run, if it doesn't see the marker file, runs based on the settings.

The design I described comes with a very small boot time penalty but imo is cleaner. Your design works too.

vigh-m (approved):

piyush-jena (approved):

PR #1041, opened 2026-09-09
Add support for nvidia driver selection at boot time

vigh-m:
**Description of changes:**
Adds boot-time NVIDIA driver-branch selection for dual-branch images. 

The kernel-kit change ([bottlerocket-kernel-kit#536](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/536)) makes the `lts` and `pb` driver branches co-installable and namespaces each branch's units as `nvidia-<branch>-*`

  **ghostdog `match-nvidia-branch`**
  - Implement new `match-nvidia-branch` subcommand. It resolves the target branch from `settings.kernel.drivers.nvidia.branch` (added in [bottlerocket-settings-sdk#154](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/154)), falling back to instance-type detection when the setting is unset
  -  writes a `/run/nvidia/branch-<branch>` marker file

  **os package wiring**
  - New `nvidia-select-branch` subpackage requiring `nvidia-lts`, `nvidia-pb`, only installed when `nvidia-dual-branch` image feature is enabled ([twoliter#747](https://github.com/bottlerocket-os/bottlerocket/pull/4939)), and building and nvidia variant
  - Adds `nvidia-select-branch.service` to execute `ghostdog match-nvidia-branch` in a dual driver image

  **Service reordering**
  - `ecs-gpu-init.service` and `generate-cdi-specs.service` now order `After=drivers.target` instead of the branch-specific `load-*-kernel-modules.service` units, since those unit names are now per-branch.

#### Backward compatibility (namespaced kernel-kit + existing core-kit/Twoliter)

- [x] nvidia-r580 namespaced kernel-kit; boot on:
  - [x] g4dn
  - [x] g5
  - [x] g6
  - [x] g6f
  - [x] g7
  - [x] g7f
- [x] nvidia-r595 namespaced kernel-kit; boot on:
  - [x] g4dn
  - [x] g5
  - [x] g6
  - [x] g6f
  - [x] g7
  - [x] g7f

#### Image feature — happy path

- [x] Enable image feature; boot on:
  - [x] g4dn
  - [x] g5
  - [x] g6
  - [x] g6f
  - [x] g7
  - [x] g7f.xlarge

**Terms of contribution:**


sources/ghostdog/src/nvidia.rs
```diff
@@ -0,0 +1,516 @@
...
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .context(error::TokioRuntimeSnafu)?;
+    let uri = "/?prefix=settings.kernel.drivers.nvidia.branch".to_string();
+    let value = match rt.block_on(apiclient::get::get_uri(constants::API_SOCKET, uri)) {
+        Ok(value) => value,
+        Err(e) => {
+            log::warn!("Failed to read driver-branch setting from API, using PCI detection: {e}");
+            return Ok(None);
```
qianxjcraig: Could we distinguish an API read failure from a successfully read, unset branch setting?

Returning Ok(None) here allows selection to succeed using PCI detection. If a user has pinned PB on a GPU supported by both branches, an API failure will therefore select LTS and write its marker despite the configured pin.

Could we retry transient failures and return an error if the setting remains unreadable, reserving PCI fallback for a successful response with no pin?

packages/os/nvidia-select-branch-ordering.conf
piyush-jena, line 1: is there a reason for this interesting naming convention? I think such names are given in the spec file when installing.

packages/os/os.spec
```diff
@@ -91,6 +91,11 @@ Source400: COPYRIGHT
 Source401: LICENSE-MIT
 Source402: LICENSE-APACHE
 
+# 5xx sources: NVIDIA dual-branch activation assets (shipped by ghostdog-nvidia)
+Source500: nvidia-select-branch.service
+Source510: nvidia-branch-gate.conf.in
+Source540: drivers.target-10-nvidia-select-branch.conf
```
piyush-jena: nit: why not just 500, 501, 502?

qianxjcraig (approved):

sources/ghostdog/src/main.rs
```diff
@@ -411,17 +322,14 @@ fn emit_device_name(device_name: &str) {
...
-        error::DriverMismatchSnafu {
-            requested: driver_name,
-            preferred: preferred_driver
-        }
-    );
-    Ok(())
+/// A `match-nvidia-driver` `ExecCondition` returning `false` means "not this
+/// flavour" -- the normal skip path, hit by most invocations. Exit non-zero
+/// quietly so systemd skips the unit without logging a spurious failure; a
+/// matched flavour returns and the process exits 0.
```
ginglis13:
```suggestion
/// Evaluate it the `match-nvidia-driver` `ExecCondition`
```

sources/ghostdog/src/main.rs
```diff
@@ -411,17 +322,14 @@ fn emit_device_name(device_name: &str) {
...
+fn exit_if_no_match(matched: bool) {
+    if !matched {
+        std::process::exit(1);
```
ginglis13: Seems the pattern here should be to return a `Result` and allow callers to handle exit

sources/ghostdog/src/nvidia.rs
```diff
@@ -0,0 +1,522 @@
...
+//! Two independent decisions live here, each exposed as one entry point:
+//!
+//! - **Branch (version): `lts` vs `pb`** — [`match_branch`] backs the
+//!   `match-nvidia-branch` subcommand. It picks the branch and writes
+//!   `/run/nvidia/branch-<branch>`, a marker that systemd `ConditionPathExists`
+//!   drop-ins use to gate the per-branch overlay and module-load units. Resolution
+//!   order: an explicit `settings.kernel.drivers.nvidia.branch` pin, else the PCI devices
+//!   present, else `lts`.
+//!
+//! - **Flavour (role): `tesla`/`open-gpu`/`grid`** — [`match_flavour`] backs the
```
ginglis13: seems we've switched to the UK spelling :) 🍵 nitpick but for consistency's sake, please revert `flavour` -> `flavor` in these changes (that's the spelling we use elsewhere in bottlerocket)

ginglis13 (commented):
still working through the PR, just two initial nits

sources/ghostdog/src/main.rs
```diff
@@ -411,17 +322,14 @@ fn emit_device_name(device_name: &str) {
...
+    }
```
ginglis13: I don't love that we're now embedding this in a helper method. It seems all this buys us, along with the separation of return type on match check to a Result<bool> is an extra log line for an `Error`. WDYT about reverting back to the previous logic here of "match -> exit 0, mismatch OR error -> exit 1" ?

sources/ghostdog/src/nvidia.rs
```diff
@@ -0,0 +1,522 @@
...
+}
+
+/// The NVIDIA driver branch (version) to activate.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum NvidiaBranch {
+    /// Long-Term Support branch (currently r580).
+    Lts,
+    /// Product Branch (currently r595).
+    Pb,
+}
```
ginglis13, line 114:
```suggestion
enum NvidiaBranch {
    /// Long-Term Support branch
    Lts,
    /// Product Branch
    Pb,
}
```

sources/ghostdog/src/nvidia.rs
```diff
@@ -0,0 +1,522 @@
...
+    #[serde(rename = "devid")]
+    device_id: String,
+    #[serde(rename = "subdevid")]
+    subdevice_id: Option<String>,
+    #[serde(rename = "subvendorid")]
+    subvendor_id: Option<String>,
+    name: String,
+    /// List of features the device supports. We are looking for "kernelopen" to match the driver
+    features: Vec<String>,
+}
```
ginglis13, line 105: this copied over but dropped comments

```suggestion
#[derive(Eq, Debug, Deserialize, Hash, PartialEq)]
/// The GPU Device Data contains various features of the device. Only Name, Device ID, and Features are required
/// for a particular device
struct GpuDeviceData {
    #[serde(rename = "devid")]
    /// PCI Device ID
    device_id: String,
    #[serde(rename = "subdevid")]
    /// PCI Subdevice ID
    subdevice_id: Option<String>,
    #[serde(rename = "subvendorid")]
    /// PCI Subvendor ID
    subvendor_id: Option<String>,
    /// Name of the device
    name: String,
    /// List of features the device supports. Noteably we are looking for "kernelopen" to match the driver
    features: Vec<String>,
}
```

sources/ghostdog/src/nvidia.rs
```diff
@@ -0,0 +1,522 @@
...
+        present_devices.len(),
+        open_gpu_device_set.len()
+    );
+
+    // Return early with grid if a match is made for these subdevices
+    if subdevice_ids.any(|subdevice| NVIDIA_GRID_SUBDEVICES.contains(subdevice.as_str())) {
+        log::debug!("flavour detection: matched a GRID subdevice");
+        return Ok("grid".to_string());
+    }
+
```
ginglis13, line 218: could optimize by doing this check in the iteration of `present_devices`, rather than constructing the `subdevice_ids` map and then iterating. Thinking like:

```
let grid_present = present_devices
    .iter()
    .filter(|x| NVIDIA_GRID_DEVICE_IDS.contains(x.device().as_str()))
    .any(|x| {
        let sub = format!("0x{}", x.subsystem_device().as_deref().unwrap_or("").to_uppercase());
        NVIDIA_GRID_SUBDEVICES.contains(sub.as_str())
    });
if grid_present {
    log::debug!("flavour detection: matched a GRID subdevice");
    return Ok("grid".to_string());
}
```

packages/ecs-gpu-init/ecs-gpu-init.service
```diff
@@ -3,7 +3,7 @@ Description=Initialize ECS GPU config
 # This has to be executed after the kernel modules are loaded
 # otherwise the userspace component of the driver will fail to
 # query the /dev devices
-After=load-tesla-kernel-modules.service load-open-gpu-kernel-modules.service
+After=drivers.target
```
ginglis13, line 6: let's update the comment here to reflect this change

packages/nvidia-container-toolkit/generate-cdi-specs.service
```diff
@@ -3,7 +3,7 @@ Description=Generate CDI specifications
 # This has to be executed after the kernel modules are loaded
 # otherwise the userspace component of the driver will fail to
 # query the /dev devices
-After=load-tesla-kernel-modules.service load-open-gpu-kernel-modules.service
+After=drivers.target
```
ginglis13, line 6: let's update the comment here to reflect this change

packages/release/nvidia-driver-branch-deprecation
```diff
@@ -0,0 +1,6 @@
+[required-extensions]
+kernel = { version = "v1", optional = true }
++++
+{{~#if settings.kernel.drivers.nvidia.branch~}}
```
maherthomsi: Should this be gated behind a variant that doesn't support the `nvidia-dual-branch` feature or would this be gated elsewhere?

packages/ecs-gpu-init/ecs-gpu-init.service
vigh-m: I think the comment still applies. We're just waiting for the target to complete instead of specific services

ginglis13 (approved):
LGTM, thanks for the testing deets

ginglis13 (approved):

PR #1042, opened 2026-09-10
Update to boot UKIs directly from ESP

piyush-jena:
**Description of changes:**
- **release:** The UKI FIPS HMAC path moves to `/boot/EFI/BOOT/.bottlerocket.efi.hmac` (the UKI now lives at the ESP fallback path). Since, `shim` is not required for UKIs, it is made conditional. 
- **prairiedog:** `prepare_boot()` mounts ESP instead of XBOOTLDR on UKI images.
- **signpost:** Replace the `XBOOTLDR` type-GUID constant and its resolvers with ESP equivalents.

Use twoliter PR https://github.com/bottlerocket-os/twoliter/pull/742 for reference

**Testing done:** See twoliter https://github.com/bottlerocket-os/twoliter/pull/745.
- Build this with old twoliter and then build aws-k8s-1.36 AMI ✅ 

**Terms of contribution:**


packages/release/release.spec
```diff
@@ -218,7 +218,7 @@ Requires: %{_cross_os}netdog
 Requires: %{_cross_os}os
 Requires: %{_cross_os}pciutils
 Requires: (%{_cross_os}rdma-core if %{_cross_os}variant-platform(aws))
-Requires: %{_cross_os}shim
+Requires: (%{_cross_os}shim or %{_cross_os}image-feature(uki-image))
```
arnaldo2792, line 221: To test this, build with an older version of twoliter. It should just work but just in case.

sources/api/prairiedog/src/error.rs
```diff
@@ -44,8 +44,8 @@ pub(super) enum Error {
     #[snafu(display("Failed to scan the OS disk's partition table: {}", source))]
```
arnaldo2792, line 44: Commit message is wrong: `prepare_boot` doesn't prepare the `/boot` partition to boot, it just mounts it.

packages/release/release.spec
piyush-jena: works

sam-berning (approved):

maherthomsi (approved):

piyush-jena:
I will merge it post twoliter merge.

vyaghras (approved):

